PR #3, opened 2013-11-20
Correct parameter types for Nested field in docs

roadsideseb:
I've just started using your awesome little library. I'm really liking it so far :thumbsup:. Reading through the docs I found those two typos and fixed them.


sloria:
Good catch. Thanks!

PR #4, opened 2013-11-20
Pass attribute name through to method and function field callables

roadsideseb:
As the branch name says, this is a suggestion. I've been looking into a use case for a serializer that has several `Method` fields that serialise different attributes but basically do the same thing. Unfortunately, within the callables used by the `Method` and `Function` field, I don't know anything about the attribute name of the field. It would be nice to have that, especially since the `key` value is available in the `to_output` methods of the fields and just needs to be passed in. A simple example would be something like this:

```
class CustomSerializer(Serializer):
    first_name = fields.Method('get_lowercase_attribute')
    last_name = fields.Method('get_lowercase_attribute')

    def get_lowercase_attribute(self, key, obj):
        return getattr(obj, key)
```

I've added the argument to both fields in this PR. This will cause problems for existing code that uses these fields so I would be good to get your input on whether you want backwards compatibility or not. What are your thoughts?

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -400,8 +400,10 @@ class Method(Raw):
     """A field that takes the value returned by a Serializer method.
 
     :param str method_name: The name of the Serializer method from which
-        to retrieve the value. The method must take a single argument ``obj``
-        (in addition to self) that is the object to be serialized.
+        to retrieve the value. The method must take two arguments, the first is
+        ``key`` identifying the name of the attribute to be serialised and the
+        second is ``obj`` which is the object (in addition to self) that is the
+        object to be serialized.
     """
 
     def __init__(self, method_name):
@@ -410,7 +412,7 @@ class Method(Raw):
 
     def output(self, key, obj):
         try:
-            return getattr(self.parent, self.method_name)(obj)
+            return getattr(self.parent, self.method_name)(key, obj)
         except AttributeError:
             pass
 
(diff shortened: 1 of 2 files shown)


sloria:
Thanks for the suggestion. I appreciate that you took care to change the docs and tests.

I don't really have a problem with this, but I'd like to see a more realistic use case (the example above is the same as using `fields.Raw()` for both fields).

The preferred way to serialize multiple fields the same way is to create a custom field.

``` python
class MyField(fields.Raw):
    def format(self, value):
        # ...Do stuff to value...
        return formatted_value

class CustomSerializer(Serializer):
    first_name = MyField()
    last_name = MyField()
```

roadsideseb:
You are right, this seems to be a much better way to handle it. And more explicit. The use case I was talking about is basically using a metaclass to generate a default serialiser based on a class definition. Basically something like this:

``` python
class SomeModel(JsonModel):
    title = json_fields.CharField()
    text = json_fields.CharField()
    nested_models = json_fields.ModelList()

# serialiser generated by a metaclass
class SomeModelSerializer(MyModelSerializer):
    title = fields.String()
    text = fields.String()
    # serialise_nested_models is defined in the MyModelSerialiser base class
    nested_models = fields.Method('serialise_nested_models')

m = SomeModel(title='The title', text="Some text to go with the title")
print m.serializer(m).data
```

The above example doesn't really work in this instance because I don't know in the `serialise_nested_models` method which attribute on the object I have to serialise.

But the above is just for reference, your suggestion makes a lot more sense to use for what I am trying to achieve. Thanks for taking the time and thank for some great work!

sloria:
Indeed, a custom field is more explicit and more flexible than using inheritance. I appreciate the suggestion, though!

PR #7, opened 2013-12-19
Fields selection

ecarreras:
Add a selection field, wich only let you to put a value that is in the selection.


sloria:
I like this idea. Thanks for the contribution.

PR #14, opened 2014-02-17
Fix integer default

kalasjocke:
This pull requests fixes the bug where serializing `None` on integer fields without a default yielded a `0.0` float instead of a 0 integer.


sloria:
Thanks for this patch!

PR #15, opened 2014-02-17
PEP8

kalasjocke:
When adding tests for the failing integer default my PEP8 linter was quite annoying. I decided to go over the tests and fix whatever it complained about.

Thanks for a great library :exclamation: :+1:


sloria:
Looks good. Thanks for your contributions!

PR #17, opened 2014-04-10
added required fields flag, fixes #16

### LINKED ISSUE #16: Required fields
I can't seem to find in the docs if there is a way to make certain fields required?  Is there no present implementation that marks the `is_valid` call as invalid if a certain field is missing?  Would be willing to contrib :smile:

DeaconDesperado:
Added a required flag that checks a field is defined on input before defaults are returned.


sloria:
Looks good. I've updated the tests to use the new pytest style. Thanks for the contribution!

PR #20, opened 2014-05-07
Fix attribute access for SQLAlchemy tuple types

hcarvalhoalves:
Fixes serialization for SQLAlchemy's KeyedTuple, and other types that look like dicts but don't respond to obj["key"].

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -151,6 +151,10 @@ def _get_value_for_key(key, obj, default):
             return obj[key]
         except KeyError:
             return default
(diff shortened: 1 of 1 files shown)


sloria:
@hcarvalhoalves 
There are failing tests because there are some situations when a `TypeError` should _not_ be caught. 

I think the simple solution for handling KeyedTuples and namedtuples would be to modify `_get_value_for_key` to check `hasattr(obj, key)` before trying to do dict access.

Thanks for catching this.

PR #25, opened 2014-06-20
Move required validation to avoid custom validation messages overriding required message

asteinlein:
When having a custom validator with a custom error message, this error message would be used even if the validation failed due to required field validation error. The same applies for built-in fields where you want custom error messages as well, for instance:

``` python
email = fields.Email(required=True, error='Invalid email address. Try again.')
```

When validating this without supplying a value for the field, the message "'Invalid email address. Try again." is the one used, while I would expect a message about the missing required value (as is common in forms -- required fields is usually a message separate from invalid input).

To solve this, some refactoring was needed, moving the required validation to validated(), which I feel is a more correct location for validation anyhow. This also led to some cleanup of where validated() is wrapped.

Finally, the required validation error message is changed to what I feel better resembles the other messages in the library: The error dictionary already contains the field name as the key, so it is redundant to repeat it in the message (and consistent with other error messages).

Please let me know if you have any input or questions. Updated tests and docs included. We would love to have this merged so we can avoid using our own fork in production. :)

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -156,7 +156,7 @@ You can make a field required by passing ``required=True``. An error will be sto
     user = User(name=None, email='foo@bar.com')
     serialized = UserSerializer(user)
     serialized.is_valid()  # False
-    serialized.errors  # {'name': '"name" is a required field.'}
+    serialized.errors  # {'name': 'Missing data for required field.'}
 
 
 Specifying Attribute Names
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -48,7 +48,7 @@ __all__ = [
 
 
 def validated(f):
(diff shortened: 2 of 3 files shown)


sloria:
I like this patch. Thanks for taking the time to contribute back.  

The changes seem to fail the tests on Python 3; I cannot merge this until those breakages are addressed.

asteinlein:
Thanks for the quick reply!

And indeed, I forgot to test on Python 3, even though we use Py3 in production ourselves, heh. Will fix ASAP.

asteinlein:
Fixed and updated. :)

sloria:
Looks good! Thanks

sloria:
This will be in the next release, which I plan to publish within the next week.

PR #36, opened 2014-08-12
Issue/#34 - Passing a list of callables for validation

Ketouem:
Added the possiblity to pass a list of callable for validation as asked here: https://github.com/sloria/marshmallow/issues/34

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -183,6 +183,18 @@ You can perform additional validation for a field by passing it a ``validate`` c
     result, errors = ValidatedUserSerializer().dump(jagger)
     errors  # => {'age': 'User is over the hill'}
 
+If you have multiple validations to perform, you can even pass a list of callables. So with a slight modification our previous example would become:
+
+.. code-block:: python
+
+    class ValidatedUserSerializer(UserSerializer):
+        age = fields.Number(validate=[lambda n: n >= 18, lambda n: n <= 40],
+                            error='User is over the hill')
+
+    jagger = User(name="Mick", email="mick@stones.com", age=71)
+    result, errors = ValidatedUserSerializer().dump(jagger)
+    errors  # => {'age': 'User is over the hill'}
+
 .. note::
 
     If you set ``strict=True`` in either the Serializer constructor or as a ``class Meta`` option, an error will be raised when invalid data are passed in.
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -236,7 +236,7 @@ class Raw(FieldABC):
     _CHECK_REQUIRED = True
 
     def __init__(self, default=None, attribute=None, error=None,
-                validate=None, required=False):
+                 validate=None, required=False):
         self.attribute = attribute
         self.default = default
         self.error = error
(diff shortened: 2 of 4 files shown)


marshmallow/fields.py
```diff
@@ -261,12 +261,17 @@ def _call_with_validation(self, method, exception_class, *args, **kwargs):
...
             func = getattr(self, method)
             output = func(*args, **kwargs)
-            if callable(self.validate):
-                if not self.validate(output):
-                    msg = 'Validator {0}({1}) is not True'.format(
-                        self.validate.__name__, output
-                    )
-                    raise exception_class(self.error or msg)
+            if self.validate is not None:
+                validators = [i for i in self.validate] if type(self.validate) == list else [self.validate]
```
sloria, line 265: Use `isinstance` for type-checking here. It should be possible to pass a tuple of functions. You can therefore use `isinstance(self.validate, (list, tuple))` or even `utils.is_iterable_but_not_string(self.validate)` to be even more robust (that would handle generators, for example).

Also, the list comprehension is unnecessary here. `[i for i in self.validate]` is the same as `self.validate` in this context.

marshmallow/fields.py
```diff
@@ -261,12 +261,17 @@ def _call_with_validation(self, method, exception_class, *args, **kwargs):
...
+            else:
+                validators = []
+            if len(validators) > 0:
+                msg = lambda v: 'Validator {0}({1}) is not True'.format(
```
sloria, line 269: Why not create this message in the for loop instead of creating a `lambda` here?

``` python
for validator in validators:
    if not validator(output):
        msg = 'Validator {0}({1}) is not True'.format(validator.__name__, output)
        raise exception_class(self.error or msg)
```

sloria:
Thank you for your contribution. I like that you were thorough in adding tests and docs. 

I think you your implementation can be made simpler and more robust to different collection types by generating the list of validators like so:

``` python
if self.validate:
    if callable(self.validate):
        validators = [self.validate]
    elif utils.is_iterable_but_not_string(self.validate):
        validators = self.validate
    else:
        raise ValueError('validate must be a callable or collection of callables')
    for func in validators:
        if not func(output):
            msg = 'Validator {0}({1}) is not True'.format(
                self.validate.__name__, output
            )
            raise exception_class(self.error or msg)
```

docs/quickstart.rst
```diff
@@ -183,6 +183,18 @@ You can perform additional validation for a field by passing it a ``validate`` c
     result, errors = ValidatedUserSerializer().dump(jagger)
     errors  # => {'age': 'User is over the hill'}
 
+If you have multiple validations to perform, you can even pass a list of callables. So with a slight modification our previous example would become:
+
+.. code-block:: python
+
+    class ValidatedUserSerializer(UserSerializer):
+        age = fields.Number(validate=[lambda n: n >= 18, lambda n: n <= 40],
```
sloria, line 191: Since this example isn't really an improvement of the previous example, I think this code example can be replaced with a simple note after the previous example.

```
.. note::

    You may also pass a list of callables to the ``validate`` parameter.
```

I would like to keep the Quickstart guide as concise as possible without sacrificing completeness.

Ketouem:
I totally see your point, I'll make the implementation more robust and I'll write unit test cases for different collection types (and update the documentation). I close this PR, I'll open another one with the modified code.

PR #37, opened 2014-08-13
Issue/#34 - Passing a list of callables for validation

Ketouem:
I took sloria's comment into consideration and refactored slightly the code to handle collections and generator (+expanding the unit test cases)

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -261,19 +261,16 @@ class Raw(FieldABC):
         try:
             func = getattr(self, method)
             output = func(*args, **kwargs)
-
-            if utils.is_iterable_but_not_string(self.validate):
-                if not utils.is_generator(self.validate):
-                    validators = self.validate
+            if self.validate:
+                if utils.is_iterable_but_not_string(self.validate):
+                    if not utils.is_generator(self.validate):
+                        validators = self.validate
+                    else:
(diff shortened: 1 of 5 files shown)


marshmallow/fields.py
```diff
@@ -261,12 +261,25 @@ def _call_with_validation(self, method, exception_class, *args, **kwargs):
...
-                if not self.validate(output):
+
+            if utils.is_iterable_but_not_string(self.validate):
+                if not utils.is_generator(self.validate):
+                    validators = self.validate
+                else:
+                    validators = [i for i in self.validate()]
+            elif callable(self.validate):
+                validators = [self.validate]
+            elif self.validate is None:
```
sloria: Instead of checking against `None` here, why not check the truthiness of `self.validate` at the beginning? That way you can remove this branch as well as check for `if len(validators) > 0`

``` python
if self.validate:
    if utils.is_iterable_but_not_string(self.validate):
        # ...
    for validator in validators:
        # ...
```

sloria:
Thanks for making those changes. Just a minor change to make and this should be good to merge.

Ketouem:
I've just made the tweaks according to your advices.

sloria:
Looks good. Thanks for your contribution! :octopus:

PR #38, opened 2014-08-13
Encapsulate data

ecarreras:
Add a feature for encapsulate data in the JSON and generate a JSON with ember compatibility. The key of encapsulate is pluralized automatically if is used with `many=True`.

``` python
user1 = User(name="Mick", email="mick@stones.com")
user2 = User(name="Keith", email="keith@stones.com")
users = [user1, user2]
serializer = UserSerializer(many=True, encapsulate='user')
results, errors = serializer.dump(users)
# {'users': [
# {'created_at': 'Fri, 08 Nov 2013 17:02:17 -0000',
#   'email': u'mick@stones.com',
#   'name': u'Mick'},
#  {'created_at': 'Fri, 08 Nov 2013 17:02:17 -0000',
#   'email': u'keith@stones.com',
#   'name': u'Keith'}]}
```

I don't know if the name `encapsulate` is right or if you will like this feature. If is accepted then I'll write the documentation.

And the examples can be changed to use this feature.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -19,6 +19,9 @@ from marshmallow.exceptions import (
     ForcedError,
 )
 
+from inflection import pluralize
+
+
 __all__ = [
     'Marshaller',
     'UnMarshaller',
@@ -101,7 +104,7 @@ class Marshaller(object):
         #: Dictionary of errors stored during serialization
         self.errors = {}
 
-    def serialize(self, obj, fields_dict, many=False):
+    def serialize(self, obj, fields_dict, many=False, encapsulate=None):
         """Takes raw data (a dict, list, or other object) and a dict of
         fields to output and serializes the data based on those fields.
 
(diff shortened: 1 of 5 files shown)


sloria:
I am hesitant to add complexity and make assumptions about data format for a relatively specific use case (supporting Ember's data format), so I'm closing this. 

That said, you could use use the `data_handler` method to add a root namespace to your final serialized output.

Something like:

``` python
class UserSerializer(Serializer):
    NAME = 'user'
    name = fields.String()

@UserSerializer.data_handler
def add_namespace(serializer, data, obj):
    if serializer.many:
        namespace = inflect.pluralize(serializer.NAME)
    else:
        namespace = serializer.NAME
    return {
         namespace: data
     }
```

Thanks for sending this PR!

ecarreras:
Thank you for your tips, what do you think about having a contrib dir with a NamespacedSerializer. Because I have to apply this in all my serializers and all my projects. If not, maybe is better to create a separate python package?

sloria:
@ecarreras One way to apply namespacing to all your serializers would be to create a subclass of  `Serializer` that overrides the `_postprocess` method like so:

``` python
from marshmallow import Serializer, fields

class NamespacedSerializer(Serializer):
    NAME = None

    def _postprocess(self, data, obj):
        data = Serializer._postprocess(self, data, obj)
        if self.NAME:
            namespace = self.NAME
            if self.many:
                namespace = inflect.pluralize(self.NAME)
            data = {namespace: data}
        return data
```

You could then declare your concrete serializers like so:

``` python
class UserSerializer(NamespacedSerializer):
    NAME = 'user'
    name = fields.String()
    email = fields.Email()

ser = UserSerializer()
user = User('Keith', email='keith@stones.com')
result = ser.dump(user)
result.data
# {'user': OrderedDict([('name', 'Keith'), ('email', 'keith@stones.com')])}
```

sloria:
I've also just added documentation that shows how you can do this using `class Meta` options: http://marshmallow.readthedocs.org/en/dev/extending.html#extending-class-meta-options

PR #46, opened 2014-09-27
Flexibly inherit fields following method resolution order.

jmcarp:
Allow schema subclasses to inherit fields from plain base classes.
Inherit fields according to MRO rather than recursing over bases.
Add tests for above.

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -20,6 +20,44 @@ MarshalResult = namedtuple('MarshalResult', ['data', 'errors'])
 #: Return type of :meth:`Schema.load`
 UnmarshalResult = namedtuple('UnmarshalResult', ['data', 'errors'])
 
+
+def get_fields(attrs, field_class, pop=False):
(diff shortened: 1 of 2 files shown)


marshmallow/schema.py
```diff
@@ -20,6 +20,44 @@
...
+    :param type klass: Class whose fields to retrieve
+    :param type field_class: Base field class
+    """
+    return sum(
+        (
+            get_fields(
+                getattr(base, '_declared_fields', base.__dict__),
+                field_class
+            )
+            for base in reversed(klass.mro()[1:])
```
sloria, line 55: Could this be `klass.mro()[:0:-1]`?

sloria:
This is an excellent PR. Thanks for the contribution!

marshmallow/schema.py
jmcarp, line 55: Sure, whichever you think is more readable.

PR #48, opened 2014-10-02
Handle unicode output during validation

amikholap:
Hi!

This tiny fix allows to validate non-ASCII output. The problem is in interpolation of `str` with `unicode`.
It can also be `from __future__ import unicode_literals` but I'm not sure about side effects.

A snippet to reproduce the error:

```
from marshmallow.fields import String

def validator(output):
    return True

f = String(validate=validator)
f.deserialize(u'привет')
```


amikholap:
Catched and fixed one more unicode issue in _WrappingException.

sloria:
Thanks for the fix!

PR #50, opened 2014-10-09
Fix for SQLAlchemy

hcarvalhoalves:
Fixes serialization for SQLAlchemy's KeyedTuple, and other types that look like dicts but don't respond to obj["key"].


sloria:
This has already been fixed on the latest `dev`. You can install it with:

```
pip install -U git+https://github.com/sloria/marshmallow.git@dev
```

sloria:
Just an update: marshmallow>=1.0.0-a will correctly serialize `KeyedTuple` objects.

PR #66, opened 2014-11-23
Fix #65

### LINKED ISSUE #65: Inconsistent behavior between dump(obj, many=True) and Schema(many=True)
Doing

```
    schema = MySchema(many=True)
    print(schema.dump(mythings))
```

I get the correct behavior and everything works fine. However, doing

```
    schema = MySchema()
    print(schema.dump(mythings, many=True))
```

results in

`myfile.py:29: in get
    print(schema.dump(mythings, many=True))
env/lib/python3.4/site-packages/marshmallow/schema.py:435: in dump
    self._update_fields(obj)
env/lib/python3.4/site-packages/marshmallow/schema.py:583: in _update_fields
    ret = self.__filter_fields(field_names, obj)
env/lib/python3.4/site-packages/marshmallow/schema.py:630: in __filter_fields
    attribute_type = type(obj_dict[key])
E   TypeError: list indices must be integers, not str`

Something's up here.

svenstaro:
This is actually a pretty critical thing and probably the tests should have caught this! I recommend merging and making 1.0.2 immediately afterwards.

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -424,8 +424,9 @@ class BaseSchema(base.SchemaABC):
 
         .. versionadded:: 1.0.0
         """
-        many = many or self.many
-        if not many and utils.is_collection(obj) and not utils.is_keyed_tuple(obj):
+        if many:
+            self.many = many
+        if not self.many and utils.is_collection(obj) and not utils.is_keyed_tuple(obj):
             warnings.warn('Implicit collection handling is deprecated. Set '
                             'many=True to serialize a collection.',
                             category=DeprecationWarning)
@@ -436,7 +437,7 @@ class BaseSchema(base.SchemaABC):
         preresult = self._marshal(
             obj,
             self.fields,
-            many=many,
+            many=self.many,
             strict=self.strict,
             skip_missing=self.skip_missing,
             accessor=self.__accessor__,
@@ -475,7 +476,8 @@ class BaseSchema(base.SchemaABC):
 
         .. versionadded:: 1.0.0
         """
-        many = many or self.many
+        if many:
+            self.many = many
         # Bind self as the first argument of validators and preprocessors
         if self.__validators__:
             validators = [partial(func, self)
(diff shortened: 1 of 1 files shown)


svenstaro:
Explanation: In some places, you only check for self.many but you don't set self.many=many in dump. As such, it uses the default value of self.many in other places.

sloria:
This solution is not ideal. Calling `load` or `dump` should not have any side-effects (i.e. setting the schema's `many` attribute other than populating the marshaller's errors dictionary.

I am unable to reproduce the original issue. Can you please provide a code example that is breaking, so I can help diagnose this?

svenstaro:
This is my current suggestion. It does fail at some places though and in fact I'm not sure whether many should be optional or not (defaulting to False). Could you take a look and see whether you like this direction? I will then debug and we can merge this proper.

marshmallow/schema.py
```diff
@@ -424,7 +424,8 @@ def dump(self, obj, many=False, update_fields=True, **kwargs):
 
         .. versionadded:: 1.0.0
         """
-        many = many or self.many
+        if many:
```
sloria, line 427: Why was this changed? 

I believe this is a logical error. If `many=False` is passed in, the value for `self.many` should be used.

marshmallow/schema.py
```diff
@@ -557,11 +559,11 @@ def _update_data(self):
 
     ##### Private Helpers #####
 
-    def _update_fields(self, obj=None):
+    def _update_fields(self, many, obj=None):
```
sloria, line 562: I would rather have `many` as an optional argument after `obj` in order to maintain backwards compatibility. Same goes for `__filter_fields`.

sloria:
`self.many` will also have to passed to `_update_fields` here: https://github.com/svenstaro/marshmallow/blob/f91fce2613b16de0a8db7d4d9e5acbcc8da69bfe/marshmallow/fields.py#L556

sloria:
I've made the above fixes and merged into `dev`. Thanks for helping out with this!

PR #67, opened 2014-11-24
Update docs for fields.DateTime default format

ewang:
Current docs shows DateTime's default format as rfc; should be iso.


sloria:
Thanks!

PR #72, opened 2014-11-29
QuerySelect field

philtay:


philtay:
@sloria any word on these 3 new fields?
thanks :)

sloria:
Thank you for your contribution.

In order to reduce maintenance burden, I've put a hold on adding any new Fields to the fields module for the time being, so I won't merge this quite yet.

Just an idea: I'm wondering if `QuerySelect` and `QuerySelectList` might belong in a separate library that tightly integrates marshmallow with popular ORMs. Along with the generic fields in this PR, it could include fields which know how to query SQLAlchemy objects, for example.

This is worth discussion.

philtay:
Sure, but why a separate library? It's better to have everything you need to use marshmallow directly in marshmallow! We could do this like wtforms has done, it's a nice middle ground. Have a look here: https://github.com/wtforms/wtforms/tree/master/wtforms/ext
If you give me green light I will add a /ext directory (with generic ORM related stuff for now). Now that marshmallow supports deserialization, QuerySelect and QuerySelectList are two essential fields. They are going to get implemented again and again. It would be far better to have a single robust implementation in the library. I'm also planning to add a "unique" validator (e.g. to make sure that a user name is not already taken). With these 3 things the ORM support is basically complete and I will be a very happy user :)

sloria:
The reasons for a separate library for ORM integration are the same reasons WTForms [deprecated all packages](https://github.com/wtforms/wtforms/blob/master/wtforms/ext/sqlalchemy/__init__.py#L3-L8) in `wtforms.ext`. 

From the [README](https://github.com/wtforms/wtforms-sqlalchemy#rationale) of wtforms-sqlalchemy:

> The reasoning for splitting out this package is that WTForms 2.0 has deprecated all its wtforms.ext.<library> packages and they will not receive any further feature updates. The authors feel that packages for companion libraries work better with their own release schedule and the ability to diverge more from WTForms.

philtay:
Well, my opinion is the opposite, especially in light of the fact that here we are just talking about two generic ORM fields, not an entire set of tools to integrate with SQLAlchemy (like WTForms-Alchemy). Anyway, no reason to argue, your house, your rules!

sloria:
@philtay My initial thought was that the generic ORM fields could provide the basis (possibly serve as abstract classes) for a library which included modules for integrating with SQLAlchemy, Django ORM, etc.

philtay:
@sloria Indeed, they provide the basis to integrate with other ORMs, but they are also usable on their own (and sometimes it's better to use a more abastract integration like this one). For instance, if you pass the right function to the 'key' argument of the two fields you can easily integrate with SQLAlchemy (like WTForms does). By the way this is exactly the reason why you should include them in your library and why I have not added the specific function for SQLAlchemy or any other ORM. Other people then will have a common and robust starting point to create their own specialized functions and classes (in separate libraries). I thought about an /ext package because, in my opinion, the next logical step is integration with webargs (i.e. /ext/orm and /ext/webargs), but this is a another story. In case you want to merge these two fields, please consider that I have found a way to avoid the quadratic behavior of QuerySelectList, thus I will open a new pull request.

sloria:
I see the value in having these fields, as they meet a very common use case. I also think they are general enough to justify having them in marshmallow core. 

I will likely merge this for the next major feature release. Feel free to update this with your improvements to `QuerySelectList`.

I still believe that maintaining bundled `/ext` packages imposes unnecessary constraints on development. Auxiliary packages should have the freedom to iterate independently of marshmallow's release schedule. Consider, as an example, that webargs is pre-1.0 and does not have the same guarantees for stability that marshmallow does.

Reducing coupling with other frameworks and libraries is, after all, a major part of marshmallow's raison d'être.

A related sidenote: I started a repo [here](https://github.com/sloria/restkit) for an experimental library integrating marshmallow and webargs. The scope of the library is still to be decided, but it illustrates my point: the library will be experimental, with none of the promises for backwards-compatibility that marshmallow has.

philtay:
There is a new pull request with QuerySelect and QuerySelectList ready to be merged.
Regarding restkit, I'd like to collaborate. I think it will be a very interesting library.
Not sure if you already know it, but Cornice could be an inspiration.

http://cornice.readthedocs.org/en/latest/index.html

sloria:
Yes, I've seen Cornice before, and I think we might be able to draw ideas from it. Thanks for offering to collaborate on restkit. I will ping you when I get back to the project.

marshmallow/fields.py
```diff
@@ -1176,6 +1177,145 @@ def _serialize(self, value, attr, obj):
...
+                error = getattr(self, 'error', None) or 'Invalid values.'
+                raise UnmarshallingError(error)
+            else:
+                keys.pop(index)
+                items.append(results.pop(index))
+
+        return items
+
+
+class Numeric(Field):
```
sloria, line 1278: Could you rename this to `Decimal`?

marshmallow/fields.py
```diff
@@ -1176,6 +1177,145 @@ def _serialize(self, value, attr, obj):
...
+        if value is None:
+            return self.default
+
+        try:
+            num = Decimal(value)
+        except (TypeError, ValueError, InvalidOperation) as err:
+            raise exception_class(getattr(self, 'error', None) or err)
+        else:
+            if self.places is not None:
+                if self.rounding is None:
```
sloria, line 1307: Is this check necessary? I believe `quantize` handles `rounding=None` as expected.

marshmallow/fields.py
```diff
@@ -1176,6 +1177,145 @@ def _serialize(self, value, attr, obj):
...
+                    num = num.quantize(self.places)
+                else:
+                    num = num.quantize(self.places, rounding=self.rounding)
+            return num
+
+    def _serialize(self, value, attr, obj):
+        return self._validated(value, MarshallingError)
```
sloria, line 1314: This should serialize to `text_type`.

PR #82, opened 2014-12-06
Fix _postprocess call in extending doc

Ch00k:
The example in this doc http://marshmallow.readthedocs.org/en/latest/extending.html#example-adding-a-namespace-to-serialized-output worked for me only after I fixed the arguments in `_postprocess` call.


sloria:
Thanks!

PR #89, opened 2014-12-08
First CLASS validators

philtay:
I'd appreciate an early code review. Then I'll write the unit tests. Spotted and fixed a bug in the email validator. Added Function and NoneOf validators. With these two I consider the validate module complete and ready for battle!

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -6,7 +6,7 @@ from __future__ import absolute_import
 import datetime as dt
 import uuid
 import warnings
-from decimal import Decimal
+import decimal
 from functools import partial
 from operator import attrgetter
 
@@ -44,6 +44,7 @@ __all__ = [
     'TimeDelta',
     'Fixed',
     'Price',
+    'Decimal',
     'Url',
     'URL',
     'Email',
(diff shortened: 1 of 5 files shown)


philtay:
The unit tests for all of the class-based validators are in the commit above.
It includes the previously missing tests for URL and Email validators.
It also expands and makes more robust several other tests.
In my opinion, the class-based validators are in a mergeable state now.

sloria:
Thank you for your contribution.

This PR contains unrelated changes. Can you please separate this into 3 PRs?
- `Decimal` field
- Convertion to class-based validators
- Two added validators (it is fine if this one is dependent on the 2nd PR)

Once those are sent, I'd be glad to review them!

philtay:
@sloria if it's not a problem for you I'd prefer to keep them in a single pr. it's more easy for me and the same for you. these unrelated changes are all in different files. you should not have any problem in reviewing them separately. thank you

sloria:
I would much prefer to review them as separate PRs. Having the separate branches not only helps mergeability, but also adds useful historical information to the commit log.

Since the changes are in separate files, it should be straightforward to do: 

```
git checkout -b decimal origin/dev
# add decimal field; commit; push
git checkout -b class-validators origin/dev
# add validators; commit; push
git checkout -b new-validators validators
# add Function and NoneOf; commit; push
```

philtay:
Ok. I'll open the PRs. Just a few minutes.

PR #91, opened 2014-12-10
Decimal

philtay:
I was in the middle of creating the 3 branches. Here's Decimal.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -6,7 +6,7 @@ from __future__ import absolute_import
 import datetime as dt
 import uuid
 import warnings
-from decimal import Decimal
+import decimal
 from functools import partial
 from operator import attrgetter
 
@@ -44,6 +44,7 @@ __all__ = [
     'TimeDelta',
     'Fixed',
     'Price',
+    'Decimal',
     'Url',
     'URL',
     'Email',
@@ -995,7 +996,7 @@ class Fixed(Number):
                  *args, **kwargs):
         super(Fixed, self).__init__(default=default, attribute=attribute, error=error,
                             *args, **kwargs)
-        self.precision = Decimal('0.' + '0' * (decimals - 1) + '1')
+        self.precision = decimal.Decimal('0.' + '0' * (decimals - 1) + '1')
 
     def _serialize(self, value, attr, obj):
         return self._validated(value, MarshallingError)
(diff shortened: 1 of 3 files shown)


marshmallow/fields.py
```diff
@@ -1025,6 +1026,55 @@ class Price(Fixed):
     def __init__(self, decimals=2, default='0.00', **kwargs):
         super(Price, self).__init__(decimals=decimals, default=default, **kwargs)
 
+
+class Decimal(Field):
```
sloria, line 1030: `Decimal` should subclass `Number` with a `num_type` of `decimal.Decimal`. You could then override `_format_num` and handle the `decimal.InvalidOperation` as well as do the quantization. There would be no need to override `_serialize` and `deserialize`

sloria:
The tests could use refactoring, but it isn't necessary before this gets merged.

For example, I would change `TestFieldSerialization.test_decimal_field` into 3 tests and use a fixture like so:

``` python
    @pytest.fixture()
    def duser(self, user):
        user.m1 = 12
        user.m2 = '12.355'
        user.m3 = decimal.Decimal(1)
        user.m4 = None
        user.m5 = 'abc'
        user.m6 = [1, 2]
        return user

    def test_decimal_field_defaults(self, duser):
        field = fields.Decimal()
        assert isinstance(field.serialize('m1', duser), decimal.Decimal)
        assert field.serialize('m1', duser) == decimal.Decimal(12)
        assert isinstance(field.serialize('m2', duseer), decimal.Decimal)
        assert field.serialize('m2', duser) == decimal.Decimal('12.355')
        assert isinstance(field.serialize('m3', duser), decimal.Decimal)
        assert field.serialize('m3', duser) == decimal.Decimal(1)
        assert field.serialize('m4', duser) is None
        with pytest.raises(MarshallingError):
            field.serialize('m5', duser)
        with pytest.raises(MarshallingError):
            field.serialize('m6', duser)

    def test_decimal_field_with_places(self, duser):
        field = fields.Decimal(1)
        # ...
    def test_decimal_field_with_rounding(self, duser):
        field = fields.Decimal(1, decimal.ROUND_DOWN)
        # ...
```

If you'd like to do this, let me know; otherwise I can do it myself when I merge.

sloria:
Pass finished. :+1:  Thank you for being thorough with testing and documentation. Let me know when this is ready for another pass.

philtay:
Ok. I'll do these things, unit tests included. It's better if you use your time reviewing the other 2 PRs than doing these tweaks.

philtay:
Two things here:
- If I just override _format_num then I can't handle the decimal.InvalidOperation exception. _format_num doesn't know what kind of exception it has to raise (if MarshallingError or UnmarshallingError). We have to add decimal.InvalidOperation to the except in Number._validated
- In Number._validated the default value (in this case decimal.Decimal()) is returned as Decimal even if as_string is True. Is this behavior correct?

sloria:
> We have to add decimal.InvalidOperation to the except in Number._validated

That is fine.

> In Number._validated the default value (in this case decimal.Decimal()) is returned as Decimal

If `as_string=True`, the string representation of `Decimal()` should be returned.

philtay:
> If as_string=True, the string representation of Decimal() should be returned.

You (we) should change Number._validated to get this behavior. It's not something I can do in Decimal without overriding it.

sloria:
The change made in https://github.com/sloria/marshmallow/commit/e9e5bbc4eb8e03bc9abcde2f4b541fc1afcb4f09 should accommodate the `Decimal` field.

philtay:
There is still a problem in Number._validated. if value is None self.default is returned without being converted to a string. In Decimal the default is decimal.Decimal() and you get it as a Decimal even if you have specified as_string=True. Could you please change it? Thanks

sloria:
Fixed the `as_string` behavior. Good catch.

philtay:
Now it doesn't deserialize to a string anymore. Is this an intentional change?

sloria:
Yes, it should not deserialize to a string, but it should deserialize "from" a string to a number.

philtay:
I agree. As soon as you merge the class-based validators, I'll open a new PR with the Decimal field. Please also let me know about Function and I'll open a PR for it and NoneOf as well. Thanks

sloria:
Was just about to look at the class validators. Why send another PR for Decimal field? You can just merge or rebase `dev` with this branch.

sloria:
Replaced by #103

Issue #85, opened 2014-12-07
Provide generic field validators


sloria:
To continue the discussion from https://github.com/sloria/marshmallow/pull/84#issuecomment-65921413:

Before release 1.1.0, the `validate` module was used mainly for internal use, i.e. as helpers for the `URL` and `Email` fields. Users were not expected to use the functions directly.

In release 1.1.0, @philtay added a few user-facing validators: `ranging`, `equal`, `regexp`, and `predicate`. 

Currently, using these validators is bit awkward because you either have to define an outer function to pass them to a field.

``` python
password = fields.Str(validate=[lambda v: validate.length(v, min=6)])
```

This can be improved by writing the validators as classes or function factories:

``` python
class Length(object):
    def __init__(self, min=None, max=None, error=None):
        #...
    def __call__(self):
        #...
```

However, before we refactor the `validate` module with potentially breaking changes, we need to decide whether marshmallow should include these user validators in the first place.
### Should marshmallow provide a set of generic user validators?

I'm not completely opposed to the idea, but my concern is that these validators have been implemented many times in other libraries. I wonder if marshmallow should integrate with existing solutions rather than reimplement them.

@philtay raised his concerns about integrating with other libraries in https://github.com/sloria/marshmallow/pull/84#issuecomment-65934045:

> ...every library raises its own exception and, in general, does things in a slightly different way. Moral of the story: we have to provide at least a minimum set of generic essential validators.

Opening this up for suggestions, opinions, and discussion.

EDIT: Fix code example.

philtay:
I'd like to add just one thing. It's very easy to maintain backward compatibility if we switch to class based validators.

``` python
class Url(ValidatorABC):
    """New url validator."""
    def __init__(self, relative=False, error=None):
        ...

    def __call__(self, value):
        ...

def url(value, relative=False, error=None):
    """Old function still works. Keep this for 2 versions then deprecate."""
    return Url(relative, error)(value)
```

sloria:
@philtay Yes, you are correct re: backwards-compatibility.

I've drafted a quick proof-of-concept to show how 3rd-party validators (WTForms and colander as examples) could be integrated with marshmallow: https://gist.github.com/sloria/edffc11fac7ee680e544.

I could see something like this going into our marshmallow utility library (restkit).

philtay:
[...]

PR #92, opened 2014-12-10
Class validators

philtay:


marshmallow/validate.py
```diff
@@ -213,13 +235,47 @@ def predicate(value, method, error=None, **kwargs):
...
 
-    return value
+        return value
+
+
+#
+# Backward compatibility
+#
+
+def url(value, relative=False, error=None):
```
sloria, line 275: As discussed in #85, these functions should be marked for deprecation You could use a decorator, like so:

``` python
import functools

def _deprecated(func):
    @functools.wraps(func)
    def new_func(*args, **kwargs):
        warnings.warn('{0} is deprecated. Use the class-based '
                      'equivalent instead.'.format(func.__name__),
                      category=DeprecationWarning)
        return func(*args, **kwargs)
    return new_func
```

EDIT: language

marshmallow/validate.py
```diff
@@ -169,18 +185,22 @@ def equal(value, comparable, error=None):
         Error message to raise in case of a validation error.
         Can be interpolated using `{cmp}`.
     """
-    if value is None:
-        return None
+    def __init__(self, comparable, error=None):
+        self.comparable = comparable
+        self.error = error or 'Must be equal to "{cmp}".'
```
sloria: Could you just do the `format` call here?

``` python
self.error = error or 'Must be equal to {0}.'.format(comparable)
```

sloria: The double quotes around `{cmp}` may be misleading for some types, e.g. numbers. I think they can be removed.

tests/test_validate.py
```diff
@@ -4,108 +4,245 @@
...
+        'http://xn--mgbh0fb.xn--kgbechtv/',
+        'http://example.com/blue/red%3Fand+green',
+        'http://www.example.com/?array%5Bkey%5D=value',
+        'http://xn--rsum-bpad.example.org/',
+        'http://123.45.67.8/',
+        'http://2001:db8::ff00:42:8329',
+        'http://www.example.com:8000/foo',
+    )
+
+    for valid_url in valid_urls:
```
sloria: Prefer pytest's parametrization feature to `for` loops, so that all the test cases can run, even if one of them fails.

sloria: Here's an example: https://github.com/philtay/marshmallow/blob/bc8a11fa3b7bad39109a4e75b3ff74f36d08afed/tests/test_serialization.py#L88-92

sloria:
Pass finished. :+1: Just a few minor changes.

philtay:
Done. The build failure seems unrelated to this PR. It should be mergeable now.

marshmallow/validate.py
```diff
@@ -142,45 +123,84 @@ def ranging(value, min=None, max=None, error=None):
...
     :param str error:
         Error message to raise in case of a validation error.
-        Can be interpolated using `{cmp}`.
+        Can be interpolated using `{0}`.
     """
-    if value is None:
-        return None
+    def __init__(self, comparable, error=None):
+        self.comparable = comparable
+        self.error = error or 'Must be equal to {0}.'
```
sloria: Ohh, I wasn't asking you to change `cmp` to `0`, I was asking why you couldn't just to the interpolation in `__init__` rather than doing it in`__call__`, but I just read the docstring more carefully and it makes sense now. Sorry for the misunderstanding.

`cmp` was better than `0`, no doubt. But `cmp` is typically used for comparator functions in the Python builtins. So how about `other` instead?

philtay:
5 min and it will be done

philtay:
Done.

sloria:
Now that the `validate.email` and `validate.url` are deprecated, `fields.Email` and `fields.URL` should use the new class-based validators rather than the partialed functions. Can you change those as well?

After merging or rebasing with dev, the Travis build should be fixed.

philtay:
I've messed with my repo. Need a fresh one. I'll do this immediately after the merge. Ok?

sloria:
Just had a thought: it is strange that the current validators treat `None` as a special value. This seems incorrect. `None` is not a valid email, or a URL, number, etc, and should not validate for those respective validation callables.

I believe the motivation for allowing `None` for the original `url` and `email` functions goes back to marshmallow's early iterations as a _serialization-only_ library. It was a shortcut which allowed objects with email and URL attributes set to `None` to serialize successfully.

Now that marshmallow does validation/serialization, that API doesn't make much sense. It leads to grossly incorrect behavior

``` python
class UserSchema(Schema):
    email = fields.Email(required=True)

s = UserSchema()
s.load({'email': None})  # no errors! wuh?
```

I propose to remove the special-casing of `None` from the new class-based validators, and leave the [incorrect] behavior in the deprecated function validators.

Thoughts?

philtay:
I agree. It doesn't make sense. This after merge as well.

sloria:
I will not not merge until this is completely ready. What exactly is wrong with your repo? Is there any way I can help resolve it? If you've made commits on this branch that you haven't pushed yet, you can just checkout the latest commit that's been pushed here (48e6f47, in this case), remove the invalid branch, and continue where you left off.

```
# rename defunct branch
git branch -m class-validators-old
git checkout -b class-validators 48e6f47
git branch -D class-validators-old
```

sloria:
Oh, I see: your fork's `dev` got ahead of upstream `dev`. No prob:

``` bash
# rename defunct dev
git checkout dev
git branch -m dev-old
# create new dev from upstream
git fetch
git checkout -b dev upstream/dev
# get class-validators up to date
git checkout class-validators
git merge dev
```

philtay:
Yes, I know, but I had another problem as well. Anyway just deleted it, no problem. Few minutes and you'll have a new PR.

sloria:
OK, thanks

sloria:
replaced by #94

Issue #33, opened 2014-08-12
Deprecate legacy serialization API


sloria:
As of 1.0.0 the correct way to serialize objects is to use the `Serializer.dump` method.

Usage of `Serializer(some_obj).data` will be deprecated, as will the related `Serializer.errors` and `Serializer.is_valid` members (`dump` returns both the serialized data and a dictionary of errors, so these validation methods are redundant).

For the 1.0.0 release, deprecation warnings should be raised.

sloria:
Resolved by 62e5cebd

PR #93, opened 2014-12-10
Function and NoneOf

philtay:


marshmallow/validate.py
```diff
@@ -213,13 +235,105 @@ def predicate(value, method, error=None, **kwargs):
...
+
+        method = getattr(value, self.method)
+
+        if not method(**self.kwargs):
+            raise ValidationError(self.error)
+
+        return value
+
+
+class Function(object):
```
sloria, line 258: Out of genuine curiousity: can you give an example use case of this validator? I find it conceptually difficult to explain; I think an example might clear things up.

sloria, line 258: Is the common use case for this to be able to pass a predicate `lambda`? If so, this is redundant with the existing field API:

``` python
foo = fields.Field(validate=lambda x: len(x) > 8, error='Must be > 8')
```

Granted, I have been considering deprecating the `error` parameter (see #33). However, it seems that this validator class would add a third way to apply a user validator with a custom message (the first being a named function that may raise a `ValidationError`).

marshmallow/validate.py
```diff
@@ -213,13 +235,105 @@ def predicate(value, method, error=None, **kwargs):
...
+    """
+    def __init__(self, iterable, error=None):
+        self.iterable = iterable
+        self.error = error or 'Invalid input.'
+
+    def __call__(self, value):
+        if value is None:
+            return None
+
+        for item in self.iterable:
```
sloria, line 309: Why not 

``` python
if item in self.iterable:
```

marshmallow/validate.py
```diff
@@ -213,13 +235,105 @@ def predicate(value, method, error=None, **kwargs):
...
+        A sequence of invalid values.
+    :param str error:
+        Error message to raise in case of a validation error.
+    """
+    def __init__(self, iterable, error=None):
+        self.iterable = iterable
+        self.error = error or 'Invalid input.'
+
+    def __call__(self, value):
+        if value is None:
```
sloria, line 306: What if `None` is in `self.iterable`?

sloria:
Pass finished :ocean:

philtay:
> Out of genuine curiousity: can you give an example use case of this validator? I find it conceptually difficult to explain; I think an example might clear things up.

The use case is when you already have a callable that performs the validation and returns an error message. Think about integration with other libraries. There are tons of functions like that out there.
For instance, you could write a function that catches the validation exception of another library (WTForms, Colander, etc) and returns the error message. Pass this function to the Function validator and you have your Converter validator!

sloria:
But why use the `Function` class when you directly pass a validator function to a field?

philtay:
Because the validator function will raise an exception or return an error message as a string.

Library X, or my own project, has these 2 functions:

``` python
def valid1(foo):
    raise MyOwnExc('bar is better than foo!')

def valid2(foo):
    return 'uhm, this foo is not very good'
```

You want marshmallow.exceptions.ValidationError('bar is better than foo!') and marshmallow.exceptions.ValidationError('uhm, this foo is not very good').

sloria:
Passing `valid1` to your `Function` class  and calling it would result in `MyOwnExc` being raised: the exception is not handled by `Function.__call__`.

So why not:

``` python
def field_validator(val):
    try:
        valid1(val)
    except MyOwnExc as err:
        raise ValidationError(err)
```

Re: `valid2`: When would a user write a validation function that returned a string error message rather than raising an exception?

philtay:
> Passing valid1 to your Function class and calling it would result in MyOwnExc being raised

Yes, you need a wrapper to catch the exception and return the error as a string.
Sure, but then you can just write a small little function like this one and raise the ValidationError yourself.
Yes, but this exactly the point. I don't want to raise the ValidationError myself. This the reason I use a library.
But look, only 4 lines of code to do it yourself.
Oh, really. Only 4? So why not put them in the library, it won't be a huge maintaince effort.
We can go on forever, you know :)

> When would a user write a validation function that returned a string error message rather than raising an exception?

Depends. Now you are thinking only about functions that strictly perform a validation. It's not always the case.
Uhm, I'm not convinced. Gimme an example on a real function that returns a string.
Dunno right now. I'll let you know however! In the meantime note that the fact you can't find a specific narrow use case is exactly the point. It's a Swiss Army knife.
Uhm, no. I need a specific use case to put it in the library.
It's the most generic validator of the library you'll never find a very specific use case. It's all and nothing at the same time.
Uhm, but I can do it in another way.
Sure, you can almost always find another way.
You see, another loop :)

philtay:
> Why not "if item in self.iterable:"

Because you risk a TypeError. For instance, my iterable is a string and the value to validate happens to be an integer:

``` python
>>> 1 in 'abc'
TypeError: 'in <string>' requires string as left operand, not int
```

philtay:
> What if None is in self.iterable?

We can decide to raise ValidationError if None is in self.iterable and the value to validate is None. Or not. For consistency it should behave as it is now (always return None is value is None). But of course we can change this behavior. Let me know what you prefer.

sloria:
I'm missing something.

Why is this:

``` python
def validate(val):
    try: 
        lib_valid(val)
    except LibError as err:
        return str(err)

myfield = fields.Field(validate=validate.Function(validate))
```

preferable to this:

``` python
def validate(val):
    try:
        lib_valid(val)
    except LibError as err:
        raise ValidationError(str(err))

myfield = fields.Field(validate=validate)
```

> Because you risk a TypeError. For instance, my iterable is a string and the value to validate happens to be an integer:

Thanks for the clarification. In this case, it would be clearer to use the `in` operator and handle the TypeError

``` python
err = ValidationError(self.error)
try:
    if value not in self.iterable:
        raise err
except TypeError:
    raise err
```

> We can decide to raise ValidationError if None is in self.iterable and the value to validate is None.

Yes, that is the preferred and expected behavior.

philtay:
> I'm missing something.

Ok, you know what, we can do in this way:

``` python
def __call__(self, value):
    if value is None:
        return None

    try:
        result = self.function(value, **self.kwargs)
    except Exception as err:
        raise ValidationError(self.error or str(err))

    if not result:
        raise ValidationError(self.error)

    if isinstance(result, basestring):
        raise ValidationError(result)

    return value
```

Now it covers this use case as well and it can work as a Converter out of the box. Let me know if you like it.

P.S. The other class based validators are ready to be merged.

philtay:
Could you please let me know about the solution above as well? Thanks.

sloria:
You could certainly do that, but I still find it to be a confusing interface: return an error if the passed function returns a falsy value OR raises an exception OR returns a string... I also don't like the idea of implicitly swallowing all exceptions nor condoning validator functions that return strings. 

I'm going to hold off on merging this, as I don't see `Function`  as an essential validator with a stable interface (i.e. one for which we can promise backwards-compatiblity in future releases).

philtay:
Ok, do you want NoneOf?

sloria:
`NoneOf` is fine. Can you make the modifications from my previous comment and send a new PR?

philtay:
Ok, as soon as the class-based validators are merged I will send a new PR for it.

PR #94, opened 2014-12-11
Class-based validators

philtay:


sloria:
Deserializing `None` with the `URL` field currently throws  a `TypeError`:

```
     45         regex = self.RELATIVE_URL_REGEX if self.relative else self.URL_REGEX
     46 
---> 47         if not regex.search(value):
     48             message = '"{0}" is not a valid URL.'.format(value)
     49             if regex.search('http://' + value):

TypeError: expected string or buffer
```

unlike the `Email` field, which handles`None` correctly. Can you please fix that and add a test case for it?

marshmallow/validate.py
```diff
@@ -213,13 +220,63 @@ def predicate(value, method, error=None, **kwargs):
...
 
-    if not method(**kwargs):
-        raise ValidationError(error or 'Invalid input.')
 
-    return value
+#
+# Backward compatibility
+#
+
+import functools
```
sloria: All imports at the top of the file, please.

philtay:
> Deserializing None with the URL field currently throws a TypeError.

Do you want it to deserialize to None or to raise a ValidationError? I'd lean toward the second option both for None and the empty string. Ok?

sloria:
Correct, `URL` should raise a `ValidationError` for any falsy values.

philtay:
Done.

linbrandon4:
I went through this PR and it’s a pretty big refactor (+422/-256). Moving to class-based validators looks clean overall, and I like that there are tests included.

One thing I’m glad got addressed: URL deserialization on None was throwing a TypeError. Making URL raise a ValidationError for falsy values (like None / empty string) matches Email and feels consistent. Also good call on adding a test for that.

Only things I’d double-check before calling it done:
- make sure falsy handling is consistent across validators (None vs "")
- make sure the new tests cover a couple basic URL cases (valid + invalid)

Issue #99, opened 2014-12-13
How to properly run the tests?


svenstaro:
Running tox locally for my python version yields 18 errors as compared to running tox on travis. What is different running tox locally as compared to travis that I would get errors here? How do I run the tests locally?

sloria:
@svenstaro Do you have each of the interpreters installed locally? Also, what errors are you seeing?

svenstaro:
Running

$ tox --skip-missing-interpreters &> /tmp/log

gives me this: https://gist.github.com/ca7bb1b6a8c339052e27

svenstaro:
It seems like it's pulling in older versions of marshmallow which are obviously not compatible. I wonder why it does that since obviously we want to be testing the local version of marshmallow.

svenstaro:
Ah, my problem is that marshmallow does not expect the virtualenv to be in the same dir as the git repo. This kind of sucks for me, because I always put my env there, as in marshmallow/env. Would you mind if I fixed marshmallow to disregard the popular "env" name for the virtualenv inside the git root?

svenstaro:
I made a pr to fix the specific issues that I experienced here.

PR #101, opened 2014-12-13
Ignore popular virtualenv names to fix issues like #99

svenstaro:
The reasoning here is that it's fairly common practice to create virtualenvs that are named either 'env' or 'venv' in the project root.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -9,12 +9,12 @@ build_dir = os.path.join(docs_dir, '_build')
 @task
 def test():
     """Run the tests."""
-    flake()
     run('python setup.py test', echo=True, pty=True)
 
 @task
 def flake():
-    run('flake8 .', echo=True)
+    """Run flake8 on codebase."""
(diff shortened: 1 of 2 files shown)


tasks.py
```diff
@@ -9,12 +9,12 @@
 @task
 def test():
     """Run the tests."""
-    flake()
```
sloria, line 12: Why was this removed?

tasks.py
```diff
@@ -9,12 +9,12 @@
...
     run('python setup.py test', echo=True, pty=True)
 
 @task
 def flake():
-    run('flake8 .', echo=True)
+    """Run flake8 on codebase."""
+    run('flake8 --exclude=.svn,CVS,.bzr,.hg,.git,__pycache__,env,venv .', echo=True)
```
sloria, line 17: Would rather have the configuration in the `setup.cfg` file, so a user can just run flake8 without having to remember command-line args.

Issue #96, opened 2014-12-13
Error when deserializing empty Date()


svenstaro:
So when deserializing an empty (but non-required) DateTime, everything works fine. When deserializing an empty (but non-required) Date(), I get:

`File "/home/svenstaro/prj/blah/blah.py", line 181, in patch
    result, errors = schema.load(request.json)
  File "/home/svenstaro/prj/blah/env/lib/python3.4/site-packages/marshmallow/schema.py", line 481, in load
    result, errors = self._do_load(data, many, postprocess=True)
  File "/home/svenstaro/prj/blah/env/lib/python3.4/site-packages/marshmallow/schema.py", line 589, in _do_load
    dict_class=self.dict_class
  File "/home/svenstaro/prj/blah/env/lib/python3.4/site-packages/marshmallow/fields.py", line 282, in deserialize
    strict=strict
  File "/home/svenstaro/prj/blah/env/lib/python3.4/site-packages/marshmallow/fields.py", line 107, in _call_and_store
    value = getter_func(data)
  File "/home/svenstaro/prj/blah/env/lib/python3.4/site-packages/marshmallow/fields.py", line 453, in deserialize
    return self._call_and_reraise(do_deserialization, UnmarshallingError)
  File "/home/svenstaro/prj/blah/env/lib/python3.4/site-packages/marshmallow/fields.py", line 403, in _call_and_reraise
    return func()
  File "/home/svenstaro/prj/blah/env/lib/python3.4/site-packages/marshmallow/fields.py", line 450, in do_deserialization
    output = self._deserialize(value)
  File "/home/svenstaro/prj/blah/env/lib/python3.4/site-packages/marshmallow/fields.py", line 960, in _deserialize
    return utils.from_iso_date(value)
  File "/home/svenstaro/prj/blah/env/lib/python3.4/site-packages/marshmallow/utils.py", line 262, in from_iso_date
    return parser.parse(datestring).date()
  File "/home/svenstaro/prj/blah/env/lib/python3.4/site-packages/dateutil/parser.py", line 743, in parse
    return DEFAULTPARSER.parse(timestr, **kwargs)
  File "/home/svenstaro/prj/blah/env/lib/python3.4/site-packages/dateutil/parser.py", line 300, in parse
    res = self._parse(timestr, **kwargs)
  File "/home/svenstaro/prj/blah/env/lib/python3.4/site-packages/dateutil/parser.py", line 358, in _parse
    l = _timelex.split(timestr)
[...]

PR #102, opened 2014-12-13
Add failing test cases for empty date stuff

svenstaro:
This shows failures in two cases for #96.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -933,7 +933,7 @@ class Time(Field):
         """Deserialize an ISO8601-formatted time to a :class:`datetime.time` object."""
         try:
             return utils.from_iso_time(value)
-        except (TypeError, ValueError):
+        except (AttributeError, TypeError, ValueError):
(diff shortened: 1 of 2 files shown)


svenstaro:
I also attached a fix that you may or may not consider a proper fix. :P

sloria:
Thanks for this. I think the fix is fine. In reviewing this PR I found another related bug: the empty string passes validation for `DateTime`, `Date`, and `Time` fields and deserializes to the current datetime/date/time. This behavior is incorrect.

 So I generalized the solution further by checking for falsy values in the time-related fields, e.g. https://github.com/sloria/marshmallow/blob/2c3f701b319de8ccbf4b7d92543d86a512cc67d7/marshmallow/fields.py#L884-885 .

I also DRY'd up your tests a bit: https://github.com/sloria/marshmallow/blob/2c3f701b319de8ccbf4b7d92543d86a512cc67d7/tests/test_deserialization.py#L90-100

PR #103, opened 2014-12-14
Decimal field

philtay:
Please give me a hand with this one. After merge you should:
- Deprecate Fixed, Price and Arbitrary fields
- Split the unit tests as you prefer

Thanks


philtay:
Do you need something else from me to merge this patch?

sloria:
Will review this over the weekend when I'll have some time to do the follow up actions. Thanks for the contribution

PR #104, opened 2014-12-14
NoneOf validator

philtay:


philtay:
Please also let me know if you want an IP address validator. I have it ready and it supports both IPv4 and IPv6. I think it would nicely complement URL and Email validators.

sloria:
Thanks for this. Let's hold off on  the IP validators for the next release. They aren't essential, generic validators. Perhaps for a future release.

Issue #106, opened 2014-12-17
UUID field raises ValueError on malformed UUID when deserializing


vesauimonen:
The following code results in a `ValueError: badly formed hexadecimal UUID string` exception.

``` python
from marshmallow import fields, Schema

class ExampleSchema(Schema):
    id = fields.UUID()

schema = ExampleSchema()
schema.load({'id': 'malformed'})
```

I'm working on a bug fix. Do you want it to be sent to `dev` branch? I'm not seeing a maintenance branch as described in https://github.com/sloria/marshmallow/blob/dev/CONTRIBUTING.rst#git-branch-structure.

sloria:
Thank you for reporting and fixing this. Yes, you can send the PR to `dev`; I haven't yet finalized the branch structured described in the contributing docs.

sloria:
Fixed by #107

PR #107, opened 2014-12-17
Fix uuid error handling

vesauimonen:
See issue #106.


sloria:
Thanks for the patch!

PR #108, opened 2014-12-19
loads should use the schema's many value by default

davidism:
When a schema is initialized with `many=True`, `dump`, `load`, and `dumps` will use that value by default, but `loads` defaults to `False`.  The user must explicitly pass `True` in order to use the same schema instance for dumping and loading.  All the methods should present the same default behavior to avoid confusion.

No tests failed by making this change, but I added a simple test anyway.  There are no asserts because simply trying to call `loads` on data dumped by the same schema instance will cause an exception.


sloria:
Good catch!

PR #109, opened 2014-12-21
Import unicode_literals from the __future__

philtay:
Make sure all string literals are unicode on Python 2 too.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """Field classes for various types of data.
 """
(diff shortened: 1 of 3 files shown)


sloria:
+1. If we decide on using `unicode_literals`, it should be done in every module. Can you also add it in `schema.py` and `utils.py`?

philtay:
Done.
P.S. A new release (1.2.0) would be very much appreciated. Thanks.

sloria:
Rebased and merged in f4c7cc2f6896d5f7b3e9f65349589018617b6c79

Yes, I was planning on release 1.2.0 today.

PR #122, opened 2015-01-05
Implementing customization of required message

0xDCA:


.directory
```diff
@@ -0,0 +1,6 @@
+[Dolphin]
```
sloria: I assume this file was added inadvertently. Please remove it.

marshmallow/fields.py
```diff
@@ -456,7 +462,9 @@ def deserialize(self, value):
         def do_deserialization():
             if value is missing:
                 if hasattr(self, 'required') and self.required:
-                    raise ValidationError('Missing data for required field.')
+                    message = self.required if isinstance(self.required, basestring) else \
```
sloria: Please use parens rather than `\` for breaking long lines:

``` python
                    message = (self.required if isinstance(self.required, basestring) else
                        'Missing data for required field.')
```

marshmallow/fields.py
```diff
@@ -360,6 +362,10 @@ def __init__(self, default=None, attribute=None, error=None,
         else:
             raise ValueError("The 'validate' parameter must be a callable "
                             'or a collection of callables.')
+
+        if not (isinstance(required, bool) or isinstance(required, basestring)):
```
sloria: I'm not sure that this check is necessary. Given the implementation of `ValidationError`, the user could pass a `str`, `dict`, or `list` and the result would be valid.

sloria:
Thanks for the contribution. I appreciate that you updated the docs. Just a few minor changes to make and this should be good to merge. :octopus:

0xDCA:
Hi. Ok, I'll fix those minor details.

> I'm not sure that this check is necessary. Given the implementation of ValidationError, the user could pass a str, dict, or list and the result would be valid.

Mmm, how would passing dict or list make any sense? The required parameter originally could be True or False, I only extended it to accept a string too. I don't see any case where accepting a list as "implicitly" bool is useful, and in those cases, one can always use bool()

sloria:
@0xDCA One use case I've seen for passing a `dict` to a `ValidationError` is to store error codes and other metadata about an error. For example, given a required field called `name`, a user might want to receive an error object like:

``` js
errors = {'name': {'error_code': 'missing-required', 'message': 'Missing data for required field.', 'code': 400}}
```

This can be handled in your code by checking for a `bool` rather than `basestring` when computing the error message.

``` python
default_msg = 'Missing data for required field.'
msg = default_msg if isinstance(self.required, bool) else self.required
raise ValidationError(msg)
```

0xDCA:
Oh, that's a good example. I updated the code and the docs accordingly, please take a look at it

sloria:
Thanks for making those changes

Issue #125, opened 2015-01-08
Allow field.attribute for both serializing and deserializing


hakjoon:
I'm happy to take a stab at coding this but wanted to get a sense if it would be a useful feature for others, or if there is way to handle this in the current tooling that I am missing.  

I have to consume a number of .NET created apis which provide their keys in PascalCase.  As a result I need to specify how attributes would map a lot.  It would be great it Marshmallow allowed you specify something like field.attribute to be use for serialization and one to be used for deserialization.  For example:

```
class User(object):
    def __init__(self, name, email, age=None):
        self.name = name
        self.email = email

class UserSchema(Schema):
    name = fields.String(deserialize_from="Name")
    email_addr = fields.String(attribute="email", deserialize_from="EmailAddress")

user = User('Keith', email='keith@stones.com')
ser = UserSchema()
result, errors = ser.dump(user)
pprint(result)
# {'name': 'Keith',
#  'email_addr': 'keith@stones.com'}

payload = {
                   'Name': 'Keith',
                   'EmailAddress': ''keith@stones.com',
}

ser = UserSchema()
result, errors = ser.load(payload)
pprint(result.data)
# {'name': 'Keith',
# 'email': ''keith@stones.com'}
```

Maybe make `deserialize_from` be an additional search key that way you could still deserialize a previously serialized instance.

Any interest in this idea?

sloria:
I think this feature makes sense. I would gladly review and merge a PR for this.

Perhaps the new parameter could be called `load_from`, for brevity and to make clear that applies to `Schema#load`.

hakjoon:
I welcome any suggestions on the code.  Also documentation is not my strength so please let me know if I can make any of it clearer.

sloria:
PR looks great! Thanks @hakjoon for your contribution.

ShayanArmanPercolate:
What do you think of a dump_to parameter? In order to make the fields camelcase after dumping?

hakjoon:
I think it makes sense. I'd be happy to take a stab at it if there is interest.

PR #126, opened 2015-01-13
load_from option for deserializing

hakjoon:
This adds a load_from option to Field which will be used as an additional key lookup when deserializing a payload as discussed in #125


bencharb:
Looking forward to v2.0 so for this feature.

PR #131, opened 2015-01-21
Add load_only and dump_only options

philtay:
Docstrings and unit tests are still missing. Let me know if these changes are ok and I'll add them.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -279,6 +279,8 @@ class Unmarshaller(object):
             for attr_name, field_obj in iteritems(fields_dict):
                 if attr_name not in fields_dict:
                     continue
+                if field_obj.dump_only:
+                    continue
                 key = fields_dict[attr_name].attribute or attr_name
                 raw_value = data.get(attr_name, missing)
                 if raw_value is missing and field_obj.load_from:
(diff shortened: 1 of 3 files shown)


marshmallow/fields.py
```diff
@@ -290,6 +294,8 @@ def deserialize(self, data, fields_dict, many=False, validators=None,
                     exception_class=UnmarshallingError,
                     strict=strict
                 )
+                if field_obj.dump_only:
```
sloria: This should go before the call to `_call_and_store`. If the field dump-only, no need to do the work of deserializing the value and validating it.

sloria:
Thanks! Just a minor change to make. Once the tests and docs are added, this is good to merge.

philtay:
Done.

marshmallow/fields.py
```diff
@@ -191,8 +191,12 @@ def serialize(self, obj, fields_dict, many=False, strict=False, skip_missing=Fal
                 exception_class=MarshallingError,
                 strict=strict
             )
-            if (value is missing) or (skip_missing and
-                                      value in field_obj.SKIPPABLE_VALUES):
+            skip_conds = (
+                field_obj.load_only,
```
sloria, line 195: Actually, we can check `field_obj.load_only` before running `_call_and_store` as well.

sloria:
Thanks for adding tests and docs. Just one last change and this is good to merge.

philtay:
Uhm, no, we can't. This test is going to fail:

https://github.com/marshmallow-code/marshmallow/blob/dev/tests/test_schema.py#L362

This is the reason why I initially put them both after _call_and_store.

sloria:
OK. I see. Thanks

Issue #115, opened 2014-12-27
Default for deserialization parameter ("missing")?


sloria:
There was some discussion in #45 over the addition of a field parameter (called `missing`) that would specify a default value for deserialization. We decided not to add the parameter because it would add complexity to the API when there are already hooks--including `make_object` and `preprocessor`--that could meet common use cases.

Some comments from that thread, snipped for brevity:

Me: 

> I think there may be a use case for a second default parameter for missing input values, similar to the missing parameter in colander: http://colander.readthedocs.org/en/latest/null.html.

@zerodivision:

>  ...It seems to me that additional missing parameter is over-engineering. But setting default value for single field in preprocessor is bad too. In the case when there is a lot of such fields it requires repeated code for each field. So I guess that missing parameter is better for these cases.

Me:

> I'm going to hold off on adding the missing parameter for now, to avoid the possibility of adding unneeded code...This isn't to say that the parameter won't be added in future releases. A compelling use case may arise that cannot be cleanly addressed using the existing hooks. Until we identify that use case, it will not be included.

Thanks to @rastikerdar, the issue is revived. I think there may be added value in the `missing` parameter. Opening this up for discussion.

rastikerdar:
Sometimes we need a callable missing:

``` python
afield = fields.DateTime(missing=lambda: datetime.datetime.utcnow())
```

And also needs to be aware of `only` and `exclude` list defined at schema initialization:

``` python
# for ...
    if name in self.exclude or (self.only and name not in self.only):
        continue # skip processing this missing field
```

sloria:
I agree. Thanks for the suggestions, @rastikerdar .

sloria:
Closed by #134

PR #134, opened 2015-01-27
Add missing parameter

philtay:
Implements #115

If it's ok tests and docs will be added.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -285,7 +285,7 @@ class Unmarshaller(object):
                 raw_value = data.get(attr_name, missing)
                 if raw_value is missing and field_obj.load_from:
                     raw_value = data.get(field_obj.load_from, missing)
-                if raw_value is missing and field_obj.missing is not None:
+                if raw_value is missing and field_obj.missing is not null:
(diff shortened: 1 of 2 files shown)


marshmallow/fields.py
```diff
@@ -371,7 +374,7 @@ class Field(FieldABC):
 
     def __init__(self, default=None, attribute=None, load_from=None, error=None,
                  validate=None, required=False, allow_none=False, load_only=False,
-                 dump_only=False, **metadata):
+                 dump_only=False, missing=None, **metadata):
```
sloria: `None` should be a valid value for `missing`. Perhaps you can make the default `null` here to allow `None` to be used as a value for missing data.

philtay:
- Used null instead of none
- Added docs
- Added tests

P.S. Have you an idea on when you'll have the time to release 2.0?

sloria:
Thanks @philtay.

I don't have a scheduled release date for 2.0, but I'm hoping to have 2.0-a released within the next 2-3 weeks.

philtay:
@sloria don't forget to update the changelog with this new parameter

PR #137, opened 2015-02-02
Fix "many" serialization with custom accessor, skip_missing

3rdcycle:
Make sure the 'accessor' parameter is passed on when
the serialize method in Field calls itself recursively.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -174,7 +174,8 @@ class Marshaller(object):
         if many and obj is not None:
             self.__pending = True
             ret = [self.serialize(d, fields_dict, many=False, strict=strict,
-                                    dict_class=dict_class, accessor=accessor)
+                                    dict_class=dict_class, accessor=accessor,
+                                    skip_missing=skip_missing)
                     for d in obj]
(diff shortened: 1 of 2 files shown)


sloria:
@3rdcycle Nice catch and patch. Could you please a test to this test class: https://github.com/marshmallow-code/marshmallow/blob/f33375dc3a52994268bc4db80d1f396fca549e7e/tests/test_schema.py#L1876? Once the test is added, this is good to merge. If you don't have time to do this, let me know and I can go ahead and do it, probably some time this weekend.

3rdcycle:
I added the test and allowed myself to make a small change to the other accessor tests. The reason is that I noticed the asserts wouldn't trigger even if the custom accessor was missing. I guess that's because the default accessor in utils.py is smart enough to handle dicts correctly anyway.

3rdcycle:
Same issue with 'skip_missing'. I hope it's ok I bundled these together.

sloria:
@3rdcycle Thanks for fixing the tests and bugs!

sloria:
Backported this fix to `1.2-line`

Issue #127, opened 2015-01-16
Turn off automatic skipping of unknown fields


ryanlowe0:
When loading/deserializing, unknown field names are skipped. It would be nice to be able to raise an error instead, particularly when calling `.validate()`. Currently this doesnt seem possible, even with a custom validator, because the field filtering appears to already have occurred by the time the validator gets a hold of it, e.g.

``` python
class MySchema:
    foo = field.String()

@MySchema.validator
def check_unknown_fields(schema, data):
    for k in data:
        if k not in schema.fields:
            raise ValidationError("Unknown field name: '{}'".format(k))

errors = MySchema().validate({'bar': 'abc'})
```

doesnt work because data will be an empty `{}` by the time `check_unknown_fields()` is called.

sloria:
@ryanlowe0 Thanks for the suggestion.

One simple change would be to pass the raw input dict to schema-level validators

``` python
@MySchema.validator
def check_unknown_fields(schema, data, raw_data):
    for k in raw_data:
        if k not in schema.fields:
            raise ValidationError("Unknown field name: '{}'".format(k))
```

If this were implemented, would this meet your use case?

ryanlowe0:
Yes indeed, that is pretty much just what I was thinking!

sloria:
Great. Then I think we can move forward with this. Pull requests welcome!

ryanlowe0:
@sloria I started to implement adding the proposed raw_data argument, however I have a concern before putting it into a PR.

Sending a second `raw_data` arg rather than changing the behavior of `data` arg sounds great from a backwards compatibility standpoint, but only if it were optional. Unfortunately, this is proving tricky due to the way validators are collected and stored before being called. I thought maybe I could use `inspect.getargspec` to detect whether the additional arg was present, but apparently it doesnt work properly with partials in Python 2.6/7 (perhaps only in 3.4).

There may yet some way to get around this, but it may get ugly. An alternative idea would be to introduce some optional meta attribute or something that would cause raw data to be sent to validators _instead_ of filtered data.

In short, which of these sounds best?
1. Figure out some way to make raw_data be an optional third arg to user-defined validators
2. Keep the `(schema, data)` arg signature but add a meta-level flag to toggle between raw/filtered data
3. Always require the third raw_data arg
4. Stop filtering data altogether

Numbers 3 and 4 are simplest but break backwards compatibility..

sloria:
[...]

PR #142, opened 2015-02-06
Raw data validation

ryanlowe0:
This pull request implements the optional raw_data parameter for custom validators, as discussed in issue #127.

### DIFF
--- a/marshmallow/utils.py
+++ b/marshmallow/utils.py
@@ -341,17 +341,6 @@ def callable_or_raise(obj):
 
 
 def get_func_args(func):
-    """Return a tuple of argument names for a function.
-
-    .. deprecated:: 1.2.0
-        Renamed to :func:`get_callable_args`.
-    """
-    warnings.warn('get_func_ars was renamed to get_callable_args. get_func_args '
-                  'will be removed in version 2.0', DeprecationWarning)
-    return get_callable_args(func)
-
-
-def get_callable_args(func):
     """Given a callable, return a tuple of argument names. Handles
     `functools.partial` objects and class-based callables.
     """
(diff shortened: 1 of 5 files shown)


marshmallow/utils.py
```diff
@@ -343,3 +343,15 @@ def callable_or_raise(obj):
 def get_func_args(func):
     """Return a tuple of argument names for a function."""
     return inspect.getargspec(func).args
+
+
+def get_callable_args(func):
```
sloria: Is there any reason we should keep both `get_func_args` and `get_callable_args`?

sloria:
Thanks for this! Looks good. Just curious whether we need both functions for getting a callable's arguments.

marshmallow/utils.py
ryanlowe0: I dont guess so. I was really just following how there was `get_func_name` and `get_callable_name` with the former being deprecated, but I dont see why all cases of `get_func_args` couldnt be replaced with `get_callable_args`. Ill give that a try.

ryanlowe0:
@sloria, Im not sure why the Travis build is failing. All tests pass for me and although I am now seeing flake8 warnings, they occur with or without the latest change..

sloria:
@ryanlowe0 I'm not quite sure how the tests are passing on dev, either. I've just fixed them on `dev`, so if you merge that you should get a passing build.

sloria:
@ryanlowe0 One last change: in order to maintain backwards-compat, why don't we just have the single `get_func_args` (with your modifications to handle partials and callable classes) instead of renaming it?

sloria:
Looks good. Thanks for adding docs and tests!

Issue #135, opened 2015-01-31
Deprecate the Select field


philtay:
I propose to deprecate the `Select` field in favor of a `OneOf` validator. Reasons:
- The `Select` field is just a validator masked as a field. It doesn't perform any real (de)serialization.
- It works only with basic data types. The field's `choices` argument can't be a list of "complex" objects (e.g. dates).
- Currently there is no built-in way to validate a field against a fixed set of values.
- A convenience method to associate choices to labels (e.g. to populate a form select field) would be quite useful. The validator could implement it in a generic way (i.e. not tied to any data type in particular).

Please let me know your opinion.

sloria:
+1 to adding a `OneOf` validator. 

I'm not sure about the point about handling "complex" objects. As long a class implements `__contains__` (which `date` does) , the the `in` operator will work. Am I missing something?

A `OneOf` validator is more flexible than the `Select` field, so we can probably deprecate `Select` for 2.0-a.

philtay:
@sloria I meant something else, but it doesn't matter anymore now that we have the `OneOf` validator.

Do you want a `ContainsOnly` validator?

http://docs.pylonsproject.org/projects/colander/en/latest/api.html#colander.ContainsOnly

It's the natural extension of `OneOf` to sequences. Colander implements it using the set's method `issubset`, but I want to make it more generic (i.e. without the requirement of hashable items). I need it especially in combination with the `List` field.

Another thing. I'm starting to think we should decouple (de)serialization logic and validation logic as a general rule. Fields like `Url` and `Email` are just `String` with a validator. In general the `ValidatedField` class should go away.

sloria:
Sure. I think a `ContainsOnly` validator would be useful.

> I'm starting to think we should decouple (de)serialization logic and validation logic as a general rule. 

Agreed. That's a good rule of thumb moving forward. Though I probably won't remove `Url` and `Email` very soon, as it would break a lot of existing applications. Also, the `validate.URL` and `validate.Email` validators only make sense with `String` fields, so at the very least, those fields serve as a convenient shorthand.

philtay:
Ok, I'll implement `ContainsOnly` ASAP.

davidism:
Rather than separate validators for `OneOf` and `ContainsOnly`, one `Choices` validator that takes a `multi=True` argument seems more straightforward.

philtay:
[...]

PR #145, opened 2015-02-10
OneOf validator

philtay:
If these changes are ok I'll add the unit tests. Ref. issue #135

### DIFF
--- a/marshmallow/validate.py
+++ b/marshmallow/validate.py
@@ -282,19 +282,25 @@ class OneOf(object):
     :param iterable labels:
         Optional sequence of labels to pair with the choices.
     :param str error:
-        Error message to raise in case of a validation error.
+        Error message to raise in case of a validation error. Can be
+        interpolated using `{value}`, `{choices}` and `{labels}`.
     """
     def __init__(self, choices, labels=None, error=None):
         self.choices = choices
         self.labels = labels if labels is not None else []
         self.error = error or 'Not a valid choice.'
 
+        choices_text = ', '.join(text_type(choice) for choice in self.choices)
+        labels_text = ', '.join(text_type(label) for label in self.labels)
+
+        self.error.format(choices=choices_text, labels=labels_text, value='{value}')
+
     def __call__(self, value):
         try:
             if value not in self.choices:
-                raise ValidationError(self.error)
+                raise ValidationError(self.error.format(value=value))
         except TypeError:
-            raise ValidationError(self.error)
+            raise ValidationError(self.error.format(value=value))
 
         return value
 
--- a/marshmallow/validate.py
+++ b/marshmallow/validate.py
@@ -287,20 +287,23 @@ class OneOf(object):
     """
     def __init__(self, choices, labels=None, error=None):
         self.choices = choices
+        self.choices_text = ', '.join(text_type(choice) for choice in self.choices)
+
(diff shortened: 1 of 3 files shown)


marshmallow/validate.py
```diff
@@ -271,6 +274,48 @@ def __call__(self, value):
...
+        A sequence of valid values.
+    :param iterable labels:
+        Optional sequence of labels to pair with the choices.
+    :param str error:
+        Error message to raise in case of a validation error.
+    """
+    def __init__(self, choices, labels=None, error=None):
+        self.choices = choices
+        self.labels = labels if labels is not None else []
+        self.error = error or 'Not a valid choice.'
```
sloria: Perhaps we can have a more descriptive error message that can be templated (like in `Range` and `Length`). Something like: `'{val} is not one of {choices}'`

sloria:
Out of curiosity, can you describe how you plan to use the `pairs` method? How will you access it and for what purpose?

marshmallow/validate.py
```diff
@@ -271,6 +274,48 @@ def __call__(self, value):
         return value
 
 
+class OneOf(object):
+    """Validator which succeeds if ``value`` is a member of ``choices``.
+
+    :param iterable choices:
```
sloria, line 280: For consistency with the rest of the codebase, don't add an extra newline after `param_name:`.

sloria:
Looks good so far. :+1: Will merge this after the tests are added.

philtay:
> Perhaps we can have a more descriptive error message that can be templated (like in Range and Length). Something like: '{val} is not one of {choices}'

I thought about this and decided to go for a basic message. Two reasons:
- It's not very useful to show {val} because the end user already see what is selected in the UI (e.g. a select field). Furthermore, more often than not, {val} is an opaque value to the end user (e.g. you select United Kingdom and the error message would be "GB is not one of ...").
- Listing all the available {choices} can be a security vulnerability in some circumstances. Not something to do by default.

> Out of curiosity, can you describe how you plan to use the pairs method? How will you access it and for what purpose?

The purpose is to generate a select form field. Something like that:

``` html
<select>
  <option value="{value}">{label}<option>
  ...
  ...
</select>
```

The options can be generated using the `pairs` method. It's useful if you want to keep together valid values and their respective labels. If you can't specify the labels somewhere in Python you'll end up to hardcode them in the UI. To use the method you can instantiate the validator and pass it both to a schema field and to your HTML template. You can also generate the list of valid values and labels, convert to JSON and throw it at your preferred client side framework (e.g. Angular or Ember) that will take care of displaying a nice select for the happiness of your users! ;)

> For consistency with the rest of the codebase, don't add an extra newline after param_name:

Ok, but all the other validators' docs have the extra newline.

Please let me know how to move forward.

sloria:
It would still be nice to have the option to interpolate `val` and `choices` if the user wanted to do that (even if the default message didn't use those values).

Thanks for the clarification re: `pairs`.

Yes, I realize now the validators follow a different convention from the rest of the codebase. I can make those consistent in a separate commit after this is merged.

philtay:
Done. Let me know if it's ok and I'll add the tests. Tnx

sloria:
Looks good. Once tests are added, this should be good to merge.

philtay:
Done.

sloria:
Thanks!

PR #147, opened 2015-02-15
1 fix + tests

philtay:


sloria:
Good catch. Since this is a bug fix, can you please resend this with the `1.2-line` branch as 
the base? I will forward port it to `dev` once it's merged.

philtay:
The bug affects only 2.0-dev. {{input}} is not supported in 1.2.

sloria:
Ah, you're true.

Thanks for adding those tests as well!

Issue #105, opened 2014-12-14
TimeDelta field bug


philtay:
Float strikes again:

``` python
from decimal import Decimal as D
from datetime import timedelta as td
from marshmallow.compat import total_seconds
from marshmallow.fields import TimeDelta

a = td(microseconds=1)
D(total_seconds(a))

# Decimal('9.99999999999999954748111825886258685613938723690807819366455078125E-7')

a = TimeDelta()
b = a.deserialize(0.1)
c = total_seconds(b)
D(c)

# Decimal('0.1000000000000000055511151231257827021181583404541015625')
```

philtay:
IMO the solution is to use integer instead of float and drop microseconds support. Here the standard to stick to is ISO 8601 (durations).

http://en.wikipedia.org/wiki/ISO_8601#Durations

To be really useful the field should also support ISO 8601 duration strings. But this requires an external library. Conversion from Python timedelta to ISO 8601 duration string and vice versa, although certainly doable, is a pain you don't easily forget.

philtay:
@sloria Are you ok with using integers instead of floats for this field? Have a look here:

https://docs.python.org/3.5/library/datetime.html#timedelta-objects

`If no argument is a float, the conversion and normalization processes are exact (no information is lost).`

I will add a `microseconds` boolean attribute to the field's **init**. If `true` the integer value is interpreted as microseconds during (de)serialization, otherwise as seconds. Defaults to `False`.

Proof of concept:

``` python
class TimeDelta(Field):
    def __init__(self, microseconds=False, error=None, **kwargs):
        self.microseconds = microseconds
        super(TimeDelta, self).__init__(error=error, **kwargs)

    def _serialize(self, value, attr, obj):
        seconds = value.days * 86400 +  value.seconds
        return seconds * 10**6 + value.microseconds if self.microseconds else seconds

    def _deserialize(self, value):
        if self.microseconds:
            return dt.timedelta(microseconds=int(value))
        else:
            return dt.timedelta(seconds=int(value))
```

sloria:
@philtay +1. Let's do this for 2.0-a (since this is technically a backwards-compatible change).

philtay:
Ok, add the 2.0-a milestone to this issue. I just need to understand why this

https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/compat.py#L61

[...]

PR #148, opened 2015-02-15
TimeDelta

philtay:
#105

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
-"""Field classes for various types of data.
-"""
+"""Field classes for various types of data."""
+
 from __future__ import absolute_import, unicode_literals
 
 import datetime as dt
@@ -11,7 +11,7 @@ from operator import attrgetter
 
 from marshmallow import validate, utils, class_registry
 from marshmallow.base import FieldABC, SchemaABC
-from marshmallow.compat import text_type, iteritems, total_seconds, basestring
+from marshmallow.compat import text_type, iteritems, basestring
 from marshmallow.exceptions import (
     MarshallingError,
     UnmarshallingError,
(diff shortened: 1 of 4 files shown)


philtay:
The Travis failure is not caused by my code. Can I increase to 110 the line length? It's better.

marshmallow/fields.py
```diff
@@ -1127,27 +1127,54 @@ def _deserialize(self, value):
...
     def _deserialize(self, value):
-        """Deserialize a value in seconds to a :class:`datetime.timedelta`
-        object.
-        """
         try:
-            return dt.timedelta(seconds=float(value))
-        except (TypeError, AttributeError, ValueError):
+            value = int(value)
+        except (TypeError, ValueError):
+            msg = '{0!r} cannot be interpreted as a valid period of time.'.format(value)
```
sloria, line 1170: Can you please add a test that reaches this branch?

marshmallow/fields.py
```diff
@@ -1127,27 +1127,54 @@ def _deserialize(self, value):
...
+            raise UnmarshallingError(getattr(self, 'error', None) or msg)
+
+        kwargs = {self.precision: value}
+
+        try:
+            return dt.timedelta(**kwargs)
+        except OverflowError:
             msg = 'Could not deserialize {0!r} to a timedelta object.'.format(value)
```
sloria: Can you please add a test that tests this branch?

sloria:
Looks good. Just a couple more tests, and this should be good to merge.

philtay:
The `test_invalid_timedelta_field_deserialization` test already reaches the branches you indicated.

``` python

# ValueError
int('')

# TypeError
int([])

# OverflowError
field = fields.TimeDelta(fields.TimeDelta.DAYS)
field.deserialize(9999999999)
```

marshmallow/fields.py
sloria: Perhaps we should change this error message to be the same as the other message. Often, the user (of the library) wants to display these messages to the client, which doesn't care about "deserializing to timedelta objects". I think your more generic message about "interpreting values as valid period of time." makes more sense.

Alternatively, we could have a more specific message about overflow, such as `Value must have magnitude <= 999999999`.

EDIT: fix typo.

sloria:
Right, those branches are reached; could you also make assertions about the error messages?

e.g.

``` python
        with pytest.raises(UnmarshallingError) as excinfo:
            field.deserialize(in_value)
        expected = ('{0!r} cannot be interpreted as '
                    'a valid period of time.').format(in_value)
        assert expected in str(excinfo)
```

philtay:
Done

marshmallow/fields.py
```diff
@@ -1127,28 +1127,55 @@ def _deserialize(self, value):
...
+
+        self.precision = precision
+        super(TimeDelta, self).__init__(error=error, **kwargs)
+
     def _serialize(self, value, attr, obj):
         try:
-            return total_seconds(value)
+            days = value.days
+            seconds = days * 86400 + value.seconds
+            microseconds = seconds * 10**6 + value.microseconds
```
sloria, line 1159: I wonder if we can save unneeded computation here by only computing to the requested precision. Something like:

``` python
        try:
            days = value.days
            if self.precision == self.DAYS:
                return days
            else:
                seconds = days * 86400 + value.seconds
                if self.precision == self.SECONDS:
                    return seconds
                else:  # microseconds
                    return seconds * 10**6 + value.microseconds
        except AttributeError:
            msg = '{0!r} cannot be formatted as a timedelta.'.format()
            raise MarshallingError(getattr(self, 'error', None) or msg)
```

Perhaps there is a more elegant way, but it would be good remove the extra computation.

philtay:
I thought about it. The CPU cycles waste is really negligible (integer math). Saving them would mean uglier code, like the big `try` block or a lot of `if-else`. It doesn't worth it.

philtay:
And I'd be curious to benchmark them: full C speed integer math vs big `try` and nested `if-else`. Who will win? :)

sloria:
Here's a quick-and-dirty comparison of serializing a bunch of `timedeltas` to microsecond values: http://nbviewer.ipython.org/gist/sloria/5c1b845a6ec59c2404d0

philtay:
LOL, you really have benchmarked them! However, this confirms that we don't have to worry about performance in this case and that we can avoid ugly code. The difference is negligible and probably within statistical error. Moreover, considering that you have tested with `precision='microseconds'`, the math involved in both `TimeDelta1` and `TimeDelta2` is exactly the same. Thus the difference in performance can't be attributed to multiplications and additions.

philtay:
You should use `precision='days'` to get the real difference and then subtract 0.64 sec (5.99 - 5.35). Anyway, if you really prefer the `try-if-else` approach, please feel free to uglify the code yourself after merge ;)

sloria:
Here are comparisons with each level of precision: http://nbviewer.ipython.org/gist/sloria/6229107fa5cea79c80a2. The try-if-else implementation is consistently faster. As expected, the difference is the largest with `days`.

If you don't want to make the change, I'll go ahead and make it when I merge.

philtay:
ok, make the change when you merge

PR #152, opened 2015-02-19
add new field called 'Binary'

mirko:
This field (de)serializes from/to base64 encoded data.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -8,6 +8,7 @@ import uuid
 import warnings
 import decimal
 from operator import attrgetter
+from base64 import b64encode, b64decode
 
 from marshmallow import validate, utils, class_registry
 from marshmallow.base import FieldABC, SchemaABC
@@ -54,6 +55,7 @@ __all__ = [
     'Str',
     'Bool',
     'Int',
+    'Binary',
 ]
 
 class Field(FieldABC):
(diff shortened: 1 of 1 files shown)


philtay:
@sloria Of course this PR is not in a mergeable state, but what do you think about the usefulness of a Binary field? For a newbie it can be tricky to implement this right, but dealing with binary data is fairly common.
Below an approximation of a correct implementation in Python 3 (rough, improvable and done in 5 min without testing it):

``` python
class Base64(Field):
    def __init__(self, urlsafe=False, error=None, **kwargs):
        if urlsafe:
            self.b64encode = urlsafe_b64encode
            self.b64decode = urlsafe_b64decode
        else:
            self.b64encode = standard_b64encode
            self.b64decode = standard_b64decode

        super(Base64, self).__init__(error=error, **kwargs)

        try:
            b64_bytes = self.b64encode(value)  # value is a bytes object
        except TypeError:
            raise MarshallingError('oooops, not a bytes object')

        return b64_bytes.decode(encoding='ascii', errors='strict')  # return a Unicode string

        try:
            return self.b64decode(value)  # return a bytes object
        except (TypeError, ValueError, binascii.Error):
            raise UnmarshallingError('oooops, cannot decode')
```

sloria:
Thank you @mirko and @philtay. I'm going to hold off on adding a `Binary` field for the time being; it is more maintenance burden that it is worth right now.

PR #158, opened 2015-02-22
PEP 257 and PEP 8

philtay:
- Separate class definitions with two blank lines. (PEP 8)
- Insert a blank line after all docstrings (one-line or multi-line) that document a class. (PEP 257)

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -56,6 +56,7 @@ __all__ = [
     'Int',
 ]
 
+
 class Field(FieldABC):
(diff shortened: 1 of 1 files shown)


sloria:
This project does not follow those specific guidelines--I believe it adds too much unnecessary whitespace; hence, the flake8 config in `setup.cfg` explicitly ignores this type of error.

philtay:
Ok, but please note that currently `fields.py` uses an inconsistent style. Sometimes we have two blank lines between classes and sometimes not. The same thing for the blank line after class docstring.

Issue #149, opened 2015-02-17
ContainsOnly validator


sloria:
From the thread in https://github.com/marshmallow-code/marshmallow/issues/135#issuecomment-74391484 (trimmed for brevity):

@philtay 

> Do you want a ContainsOnly validator? [...] It's the natural extension of OneOf to sequences. Colander implements it using the set's method issubset, but I want to make it more generic (i.e. without the requirement of hashable items). I need it especially in combination with the List field.

Further comments

@davidism 

> Rather than separate validators for OneOf and ContainsOnly, one Choices validator that takes a multi=True argument seems more straightforward.

@philtay 

> My idea is to subclass OneOf and override **call**. I prefer to not put too much logic in a single validator. They must be very robust and thus very easy to debug. Anyway your idea is certainly doable.

philtay:
@sloria what approach do you prefer?

sloria:
I agree with @philtay on having separate validators. The meaning of `ContainsOnly` seems clearer than `Choices(multi=True)` and better to keep the validators small and focused.

sloria:
Closed by #159

PR #159, opened 2015-02-22
ContainsOnly

philtay:
#149


sloria:
Thanks for this! I just made a couple small changes when I merged:
- Break up tests into smaller units
- Add a comment explaining why `list.index` is used.

Issue #173, opened 2015-03-17
as_string doesn't work on fields.Integer


taion:
`fields.Integer(as_string=True).as_string == False`

This is a trivial fix and I'll submit a preliminary PR shortly.

I am confused by the following things, however:
1. Why are the non-`kwargs` arguments on `Integer` other than `default` duplicated with the same value as on `Number`?
2. Why is the documentation for `as_string` not on `Number`?
3. Is it intentional that `attribute` and `error` are not accepted as positional args for `Decimal`?

PR #175, opened 2015-03-17
Fix as_string on fields.Integer

taion:
For marshmallow-code/marshmallow#173

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -745,10 +745,8 @@ class Integer(Number):
 
     num_type = int
 
-    def __init__(self, default=0, attribute=None, as_string=False, error=None, **kwargs):
-        self.as_string = as_string
-        super(Integer, self).__init__(default=default, attribute=attribute,
-            error=error, **kwargs)
+    def __init__(self, default=0, *args, **kwargs):
+        super(Integer, self).__init__(default=default, *args, **kwargs)
 
 
 class Decimal(Number):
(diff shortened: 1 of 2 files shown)


taion:
It so happens that the other arguments on `Integer` are an exact match to the ones on `Number`, so it seems like it'd be cleaner all-around to just pass those up to `Number.__init__`.

I do think that `as_string` should be documented on `Number`, and that the signature on `Decimal.__init__` is very odd.

I've verified that the new test case fails against the old code, but works against the new code.

sloria:
Thanks for the patch! I agree; it makes sense to document `as_string` on Number. What do you find strange about the `Decimal` signature?

taion:
The positional args on `Decimal.__init__` are: `places`, `rounding`, `default`, `as_string`. However, on `Number.__init__` (and for `Integer` and `Float`), the args are: `default`, `attribute`, `as_string`, `error`.

It's a pretty minor mismatch - the only issue is that `as_string` is re-ordered relative to `attribute`. On the other hand, I can't imagine why any sane person would pass `as_string` in as a positional arg, so this is a non-issue.

It only matters in that maybe you'd want to drop `as_string` as an explicit arg on `Decimal.__init__` and just have it handled via `args` or `kwargs`, but this has a slight chance of changing the existing behavior, if anybody is doing the silly thing of passing in `as_string` as a positional arg here.

PR #182, opened 2015-03-21
Schema.py - only respects exclude

lustdante:
I did make an issue a week ago but couldn't resist to try out this pull request.

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -589,13 +589,10 @@ class BaseSchema(base.SchemaABC):
         """Update fields based on the passed in object."""
         # if only __init__ param is specified, only return those fields
         if self.only:
-            ret = self.__filter_fields(self.only, obj, many=many)
-            self.__set_field_attrs(ret)
-            self.fields = ret
-            return self.fields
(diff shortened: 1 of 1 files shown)


sloria:
Thanks @lustdante for tackling this.

I think the `only` parameter should still be respected, but `exclude` should also be respected. The final fields used should be the set difference between `only` and `exclude`. Could you please update your PR accordingly?

lustdante:
Hmm, my PR was about removing self.only's short-circuiting so that it goes through field_names = field_names - excludes just like any other conditions do. I tried it on my forked repo and it's quite functional. Please let me know if any more change is required.

Thanks!

sloria:
@lustdante Ok thanks. I will give this a closer look within the next few days.

sloria:
This looks good. :+1: The failing test can be removed as it is no longer correct. It would be great to expand the tests to account for this API change; would you like to add those? If not, I can certainly add them when I merge.

sloria:
Went ahead and fixed the tests. Thanks @lustdante for the fix!

Issue #185, opened 2015-03-31
fields.List and generators


aganezov:
Hi,

First, I would like to thank you for this wonderful library! Using it and getting happier and happier the more I use it.

Set up for the question:
Assume I have the following Schema

```
class MySchema(Schema)
    values = fields.List(fields.Int, attribute="my_generator", allow_none=False)
```

and a class

```
class MyClass(object):
    def __init__(pipe):
        self.pipe = pipe

    def my_generator(self):
        yield from self.pipe
```

and then I do the following:

```
schema = MySchema()
m_obj = MyClass([1, 2, 3, 4])
print(schema.dump(m_obj).data)
```

I get something like {"values": []}

but is I change generator to list

```
def my_generator(self):
        return list(self.pipe)
```

it works fine. 

The question:
Is there some support for `fields.List` to work with generators?

sloria:
@sergey-aganezov-jr Thanks for kind words and for reporting this issue.

Yes, I think some modifications to `fields.List#_serialize` would allow generators to be serialized.

PRs welcome!

davidism:
Why not just take out both checks (generator and the one below it about index/string) and just allow any iterable?  This would be _way_ more convenient when, for example, you have a set that you want to serialize, since there is no set in JSON and a list is probably good enough.

The one downside is that it will now accept a bare string, but that "problem" will pretty much be immediately apparent when you get a list of single characters.

This would also do away with the Python anti-pattern of iterating over a range the length of the list and indexing.

aganezov:
I don't think `string` issue will be that big, as in this case it would be a "hack" to go from string to a list of chars, by simply specifying `fields.List(fields.String())` on the string value in the serialized class. And if people want to serialize string, they shall really use `fields.String()`.

I believe this pull request is the first step, and then `fields.List` can be generalized into supporting all iterables. On the other hand, it is not that easy to check if something is `iterable` in python ([link](http://stackoverflow.com/questions/1952464/in-python-how-do-i-determine-if-an-object-is-iterable)).

It seems that something like this can be done:

``` python
    if utils.is_indexable_but_not_string(value) and not isinstance(value, dict):
        return [self.container.serialize(idx, value) for idx in range(len(value))]
    if utils.is_iterable_but_not_string(value) and not isinstance(value, dict):
        return [self.container.serialize(0, [entry]) for entry in value]
    ....
```

and `utils.is_iterable_but_not_string(value)` can be implemented something like this:

``` python
[...]

PR #186, opened 2015-04-02
added support for fields.List.serialize to correctly process generators (

aganezov:
following our discussion on #185 I've made small workaround changes to the fields.List.serialize method to support generators.

additional test cases were added and successfully passed, as well as the rest of the test suite

while my workaround is simple, it does not support the generator protocol in full (we create dummy one entry list every time we go through generator, and no "exception chaining", no "send()", etc is provided), but I don't think this is that important at the moment.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -4,6 +4,7 @@
 from __future__ import absolute_import, unicode_literals
 
 import datetime as dt
+import types
 import uuid
 import warnings
 import decimal
@@ -408,6 +409,8 @@ class List(Field):
             self.container = cls_or_instance
 
     def _serialize(self, value, attr, obj):
+        if isinstance(value, types.GeneratorType):
+            return [self.container.serialize(0, [entry]) for entry in value]
         if utils.is_indexable_but_not_string(value) and not isinstance(value, dict):
             return [self.container.serialize(idx, value) for idx
                     in range(len(value))]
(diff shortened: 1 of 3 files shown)


davidism:
Why not just take out both checks (generator and the one below it about index/string) and just allow any iterable?  This would be *way* more convenient when, for example, you have a set that you want to serialize, since there is no set in JSON and a list is probably good enough.

The one downside is that it will now accept a bare string, but that "problem" will pretty much be immediately apparent when you get a list of single characters.

sergey-aganezov-jr:
I don't think `string` issue will be that big, as in this case it would be a "hack" to go from string to a list of chars, by simply specifying `fields.List(fields.String())` on the string value in the serialized class. And if people want to serialize string, they shall really use `fields.String()`.

I believe this pull request is the first step, and then `fields.List` can be generalized into supporting all iterables. On the other hand, it is not that easy to check if something is `iterable` in python ([link](http://stackoverflow.com/questions/1952464/in-python-how-do-i-determine-if-an-object-is-iterable)).

It seems that something like this can be done:

```python
    if utils.is_indexable_but_not_string(value) and not isinstance(value, dict):
        return [self.container.serialize(idx, value) for idx in range(len(value))]
    if utils.is_iterable_but_not_string(value) and not isinstance(value, dict):
        return [self.container.serialize(0, [entry]) for entry in value]
    ....
````

and `utils.is_iterable_but_not_string(value)` can be implemented something like this:
```python
    def is_iterable_but_not_string(obj):
        return (hasattr(obj, "__iter__") or isinstance(obj, types.GeneratorType)) and not not hasattr(obj, "strip")

Issue #153, opened 2015-02-19
Processing API for dumping does not match API for loading


davidism:
When dumping, I want to manipulate some of the attributes on the object being dumped before actually dumping.  So I decorate a `preprocessor` function.  But that only gets used when _loading_.  I could use a `Method` field, but the data returned from that must be the final serialized form, so there's no way to specify that it's _actually_ a `Nested(OtherSchema, many=True)` unless I do that serialization manually at the end of the method.

When loading, I want to manipulate the loaded data in the exact opposite direction as the situation above.  So I decorate a `data_handler` function.  But that only gets used when _dumping_.  The solution is more straightforward here: I override the `make_object` method to manipulate the final output.

My point is that the names are different everywhere, some of them are decorators while others are methods, and they don't apply in both directions, when it would be very convenient to be able to do so.  The loading situation is in better shape than the dumping one, since loading has `preprocessor` and `make_object`, except there's still a weird decorator vs. method difference.

There should be a standard way to preprocess and postprocess the entire data during both loading and dumping.  One solution is adding `pre_dump`, `post_dump`, `pre_load`, and `post_load` hooks, either as methods on the schema or as decorators.

davidism:
This is also somewhat related to #139, which is looking for a way to more accurately control how a field is named when dumping *and* loading.

sloria:
You raise fair points @davidism.

The rationale for the current API is as follows:
- The Schema decorators--`preprocessor` and `data_handler`--allow multiple Schemas to use the same pre- and post-processing routines without having to inherit from base classes or mixins. They also hide the fact that multiple functions can be stored within a list.
- `make_object` is a method because it is meant to be specific to a single Schema. Whereas pre- and post-processors are written as generic functions, `make_object` provides a 1-to-1 mapping between a schema and an "app-level" object.
- There is no preprocessing hook for dumping/serialization because marshmallow is not meant to make updates to app-level objects. All processing happens in your application's business logic _before_ serialization.

Here is the intended flow:

```
[client input] -> preprocess/validate -> make object -> [app-level object] -> business logic* -> [app-level-object] -> serialize -> modify serialized data
```

That said, I am not opposed to changing the API if it is unclear. Perhaps you could describe your use case in more detail and we can discuss possible solutions.

sloria:
[...]

PR #191, opened 2015-04-09
Add decorator-based pre/post-dump/load support

taion:
For #153

### DIFF
--- a/marshmallow/decorators.py
+++ b/marshmallow/decorators.py
@@ -1,0 +1,67 @@
+PRE_DUMP = 'pre_dump'
+POST_DUMP = 'post_dump'
+PRE_LOAD = 'pre_load'
+POST_LOAD = 'post_load'
+
+
+def pre_dump(*args, **kwargs):
+    return tag_processor(PRE_DUMP, *args, **kwargs)
+
+
+def post_dump(*args, **kwargs):
+    return tag_processor(POST_DUMP, *args, **kwargs)
+
+
+def pre_load(*args, **kwargs):
+    return tag_processor(PRE_LOAD, *args, **kwargs)
+
+
+def post_load(*args, **kwargs):
+    return tag_processor(POST_LOAD, *args, **kwargs)
+
+
+class _StaticProcessorMethod(staticmethod):
+    """Allows setting attributes on a staticmethod"""
+    pass
+
+
+class _ClassProcessorMethod(classmethod):
+    """Allows setting attributes on a classmethod"""
+    pass
+
+
+def tag_processor(tag_name, fn=None, raw=False):
+    """Tags decorated processor function to be picked up later
+
+    :return: Decorated function if supplied, else this decorator with its args
+        bound.
+    """
+    if fn is None:
+        return lambda fn_actual: tag_processor(tag_name, fn_actual, raw)
+
+    # Special-case rewrapping staticmethod and classmethod, because we can't
+    # directly set attributes on those.
+    if isinstance(fn, staticmethod):
+        try:
+            unwrapped = fn.__func__
+        except AttributeError:
+            # For Python 2.6.
+            unwrapped = fn.__get__(True)
+        fn = _StaticProcessorMethod(unwrapped)
+    elif isinstance(fn, classmethod):
+        try:
+            unwrapped = fn.__func__
+        except AttributeError:
+            # For Python 2.6.
+            unwrapped = fn.__get__(True).im_func
+        fn = _ClassProcessorMethod(unwrapped)
+
+    # Set a processor_tags attribute instead of wrapping in some class,
+    # because I still want this to end up as a normal (unbound) method.
+    try:
+        processor_tags = fn.__processor_tags__
+    except AttributeError:
+        fn.__processor_tags__ = processor_tags = set()
+    processor_tags.add((tag_name, raw))
+
+    return fn
--- a/marshmallow/decorators.py
+++ b/marshmallow/decorators.py
@@ -4,20 +4,20 @@ PRE_LOAD = 'pre_load'
 POST_LOAD = 'post_load'
 
 
-def pre_dump(*args, **kwargs):
-    return tag_processor(PRE_DUMP, *args, **kwargs)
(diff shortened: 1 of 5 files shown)


taion:
Two open questions:
1. What do you want to do about deprecating the old hooks? I didn't add any deprecation warnings.
2. In general the pattern for these is going to be mutating the passed-in item. I forgot to return the item a few times. It might improve ergonomics to just assume that the passed-in object was mutated in-place if the processor function returns None.

sloria:
> What do you want to do about deprecating the old hooks?

`preprocessor`, `data_handler`, and `extra` can all be deprecated, but I think we can defer that for now until this new API is finalized.

> It might improve ergonomics to just assume that the passed-in object was mutated in-place if the processor function returns None.

I agree.

sloria:
@taion Thanks for your work on this. I will look at this more in-depth over the weekend.

marshmallow/schema.py
```diff
@@ -99,6 +101,37 @@ def __new__(mcs, name, bases, attrs):
         for attr in mcs.FUNC_LISTS:
             attr_copy = copy.copy(getattr(klass, attr))
             setattr(klass, attr, attr_copy)
+
+        # Now add in the decorated processors. By doing this after constructing
+        # the class, we let standard inheritance do all the hard work.
+        mro = klass.mro()
+
+        klass.processors = defaultdict(list)
+        for attr_name in dir(klass):
```
jmcarp, line 126: `dir` apparently returns an alphabetical list of attributes, meaning that decorated methods might not be applied in order expected by the developer. You might use `inspect` or store a counter on the decorators to get the right order.

taion, line 126: Do you think we need to guarantee order beyond "raw processors last when dumping, raw processors first when loading"? I feel that this would not be an especially obvious feature, and that ordering dependencies would be better expressed by putting them in the same processor, rather than depending on some guaranteed order.

jmcarp, line 126: Agreed that users _should_ handle order by putting ordered operations in the same callback. But they might not, and if they don't, I could imagine this being a debugging headache. Anyway, if we don't want callbacks to run in any particular order, the docs should be clear about that, so that users don't make assumptions and get confused.

taion, line 126: I'd be inclined to go with telling users not to depend on any ordering guarantees in the docs. It just seems like not the best idea. @sloria what do you think?

marshmallow/schema.py
```diff
@@ -99,6 +101,37 @@ def __new__(mcs, name, bases, attrs):
...
+            for parent in mro:
+                try:
+                    attr = parent.__dict__[attr_name]
+                except KeyError:
+                    continue
+                else:
+                    break
+            else:
+                # In case we didn't find the attribute and didn't break above.
+                # This isn't necessary and we should never hit this.
```
sloria: Why include this?

marshmallow/schema.py
sloria, line 126: Yes, I think the docs should clearly state that ordering is not guaranteed.

marshmallow/schema.py
```diff
@@ -99,6 +101,37 @@ def __new__(mcs, name, bases, attrs):
         for attr in mcs.FUNC_LISTS:
             attr_copy = copy.copy(getattr(klass, attr))
             setattr(klass, attr, attr_copy)
+
+        # Now add in the decorated processors. By doing this after constructing
+        # the class, we let standard inheritance do all the hard work.
+        mro = klass.mro()
+
+        klass.processors = defaultdict(list)
```
sloria: Might be good to make this `klass.__processors__` to prevent potential clashes with fields called `processors`.

marshmallow/schema.py
taion: I need to use `clazz.__dict__` to access the descriptor. I need to explicitly traverse the MRO to handle inheritance correctly, because inherited descriptors are not on `clazz.__dict__` and need to be pulled from the parent.

The `else` clause on the `for` loop is just for completeness. Even though that branch should be impossible to hit, I still didn't want something where `attr` could potentially end up undefined.

marshmallow/schema.py
taion: Makes sense. I don't want to add too many "fix XYZ" mini-commits, so I'll fix that when I revise the PR to also include doing something sensible when the processor does not return anything.

marshmallow/decorators.py
```diff
@@ -0,0 +1,67 @@
+PRE_DUMP = 'pre_dump'
+POST_DUMP = 'post_dump'
+PRE_LOAD = 'pre_load'
+POST_LOAD = 'post_load'
+
+
+def pre_dump(*args, **kwargs):
```
sloria: For documentation purposes, it might be clearer to explicitly include `raw` in these signatures. The `tag_processor` includes the `raw` value in the tuple tags anyway.

sloria:
Looking good so far @taion . Let me know when this is ready for further review.

taion:
New commit includes (2) from my 1st comment about treating `None` return values (implicit or otherwise) from processors as meaning that the element was modified in place, in addition to what was mentioned in comments on commit.

marshmallow/schema.py
```diff
@@ -99,6 +101,39 @@ def __new__(mcs, name, bases, attrs):
         for attr in mcs.FUNC_LISTS:
             attr_copy = copy.copy(getattr(klass, attr))
             setattr(klass, attr, attr_copy)
+
+        # Now add in the decorated processors. By doing this after constructing
+        # the class, we let standard inheritance do all the hard work.
+        mro = klass.mro()
+
+        klass.__processors__ = defaultdict(list)
```
sloria, line 125: Can this "processor tag" logic be put in a separate helper function or method?

sloria:
Just a minor nitpicky refactor, and this should be good to merge!

taion:
Done. Is this for 2.x, or will there be a 1.3 line that includes new features without removing old ones?

sloria:
Thanks @taion . This will go into 2.0-a.

Issue #204, opened 2015-04-28
Add useful `__repr__` methods for validators


sloria:
When inspecting a `Field`, it's useful to have a human-friendly representation of the field's validators, e.g. seeing the `min` and `max` values for a `Range` or `Length` validator.

philtay:
What about something like this?

``` python
class Validator(object):
    """Base abstract class"""

    def __repr__(self):
        return (
            '<validate.{ClassName}({_args}, error={self.error!r})>'
            .format(ClassName=self.__class__.__name__, self=self)
        )


class Range(Validator):
    def __repr__(self):
        rep = 'min={0!r}, max={1!r}'.format(self.min, self.max)
        return super(Range, self).__repr__().format(_args=rep)
```

P.S. Is it time to start removing deprecated fields and validator functions? I could open some PR, just  to speed up a bit the final release of 2.0

P.P.S Don't forget to tag 2.0.0b1 here: https://github.com/marshmallow-code/marshmallow/releases

sloria:
@philtay Maybe make a slight modification to make the `super` call unnecessary:

``` python
class Validator(object):
    """Base abstract class"""
    def __repr__(self):
        return (
            '<validate.{ClassName}({args}, error={self.error!r})>'
            .format(ClassName=self.__class__.__name__, self=self, args=self._repr_args())
        )

class Range(Validator):
    def _repr_args(self):
        return 'min={0!r}, max={1!r}'.format(self.min, self.max)
```

> Is it time to start removing deprecated fields and validator functions? I could open some PR, just to speed up a bit the final release of 2.0

My plan was to remove deprecated fields/validators in a minor release, e.g. 2.1.0, in order to ease migration to 2.0.0. 

When that time comes, I was also thinking of turning the lowercased validator functions into aliases of the class validators. I'm not committed to it, but I think it'd make the validator implementation (function factory vs. callable class) more transparent.

> Don't forget to tag 2.0.0b1 here

Thanks. Just pushed the tag.

philtay:
Agreed, without `super` is better. I still have a doubt however. What `args` should be for a validator like `Email`? There are no `args` other than `error`.

sloria:
@philtay We could do something like

``` python
def __repr__(self):
    args = self._repr_args()
    class_name = self.__class__.__name__
    if args:
        return (
            '<validate.{ClassName}({args}, error={self.error!r})>'
            .format(ClassName=class_name, self=self, args=args)
        )
    else:
[...]

PR #205, opened 2015-05-01
__repr__

philtay:
#204

### DIFF
--- a/marshmallow/validate.py
+++ b/marshmallow/validate.py
@@ -12,7 +12,23 @@ from marshmallow.compat import basestring, text_type, zip_longest
 from marshmallow.exceptions import ValidationError
 
 
-class URL(object):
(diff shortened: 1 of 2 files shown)


marshmallow/validate.py
```diff
@@ -12,7 +12,23 @@
...
-class URL(object):
+class Validator(object):
+    """Base abstract class for validators."""
+
+    def __repr__(self):
+        args = self._repr_args()
+        args = '{0}, '.format(args) if args else ''
+
+        return (
+            '<validate.{self.__class__.__name__}({args}error={self.error!r})>'
```
sloria, line 23: I'm going to remove the `validate.` prefix to account for user validators that subclass `Validator`.

sloria:
Thanks @philtay !

PR #206, opened 2015-05-03
fields.Decimal fixes

philtay:
Fix bugs and bad behaviors.


philtay:
@sloria IMO we should deprecate Float in favour of Decimal. This field has several problems, including lack of repeatability, idempotency and special numeric values. If you know what you're doing and REALLY want floats, you can easily convert Decimal <-> float. I can't think of any use-case where you want to receive "0.100000001490116119384765625" when a client sends you "0.1". Basically Integer and Decimal should be the only two supported numeric types.

PR #210, opened 2015-05-05
Skip call to `get_value` when not `_CHECK_ATTRIBUTE`.

### LINKED ISSUE #208: Unnecessary call to `get_value` for `Function`, `Method` fields
Related to #207.

On serializing, `Function` and `Method` fields make a call to `Field#get_value`. The result of this call is then discarded, because the `_serialize` method of both fields ignores the `value` parameter that gets passed in. In most use cases, this won't have any meaningful performance penalty, but it's still unnecessary work. And incidentally, the issue I ran into in #207 wouldn't have come up if `Field#serialize` could skip the call to `Field#get_value` where that value won't be used.

How about skipping the call to `get_value` when `_CHECK_ATTRIBUTE` is `False`? It seems like that would be more consistent with the naming of the flag than the current behavior, and it would save the extra call.

jmcarp:
For `Field` subclasses with `_CHECK_ATTRIBUTE` set to `False`, the call
to `Field#get_value` in `Field#serialize` is unnecessary because its
return value is never used. This patch skips the call to
`Field#get_value` in this case.

[Resolves #208]

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -211,13 +211,16 @@ class Field(FieldABC):
         :param callable accessor: Function used to pull values from ``obj``.
         :raise ValidationError: In case of formatting problem
         """
-        value = self.get_value(attr, obj, accessor=accessor)
-        if value is missing_ and self._CHECK_ATTRIBUTE:
-            if hasattr(self, 'default'):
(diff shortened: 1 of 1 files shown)


sloria:
Thanks!

Issue #116, opened 2014-12-27
Field validators as schema methods?


sloria:
Often, validators and preprocessors only apply to a single `Schema`, so it may make sense to define them within the `Schema`--rather than as separate functions--for better cohesion.

Method names could be passed to `__validators__`, et. al, like so:

``` python
class UserSchema(Schema):
    __validators__ = ['validate_schema']

    def validate_schema(self, data):
        # ...
```

philtay:
+1

This would be very nice to have in 2.0

taion:
Have you considered making this a `@property` instead of a class field? Then you could do e.g.

``` python
@property
def __validators__(self):
    return {
        '__SCHEMA__': self.validate_schema,
        'field_1': self.validate_field_1
    }
```

Instead of using the names of the validation functions.

taion:
Actually, you ought to be able to use decorators for this. Something like

``` python
def validates(field_name):
    def wrapper(fn):
        fn._marshmallow_validates = field_name
        return fn

    return wrapper


class SchemaBase(type):
    def __new__(mcs, name, bases, dct):
        for value in dct.itervalues():
            if hasattr(value, '_marshmallow_validates'):
                print value._marshmallow_validates, value


class Schema(object):
    __metaclass__ = SchemaBase

    @validates('bob')
    def validate_bob(self):
        print self
```

Or if you wanted to be fancy, returning a descriptor from the decorator and doing an `isinstance` check or something.

I think this has better ergonomics and is more consistent with the rest of the API, though.

sloria:
Good idea, @taion .

sloria:
Now that https://github.com/marshmallow-code/marshmallow/pull/191 is merged, we can move forward with the decorator-based validators.

sloria:
The `marshmallow.validator` decorator is implemented in https://github.com/marshmallow-code/marshmallow/commit/591729993b8e2b9e12316b6889b86b3b6206354b. 

See the "Schema-level validation" section of the docs for example usage: https://marshmallow.readthedocs.org/en/dev/extending.html#schema-level-validation. 

The actual API has a minor difference from @taion 's proposal above: there is no need to pass the field name to the `validator` decorator. This is because you can specify field names in the `ValidationError` constructor:

``` python
from marshmallow import Schema, fields, validator, ValidationError

class MySchema(Schema):
    bob = fields.Field()

    @validator
    def validate_bob(self, data):
        #...
[...]

PR #218, opened 2015-05-26
@validates

philtay:
tests + docs after initial review
#116

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -18,7 +18,8 @@ from marshmallow import base, fields, utils, class_registry, marshalling
 from marshmallow.compat import (with_metaclass, iteritems, text_type,
                                 binary_type, OrderedDict)
 from marshmallow.orderedset import OrderedSet
-from marshmallow.decorators import PRE_DUMP, POST_DUMP, PRE_LOAD, POST_LOAD, VALIDATES_SCHEMA
+from marshmallow.decorators import (PRE_DUMP, POST_DUMP, PRE_LOAD, POST_LOAD,
+                                    VALIDATES, VALIDATES_SCHEMA)
 
 
 #: Return type of :meth:`Schema.dump` including serialized data and errors
@@ -352,6 +353,7 @@ class BaseSchema(base.SchemaABC):
         self.extra = extra
         self.context = context or {}
         self._update_fields(many=many)
+        self._push_field_validators()
 
     def __repr__(self):
         return '<{ClassName}(many={self.many}, strict={self.strict})>'.format(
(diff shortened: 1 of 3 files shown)


marshmallow/schema.py
```diff
@@ -721,6 +723,21 @@ def __filter_fields(self, field_names, obj, many=False):
...
+
+            try:
+                field = self.fields[validator_kwargs['field']]
+            except KeyError:
+                raise ValueError(
+                    '"{}" field does not exist.'
+                    .format(validator_kwargs['field'])
+                )
+            else:
+                field.validators.append(validator)
```
sloria, line 739: Will this result in the same validator being appended multiple times if the Schema is instantiated multiple times?

philtay:
oops... two possible solutions:
1. use the `@validates` decorator to set an attribute on the decorated function and use this attribute to avoid appending the function multiple times.
2. iterate over `cls_fields` in `SchemaMeta.__new__` and append the decorated validators.

which one do you prefer? other alternatives?

sloria:
One possible alternative would be to use `Unmarshaller._run_validator`--as is done for schema validators. Not sure if this is the best solution, but it has an advantage in that it might allow us to validate implicitly-created fields (through the `fields` or `additional` class Meta option). 

This would require a slight modification of `_run_validator` to allow for explicit specification of which field to store an error message on.

philtay:
Uhm, is there any reason for `Field.validators` to be a `list` instead of a `set`? If not, I think using a `set` is the best solution.

sloria:
Field.validators is a list because the order of error messages may be significant in some use cases.

philtay:
Out of curiosity, can you describe a real-world use case where you need or rely on validation order? In any case: currently the invocation order of `@validates_schema` methods is not guaranteed. Have you any objection to the use of an `OrderedSet` instead of a `list` as the `Field.validators` type?

http://code.activestate.com/recipes/576694/

sloria:
The use case I've run into is when you want a deterministic order of error messages displayed to a user, e.g. display more important error messages before less important ones.

`OrderedSet` is already included in the marshmallow source. It could work, but I still think modifying `_run_validator` might be better if only because it would allow validation of implicitly-created (class Meta) fields.

philtay:
Ok, but I don't want to push even more logic in an already semi-spaghettish function. It looks like `_run_validator` is meant for schema-level validation only. I can rename it to `_validate_schema` and add another function `_validate_field`. I will use the latter to implement `@validates`. Ok?

sloria:
Go for it

philtay:
Please find below the method you should review before I open a new and clean PR.

``` python
def _invoke_field_validators(self, data, many):
    for attr_name in self.__processors__[(VALIDATES, False)]:
        validator = getattr(self, attr_name)
        validator_kwargs = validator.__marshmallow_kwargs__[(VALIDATES, False)]
        field_name = validator_kwargs['field']

        try:
            field_obj = self.fields[field_name]
        except KeyError:
            raise ValueError('"{}" field does not exist.'.format(field_name))

        if many:
            for idx, item in enumerate(data):
                try:
                    value = item[field_name]
                except KeyError:
                    pass
                else:
                    self._unmarshal.call_and_store(
                        getter_func=validator,
                        data=value,
                        field_name=field_name,
                        field_obj=field_obj,
                        index=(idx if self.opts.index_errors else None)
                    )
        else:
            try:
                value = data[field_name]
            except KeyError:
                pass
            else:
                self._unmarshal.call_and_store(
                    getter_func=validator,
                    data=value,
                    field_name=field_name,
                    field_obj=field_obj
                )
```

It's a method of the `BaseSchema` class and is invoked in `_do_load` just before `_invoke_validators` in this way:

``` python
self._invoke_field_validators(data=result, many=many)
```

Let me know. If it looks good I'll go ahead with a complete PR.

sloria:
@philtay Yes, that is a reasonable approach. Feel free to proceed

PR #220, opened 2015-06-06
Added Constant field type for static values.

kevinastone:
Allows to to specify static fields that will always be included.  Useful of specify static values, especially in sub-classes of Schemas.


sloria:
Thanks!

PR #223, opened 2015-06-14
Improve @validates tests

philtay:


philtay:
3 things you should do:
1. Mention `@validates` in the changelog
2. At least a pointer to `@validates` and `@validates_schema` [here](http://marshmallow.readthedocs.org/en/latest/quickstart.html#validation)
3. Release 2.0.0.b3 or, even better, directly 2.0. I think it's time.

PR #232, opened 2015-07-06
Bugfix/return field name or load from on validation errors

alexmorken:
Fixed bug where attribute (not field_name or load_from param) was being returned on errors when deserializing data, wrote tests and updated docs.

### DIFF
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -826,6 +826,32 @@ class TestSchemaDeserialization:
         assert result['email'] == 'foo@bar.com'
         assert result['age'] == 42
 
+    def test_deserialize_with_attribute_param_error_returns_field_name_not_attribute_name(self):
+        class AliasingUserSerializer(Schema):
+            username = fields.Email(attribute='email')
+            years = fields.Integer(attribute='age')
+        data = {
+            'username': 'foobar.com',
+            'years': '42'
+        }
+        result, errors = AliasingUserSerializer().load(data)
+        assert errors
+        assert errors['username'] == [u'"foobar.com" is not a valid email address.']
+
+    def test_deserialize_with_attribute_param_error_returns_load_from_not_attribute_name(self):
+        class AliasingUserSerializer(Schema):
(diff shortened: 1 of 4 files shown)


tests/test_deserialization.py
```diff
@@ -826,6 +826,32 @@ class AliasingUserSerializer(Schema):
...
+            name = fields.String(load_from='Name')
+            username = fields.Email(attribute='email', load_from='UserName')
+            years = fields.Integer(attribute='age', load_from='Years')
+        data = {
+            'Name': 'Mick',
+            'UserName': 'foobar.com',
+            'years': 'abc'
+        }
+        result, errors = AliasingUserSerializer().load(data)
+        assert errors['UserName'] == [u'"foobar.com" is not a valid email address.']
```
sloria, line 852: I'm not sure about this behavior: why should the key on errors change if the input value is missing vs. present? My expectation (as a user) would be that the key on errors is always the same as the name of the field.

alexmorken, line 852: Yes, I had the same concern, but decided to do it this way because it is
the actual field name the data was loaded from (being that it only gets set
if it can't load from the attribute or field name).

I am fine with only passing back the actual field name, but I do prefer the
clear reflection of what was actually happening when loading the data.

sloria, line 852: OK, I see now why you'd want the error keys to be the same as `load_from` because the field names could be considered internal aliases for the value passed by the client. However, I'm still not clear on why you would only want to use `load_from` when the value is missing from the input. It seems more consistent to always use `load_from` if it is set.

alexmorken, line 852: What I did was just follow the behavior of where the data is actually
loaded from.  I thought it would be easier for the client to introspect if
we handed back the name of the field they specified instead of the one
specified by load_from.  `load_from` only gets used as fallback and I
thought I would reflect that in the error handling.

sloria, line 852: Ah ok. Yes, that seems right. Thanks for the clarification!

PR #234, opened 2015-07-07
fixing set serialization issue

### LINKED ISSUE #233: fields.Nested does not support sets
Currently `fields.Nested` assumes that the value of the field is a list - https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/schema.py#L702 - and fails for `set` during serialization

traut:
fixes #233


sloria:
Thanks!

Issue #188, opened 2015-04-06
Nested field is not being validated correctly


juanrossi:
Hello, I'm using marshmallow as a validation and marshalling library. In some validation classes I use Nested field.

Part of my test suite tests that the validations are returning the corrected error when some unexpected data is sent. If I try to send a string to a Nested field I get an exception:

``` python
marshmallow/fields.pyc in deserialize(self, data, fields_dict, many, validators, preprocess, postprocess, strict, dict_class)
    283                     continue
    284                 key = fields_dict[attr_name].attribute or attr_name
--> 285                 raw_value = data.get(attr_name, missing)
    286                 if raw_value is missing and not field_obj.required:
    287                     continue

AttributeError: 'str' object has no attribute 'get'
```

The example code I'm using to replicate this error is:

``` python
from marshmallow import Schema, fields, validate


class ValidateClass(Schema):
    item = fields.Nested('NestClass', many=True, required=True)

class NestClass(Schema):
    item1 = fields.Integer(required=True, validate=validate.Range(min=2, max=24))
    item2 = fields.String(required=True, validate=validate.Length(min=0))


ValidateClass().load({'item': 'a'})
```

I also tried adding a validator to item field where I check that I'm receiving a valid list of objects, but it's still raising the exception.

Any thoughts? Thanks.

sloria:
Nested objects are expected to be a dictionaries; you are passing a string. The following would be valid according to your `ValidateClass`:

```
{'item': [{'item1': 2, 'item2': 'foo'}]}
```

juanrossi:
As a dictionary works great, but what I'm trying to do is break the validation. 

If someone sents me an invalid value I should handle it, shouldn't it receive an error similar to the one int returns if you send a string?

``` python
from marshmallow import Schema, fields

class Test(Schema):
    test_int = fields.Integer(required=True)

print Test().load({'test_int': 'value'}).errors

# Returns {'test_int': [u"invalid literal for int() with base 10: 'value'"]}
```

Maybe we should return something similar with an error in errors dict, what do you think?

sloria:
@juanrossi . I agree; a validation error should be raised if an invalid type is passed to a `Nested` field.

Pull requests welcome!

juanrossi:
:+1:

sloria:
[...]

PR #235, opened 2015-07-07
Type error and missing required data error handling for nested fields.

max-orhai:
Although partially fixed already in issue #188, it is still possible for nested schemas to incorrectly deserialize wrongly-typed data. Moreover, the changes introduced by the fix for #188 generate an error message that is confusing in the case of a string in the place of a (`many=True`) list of nested schemas: the message claims that a dict (i.e. the nested schema) was expected when what really was expected was a _list_ of dicts. This fixes that, and still ensures test coverage for the `many=False` case when deserializing data with something other than the expected dict.

Also, Marshmallow currently doesn't generate an error when deserializing data with missing nested schema containing required fields (whether or not `many=True`). These changes fix that too.

### DIFF
--- a/marshmallow/__init__.py
+++ b/marshmallow/__init__.py
@@ -13,7 +13,7 @@ from marshmallow.decorators import (
 from marshmallow.utils import pprint, missing
 from marshmallow.exceptions import MarshallingError, UnmarshallingError, ValidationError
 
-__version__ = '2.0.0b4'
+__version__ = '2.0.0b4.idealist.2'
 __author__ = 'Steven Loria'
 __license__ = 'MIT'
 
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -379,9 +379,9 @@ class Nested(Field):
         return ret
 
     def _deserialize(self, value):
-        if self.many and not isinstance(value, list):
+        if self.many and not utils.is_collection(value):
             raise ValidationError(
-                'Expected a list, got a {0}.'.format(value.__class__.__name__))
+                'Expected a collection of dicts, got a {0}.'.format(value.__class__.__name__))
 
         data, errors = self.schema.load(value)
         if errors:
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -418,7 +418,7 @@ class Nested(Field):
                     try:
                         field._validate_missing(field.missing)
                     except ValidationError as ve:
-                        errors[field_name] = [ve.message]
+                        errors[field_name] = ve.messages
         return errors
 
 
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -374,6 +374,13 @@ class Nested(Field):
         return ret
 
     def _deserialize(self, value):
+        if self.required is True and not value:
+            raise ValidationError(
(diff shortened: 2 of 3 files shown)


sloria:
I'm not quite sure this is the desired behavior. Marshmallow 2 eschews arbitrary empty values (see https://marshmallow.readthedocs.org/en/latest/upgrading.html#default-values); nested values are no exception. If the client provides a value (even an empty dict), it is not missing.

If you want to validate against empty input, I suggest using a validator instead.

``` python
def must_not_be_empty(val):
    if not val:
        raise ValidationError('Value cannot be empty')

class Inner(Schema):
    foo = fields.Field()

class Outer(Schema):
    inner = fields.Nested(Inner, validate=[must_not_be_empty])
```

max-orhai:
Thanks for the prompt response!

There are two issues that I was trying to address with this pull request, and neither are about validating _arbitrary_ empty values.

The first new conditional in Nested's `_deserialize` method is to ensure that `required=True` means the same thing for Nested schemas as for the other field types. Currently, if an inner nested schema contains fields marked as `required`, and it is itself `required` within an outer schema, neither schema generates any validation errors when deserializing an empty dict. Is this really the desired behavior?

Perhaps it would be more correct to check that the inner Nested schema does indeed have `required` fields, and raise exceptions on those specific required fields rather than the Nested field, allowing an empty dict if the inner schema doesn't have any `required` fields? I'd be happy to implement that if you'd like.

The second conditional is to correct a type error that currently can happen when deserializing an iterable (such as a string) using a schema that expects a list of nested schemas. The error message attached to the exception generated in this case currently says _(e.g.)_ `'Data must be a dict, got a str'`, because it is not generated until `_invoke_processors` is run, which pulls values from the iterable using a list comprehension. By checking a little earlier, we can provide a type-correct error message.

sloria:
Thanks for the clarification. Empty dicts should not be considered a special empty value, but I do think it would be useful to perform validation on inner required fields.

> Perhaps it would be more correct to check that the inner Nested schema does indeed have required fields, and raise exceptions on those specific required fields rather than the Nested field, allowing an empty dict if the inner schema doesn't have any required fields?

Yes, I think this is desirable.

sloria:
> The second conditional is to correct a type error that currently can happen when deserializing an iterable (such as a string) using a schema that expects a list of nested schemas

OK, but the solution should account for any collection type (sets, tuples, etc).

max-orhai:
Now recursively checking the schema of a required Nested field on deserialization if the field is not present in the input data, and storing errors for each required (non-Nested) field; the test cases should make the behavior clear. Notice that if `many=True` we store the nested error under a placeholder 0 key corresponding to the required list item.

max-orhai:
> OK, but the solution should account for any collection type (sets, tuples, etc).

Commit https://github.com/marshmallow-code/marshmallow/commit/abe114a15cb5195d9df79939f4ff3040099fa405 fixes that.

alexmorken:
Any update on getting this merged in?

sloria:
Sorry for the delay on this. I will try to get to this sometime this week or over the weekend.

sloria:
This looks good. Thanks for writing this and for your patience!.

PR #239, opened 2015-07-13
Fixed boolean field serialization.

stas:
I updated the Boolean field behavior to be as similar as possible with the de-serialization of the same field.

This breaks a test that verifies the field serialization of `None` to a `None`. I'm not sure what is the purpose of that, but it doesn't really make sense in the context of a boolean field.

Please help me understand if there's some backwards-compatibility or something with it, I will be happy to revert it back or fix the relevant test.

Thanks in advance for reviewing this.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -648,15 +648,20 @@ class Boolean(Field):
     :param kwargs: The same keyword arguments that :class:`Field` receives.
     """
 
-    #: Values that will deserialize to `True`. If an empty set, any non-falsy
+    #: Values that will (de)serialize to `True`. If an empty set, any non-falsy
     #  value will deserialize to `True`.
     truthy = set()
-    #: Values that will deserialize to `False`.
+    #: Values that will (de)serialize to `False`.
     falsy = set(['False', 'false', '0', 'null', 'None'])
 
     def _serialize(self, value, attr, obj):
         if value is None:
             return None
+        elif value in self.truthy:
+            return True
+        elif value in self.falsy:
+            return False
+
         return bool(value)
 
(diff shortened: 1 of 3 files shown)


sloria:
All "simple" fields (`fields.Str`, `fields.Int`, etc.) serialize `None` to `None` if `allow_none=True`. I'm not sure that `fields.Boolean` should be an exception, though I can see the argument for symmetry with deserialization.

I'l have to think further about this. Feedback welcome.

stas:
Hey @sloria, I don't want this to be a blocker. I reverted back the `allow_none` behavior and added an issue to be able to track this concern and get more feedback #241.

Let me know if you need anything else in order to merge this.

stas:
@sloria I will be forever grateful if you could make some time and review this again. :bow:

marshmallow/fields.py
```diff
@@ -607,6 +607,9 @@ class Boolean(Field):
     def _serialize(self, value, attr, obj):
         if value is None:
             return None
+        elif value in self.falsy:
```
sloria: For consistency, we should also check `if value in self.truthy` as well.

sloria: Also, can you update the comments on ~L601 and ~L604 to reflect this change? Should just be a matter of changing "deserialize" to "(de)serialize".

sloria:
Just a few changes, and this should be good to merge. :+1: 

Also, don't forget to add yourself to AUTHORS.rst

stas:
Thanks @sloria 
I added your suggestions squashed and rebased the changes.

sloria:
Thanks!

PR #242, opened 2015-07-15
Updated utils._get_value_for_key return the called attribute if it is callable

alexmorken:
I believe this should be the behavior expected.  While most things can be made an @property, sometimes they aren't and it is great to be able to get the result back.


sloria:
I agree. Thanks!

Issue #240, opened 2015-07-13
Decimal objects are not serialized correctly with implicit field creation


seedifferently:
Steps to reproduce (I tested on version 1.2.6):

```
from marshmallow import Schema
from datetime import datetime
from decimal import Decimal

obj = {'int': 1, 'decimal': Decimal('1.23'), 'datetime': datetime.now()}

class MySchema(Schema):
    class Meta:
        fields = ['int', 'decimal', 'datetime']

json = MySchema()
json.dump(obj).data
```

The resulted object does not properly serialize the decimal.

seedifferently:
Closing after reading the warning about the decimal field: https://marshmallow.readthedocs.org/en/latest/api_reference.html?highlight=function#marshmallow.fields.Decimal

PR #250, opened 2015-07-27
setuptools extras

philtay:
1. recommended dependencies are now listed in a single standard place
2. users can easily install marshmallow + recommended dependencies: `pip install marshmallow[reco]`
3. you can remove the "soft dependencies" from dev-requirements.txt
4. 100% backward compatible
5. if properly documented can help to prevent problems like this #240

Refs:
- http://packages.python.org/setuptools/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies
- https://pip.pypa.io/en/latest/reference/pip_install.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ python:
 
 install:
   - pip install -U -r dev-requirements.txt
-  - pip install -U .
+  - pip install -U .[reco]
 
 before_script:
   - flake8 .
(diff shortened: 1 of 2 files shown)


philtay:
ah, travis ci uses an old pip version. it fails for this reason.

https://github.com/pypa/pip/issues/1236

the workaround is to upgrade pip before everything else in .travis.yml: `pip install --upgrade pip`

if you're interested in this PR i'll fix the problem

sloria:
Yes, this looks good. Thanks! Once the Travis build is passing, this is good to merge.

philtay:
done

PR #252, opened 2015-07-28
DateTime uses strptime during deserialization

bartaelterman:
Note that this fix will solely rely on the DateTime format given. No deserialization attempt is made using dateutil.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -710,7 +710,8 @@ class DateTime(Field):
     Example: ``'2014-12-22T03:12:58.019077+00:00'``
 
     :param str format: Either ``"rfc"`` (for RFC822), ``"iso"`` (for ISO8601),
-        or a date format string. If `None`, defaults to "iso".
+        or a date format string. If `None`, "iso" is used for serialization while deserialization will be attempted with
+        dateutil.
     :param kwargs: The same keyword arguments that :class:`Field` receives.
 
     """
@@ -762,18 +763,26 @@ class DateTime(Field):
         err = ValidationError(getattr(self, 'error', None) or msg)
         if not value:  # Falsy values, e.g. '', None, [] are not valid
             raise err
-        self.dateformat = self.dateformat or self.DEFAULT_FORMAT
         func = self.DATEFORMAT_DESERIALIZATION_FUNCS.get(self.dateformat)
         if func:
             try:
                 return func(value)
             except (TypeError, AttributeError, ValueError):
                 raise err
-        else:
+        elif type(value) is str:
             try:
                 return dt.datetime(*time.strptime(value, self.dateformat)[0:6])
             except TypeError:
                 raise err
+        elif utils.dateutil_available:
+            try:
+                return utils.from_datestring(value)
+            except TypeError:
+                raise err
+        else:
+            warnings.warn('It is recommended that you install python-dateutil '
+                          'for improved datetime deserialization.')
+            raise err
 
 
 class LocalDateTime(DateTime):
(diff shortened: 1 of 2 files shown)


sloria:
It'd be nice to have the option of using dateutil for parsing. Perhaps we should only use strptime if `dateformat` is a string that isn't one of the keys of `DATETIME_DESERIALIZATION_FUNCS`.

bartaelterman:
That is what happens now. The check is still present [here](https://github.com/bartaelterman/marshmallow/blob/dev/marshmallow/fields.py#L766). However, I think what we really want is (in pseudo code):

```
if value is in DATETIME_DESERIALIZATION_FUNCS:
    parse with that function
elif value is a valid strptime time formatting string:
    parse with strptime
else
    parse with dateutil
```

However:
1. From the API perspective, I think that results in weird behaviour. Either the user sets the `format` parameter to `rfc` or `iso` (which will result in parsing with the function in `DATETIME_DESERIALIZATION_FUNCS`), or he gives a time formatting string, which will result in parsing with `strptime`. If the parameter is not set, `iso` is used as default. So from a user perspective, to what value should I set the `format` parameter to have dateutil do the parsing? Some random string not being `iso`, `rfc` or a time formatting string and not `None`? Maybe the best option would then be to override the behaviour in case the parameter is `None`.
2. I'm not sure how to check whether the given string is a valid strptime formatting string anyway.

So I agree that parsing with dateutil could be useful, but we should give it a proper place in the API.

sloria:
I think it makes sense to parse with dateutil if `format=None`. So the pseudocode becomes:

```
if value is in DATETIME_DESERIALIZATION_FUNCS:
    parse with that function
elif value is a string:
    parse with strptime
else
    parse with dateutil
```

The default `format` would be changed to `None`.

bartaelterman:
Ok, I'll change it.

sloria:
@bartaelterman Can you look into the failing tests?

bartaelterman:
Will do so asap.

bartaelterman:
Fixed two tests. Three yet to fix.

marshmallow/fields.py
```diff
@@ -803,17 +805,21 @@ def _deserialize(self, value):
...
-        self.dateformat = self.dateformat or self.DEFAULT_FORMAT
         func = self.DATEFORMAT_DESERIALIZATION_FUNCS.get(self.dateformat)
         if func:
             try:
                 return func(value)
             except (TypeError, AttributeError, ValueError):
                 raise err
+        elif type(value) is str and self.dateformat:
+            try:
+                return dt.datetime(*time.strptime(value, self.dateformat)[0:6])
```
sloria, line 816: Why not use `dt.datetime.strptime`?

marshmallow/fields.py
```diff
@@ -803,17 +805,21 @@ def _deserialize(self, value):
...
         if not value:  # Falsy values, e.g. '', None, [] are not valid
             raise err
-        self.dateformat = self.dateformat or self.DEFAULT_FORMAT
         func = self.DATEFORMAT_DESERIALIZATION_FUNCS.get(self.dateformat)
         if func:
             try:
                 return func(value)
             except (TypeError, AttributeError, ValueError):
                 raise err
+        elif type(value) is str and self.dateformat:
```
sloria, line 814: I'm not sure that the typechecking is necessary. We can assume that `self.dateformat` is either a string or `None`.

justanr, line 814: Especially because of the difference between `""` and `u""` in Python 2.

sloria:
It appears that https://github.com/marshmallow-code/marshmallow/pull/262 attempts to solve the same issue.

@bartaelterman Would that PR solve the issue effectively? Is there anything present in this approach that is missing in #262 ?

bartaelterman:
Nope, that's exactly the same thing.

Issue #221, opened 2015-06-10
Email address validation fails when mailbox name contains non-ASCII characters


lextoumbourou:
Hi there,

It appears email address validation fails when non-ASCII characters are present in the mailbox name, which should be permitted as per [RFC 6530](http://en.wikipedia.org/wiki/Email_address#Internationalization). See example (tested on `dev` branch):

```
In [1]: from marshmallow import fields, Schema

In [2]: class Test(Schema):
   ...:     email = fields.Email()
   ...:

In [3]: t = Test()

In [4]: t.load({'email': u'josé@blah.com'})
Out[4]: UnmarshalResult(data={}, errors={'email': [u'"jos\xe9@blah.com" is not a valid email address.']})
```

Chars outside of ASCII seem to work fine in the domain though.

Thanks,

Lex

sloria:
Thanks for reporting @lextoumbourou . Pull requests welcome!

PR #256, opened 2015-08-06
Fix errors on non-ASCII local email part

mwstobo:
Add the `re.UNICODE` flag and allow for all word characters as opposed to `A-Z`

Fixes https://github.com/marshmallow-code/marshmallow/issues/221


sloria:
Thank you!

sloria:
I realize now that this may break users' code that relies on the `Email` validator _disallowing_ non-ASCII characters, so I'm going to revert this on the 1.2 line and cherry pick this into the 2.0 branch.

PR #260, opened 2015-08-09
Simplify USER_REGEX

philtay:


philtay:
A probably unnecessary clarification: `\w` already includes `0-9` and `_`, thus specifying them again is redundant.

sloria:
Indeed. Thanks

Issue #172, opened 2015-03-16
Deserialize field that depends on other field?


alexmic:
Hey,

How can I deserialize a field which depends on the value of another field? For example, imagine this input: 

``` json
{
    "currency": "USD",
    "amount": 1000
}
```

My application looks at the amount and constructs an appropriate `Money` object which translates the amount (in cents) to a `Decimal` according to the currency (different currencies have different decimal digits). I'm failing to see how I can do this since `deserialize()` only gets the value. Is there a reason for not providing the actual object to be deserialized as well?

sloria:
You could use the `context` of the schema:

``` python
schema.context['currency'] = input_data['currency']

class MoneyField(fields.Field):
    def _deserialize(self, val):
        currency = self.context.get('currency', 'USD')
        # ... deserialize based on currency
```

@alexmic Does this meet your use case?

kevinastone:
Same issue (in fact, amount + currency).  Why doesn't `_deserialize` also pass back the `attr`and `obj`?

sloria:
@alexmic @kevinastone I've added the `attr` and `data` parameters to `_deserialize` on the [`deserialize-signature`](https://github.com/marshmallow-code/marshmallow/tree/deserialize-signature) branch. 

Would love if you tried it out to see if it meets your use case.

alexmic:
Works like a charm, thanks!  :star:

sloria:
Thank you for the suggestion. Merged.

PR #264, opened 2015-08-25
_deserialize receives attr and input data

sloria:
Passing `attr` and `data` allows custom fields to deserialize values based on other values in the input data.

see #172


sloria:
Merged into `dev`.

PR #266, opened 2015-08-28
Unmarshaller: ask for attribute-key on demand

floqqi:
I need this because I created a custom field which accepts multiple input values in `_deserialize`. Based on the input value `self.attribute` maybe changes.

With this change the key is request after the call of `my_custom_field.deserialize()`.

Example:
In my REST-API I want to support both foreign keys and nested objects. If someone would like to pass a foreign key instead of a nested object I need to set `field_id` instead of `field` (in my SQLAlchemy-model). But both should be setable (in the API) using `field`.


sloria:
A fine change. Thanks!

Issue #195, opened 2015-04-13
Serializing a mixed list


kishinmanglani:
I have a list with two types of objects A and B with two different Schemas, ASchema and BSchema. I want to serialize this list, so that depending on the type of object it serializes it using the correct schema. Is this possible?

sloria:
@kishinmanglani You could define a `Field` that is able to handle both types `A` and `B` and pass that to the `List` field.

justanr:
@kishinmanglani I ran into this problem a while back while trying to handle serializing an object that could have two different kind of "owners". Luckily, @sloria provided a base recipe in a [similar issue](https://github.com/marshmallow-code/marshmallow/issues/42#issuecomment-54719736).

I have a [spin on it](https://github.com/justanr/owa/blob/dbaefee80d812a676c3d172818783552559caae7/owa/utils/schema.py#L79) in one of my projects, too. [Example usage](https://github.com/justanr/owa/blob/dbaefee80d812a676c3d172818783552559caae7/owa/schemas.py#L40)

sloria:
Support for polymorphic fields won't be added to marshmallow core. See the implementations suggested above, as well as [marshmallow-polyfield](https://github.com/Bachmann1234/marshmallow-polyfield).

ramnes:
For anyone out there that would like something simpler than marshmallow-polyfield, I overloaded `Nested` so that it also accept a list of schemas and iterates through them to check the data until it find a corresponding schema. It's kind of dumb (and almost surely very bad on performances), but it does the job.

https://gist.github.com/ramnes/89245fbd9f2dfff52a78

jkryanchou:
Awesome Job I have do a lot research on the issue. Thanks all

sampart:
I had a similar problem and solved it using the [Function](https://marshmallow.readthedocs.io/en/3.0/api_reference.html#marshmallow.fields.Function) field type and creating appropriate serialisation/deserialisation methods which checked the types.  Thought it worth mentioning in case it helps anyone - it's simpler than making your own field subclass.

jkryanchou:
Cool

kathur2rain:
finally I solved this problem by override fields.nested.  but I just find here's a, plugin? called marshmallow-oneofschema. which just might help.

sloria:
[...]

Issue #42, opened 2014-09-05
How to handle (polymorphic) subclasses


ghost:
I'm using SQLAlchemy's [polymorphic identities](http://docs.sqlalchemy.org/en/latest/orm/inheritance.html) and have been trying to figure out how to get the `UserMarshal` to use the `BusinessProfileMarshal` if the `Profile` attached to the `User` is actually a `BusinessProfile`

```
class User(db.Model):
    profile = db.relationship('Profile', backref='users')

class Profile(db.Model):

    __mapper_args__ = {
        'polymorphic_identity': 'profile',
        'polymorphic_on': type
    }
    ...

class BusinessProfile(Profile):

    __tablename__ = 'profile_business'
    __mapper_args__ = {
        'polymorphic_identity': 'business',
    }
    ...


class UserMarshal(ma.Serializer):
    class Meta:
        fields = (
            'email',
            'profile',
        )

    profile = fields.Nested(ProfileMarshal)


class ProfileMarshal(ma.Serializer):
    class Meta:
        fields = (
            'first_name',
        )

class BusinessProfileMarshal(ma.Serializer):
    class Meta:
        fields = (
            'first_name',
            'company_name',
        )
```

sloria:
How about a custom `Polymorphic` field that allows you to pass a `dict` mapping class names to their corresponding `Serializer`?

``` python
class UserMarshal(ma.Serializer):
    class Meta:
        fields = (
            'email',
            'profile',
        )

    profile = Polymorphic({
        'BusinessProfile': BusinessProfileMarshal,
        'Profile': ProfileMarshal,
    }, default_serializer=ProfileMarshal)
```

The `Polymorphic` field could be implemented like so:

``` python
class Polymorphic(fields.Raw):

    def __init__(self, mapping, default_serializer,
                 many=False, only=tuple(), exclude=tuple(), **kwargs):
        fields.Raw.__init__(self, **kwargs)
        self.mapping = mapping
        self.default_serializer = default_serializer
        self.many = many
        self.only = only
        self.exclude = exclude

    def output(self, key, obj):
        nested_obj = self.get_value(key, obj)
        nested_obj_cls = nested_obj.__class__.__name__      
        serializer_cls = self.mapping.get(nested_obj_cls, self.default_serializer)
        ser = serializer_cls(nested_obj, many=self.many, exclude=self.exclude,
                             only=self.only)
        return ser.data
```

ghost:
[...]

PR #270, opened 2015-09-06
Add support for polymorphic fields

Bachmann1234:
Howdy!

I use marshmallow to communicate with a service using polymorphism. One of the fields is basically "anything that is a subtype of X" I was not sure how to implement this with what was there so I wrote this field.

I thought I would offer it to you in case you wanted it. 

If you want to support the use case but have an idea for a better design (I find the disambiguation function a tad clunky but a better way of disambiguating did not come to me...) I would gladly implement it.

If you like it ill write up some docs for it. Otherwise close and forget :-)

Thanks for this library btw. Really helped me out.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -1348,6 +1348,66 @@ class Constant(Field):
         return self.constant
 
 
+class PolyField(Field):
+    """
+    A field that (de)serializes to one of many types. a function
+    is called to disambiguate what schema to use for the (de)serialization
+    Intended to assist in working with fields that can contain any subclass
+    of a base type
+    """
+    def __init__(self, disambiguation_function, many=False, **metadata):
+        """
+        :param disambiguation_function: Function that takes in either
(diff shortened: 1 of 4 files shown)


sloria:
Thanks @Bachmann1234 . This is some good work.

There are many ways to implement polymorphic fields--see https://github.com/marshmallow-code/marshmallow/issues/42 and .https://github.com/marshmallow-code/marshmallow/issues/195, for examples--and I'm not quite set on any one. I believe it makes sense to have this in its own library.

Bachmann1234:
Oo, thanks for the examples. Sounds good. I'll take a look and package this up is anyone out there needs this.

Bachmann1234:
Went ahead and packaged it up
https://github.com/Bachmann1234/marshmallow-polyfield

https://pypi.python.org/pypi/marshmallow-polyfield

It should work if anyone out there finds this.

sloria:
Very nice!

Issue #273, opened 2015-09-11
Allow arbitrary field names in ValidationError


evgeny-sureev:
I tried to modify example from docs to use missing field name instead of default "_schema", but got an KeyError. 

``` python
class MySchema(Schema):
    foo = fields.Int()
    bar = fields.Int()

    @validates_schema(pass_original=True)
    def check_unknown_fields(self, data, original_data):
        for key in original_data:
            if key not in self.fields:
                raise ValidationError('Unknown field name', key)

schema = MySchema()
result, errors = schema.load({'foo': 1, 'bar': 2, 'baz': 3})  # throws KeyError in _run_validator
errors  # this is what I am trying to achieve: {'baz': ['Unknown field name']}
```

sloria:
Yes, I think it would be good to allow any key name to be passed and stored. Pull requests welcome!

evgeny-sureev:
Ok, I made PR #275

sloria:
Thanks again!

PR #275, opened 2015-09-14
Allow arbitrary field names in ValidationError #273

evgeny-sureev:


marshmallow/marshalling.py
```diff
@@ -174,7 +174,7 @@ def _run_validator(self, validator_func, output,
             # Store or reraise errors
             if err.field_names:
                 field_names = err.field_names
-                field_objs = [fields_dict[each] for each in field_names]
+                field_objs = [fields_dict[each] if each in fields_dict else None for each in field_names]
```
sloria: We shouldn't put `None` in this list if a field name isn't in `fields_dict`. You might need to do an old-fashioned for-loop to fix this.

sloria:
Once the minor change is made and the Travis CI build is passing, this is good to merge.

Thanks for your help!

marshmallow/marshalling.py
evgeny-sureev: This would put `fields_names` and `fields_objs` out of sync.

sloria: Ah, fair enough. I do see a use case for keeping the lists the same order and length. OK, this is fine.

Could you fix this so that it passes the flake8 test?

evgeny-sureev:
Done.

PR #279, opened 2015-09-17
Nested Field _check_required method bug formatting errors incorrectly

alexmorken:
Fixed recursive function to check and format errors for missing required, many nested fields; Fixed test in test_schema to reflect change


sloria:
Thanks!

PR #280, opened 2015-09-20
Prevent infinite loop when recursive required field is missing

Bachmann1234:
Got bit by this one when testing my error flows

When trying to get all the errors in infinitely recurses and explodes

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -50,6 +50,7 @@ MISSING_ERROR_MESSAGE = (
     'ValidationError raised by `{class_name}`, but error key `{key}` does '
     'not exist in the `error_messages` dictionary.'
 )
+_RECURSIVE_NESTED = 'self'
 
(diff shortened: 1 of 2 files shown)


marshmallow/fields.py
```diff
@@ -440,6 +441,8 @@ def _validate_missing(self, value):
         `value` should be considered missing.
         """
         if value is missing_ and hasattr(self, 'required'):
+            if self.nested == _RECURSIVE_NESTED:
+                raise ValidationError(['Missing data for required field.'])
```
sloria: Use `self.fail('required')` here, so that the error message can be configured.

Bachmann1234: Oh, did not notice that. Fixed!

sloria:
Thanks!

Bachmann1234:
w00t! No problem

PR #282, opened 2015-09-21
Fix another infinite loop

Bachmann1234:
This prevents another infinite loop when dealing with recursive fields.
The solution leads to recursing one level too deep but the answer is
not wrong, just a bit more verbose then necessary.

This fix breaks us out of the loop. However, the result is perhaps a bit misleading 

https://github.com/marshmallow-code/marshmallow/compare/dev...Bachmann1234:another-infinite-loop?expand=1#diff-e1bc6e780c4b78fd3a4a85bf7b8ae0acR1274

```
'basic': {'sub_basics': [u'Missing data for required field.']}
```

Is probably more correct but I did not see a way of doing that without breaking the existing logic as it would require me not to recurse down nested objects if they are missing.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -451,7 +451,11 @@ class Nested(Field):
             for field_name, field in self.schema.fields.items():
                 if not field.required:
                     continue
-                if isinstance(field, Nested):
+                if (
+                    isinstance(field, Nested) and
+                    self.nested != _RECURSIVE_NESTED and
+                    field.nested != _RECURSIVE_NESTED
+                ):
                     errors[field_name] = field._check_required()
                 else:
                     try:
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -1262,6 +1262,46 @@ class TestSelfReference:
             'sub_basics': ['Missing data for required field.']
         }
 
+    def test_recursive_missing_required_field_one_level_in(self):
+        class BasicSchema(Schema):
+            sub_basics = fields.Nested("self", required=True, exclude=('sub_basics', ))
+            simple_field = fields.Str(required=True)
+
+        class DeepSchema(Schema):
+            basic = fields.Nested(BasicSchema(), required=True)
+
+        data, errors = DeepSchema().load({})
+        assert data == {}
+
+        assert errors == {
+            'basic': {
+                'sub_basics':  [u'Missing data for required field.'] ,
+                'simple_field': [u'Missing data for required field.'],
+            }
+        }
+
+        partially_valid = {
+            'basic': {'sub_basics': {'simple_field': 'foo'}}
+        }
+        data, errors = DeepSchema().load(partially_valid)
+        assert data == partially_valid
+        assert errors == {
+            'basic': {
+                'simple_field': [u'Missing data for required field.'],
+            }
+        }
+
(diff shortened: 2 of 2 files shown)


sloria:
Thanks again for the patch.

I think we get the desired result if we also check `field.nested != _RECURSIVE_NESTED`:

``` python
    def _check_required(self, value):
         #...
            for field_name, field in self.schema.fields.items():
                #....
                if (
                    isinstance(field, Nested) and
                    self.nested != _RECURSIVE_NESTED and
                    field.nested != _RECURSIVE_NESTED  # Added to your patch
                ):
                    errors[field_name] = field._check_required(value)
#...
```

When I expand your test case to

``` python
    def test_recursive_missing_required_field_one_level_in(self):
        class BasicSchema(Schema):
            sub_basics = fields.Nested("self", required=True, exclude=('sub_basics', ))
            simple_field = fields.Str(required=True)

        class DeepSchema(Schema):
            basic = fields.Nested(BasicSchema(), required=True)

        data, errors = DeepSchema().load({})
        assert data == {}

        assert errors == {
            'basic': {
                'sub_basics':  [u'Missing data for required field.'] ,
                'simple_field': [u'Missing data for required field.'],
            }
        }

        partially_valid = {
            'basic': {'sub_basics': {'simple_field': 'foo'}}
        }
        data, errors = DeepSchema().load(partially_valid)
        assert data == partially_valid
        assert errors == {
            'basic': {
                'simple_field': [u'Missing data for required field.'],
            }
        }

        partially_valid2 = {
            'basic': {'simple_field': 'foo'}
        }
        data, errors = DeepSchema().load(partially_valid2)
        assert data == partially_valid2
        assert errors == {
            'basic': {
                'sub_basics': ['Missing data for required field.'],
            }
        }
```

The test passes.

Are there other edge cases I'm missing here?

Bachmann1234:
@sloria Ill give that a shot

For other edge cases I think this covers all the ones i'm gonna hit anytime soon. Sorry I did not combine these but honestly I made the patches as the issues came up. Im dealing with a series of highly recursive objects right now. 

With this patch the project im working on seems to work. But it will be getting more testing in the coming weeks.

sloria:
No need to apologize! I'm just glad that you caught the bug.

I thiink the modification above gives the correct output. Note that I changed the test to make the self-nested field not recurse

``` python
sub_basics = fields.Nested("self", required=True, exclude=('sub_basics', ))
```

I'm not sure that a self-nested field that includes itself should ever be `required`; it would never be valid. Perhaps marshmallow should prevent the user from making this mistake.

Bachmann1234:
Fair enough. I think in my case I had it required as it was a list so I required an empty list to be in there. Though I guess I could have simply had the post_process convert a missing value to an empty list

sloria:
Could you set `missing=list` on the field?

Bachmann1234:
@sloria ah, that would probably work. Still new to the library :-P

Bachmann1234:
I applied you changes and squashed it in. As for disallowing the impossible to fulfill schema that seems like a separate patch. 

I like the idea, some high level metaschema :-P Basically applying logic and throwing errors quickly when schemas are invalid. Though I suspect it can get hairy.

sloria:
Thanks a lot!

Issue #251, opened 2015-07-28
requesting standard support of Dict fields type


Dowwie:
Please accept this request to adopt native support of a dictionary field type

sloria:
What would a `Dict` field support that isn't already supported by a `Nested` field?

Dowwie:
Nested is sufficient for dicts whose schema is known in advance of use, but what about cases where the elements to serialize aren't known in advance?

imhoffd:
Sounds like `Dict` would just be `Raw` with a type check.

mcandre:
+1

sloria:
I would be open to a PR adding this. So would the interface be similar to DRF, where you can pass an inner field for validating values?

``` python
foo = fields.Dict(fields.Int())
```

imhoffd:
But, as @Dowwie said, there are cases where the schema of the elements inside the dict aren't known in advance. I could take a look at creating a PR.

sloria:
The Dict field has been added on `dev`.

PR #286, opened 2015-09-26
Dict

imhoffd:
Should provide what is described here: https://github.com/marshmallow-code/marshmallow/issues/251

The test base for marshmallow is huge and awesome, but I found it slightly overwhelming. If there are other tests needed, let me know. I think I covered everything, but there may be integrations or something I am missing.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,8 +1,8 @@
 Changelog
 ---------
 
-2.1.0 (date pending)
-++++++++++++++++++++
+2.1.0 (unreleased)
+++++++++++++++++++
 
 Features:
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,13 @@
 Changelog
 ---------
 
+2.1.0 (date pending)
+++++++++++++++++++++
+
+Features:
+
+- Add ``Dict`` field for arbitrary mapping data (:issue:`251`).
+
 2.0.0 (2015-09-25)
 ++++++++++++++++++
 
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -12,6 +12,7 @@ twine
 # Testing
 pytest>=2.7.3
 tox>=1.5.0
+simplejson
 
 # Syntax checking
 flake8==2.4.1
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -10,7 +10,7 @@ wheel
 twine
 
 # Testing
-pytest>=2.7.3
+pytest==2.8.0
 tox>=1.5.0
 simplejson
 
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -3,6 +3,7 @@
 
 from __future__ import absolute_import, unicode_literals
 
+import collections
 import datetime as dt
 import uuid
 import warnings
@@ -1006,7 +1007,7 @@ class Dict(Field):
     }
 
     def _validated(self, value):
-        if utils.is_dict(value):
+        if isinstance(value, collections.Mapping):
             return value
         else:
             self.fail('invalid')
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -19,6 +19,7 @@ __all__ = [
     'Field',
     'Raw',
     'Nested',
+    'Dict',
     'List',
     'String',
     'UUID',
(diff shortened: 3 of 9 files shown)


marshmallow/utils.py
```diff
@@ -47,6 +47,13 @@ def is_generator(obj):
     return inspect.isgeneratorfunction(obj) or inspect.isgenerator(obj)
 
 
+def is_dict(obj):
```
sloria: This function probably isn't necessary.

dev-requirements.txt
```diff
@@ -12,6 +12,7 @@ twine
 # Testing
 pytest>=2.7.3
 tox>=1.5.0
+simplejson
```
sloria, line 15: Good catch.

sloria:
Please add `fields.Dict` to `tests.base.ALL_FIELDS` so that it will be included in the parametrized tests.

sloria:
There may be use cases where a user might want to validate the values in a Dict field, in which case we could adapt from DRF's `DictField`,

``` python
various_data = fields.Dict(fields.Str())
```

If you don't pass an inner field, there would be no additional validation of values.

That said, I'm inclined toward YAGNI; let's stick with the simple approach you have here. :+1: 

Tests look good; thank you for being thorough. Just a few minor changes noted above, and this is good to merge.

CHANGELOG.rst
```diff
@@ -1,6 +1,13 @@
 Changelog
 ---------
 
+2.1.0 (date pending)
```
sloria: By convention, this should be  `2.1.0 (unreleased)`.

imhoffd:
:+1: 

I think for validation of values within the dict, where the schema of the inner fields is at all known, the user should consider using `fields.Nested` with an additional `Schema`.

The use cases for `fields.Dict` are hopefully rare. My particular use case is I am using marshmallow for marshalling responses in a web service that offers a presigned file upload URL from S3. S3 says "take this JSON and plug it in to your POST values with this URL." And so I basically want a schemaless pass-through of those values in my response. I am using `fields.Raw`, which does the trick, of course, but since I know I am expecting a dict, but don't care what's inside, a `fields.Dict` makes sense.

marshmallow/fields.py
```diff
@@ -994,6 +996,31 @@ def _deserialize(self, value, attr, data):
...
+    def _validated(self, value):
+        if isinstance(value, collections.Mapping):
+            return value
+        else:
+            self.fail('invalid')
+
+    def _serialize(self, value, attr, obj):
+        if value is None:
+            return None
+        return self._validated(value)
```
sloria: I think we can remove validation from `_serialize`. Data to be serialized is assumed to be valid.

sloria:
I agree with you re: use cases for `Dict`. Just one last change suggested above, and this is good to go.

marshmallow/fields.py
imhoffd: I was modeling it after `fields.Number`, which validates the type during serialization. It will try to turn the thing into a float.

In fact I may want to `return dict(value)` here, to convert dict-like things into true Python dictionaries. Marshmallow is for converting complex data types to primitive Python data types. This would ease any further serialization, such as jsonifying it.

sloria: `Number` actually does formatting in its `_validated` method, but `Dict` does not. 

I don't think it's necessary to convert to a `dict`, I think it would be desirable for an `OrderedDict`, for example, to serialize to an `OrderedDict`.

marshmallow/fields.py
```diff
@@ -994,6 +996,31 @@ def _deserialize(self, value, attr, data):
...
 
 
+class Dict(Field):
+    """A dict field. Supports dicts and dict-like objects.
+
+    .. versionadded:: 2.1.0
+    """
+
+    default_error_messages = {
+        'invalid': 'Not a valid dict.'
```
sloria: We should try to avoid Python-specific language when possible. marshmallow is most commonly used for HTTP APIs, where clients may not "speak" Python. Maybe "Not a valid mapping type."

imhoffd:
I made those changes, tests are passing with `tox` but failing in travis.

sloria:
Travis is failing due to what appears to be a regression in the latest release of pytest. I've reported an issue here https://github.com/pytest-dev/pytest/issues/1086. For now, pinning to pytest==2.8.0 should fix the build

tests/test_schema.py
```diff
@@ -581,6 +582,11 @@ def test_invalid_url():
     assert 'homepage' in s.errors
     assert 'Not a valid URL.' in s.errors['homepage'][0]
 
+def test_invalid_dict_but_okay():
+    u = User('Joe', various_data='baddict')
```
imhoffd, line 586: @sloria Just clarifying that this is the functionality desired.

sloria, line 586: Yes, this is the expected behavior. In practice, this will not happen, because only data are expected to be validated _before_ serialization.

sloria:
Thanks again for this!

imhoffd:
You're welcome! Thanks for marshmallow--it has eased a lot of headaches. :+1:

Dowwie:
and thanks to you both for making this update happen :+1:

PR #287, opened 2015-09-30
Run all schema-level validators without stopping at first error

evgeny-sureev:
Thought it would be nice to perform all validations if there are many registered.


sloria:
Good catch! This is actually the intended behavior. A regression may have been introduced in  68d2dc6917d7a8191177e8ce4195235ee0d4720c. I will look at this more closely later today.

PR #293, opened 2015-10-07
Add support for partial loading

### LINKED ISSUE #290: "Partial" deserialization support
For implementing `PATCH` handlers on REST endpoints, it would be useful to have a concept of partial deserialization.

This would mean ignoring missing required fields and default values for missing fields.

I know this sounds a bit weird, but it matches a standard CRUD endpoint fairly well - `POST` or `PUT` to that endpoint should use the full validation w/r/t required fields or defaults, but `PATCH` is intended to apply a partial update and only modify what was actually changed.

I can handle this in userspace by catching `ValidationError`s for missing fields, but I don't think I can do the same for ignoring default field values.

Here's the equivalent API in DRF: http://www.django-rest-framework.org/api-guide/serializers/#partial-updates. From my POV I think any `partial` arg (if you think it makes sense) would best be positioned as a named argument on `load`, though.

taion:
Closes #290

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -303,6 +303,22 @@ Dictionaries or lists are also accepted as the custom error message, in case you
     #  'age': ['Age is required.'],
     #  'city': {'message': 'City required', 'code': 400}}
 
+Partial Loading
++++++++++++++++
+
+When using the same schema in multiple places, you may only want to check required fields some of the time when deserializing. You can ignore missing fields entirely by setting ``partial=True``.
+
+.. code-block:: python
+    :emphasize-lines: 5,6
+
+    class UserSchema(Schema):
+        name = fields.String(required=True)
+        age = fields.Integer(required=True)
+
+    data, errors = UserSchema().load({'age': 42}, partial=True)
+    # OR UserSchema(partial=True).load({'age': 42})
+    data, errors  # => ({'age': 42}, {})
+
 Schema.validate
 +++++++++++++++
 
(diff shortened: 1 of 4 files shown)


marshmallow/schema.py
```diff
@@ -540,13 +545,20 @@ def loads(self, json_data, many=None, *args, **kwargs):
...
+            value for `self.partial` is used.
         :return: A tuple of the form (``data``, ``errors``)
         :rtype: `UnmarshalResult`, a `collections.namedtuple`
 
         .. versionadded:: 1.0.0
         """
+        # TODO: This avoids breaking backward compatibility if people were
+        # passing in positional args after `many` for use by `json.loads`, but
+        # ideally we shouldn't have to do this.
+        partial = kwargs.pop('partial', None)
```
taion, line 558: This is ugly. I'd prefer not to do this.

sloria, line 558: This is ok for now. It doesn't significantly reduce readability. We can remove it in marshmallow 3.

docs/quickstart.rst
```diff
@@ -303,6 +303,22 @@ Dictionaries or lists are also accepted as the custom error message, in case you
     #  'age': ['Age is required.'],
     #  'city': {'message': 'City required', 'code': 400}}
 
+Partial Loading
```
sloria, line 306: :+1:

sloria:
Very nice. Thanks!

Bala2105 (commented):
I reviewed PR #293 and it’s a foundational update for marshmallow that introduces the partial loading feature. This addition allows schemas to bypass the required=True validation on fields, which is essential for handling PATCH requests where only a subset of data is provided. The implementation correctly adds the partial parameter to the Schema constructor and load method, ensuring flexibility between persistent schema settings and one-off execution needs.

PR #305, opened 2015-10-21
Cleaner implementation of tag_processor

taion:
This is pure yak shaving, but I was writing another decorator with the same pattern, and I realized that I originally implemented this in a really awkward way.


sloria:
Thanks!

PR #311, opened 2015-10-28
pytest: don't recurse into virtualenv + other dirs

franciscod:
While developing, `invoke test` failed due to some tests inside virtualenv.

This should prevent pytest to run tests outside marshmallow source code.

https://pytest.org/latest/customize.html#confval-norecursedirs

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,3 +5,6 @@ universal = 1
 ignore = E127,E128,E265,E302,N803,N804,N806,E266,E731
 max-line-length = 100
 exclude = .git,.ropeproject,.tox,docs,.git,tests/,examples/,marshmallow/compat.py,marshmallow/ordereddict.py,build,setup.py,env,venv
+
+[pytest]
+norecursedirs = .git .ropeproject .tox docs env venv


sloria:
Thanks!

franciscod:
np! :dancer: :D

ShayanArmanPercolate:
So here's an issue that I just noticed.
https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/marshalling.py#L134

if field_obj.dump_to:
    key = field_obj.dump_to
elif not self.prefix:
    key = attr_name
else:
    key = ''".join([self.prefix, attr_name])

What if there's a case where field_obj.dump_to is set AND self.prefix is set. Then, do we want key = ''.join([self.prefix, field_obj.dump_to])?

so maybe:

key = "" if not self.prefix else self.prefix
if field_obj.dump_to:
    key += field_obj.dump_to
else:
    key += attr_name

What do you think?
@sloria @franciscod

sloria:
@ShayanArmanPercolate Good catch. Yes, I think the prefix should be included if it is set. Anyone care to send a PR for this?

sloria:
I've opened a new issue: https://github.com/marshmallow-code/marshmallow/issues/317 . We can continue discussion there.

PR #312, opened 2015-10-28
Dump to (implements #310

### LINKED ISSUE #310: Opposite functionality of 'load_from' needed. Perhaps 'dump_to'?
Currently, there is an attribute on the Field class called load_from. This is useful when a schema pythonic style fields, but the service could accept CamelCase field names.

ex.
class JobSchema(Schema):
    pipeline_id = fields.Str(load_from='PipelineId', load_only=True, required=True)

This is a great solution for deserialization. However, when serializing, it would be great to change the key from pipeline_id back to PipelineId. My current solution is to have a dump_only field that uses the attribute parameter to take the value of pipeline_id:

class JobSchema(Schema):
    pipeline_id = fields.Str(load_from='PipelineId', load_only=True, required=True)
    PipelineId = fields.Str(dump_only=True, attribute='pipeline_id')

Is there a better way to do this? 

Thanks a lot!
Shayan
@sloria

franciscod:
Hopefully fixes #310 :dancer:

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -375,6 +375,29 @@ By default `Schemas` will unmarshal an input dictionary to an output dictionary 
 
 .. _meta_options:
 
+
+Specifying Serialization Keys
+-------------------------------
+
+If you want to marshal a field to a different key than the field name you can use `dump_to`, which is analogous to `load_from`.
+
+.. code-block:: python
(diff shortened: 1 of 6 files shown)


sloria:
This looks good! Thanks.

franciscod:
:dancer:

Issue #307, opened 2015-10-23
Load Nested field with only=str gives error when sent flat data structure instead of dictionary


erlingbo:
When using a schema with the only attribute set to a string, the load method returns UnmarshalResult without errors nor data.

```
class MySchema(Schema):
    class Meta:
        fields = ('pk',)

    pk = fields.String()

MySchema(only='pk').load('an_id1234')
MySchema(only='pk').load({'pk': 'an_id1234'})
```

Both return the same UnmarshalResult
`UnmarshalResult(data={}, errors={})`

Preferably the first should work, but the second could be used with a pre load method on the schema class.

Is there anything I have done wrong or is this a bug?

Note: I use the same schema for dumping data an thus use `only='pk'` and not `only=('pk',)` for getting the correct dump output

franciscod:
Since `'pk'` may be interpreted as `('p', 'k')`, you should use `('pk',)`, right?

When using only=('pk',) it gives a more interesting result:

``` python
>>> MySchema(only=('pk',)).load('an_id1234')
UnmarshalResult(data={}, errors={'_schema': ['Invalid input type.']})
>>> MySchema(only=('pk',)).load({'pk': 'an_id1234'})
UnmarshalResult(data={'pk': 'an_id1234'}, errors={})
```

https://marshmallow.readthedocs.org/en/latest/api_reference.html#schema
The `only` attribute expects a tuple or a list, if you give it a string it will interpret it as an iterable of chars:

``` python
>>> class CharSchema(Schema):
...     class Meta:
...             fields = ('a', 'b', 'c', 'd')
...     a = fields.String()
...     b = fields.String()
...     c = fields.String()
...     d = fields.String()
... 
>>> CharSchema(only='ab').load({'a': 'asd', 'b': 'qwe'})
UnmarshalResult(data={'b': 'qwe', 'a': 'asd'}, errors={})

```

erlingbo:
My bad. I went to far when I tried to dumb down my problem. 

The real scenario is like this:

```
        class MyNestedSchema(Schema):
        class Meta:
            fields = ('pk',)

        pk = fields.String()

    class MySchema(Schema):
        class Meta:
            fields = ('pk', 'child')
        child = fields.Nested(MyNestedSchema, only='pk')

    class Obj(object):
        def __init__(self, pk, child):
            self.pk = pk
            self.child = child

    result = MySchema().dump(Obj(123, Obj(456, None)))
    print result
[...]

PR #314, opened 2015-10-29
Adding load support for only=basestring for Nested field

erlingbo:
Proposal for solving #307

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -428,6 +428,11 @@ class Nested(Field):
         if self.many and not utils.is_collection(value):
             self.fail('type', input=value, type=value.__class__.__name__)
 
+        if isinstance(self.only, basestring):  # self.only is a field name
+            if self.many:
+                value = [{self.only: v} for v in value]
+            else:
+                value = {self.only: value}
         data, errors = self.schema.load(value)
         if errors:
             raise ValidationError(errors, data=data)
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -791,6 +791,33 @@ class TestSchemaDeserialization:
         assert 'pets' in errors
         assert errors['pets'] == ['Missing data for required field.']
 
+    def test_nested_only_basestring(self):
+        class ANestedSchema(Schema):
+            pk = fields.Str()
+
+        class MainSchema(Schema):
+            pk = fields.Str()
+            child = fields.Nested(ANestedSchema, only='pk')
+
+        sch = MainSchema()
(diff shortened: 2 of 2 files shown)


sloria:
Thanks for the PR @erlingbo . I see the rationale for this change. This could, however, break backwards compatibility, so we'll have to consider whether the benefits are worth that cost. I'll have to think about this.

ghost:
I would like also to have this feature. 

It's most readable, JSON becomes most light and you manage the relationship in the server.

Choice 1:
`{
  "country": {
    "country_id": 1,
    "country": "France",
    "country_code": "FR"
  }
}
`
Choice 2:
`{
  "country": "FR"
}`

deckar01:
Serialization and deserialization should be symmetric operations. The fact that the `NestedField` serialization logic handles the `only` property, but the deserialization logic does not indicates that the implementation is asymmetric.

Deserializing data that Marshmallow itself created currently results in an error. Considering that the documentation only describes how the `only` option affects serialization, deserialization of a `Nested` field with the `only` option supplied as a string should be consider undocumented behavior.

I would argue that it can reasonably be assumed that serialization and deserialization are symmetric, therefore this change would be considered a bug fix warranting a patch bump. I can see an argument that defining undefined behavior could be considered a feature warranting a minor version bump. I would not consider this a breaking change.

I would suggest documenting the deserialization behavior in this PR:

``` py
        :emphasize-lines: 4, 11, 18

        class UserSchema(Schema):
            name = fields.String()
            email = fields.Email()
            friends = fields.Nested('self', only='name', many=True)
        # ... create ``user`` ...
        result, errors = UserSchema().dump(user)
        pprint(result)
        # {
        #     "name": "Steve",
        #     "email": "steve@example.com",
        #     "friends": ["Mike", "Joe"]
        # }
        data = UserSchema().load(result)
        pprint(data)
        # {
        #     "name": "Steve",
        #     "email": "steve@example.com",
        #     "friends": [{"name": "Mike"}, {"name": "Joe"}]
        # }
```

[docs/nesting.rst#L90](https://github.com/marshmallow-code/marshmallow/blob/d35151/docs/nesting.rst#L90)

sloria:
Let's move this forward for 3.0.0. All that's left to do is add documentation. I like @deckar01 's suggestion above.

@erlingbo Feel free to make the update (and add yourself to AUTHORS). Otherwise, I'll try to get to this over the weekend.

sloria:
This has been rebased and merged into `dev`. Thanks @erlingbo for the PR!

Issue #302, opened 2015-10-14
Allow callables for fields.Nested


taion:
Suggested earlier here: https://github.com/marshmallow-code/marshmallow/issues/19#issuecomment-43685498.

The suggestion would be to allow:

``` python
foo = fields.Nested(lambda: Foo)
```

in addition to

``` python
foo = fields.Nested('Foo')
```

Seems a bit cleaner in general.

sloria:
I'm -0 on this. I think allowing a callable is fine, but I don't think it justifies the cost of an extra typecheck and a redundant API.

taion:
Thinking about it a bit more, I probably wouldn't actually use this. This could potentially be slightly cleaner if there were name conflicts or something, because it doesn't require a global registry, but in practice it's just not that big a deal.

sloria:
Fair enough. FWIW, you can provide the full import path if there are name conflicts:

``` python
foo = fields.Nested('path.to.Foo')
foo2 = fields.Nested('path.to.other.Foo')
```

taion:
Yup - but seems a little cleaner to use whatever you have in scope. But in practice I'm just not seeing the use case I was imagining, and the `lambda:` just looks a bit uglier.

sloria:
Reopening this because I think this would enable valid use cases that are currently unmet.

(1) Passing arbitrary arguments to the nested Schema (only `many`, `exclude`, and `only` are currently supported. This could solve https://github.com/marshmallow-code/marshmallow/issues/285#issuecomment-152577723.

``` python
# Pass arbitrary arguments
foo = fields.Nested(lambda: FooSchema(partial=True))

# https://github.com/marshmallow-code/marshmallow/issues/285#issuecomment-152577723
class TreeSchema(Schema):
    tree = fields.Nested(lambda: TreeSchema(many=True, strict=True))
```

(2) With a slight modification to the OP's proposed API, users could easily implement a polymorphic nested field.

``` python
def make_schema(obj):
    if isinstance(obj, Foo):
        return FooSchema()
    elif isinstance(obj, Bar):
        return BarSchema()

class MySchema(Schema):
    poly = fields.Nested(make_schema)
```

Feedback welcome.

justanr:
Seems like a better twist on the polymorphic nested field. Would it be practical to apply this to `List` as well?

sloria:
@justanr Definitely a possibility.

justanr:
[...]

PR #318, opened 2015-11-02
Implements #302: Allow callable in fields.Nested

justanr:
This only implements allowing a callable to `fields.Nested`. I'm happy to squash commits into one if desired.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -51,3 +51,4 @@ Contributors (chronological)
 - `@Mise <https://github.com/Mise>`_
 - Taylor Edmiston `@tedmiston <https://github.com/tedmiston>`_
 - Francisco Demartino `@franciscod <https://github.com/franciscod>`_
+- Alec Reiter `@justanr <https://github.com/justanr>`_
--- a/docs/nesting.rst
+++ b/docs/nesting.rst
@@ -214,6 +214,42 @@ If the object to be marshalled has a relationship to an object of the same type,
     #     }
     # }
 
+Using A Callable To Nest
+------------------------
+
+It's also possible to use a callable to nest a `Schema`. The callable must accept
+the many, exclude, only, parent and context keywords and it must return an instance
+of `Schema` as well. The passed callable may be a function, method or object that 
+defines `__call__`.
+
+.. code-block:: python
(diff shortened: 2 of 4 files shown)


sloria:
Thanks @justanr .

I'm not sure about the signature for the callable. What if it just received the target object? Or perhaps the target object and `is_dump` (whether the target object is being serialized vs. deserialized) This would mean that the `Schema` would get instantiated with every (de)serialization, which is perhaps what a user would expect if passing a callable.

justanr:
Hmm. That could be something, I did notice without the re-check, at least one other test broke. There'd probably have to be changes made in dumps/loads. It'll probably be tomorrow before I can poke at it.

Specifically, [this test](https://github.com/justanr/marshmallow/blob/adc8605a67eacafeccd160078d32f5b3690971ef/tests/test_schema.py#L1195)

justanr:
@sloria I was just thinking this over, and would using `fields.Method` or `fields.Function` be a more appropriate way of going about this? It's already implemented, is called every time, can be nestled inside `fields.List`.

sloria:
Can you be more specific about how you would use `Method` or `Function`?

justanr:
They already receive the object being serialized as well as the context. If the expectation of passing a callable to Nested would be to generate a schema every time, Method and Function can do that. It seems like a better fit after thinking it over, at least from my POV.

Basically,

``` python
class MySchema(Schema):
     owner = Method('owner_serial', 'owner_deserial')

    def owner_serial(self, value):
        if isinstance(value, User):
            return UserSchema(...).dumps(value)
        else:
            return OtherSchema(...).dumps(value)
```

taion:
It seems like it'd be difficult in practice to handle both serialization and deserialization if the callable is receiving the nested object. Naively I'd have thought that it wouldn't receive any arguments at all, and just let the user take care of everything (and use context if necessary to pass anything down to the child schema).

sloria:
@justanr Thanks for the clarification. Even though using a Method or Function field is more verbose than the API proposed in the OP, it certainly meets the stated use cases well.

sloria:
I'm going to hold off on merging this for now. As @justanr points out, a `Method` field meets the use cases that this PR was targeting.

PR #321, opened 2015-11-05
Ensure prefix is included in key if both prefix and dump_to params are set

### LINKED ISSUE #317: `prefix` not respected when `dump_to` is set
Applies only to the latest `dev` version.

From https://github.com/marshmallow-code/marshmallow/pull/311#issuecomment-152202176:

> So here's an issue that I just noticed.
> https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/marshalling.py#L134
> 
> if field_obj.dump_to:
> key = field_obj.dump_to
> elif not self.prefix:
> key = attr_name
> else:
> key = ''".join([self.prefix, attr_name])
> 
> What if there's a case where field_obj.dump_to is set AND self.prefix is set. Then, do we want key = ''.join([self.prefix, field_obj.dump_to])?
> 
> so maybe:
> 
> key = "" if not self.prefix else self.prefix
> if field_obj.dump_to:
> key += field_obj.dump_to
> else:
> key += attr_name
> 
> What do you think?
> @sloria @franciscod

ewang:
Fixes #317

### DIFF
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -131,12 +131,9 @@ class Marshaller(ErrorStore):
         for attr_name, field_obj in iteritems(fields_dict):
             if getattr(field_obj, 'load_only', False):
                 continue
-            if field_obj.dump_to:
-                key = field_obj.dump_to
-            elif not self.prefix:
-                key = attr_name
-            else:
-                key = ''.join([self.prefix, attr_name])
+
+            key = ''.join([self.prefix or '', field_obj.dump_to or attr_name])
+
             getter = lambda d: field_obj.serialize(attr_name, d, accessor=accessor)
             value = self.call_and_store(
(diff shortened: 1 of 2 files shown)


marshmallow/marshalling.py
```diff
@@ -131,12 +131,9 @@ def serialize(self, obj, fields_dict, many=False,
...
             if getattr(field_obj, 'load_only', False):
                 continue
-            if field_obj.dump_to:
-                key = field_obj.dump_to
-            elif not self.prefix:
-                key = attr_name
-            else:
-                key = ''.join([self.prefix, attr_name])
+
+            key = ''.join([self.prefix or '', field_obj.dump_to or attr_name])
```
franciscod, line 135: awesome!

sloria:
Thanks!

PR #322, opened 2015-11-05
Issue-320: Ignore field validator if field is declared but not in "only"

eprikazc:


tests/test_schema.py
```diff
@@ -1003,6 +1003,19 @@ class MySchema(Schema):
         errors = s.validate({'foo': 42})
         assert errors['foo'] == [{'code': 'invalid_foo'}]
 
+    def test_ignored_if_not_in_only(self):
+        class MySchema(Schema):
+            a = fields.Field()
+            b = fields.Field()
+
+            @validates('a')
+            def validate_a(val):
```
carlos-alberto: This needs to take self as a first argument. (Not that it will ever be called).

marshmallow/schema.py
```diff
@@ -766,6 +766,8 @@ def _invoke_field_validators(self, data, many):
             try:
                 field_obj = self.fields[field_name]
             except KeyError:
+                if field_name in self.declared_fields:
+                    return
```
carlos-alberto: I think you want to `continue` here instead of `return`ing. Otherwise other declared validators may not be invoked.

carlos-alberto: It might also be worth adding a test that the other validators are invoked.

eprikazc: Thanks, I have updated the code according to your comments.

tests/test_schema.py
```diff
@@ -1003,6 +1003,23 @@ class MySchema(Schema):
...
+            def validate_a(self, val):
+                raise ValidationError({'code': 'invalid_a'})
+
+            @validates('b')
+            def validate_b(self, val):
+                raise ValidationError({'code': 'invalid_b'})
+
+        s = MySchema(only=('b',))
+        errors = s.validate({'b': 'data'})
+        assert errors == {'b': {'code': 'invalid_b'}}
```
eprikazc, line 1021: By the way, I was expecting list in the value, like:
`{'b': [{'code': 'invalid_b'}]}`
Is it an issue?

sloria:
Thanks!

PR #327, opened 2015-11-10
Re-work fields.Function

DamianHeard:
Argument func is now optional and deprecated in favor of serialize. The deserialize function can now optionally receive the context.

Note that deserialize is still noop by default to maintain backwards compatibility.

I'm assuming we don't want to mess with the order of arguments which is slightly untidy but I think we have to wear it for now for backwards compatibility.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -1133,40 +1133,59 @@ class Method(Field):
 class Function(Field):
     """A field that takes the value returned by a function.
 
-    :param callable func: A callable from which to retrieve the value.
+    :param callable func: This argument is to be depreciated. It exists for
+        backwards compatiblity. Use serialize instead.
+
+        A callable from which to retrieve the value.
         The function must take a single argument ``obj`` which is the object
         to be serialized. It can also optionally take a ``context`` argument,
         which is a dictionary of context variables passed to the serializer.
-    :param callable deserialize: Deserialization function that takes the value
-        to be deserialized as its only argument.
+    :param callable deserialize: A callable from which to retrieve the value.
+        The function must take a single argument ``value`` which is the value
+        to be deserialized. It can also optionally take a ``context`` argument,
+        which is a dictionary of context variables passed to the deserializer.
+        If no callable is provided then ```value``` we be passed through
+        unchanged.
+    :param callable serialize: A callable from which to retrieve the value.
+        The function must take a single argument ``obj`` which is the object
+        to be serialized. It can also optionally take a ``context`` argument,
+        which is a dictionary of context variables passed to the serializer.
+        If no callable is provided then the ```load_only``` flag will be set
+        to True.
     """
     _CHECK_ATTRIBUTE = False
 
-    def __init__(self, func, deserialize=None, **kwargs):
-        super(Function, self).__init__(**kwargs)
-        self.func = utils.callable_or_raise(func)
-        if deserialize:
-            self.deserialize_func = utils.callable_or_raise(deserialize)
-        else:
-            self.deserialize_func = None
+    def __init__(self, func=None, deserialize=None, serialize=None, **kwargs):
+        if func:
+            warnings.warn('Argument func of fields.Function is being depreciated.'
+                          ' Use the serialize argument instead.')
+            serialize = func
+
+        super(Function, self).__init__(load_only=not serialize, **kwargs)
+        self.serialize_func = serialize and utils.callable_or_raise(serialize)
+        self.deserialize_func = deserialize and utils.callable_or_raise(deserialize)
 
     def _serialize(self, value, attr, obj):
         try:
-            if len(utils.get_func_args(self.func)) > 1:
-                if self.parent.context is None:
-                    msg = 'No context available for Function field {0!r}'.format(attr)
-                    raise ValidationError(msg)
-                return self.func(obj, self.parent.context)
-            else:
-                return self.func(obj)
+            return self._call_or_raise(self.serialize_func, obj, attr)
         except AttributeError:  # the object is not expected to have the attribute
             pass
         return missing_
 
-    def _deserialize(self, value, attr, data):
+    def _deserialize(self, value, attr, obj):
         if self.deserialize_func:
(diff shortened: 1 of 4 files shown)


marshmallow/fields.py
```diff
@@ -1133,41 +1133,55 @@ def _deserialize(self, value, attr, data):
...
-                return self.func(obj, self.parent.context)
-            else:
-                return self.func(obj)
+            return self._call_or_raise(self.serialize_func, obj, attr)
         except AttributeError:  # the object is not expected to have the attribute
             pass
         return missing_
 
-    def _deserialize(self, value, attr, data):
+    def _deserialize(self, value, attr, obj):
```
sloria: Let's keep the argument name `data` here for consistency with the other fields and for backwards compat.

marshmallow/fields.py
```diff
@@ -1133,41 +1133,55 @@ def _deserialize(self, value, attr, data):
...
-    def __init__(self, func, deserialize=None, **kwargs):
-        super(Function, self).__init__(**kwargs)
-        self.func = utils.callable_or_raise(func)
-        if deserialize:
-            self.deserialize_func = utils.callable_or_raise(deserialize)
-        else:
-            self.deserialize_func = None
+    def __init__(self, func=None, deserialize=None, serialize=None, **kwargs):
+        if func:
+            warnings.warn('Argument func of fields.Function is being depreciated.'
```
sloria: Typo here: change to '"func" argument of fields.Function is deprecated. Use the "serialize" argument instead"

tests/test_deserialization.py
```diff
@@ -493,6 +494,15 @@ def test_function_field_deserialization_with_callable(self):
                                 deserialize=lambda val: val.upper())
         assert field.deserialize('foo') == 'FOO'
 
+    def test_function_field_deserialization_with_context(self):
+        field = fields.Function(
+            lambda x: None,
+            deserialize=lambda val, context: val.upper() + context['key']
+        )
+        field.parent = mock.MagicMock()
```
sloria: I don't think there's a need to use a `Mock` here. `parent` will be either another field or a Schema, so you can set `field.parent` to a Schema instance with context.

marshmallow/fields.py
```diff
@@ -1133,41 +1133,55 @@ def _deserialize(self, value, attr, data):
...
     _CHECK_ATTRIBUTE = False
 
-    def __init__(self, func, deserialize=None, **kwargs):
-        super(Function, self).__init__(**kwargs)
-        self.func = utils.callable_or_raise(func)
-        if deserialize:
-            self.deserialize_func = utils.callable_or_raise(deserialize)
-        else:
-            self.deserialize_func = None
+    def __init__(self, func=None, deserialize=None, serialize=None, **kwargs):
```
sloria: It might make more sense to have `serialize` be the first argument and `func` as the 3rd argument to encourage usage of `serialize` over `func`.

marshmallow/fields.py
```diff
@@ -1133,41 +1133,55 @@ def _deserialize(self, value, attr, data):
...
+        if func:
+            warnings.warn('Argument func of fields.Function is being depreciated.'
+                          ' Use the serialize argument instead.')
+            serialize = func
+
+        super(Function, self).__init__(load_only=not serialize, **kwargs)
+        self.serialize_func = serialize and utils.callable_or_raise(serialize)
```
sloria, line 1160: :+1:

marshmallow/fields.py
```diff
@@ -1133,41 +1133,55 @@ def _deserialize(self, value, attr, data):
 class Function(Field):
     """A field that takes the value returned by a function.
 
-    :param callable func: A callable from which to retrieve the value.
+    :param callable func: This argument is to be depreciated. It exists for
```
sloria: Typo here: "deprecated"

sloria:
Pass finished. Thanks for the PR!

marshmallow/fields.py
DamianHeard: my bad

marshmallow/fields.py
DamianHeard: Yeah good point, for some reason I was worried that would break backwards compatibility but it should be fine.

DamianHeard:
No thank you @sloria. Hopefully I've addressed all those concerns.

tests/test_serialization.py
```diff
@@ -4,6 +4,7 @@
 import datetime as dt
 import decimal
 
+import mock
```
sloria, line 7: The `mock` imports can now be removed.

sloria, line 7: Made this change when I merge.

marshmallow/fields.py
```diff
@@ -1133,41 +1133,55 @@ def _deserialize(self, value, attr, data):
...
-        super(Function, self).__init__(**kwargs)
-        self.func = utils.callable_or_raise(func)
-        if deserialize:
-            self.deserialize_func = utils.callable_or_raise(deserialize)
-        else:
-            self.deserialize_func = None
+    def __init__(self, serialize=None, deserialize=None, func=None, **kwargs):
+        if func:
+            warnings.warn('"func" argument of fields.Function is deprecated. '
+                          'Use the "serialize" argument instead.')
```
sloria, line 1156: This should be a `DeprecationWarning`.

marshmallow/fields.py
```diff
@@ -1133,41 +1133,55 @@ def _deserialize(self, value, attr, data):
...
+            serialize = func
+
+        super(Function, self).__init__(load_only=not serialize, **kwargs)
```
sloria, line 1159: `dump_only` should also be set if `deserialize` is None and `serialize` is passed.

sloria:
Caught a few small things after I merged. I will change those shortly.

marshmallow/fields.py
sloria, line 1159: Actually, I realize now that setting `dump_only` and `load_only` based on the passed arguments could potentially break users' code. I'm going to remove this for now. I've opened a new issue here: #328

PR #330, opened 2015-11-12
Optional serialization parameter for fields.Method

### LINKED ISSUE #329: Method field should allow passing deserialization method only
Proposal:
- Rename `method_name` to `serialize` for better symmetry with `deserialize`.
- `serialize` should default to `None`.

Rationale:

Consistency with `Function` field. See https://github.com/marshmallow-code/marshmallow/issues/325.

justanr:
'method_name' parameter of fields.Method is deprecated in favor of the
new 'serialize` method and is now optional. Serializing without a
provided method name returns the utils.missing value.

Implements and closes #329

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -1099,19 +1099,28 @@ class Method(Field):
 
     .. versionchanged:: 2.0.0
         Removed optional ``context`` parameter on methods. Use ``self.context`` instead.
+    .. versionchanged:: 2.3.0
+        Deprecated ``method_name`` parameter in favor of ``serialize`` and allow
+        ``serialize`` to not be passed at all.
     """
     _CHECK_ATTRIBUTE = False
 
-    def __init__(self, method_name, deserialize=None, **kwargs):
-        self.method_name = method_name
-        if deserialize:
-            self.deserialize_method_name = deserialize
-        else:
-            self.deserialize_method_name = None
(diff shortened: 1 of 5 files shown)


justanr:
I left the changelog alone pending this merge because it still contains stuff about `dump_only`/`load_only` being set by the presence of `deserialize`/`serialize` in `fields.Function` which I'd rather address in a separate commit.

marshmallow/fields.py
```diff
@@ -1099,19 +1099,28 @@ class Method(Field):
...
-        if deserialize:
-            self.deserialize_method_name = deserialize
-        else:
-            self.deserialize_method_name = None
+    def __init__(self, serialize=None, deserialize=None, method_name=None, **kwargs):
+        if method_name is not None:
+            warnings.warn('"method_name" argument of fields.Method is deprecated. '
+                          'Use the "serialize" argument instead.', DeprecationWarning)
+
+        self.serialize_method_name = serialize or method_name
```
sloria, line 1113: For complete backwards compatibility, we should also set `self.method_name`

``` python
self.serialize_method_name = self.method_name = serialize or method_name
```

sloria, line 1113: I'll take care of this post-merge.

sloria:
Thanks!

justanr:
No problem.

PR #340, opened 2015-11-19
Minor type checking fix

justanr:
Kinda pedantic, but using `isinstance(obj, cls)` is more Pythonic than `type(obj) == cls` and allows for whatever crazy subclass someone comes up with (`bool` because some people use).


sloria:
Thanks

justanr:
No problem!

Issue #323, opened 2015-11-05
Dont run validates_schema if individual fields were invalid


jjvattamattom:
I just discovered that validates_schema runs if individual fields have failed validation. I kinda expected otherwise.

I want a way to prevent validates_schema from being run if individual fields have failed validation. Please confirm if my way of comparing original and processed data is sufficient or that there is a simpler alternative.

```
@validates_schema(pass_original=True)
def validate_schema(self, data, original):
    # dont run below tests if validation on individual fields failed:
    if sorted(data.keys()) != sorted(original.keys()):
        return
    ...tests that presume individual fields are valid
```

sloria:
Sorry for the delayed response. I'm not sure there is a clean way to achieve this at the moment. I don't think we should change the default behavior; there are use cases where you want schema validators to run even if field validators fail. That said, I am open to suggestions for supporting the OP's use case.

DamianHeard:
Perhaps an alternative would be to properly expose the errors that have been captured during unmarshalling. At the moment you have to go through `self._unmarshal.errors`. Individual validate_schema methods would either check for errors before running of the decorator could be modified to automatically do this check if configured. Of course, I could be missing something and these errors might be already getting exposed at a more desirable location.

d-sutherland:
Another simple alternative would be to provide this as an argument to the validation decorators, allowing full backwards compatibility. 
Yet another alternative is to call validates_schema methods when a field has already failed only if strict mode is off. As such when strict mode is on the errors are propagated immediately (or at least it appears so to the user), which follows the same pattern users would be familiar with for exceptions. This would also follow the principle of least astonishment - if an error is raised (as occurs when strict is True) then the user can expect that code further down the pipeline won't be reached.

maximkulkin:
I just hit the same issue which got me surprised. IMO the most sane behavior is to:
- Deserialize fields and let field validations to run
- For each field run validations declared with `validates(field_name)` decorator only if that field does not have errors
- If there are no errors, run validations declared with `validates_schema` decorator.

sloria:
The `skip_on_field_errors` parameter has been added to `validates_schema` in 2.4.0.

PR #346, opened 2015-12-01
Optional skip_on_field_errors arg for validates_schema

d-sutherland:
Add 'skip_on_field_errors' parameter to validates_schema.
'skip_on_field_errors' allows schema level validation to be skipped
when errors are detected in field validation removing the need for
the user to check.
Address issue #323
Suggested that to further address issue #323 the next major release
should change the default value for this parameter from False to True.

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -800,7 +800,7 @@ class BaseSchema(base.SchemaABC):
                         field_obj=field_obj
                     )
 
-    def _invoke_validators(self, pass_many, data, original_data, many, field_errors):
+    def _invoke_validators(self, pass_many, data, original_data, many, field_errors=False):
         errors = {}
         for attr_name in self.__processors__[(VALIDATES_SCHEMA, pass_many)]:
             validator = getattr(self, attr_name)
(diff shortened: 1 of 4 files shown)


tests/test_decorators.py
```diff
@@ -429,3 +429,40 @@ def strict_fields(self, data, original_data):
         assert 'baz' in errors
         assert len(errors['baz']) == 1
         assert errors['baz'][0] == 'Unknown field name'
+
+    def test_skip_on_field_errors(self):
+
+        class MySchema(Schema):
+            foo = fields.Int(required=True)
+            bar = fields.Int(required=True)
```
sloria, line 437: Could you also test that the schema validators aren't run if a field validator (i.e. a validator callable passed as the `validate` argument) fails?

marshmallow/schema.py
```diff
@@ -797,12 +800,17 @@ def _invoke_field_validators(self, data, many):
                         field_obj=field_obj
                     )
 
-    def _invoke_validators(self, pass_many, data, original_data, many):
+    def _invoke_validators(self, pass_many, data, original_data, many, field_errors):
```
sloria: In order to maintain backwards-compat, could you make `field_errors` an optional argument?

sloria:
This is looking good. Just a few minor changes suggested above, and this should be good to merge. Thanks for the PR!

d-sutherland:
Thanks for the comments. Updates have been made.

sloria:
Thanks!

PR #355, opened 2015-12-06
Eagerly evaluate list-like iterators with `many`.

### LINKED ISSUE #353: Serialization drops first item of custom iterators
```

class Range:

    def __init__(self, max):
        self.value = 0
        self.max = max

    def __iter__(self):
        return self

    def __next__(self):
        self.value += 1
        if self.value > self.max:
            raise StopIteration
        return {'value': self.value}

class RangeSchema(Schema):
    value = fields.Int()

RangeSchema(many=True).dump(Range(5)).data
# [{'value': 2}, {'value': 3}, {'value': 4}, {'value': 5}]

list(Range(5))
# [{'value': 1}, {'value': 2}, {'value': 3}, {'value': 4}, {'value': 5}]
```

I'm guessing this is the same problem described in #343, but this should be easier to reproduce. This could be solved by detecting that the value is an iterator and casting to a list, as for generators (which might not be straightforward), or using something like `itertools.tee` to get separate iterators (one for serializing, one for inspecting the first item). Or maybe this is all overkill--maybe values serialized with `many` should have to be eager sequences, so that it isn't marshmallow's responsibility to cast to lists or inspect iterators without advancing them.

jmcarp:
Eagerly evaluate non-string iterators on `dump` when `many` is passed.
This ensures that serializing many values doesn't advance an iterator
and drop the first item. Values are only evaluated eagerly when `many`
is set so that we don't inappropriately cast other iterables (dicts,
namedtuples, etc) to lists.

[Resolves #353]

Note: it could be argued that this would be a breaking change--it's possible that users _wanted_ generators to be cast to lists even when `many` is `False`, although I kind of doubt it. Either way, I think it's desirable for the added test to pass, which it didn't before the proposed change to `Schema`.

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -8,7 +8,6 @@ import datetime as dt
 import decimal
 import inspect
 import json
-import types
 import uuid
 import warnings
 from collections import namedtuple
@@ -472,7 +471,7 @@ class BaseSchema(base.SchemaABC):
                             'many=True to serialize a collection.',
                             category=DeprecationWarning)
 
-        if isinstance(obj, types.GeneratorType):
+        if many and utils.is_iterable_but_not_string(obj):
             obj = list(obj)
 
         processed_obj = self._invoke_dump_processors(PRE_DUMP, obj, many, original_data=obj)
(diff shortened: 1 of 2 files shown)


sloria:
Looks good. I also doubt that casting to a list when `many=False` was desired behavior, and this is an important fix, so I may cut this into a 2.4.1 release. Thanks!

PR #356, opened 2015-12-07
Valiadate ws:// ad wss:// for websocket URLs

nicfit:


sloria:
Thanks for the PR. This could be considered a breaking change, so I'm going to hold off on merging it for now. We might want to mimic Django's more general approach, which allows the user to specify which schemes are considered valid: https://github.com/django/django/blob/master/django/core/validators.py#L92

nicfit:
OK, I'll provide my own validator.

sloria:
I've added a `schemes` param to `validate.URL`, so you can now do:

``` python
validate.URL(schemes=set(['http', 'https', 'ws', 'wss']))
```

PR #360, opened 2015-12-09
Remove unnecessary raising of ValidationError in unmarshaller

alexmorken:
The validation error does not need to be thrown because it is already being picked up.  Most of the time the error message was stomping on itself and it was not noticeable bug.  The only time this would come up is if a nested schema had a validation error and an index was being tacked on the beginning of the error - then the error would appear twice, once indexed and the other not indexed.  This is a tricky area that should be cleaned up more.


sloria:
Thank you. Made a hotfix release (2.4.2).

PR #363, opened 2015-12-24
Avoid use of sys.version for version checks

svisser:
The string `sys.version` should not be used for version checks so this change updates it to be done using `sys.version_info`.

### DIFF
--- a/marshmallow/compat.py
+++ b/marshmallow/compat.py
@@ -2,7 +2,7 @@
 import sys
 import itertools
 
-PY2 = int(sys.version[0]) == 2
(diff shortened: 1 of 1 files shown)


sloria:
I'm curious: what is wrong with using `sys.version`?

svisser:
It's a string with version information but it can't be relied upon to always contain the version number at the beginning of the string (across Python interpreters). See also the note in the documentation: https://docs.python.org/3/library/sys.html#sys.version.

sloria:
Ah, good to know. Thanks!

Issue #367, opened 2015-12-26
RFC: Improved API for registering field validator methods


sloria:
## Status quo

Validator methods are registered by decorating them with `@validates('my_field_name')`.

``` python
from marshmallow import fields, Schema, ValidationError, validates

class ItemSchema(Schema):
    quantity = fields.Int()

    @validates('quantity')
    def validate_quantity(self, value):
        if value > 30:
            raise ValidationError('Quantity must not be greater than 30')
```
## Proposed API

Validator methods are registered by decorating them with `@my_field.validator`.

``` python
from marshmallow import fields, Schema, ValidationError

class ItemSchema(Schema):
    quantity = fields.Int()

    @quantity.validator
    def validate_quantity(self, value):
        if value > 30:
            raise ValidationError('Quantity must not be greater than 30')
```
## Benefits
- Less prone to user error (no magic strings).
- Removes redundant code. Since `Field#validator()` decorator would just append to a field's `validators` attribute, there would be no need for [this code](https://github.com/marshmallow-code/marshmallow/blob/bad7e56cd5602daf4eea31ce3cb49f07a5d7a9ba/marshmallow/schema.py#L762-L800). 
- Better introspection of fields. Again, since `Field#validator()` appends to `Field#validators`, code that introspects fields (e.g. apispec) can find all of a field's validators in one place.

justanr:
:+1: Less magic is always good.

jmcarp:
> Less prone to user error (no magic strings).

:guitar:

sloria:
A sketch of this is implemented in #368.

immerrr:
Less magic strings is good, so it seems a nice idea to replace `@validates('quantity')` with `@validates(quantity)`.

I would however like to point out that  schema-level validators may check how two or more fields interact with each other. A recently introduced feature allows skipping schema-level validation if the validation of the field itself failed, it can be generalized to multiple fields, too. There's not much sense in checking how two, say, integer parameters interact with each other if one of those is unexpectedly a dictionary. Current API could support that naturally with `@validates([quantity, quality, speed])`, but it is not that obvious to me how to do that with the proposed one.

taion:
I think @immerrr's proposal provides a good middle ground between the current API and the proposal – it avoids the issue of using a string for no clear API reason, but sidesteps some of the problems noted with #368.

deckar01:
[...]

PR #368, opened 2015-12-28
WIP: Add Field#validator for registering field validation methods

sloria:
Implements #367

``` python

class ItemSchema(Schema):
    quantity = fields.Int()

    @quantity.validator
    def validate_quantity(self, value):
        if quantity > 30:
            raise ValidationError('Must not be less than 30.')

```
### TODO
- [ ] Docs
- [ ] Mark `@validates` for PendingDeprecation
- [ ] Fix override behavior (see https://github.com/marshmallow-code/marshmallow/pull/368#issuecomment-167723615)

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -4,6 +4,7 @@
 from __future__ import absolute_import, unicode_literals
 
 import collections
+import functools
 import datetime as dt
 import uuid
 import warnings
@@ -142,6 +143,7 @@ class Field(FieldABC):
         else:
             raise ValueError("The 'validate' parameter must be a callable "
                              "or a collection of callables.")
+        self._method_validators = []
 
         self.required = required
         # If missing=None, None should be considered valid by default
@@ -263,6 +265,27 @@ class Field(FieldABC):
         self._validate(output)
         return output
 
+    def validator(self, func):
+        """Decorator to register an instance method of a `Schema` as a
+        validator. ::
+
+            from marshmallow import Schema, fields
+
+            class ItemSchema(Schema):
+                quantity = fields.Int()
+
+                @quantity.validator
+                def validate_quantity(self, value):
+                    if quantity > 30:
+                        raise ValidationError('Must not be less than 30.')
+
+        :versionadded: 2.5.0
+        """
+        # Func is an unbound method. We partial it when the field gets bound
+        # to the schema in `_add_to_schema`
+        self._method_validators.append(func)
+        return func
+
     # Methods for concrete classes to override.
 
     def _add_to_schema(self, field_name, schema):
@@ -274,6 +297,10 @@ class Field(FieldABC):
         """
         self.parent = self.parent or schema
         self.name = self.name or field_name
+        # Bind registered method validators so that the `self` argument
+        # is the Schema
+        for validator in self._method_validators:
+            self.validators.append(functools.partial(validator, schema))
 
     def _serialize(self, value, attr, obj):
         """Serializes ``value`` to a basic Python datatype. Noop by default.
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -14,7 +14,7 @@ from operator import attrgetter
 from marshmallow import validate, utils, class_registry
 from marshmallow.base import FieldABC, SchemaABC
 from marshmallow.utils import missing as missing_
-from marshmallow.compat import text_type, basestring
+from marshmallow.compat import text_type, basestring, OrderedDict
 from marshmallow.exceptions import ValidationError
 
 __all__ = [
@@ -143,7 +143,7 @@ class Field(FieldABC):
         else:
             raise ValueError("The 'validate' parameter must be a callable "
                              "or a collection of callables.")
-        self._method_validators = []
+        self._method_validators = OrderedDict()
 
         self.required = required
         # If missing=None, None should be considered valid by default
@@ -283,7 +283,7 @@ class Field(FieldABC):
         """
         # Func is an unbound method. We partial it when the field gets bound
         # to the schema in `_add_to_schema`
-        self._method_validators.append(func)
+        self._method_validators[func.__name__] = func
         return func
 
     # Methods for concrete classes to override.
@@ -299,7 +299,7 @@ class Field(FieldABC):
         self.name = self.name or field_name
         # Bind registered method validators so that the `self` argument
         # is the Schema
-        for validator in self._method_validators:
+        for validator in self._method_validators.values():
             self.validators.append(functools.partial(validator, schema))
 
     def _serialize(self, value, attr, obj):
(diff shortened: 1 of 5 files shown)


marshmallow/fields.py
```diff
@@ -274,6 +297,10 @@ def _add_to_schema(self, field_name, schema):
         """
         self.parent = self.parent or schema
         self.name = self.name or field_name
+        # Bind registered method validators so that the `self` argument
+        # is the Schema
+        for validator in self._method_validators:
+            self.validators.append(functools.partial(validator, schema))
```
justanr: Not that I forsee a huge problem, but won't this bind the validator to the first schema it's on, rather than to the inheritance chain, unless I missed some Juju in SchemaMeta. I can't test my hunch as I'm mobile, but I think this shouldn't produce the expected results...

```
class MySchema(Schema):
    bar = DecimalField()

    @bar.validator
    def error(self, *a, **k):
        raise ValidationError(self.__class__.__name__)


class SubSchema(MySchema):
    pass

SubSchema.dump(dotteddict(bar=4)).errors 
```

But I _could_ be wrong. Another sticky part could be child schemas that have stricter validation needs. How would this react to:

```
class SubSchema(MySchema):
    @MySchema.bar.validates
    def stricter a validation(...):
        ...
```

justanr: Pretend that last bit is a valid python name. Stupid autocorrect.

sloria:
@justanr You raise valid concerns in https://github.com/marshmallow-code/marshmallow/pull/368#discussion_r48520184 (which I'm moving out here so this discussion doesn't get lost if the code changes).

So we have 2 cases to address:
1. Validator methods should be bound to the schema they were defined on (not parent classes).
2. Validator methods should be override-able.

Since `Field#_add_to_schema` gets called on Schema _instantiation_--not in `SchemaMeta`--case 1 turns out to be a non-issue. I've verified the behavior in this test https://github.com/marshmallow-code/marshmallow/blob/b30b705b836a86f722a89c8cf11b9d01ac17091d/tests/test_decorators.py#L376. 

Case 2 is currently broken. Overriding a validator method in a concrete Schema class will not prevent its parent validator method from being invoked. 

Also, I had to change `SchemaMeta` such that fields were accessible as class variables (by passing `pop=False` here: https://github.com/marshmallow-code/marshmallow/blob/b30b705b836a86f722a89c8cf11b9d01ac17091d/marshmallow/schema.py#L97)This is a breaking change, but without it, we can't do `@MySchema.bar.validate`. @jmcarp Do you recall why we decided to remove fields as class variables in the first place?

justanr:
I missed that sneaky assert in that test case. I have always found it odd that fields aren't accessible as properties - though, I can see the logic about that if it's possible to access them in a broken state that way.

I guess my last question would be since one instance is shared through the inheritance chain. Would validators registered on child schemas be run when the parent runs the validate method?

Going back to my second example, would 

``` python
MySchema().dump(...)
```

Run the `stricter_validation` from the child?

Sorry to ask a bunch of questions, just trying to stem potential edge cases before they're actual issues.

jmcarp:
I submitted some test fixes in #400, but now that I'm thinking about this proposal, I have a few reservations. Or maybe I'm just confused--it's hard to tell the difference.
- I think the decision to keep fields off schema classes happened before I started using marshmallow, but it makes sense to me--if fields are available off the schema, we have to worry about conflicts with other class and instance variables. It seems like we'd have to keep users from naming fields `extra`, `loads`, `opts`, etc., since those names are reserved by `Schema`.
- As implemented in #368, a decoration like `@MySchema.bar.validates` in a class that subclasses `MySchema` can do potentially bad things, since it mutates the field object on `MySchema`. If subclass `SubSchema` adds a validator to a field in `MySchema`, and we then create another subclass of `MySchema` called `SubSchema2`, the second subclass will also use the validator added by its sibling, since all three schema classes refer to the same field object.

Assuming I'm not off-base so far, I'm wondering if we should take field objects back off the schema and not allow schema subclasses to add new validators to fields defined in their parents, since both behaviors lead to edge cases and possibly counter-intuitive behavior. Would that make this interface too hard to use?

sloria:
@jmcarp You make a good case for keeping fields off of schema classes. I am still unsure how to answer the question about validators defined on subclasses. I'll have to give this more thought, but we definitely should prevent subclasses from adding validators to parents.

justanr:
@sloria Not sure how willing you are to do this, but instance variables defined in _metaclasses_ are available to classes but not to instances of those classes. It's an interesting effect of the type system and potentially useful for this, but could cause a lot of confusion. Classes can have things defined as one name and instances can overwrite those and provide their own values for them, not affecting the class at all.

For an example: https://gist.github.com/justanr/52301d3af774f714a80d

The `Builder.of` is only available as an attribute on the Builder class but not instances of it.

Something to mull over at least. I'll also note, I've not verified that this is intended, documented or compatible between Python 2 and 3.

justanr:
@sloria Some thoughts after mulling this over for a while.

By default, child schemas inherit their parents' validators. So, what do the following semantics mean:

``` python
class MySchema(Schema):
    bar = DecimalField()

    @bar.validator
    def error(self, *a, **k):
        raise ValidationError(self.__class__.__name__)


class SubSchema(MySchema):
    @MySchema.bar.validator
    def stricter(self, *a, **k):
         pass
```

Additionally, is it possible to have multiple validators bound to a single field:

``` python
class MySchema(Schema):
    bar = DecimalField()

    @bar.validator
    def error(self, *a, **k):
        raise ValidationError

     @bar.validator
     def other(self, *a, **k):
          pass
```

I know in Python 2 this would always be non-deterministic registering, but it's possible to avoid in Python 3 by specifying `__prepare__ = lambda *a, **k: OrderedDict()` in SchemaMeta.

If the behavior is that, no you should register one and one validator callable per field, and child validators should override their parents for their own instances, I think a solution is possible using some metaclass magic.

A very basic (and potentially wrong) example could be:

``` python
class Field(object):
         self.validators = {}

    def validator(self, func):
        func.__validator_for__ = self
        return func

    def _run_validators(self, schema, value):
        # wrong for finding parent validators, we should look to py3 functools._find_impl for inspiration
        registered = self.validators.get(type(schema), lambda schema, value: value)
        registered(schema, value)

class SchemaMeta(type):
    def __init__(cls, name, bases, attrs):
        super(cls, cls).__init__(cls, name, bases, attrs)

        for possible in attrs.values():
            field = getattr(possible, '__validator_for__', None)
            if not field:
                continue

            field.validators[cls] = possible
```

If the behavior is that, yes you can register multiple validation methods for a single field, then the solution is still possible but trickier because now we also need to walk the MRO of the Schema in question and call all parent validators in MRO.

sloria:
Closing this for now, as I don't plan on working on this in the immediate future (at least for 3.0 final). Thanks @justanr and @jmcarp for the feedback you gave on this; we'll will definitely refer back to your comments if we decide to pick this back up.

PR #369, opened 2015-12-28
Support passing specific fields to `partial`

tdevelioglu:
Support passing specific fields to `partial`. This restricts the fields
allowed to be missing during partial deserialization to those specified
in the value of `partial`.
This is particularly useful in api's when dealing with non client-generated
primary key fields (at creation), where only a specific field is allowed
to be missing, and you'd still like to use the same schema.

### DIFF
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -10,7 +10,7 @@ and from primitive types.
 
 from __future__ import unicode_literals
 
-from marshmallow.utils import missing
+from marshmallow.utils import is_collection, missing
 from marshmallow.compat import text_type, iteritems
 from marshmallow.exceptions import (
     ValidationError,
@@ -213,7 +213,9 @@ class Unmarshaller(ErrorStore):
         :param dict fields_dict: Mapping of field names to :class:`Field` objects.
         :param bool many: Set to `True` if ``data`` should be deserialized as
             a collection.
-        :param bool partial: If `True`, ignore missing fields.
+        :param bool|tuple partial: Whether to ignore missing fields. If its
+            value is an iterable, only missing fields listed in that iterable
+            will be ignored.
         :param type dict_class: Dictionary class used to construct the output.
         :param bool index_errors: Whether to store the index of invalid items in
             ``self.errors`` when ``many=True``.
@@ -242,6 +244,7 @@ class Unmarshaller(ErrorStore):
             return ret
         if data is not None:
             items = []
+            partial_is_collection = is_collection(partial)
             for attr_name, field_obj in iteritems(fields_dict):
                 if field_obj.dump_only:
                     continue
@@ -263,7 +266,9 @@ class Unmarshaller(ErrorStore):
                     field_name = field_obj.load_from
                     raw_value = data.get(field_obj.load_from, missing)
                 if raw_value is missing:
-                    if partial:
+                    # Ignore missing field if we're allowed to.
+                    if partial is True or\
+                            partial_is_collection and attr_name in partial:
                         continue
                     _miss = field_obj.missing
                     raw_value = _miss() if callable(_miss) else _miss
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -250,7 +250,9 @@ class BaseSchema(base.SchemaABC):
     :param tuple load_only: A list or tuple of fields to skip during serialization
     :param tuple dump_only: A list or tuple of fields to skip during
         deserialization, read-only fields
-    :param bool partial: If `True`, ignore missing fields when deserializing.
+    :param bool|tuple partial: Whether to ignore missing fields. If its value
+        is an iterable, only missing fields listed in that iterable will be
+        ignored.
 
     .. versionchanged:: 2.0.0
         `__validators__`, `__preprocessors__`, and `__data_handlers__` are removed in favor of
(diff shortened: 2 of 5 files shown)


marshmallow/marshalling.py
```diff
@@ -263,7 +265,9 @@ def deserialize(self, data, fields_dict, many=False, partial=False,
                     field_name = field_obj.load_from
                     raw_value = data.get(field_obj.load_from, missing)
                 if raw_value is missing:
-                    if partial:
+                    # Ignore missing field if we're allowed to.
+                    if partial is True or\
+                            is_collection(partial) and attr_name in partial:
```
sloria: We can avoid type-checking here if the default for `partial` is changed to `tuple()`.

marshmallow/schema.py
```diff
@@ -583,13 +586,17 @@ def _do_load(self, data, many=None, partial=None, postprocess=True):
...
+            the value for `self.partial` is used. If its value is an iterable,
+            only missing fields listed in that iterable will be ignored.
         :param bool postprocess: Whether to run post_load methods..
         :return: A tuple of the form (`data`, `errors`)
         """
         many = self.many if many is None else bool(many)
-        partial = self.partial if partial is None else bool(partial)
+        if partial is None:
+            partial = self.partial
+        elif not utils.is_collection(partial):
```
sloria: Is this branch necessary?

tests/test_deserialization.py
```diff
@@ -1067,6 +1067,24 @@ class MySchema(Schema):
...
+        assert data['foo'] == 3
+        assert 'bar' in errors
+        assert 'baz' in errors
+
+        data, errors = MySchema().load({'foo': 3}, partial=('bar', 'baz'))
+        assert data['foo'] == 3
+        assert 'bar' not in data
+        assert 'baz' not in data
+        assert not errors
+
```
sloria, line 1086: Could you also add a test for overriding the `partial` attribute in `load`?

e.g.

``` python
schema = MySchema(partial=True)
schema.load(.... , partial=('foo', ))
```

sloria:
This looks good. Just a few suggestions above.

Thank you for adding docs and tests.

marshmallow/schema.py
tdevelioglu: Frankly, I don't know. It was coercing to bool and I didn't want to change the behavior, this way it still coerces what's left (i.e. not iterable) to bool.
Let me know what you prefer and I'll change it.

marshmallow/marshalling.py
tdevelioglu: It breaks the tests, on False values, though

tests/test_deserialization.py
tdevelioglu, line 1086: done :)

marshmallow/schema.py
justanr: It looks like - unless I missed something - that partial is already coerced into a bool with `if partial` checks and None has no special significance. So a regular tuple should maintain the behavior.

tdevelioglu: I'm a little bit confused.

If you look closely, it was assigning the coerced value to partial if partial wasn't None, and I kept that behavior (with the obvious exception, as we don't want to convert a tuple value).

Perhaps you could write down what you'd prefer and I'll change it.

justanr: https://github.com/marshmallow-code/marshmallow/blob/bad7e56cd5602daf4eea31ce3cb49f07a5d7a9ba/marshmallow/marshalling.py#L266

That's the only place where partial is referenced in Unmarshaller and it's a simple `if partial` check. Changing the default of partial to `()` -- an empty tuple -- allows doing `if field_name in partial` there instead.

I misunderstood `partial is None` before, it's a check in the schema to decide which partial value to use. Instead, if we're using a tuple to declare specific fields to partial, why not do `partial = partial or self.partial`.

sloria: I like @justanr 's suggestion. I would like to avoid type checks where possible.

tdevelioglu: @justanr
yes, but how to deal with boolean values then, are you suggesting to completely change the semantics of partial ? My intention was to keep things backward compatible since this is a public api.
I agree with the type checking thing, I just need to know if we want to change partial completely, or keep around support for boolean values before I re-factor it.
Another option would be to add another parameter, e.g. `ignore_missing=tuple()` and leave partial untouched

justanr: Hmm. That's a good point. I think @sloria should make the call if this should be broken.

sloria: I'm not understanding how changing this code to 

``` python
partial = partial or self.partial
```

results in a breaking change. I am clearly missing something. The tests pass with this change; can you give an example of code that would break?

marshmallow/marshalling.py
sloria: Ah, fair enough. This is fine as is.

marshmallow/schema.py
tdevelioglu: With `partial = partial or self.partial` if you pass `False`, `tuple()` or anything else that gets implicitly evaluated false, you'll end up using `self.partial`. So you need to explicitly check for the default value (None).

For example:

``` python
s = Schema(partial=True)
s.load({}, partial=False)  # partial == True
```

We can drop the branch if it's OK to not coerce things into bool anymore though.

sloria: > We can drop the branch if it's OK to not coerce things into bool anymore though.

Yes, that is fine.

tdevelioglu: done :)

marshmallow/marshalling.py
sloria: Actually, we can at least avoid type-checking within the for loop:

``` python
partial_is_collection = is_collection(partial)
for attr_name, field_obj in iteritems(fields_dict):
 ...
    if (
        partial is True or
        (partial_is_collection and attr_name in partial)
    )
```

tdevelioglu: good one, I've updated it.

sloria:
Thanks!

ewang:
@sloria @tdevelioglu I was just going over the changelog to upgrade and noticed this PR. Doesn't passing in MySchema(partial=True, exclude=[...]) achieve the same effect as the merged PR's change of passing in MySchema(partial=(...))

Issue #347, opened 2015-12-01
Unenveloping None values doesn't work as expected


tdevelioglu:
Trying to unenvelope serialized data when the contained data is None continues deserialization with the original data. This makes it impossible to deserialize nested fields with enveloped data where None is a valid value. (Related: http://jsonapi.org/format/)

Example (let's pretend our system allows for homeless employees):
https://gist.github.com/tdevelioglu/8e433355494ba022f48f
(I expect a deserialized Resource object, with an attribute 'address' with value 'None')

sloria:
Thanks for the report. This does appear to be a limitation of the `pre_/post_` methods. The decorators were meant to allow for in-place modifications of data and no return values (i.e. implicitly returning `None`).

We'll have to give this more thought to reach a solution.

sloria:
@taion @lafrech @deckar01 Any ideas on this?

We could make it so that `pre_`/`post_` methods *must* return the processed data and allow `None` to be returned. This would be a breaking change, which is why I'd like to resolve this for 3.0.

deckar01:
> The method receives the deserialized data and returns the processed data.

https://marshmallow.readthedocs.io/en/latest/api_reference.html#marshmallow.decorators.post_load

A return value of `None` indicating that the data was modified in place is not part of the docs. Anyone not returning a value is relying on undocumented behavior, but I can see how a lot of people could be relying on this unknowingly.

I recommend documenting the `None` behavior in the 2.x-line docs in addition to fixing it in 3.0.

lafrech:
I also think this is the right thing to do.

The change in the code should be easy: just remove `if_none` stuff.

I like this. It fixes the use case while stripping unnecessary complication from the code.

sloria:
Thanks guys. Let's make it so!

taion:
This was my fault, wasn't it? Yeah, in retrospect trying to handle "not returning" from those hooks was way too cute. Sorry!

deckar01:
It was discussed at the time, it just wasn't documented. It's never too late to add some docs. 😉 

https://github.com/marshmallow-code/marshmallow/pull/191#issuecomment-91424999

PR #371, opened 2015-12-30
Support specifying envelope for fields

tdevelioglu:
This adds support for marshalling/unmarshalling fields that are enveloped, making it simpler to deal with fields that require nesting. (relevant: http://jsonapi.org/format/)

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -58,3 +58,4 @@ Contributors (chronological)
 - Dan Sutherland `@d-sutherland <https://github.com/d-sutherland>`_
 - Jeff Widman `@jeffwidman <https://github.com/jeffwidman>`_
 - Simeon Visser `@svisser <https://github.com/svisser>`_
+- Taylan Develioglu `@tdevelioglu <https://github.com/tdevelioglu>`_
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -398,6 +398,30 @@ If you want to marshal a field to a different key than the field name you can us
     # 'CamelCasedEmail': 'foo@bar.com'}
 
 
+Enveloping Keys
+-------------------------------
+
+If you want to marshal or unmarshal fields that are enveloped, you can use `envelope` to specify a tuple of keys that a field is nested in.
+
+.. code-block:: python
+    :emphasize-lines: 2,3,4,14
+
+    class UserSchema(Schema):
+        id = fields.String(envelope=('data',))
+        name = fields.String(envelope=('data', 'attributes'))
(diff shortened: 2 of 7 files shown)


justanr:
To be curious, what is the advantage over not using `pre_load` to strip the envelope off?

tdevelioglu:
https://github.com/marshmallow-code/marshmallow/issues/347 is one issue.

Having an envelope parameter on the field is more intuitive, and it saves one from being forced to write a matching `post_dump` method to properly envelope all the relevant fields (how do we know which fields belong where ?) because `envelope` works both ways.

This is how we implement this for a resource object using `post_dump` and `pre_load`:

``` python
@pre_load
def __pre_load(self, data):
    result_data = {}
    attributes = data.get('attributes')
    relationships = data.get('relationships')
    if 'id' in data:
        result_data['id'] = data['id']
    result_data['type'] = data.get('type')

    if attributes:
        for attribute, value in attributes.items():
            result_data[attribute] = value
    if relationships:
        for relationship, value in relationships.items():
            result_data[relationship] = value
    return result_data

@post_dump
def __post_dump(self, data):
    result_data = {'attributes': {}, 'relationships': {}, 'links': links}
    for key, value in data.items():
        field = self.fields.get(key)
        if field:
            # 'id' and 'type' members belong at the top of a JSONAPI
            #  document.
            if key == 'id' or key == 'type':
                result_data[key] = value
                continue
            if isinstance(field, RelationshipObject):
                result_data['relationships'][key] = value
            else:
                result_data['attributes'][key] = value
    return result_data
```

Using `envelope`, you just specify where to envelope in each field, and nothing else:

``` python
class PoolMemberToJSONAPIResourceObject(ResourceObject):
    """
    PoolMember schema
    """
    __model__ = models.PoolMember

    name = fields.String(envelope=('attributes',), dump_only=True)
    enabled = fields.Boolean(envelope=('attributes',))
    ratio = fields.Integer(envelope=('attributes',))

    node = RelationshipObject(
        'NodeToJSONAPIResourceObject', envelope=('relationships',), required=True)
    pool = RelationshipObject(
        'PoolToJSONAPIResourceObject', envelope=('relationships',), required=True)
    properties = RelationshipObject(
        'PoolMemberPropertyToJSONAPIResourceObject', envelope=('relationships',), many=True)
```

sloria:
Sorry for the delay on this. I am hoping to review this and your other PR more in depth this weekend.

sloria:
Thank you for the PR. I'm not comfortable merging this at this point. I believe handling envelope belongs in user code and wrapper libraries.

PR #373, opened 2016-01-05
Serialize mappings

zbristow:


sloria:
So the purpose of this is to serialize non-dict mapping types to dicts?

Not quite sure whether to consider this a breaking change or not; depends on how users expect mappings to be serialized...

zbristow:
That's correct, to take objects implementing a mapping interface and serializing them to dicts.  Personally, that was my original expectation when using this field, and was surprised when json.dumps failed.

sloria:
I'm going to hold off on merging this for now, 1) to be conservative about backwards-compat and 2) to keep `Dict` simple. I suggest implementing a custom `Mapping` field in your codebase to achieve what you want here. 

Thanks anyway for the contribution.

PR #379, opened 2016-01-19
Schema.validate() should support `partial`

tdevelioglu:
We support `partial` in Schema.load(), but not in Schema.validate(). This PR adds support for passing `partial` to Schema.validate() as well.


sloria:
Thanks. I'll add some tests to this after I merge.

PR #382, opened 2016-01-21
Report validation errors for each item of List field

### LINKED ISSUE #345: Incorrect implementation of List field deserialization
Problem explanation:

``` python
from marshmallow import Schema, fields, validate, validates, validates_schema, ValidationError


class IntRangeSchema(Schema):
    first = fields.Integer(required=True)
    last = fields.Integer(required=True)


class PoolSchema(Schema):
    name = fields.String(required=True)
    ranges = fields.List(fields.Nested(IntRangeSchema))

    @validates('ranges')
    def validate_ranges(self, ranges):
        print 'validate_ranges %s' % repr(ranges)
```

When I load valid data, the validate_ranges() is called with a list of dictionaries (which is expected):

``` python
schema = PoolSchema(strict=True)
data = {
    'name': 'pool1',
    'ranges': [
        {'first': 1, 'last': 10},
        {'first': 11, 'last': 20},
    ],
}
print repr(schema.load(data).data)
# => validate_ranges [{'last': 10, 'first': 1}, {'last': 20, 'first': 11}]
# => {'ranges': [{'last': 10, 'first': 1}, {'last': 20, 'first': 11}], 'name': u'pool1'}
```

But when I load an invalid data, then the validates_ranges() receives only a non-valid dictionary, not a list:

``` python
data = {
    'name': 'pool2',
    'ranges': [
        {'first': 1, 'last': 10},
        {'last': 10},
    ],
}            
print repr(schema.load(data).data)
# => validate_ranges {'last': 10}
# => ... ValidationError
```

Moreover, if I add complete schema validation (via `validates_schema`), this is what passed to it:

```
{'ranges': {'last': 10}, 'name': u'pool2'}
```

I see multiple problems here:
1) naive deserialization in List field type
2) running validations for fields that have already failed by other means.

The second issue is addressed in https://github.com/marshmallow-code/marshmallow/issues/323, so I'll focus on the first one.

Let's take List field as example. Deserialization of list types is basically:

``` python
def _deserialize(self, value, attr, data):
    return [self.container.deserialize(each) for each in value]
```

By definition container is a field type which will report deserialization/validation errors by raise ValidationError exceptions. That means that you will get an error for the first invalid item in the list and the rest of items won't be validated at all. The proper way would be to accumulate all errors:

``` python
def _deserialize(self, value, attr, data):
    if not utils.is_collection(value):
        self.fail('invalid')
    result = []
    errors = {}
    for idx, each in enumerate(value):
        try:
            result.append(self.container.deserialize(each))
        except ValidationError as e:
            result.append(e.data)
            errors.update({str(idx): e.message})

    if errors:
        raise ValidationError(errors, data=result)

    return result
```

maximkulkin:
Current implementation of List field does deserialization
in a naive way stopping at first validation error. When
you get an error it is hard to tell which item it belongs to.

This implementation captures errors for each item of List field,
returning them keyed by item index.

Fixes #345.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -531,11 +531,22 @@ class List(Field):
         return [self.container._serialize(value, attr, obj)]
 
     def _deserialize(self, value, attr, data):
-        if utils.is_collection(value):
-            # Convert all instances in typed list to container type
-            return [self.container.deserialize(each) for each in value]
-        else:
+        if not utils.is_collection(value):
             self.fail('invalid')
+
+        result = []
+        errors = {}
+        for idx, each in enumerate(value):
+            try:
+                result.append(self.container.deserialize(each))
+            except ValidationError as e:
+                result.append(e.data)
+                errors.update({idx: e.messages})
+
+        if errors:
+            raise ValidationError(errors, data=result)
+
+        return result
 
 class String(Field):
     """A string field.
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -582,12 +582,24 @@ class TestFieldDeserialization:
         field = fields.List(fields.DateTime)
         with pytest.raises(ValidationError) as excinfo:
             field.deserialize(['badvalue'])
-        assert excinfo.value.args[0] == 'Not a valid datetime.'
+        assert excinfo.value.args[0] == {0: ['Not a valid datetime.']}
 
         field = fields.List(fields.Str())
         with pytest.raises(ValidationError) as excinfo:
             field.deserialize(['good', 42])
-        assert excinfo.value.args[0] == 'Not a valid string.'
+        assert excinfo.value.args[0] == {1: ['Not a valid string.']}
+
+    def test_list_field_deserialize_multiple_invalid_items(self):
+        field = fields.List(
+            fields.Int(
+                validate=validate.Range(10, 20, error='Value {input} not in range')
+            )
+        )
+        with pytest.raises(ValidationError) as excinfo:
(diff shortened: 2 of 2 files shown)


marshmallow/fields.py
```diff
@@ -531,12 +531,26 @@ def _serialize(self, value, attr, obj):
...
+        errors = {}
+        for idx, each in enumerate(value):
+            try:
+                result.append(self.container.deserialize(each))
+            except ValidationError as e:
+                result.append(e.data)
+                item_errors = e.messages
+                if utils.is_collection(item_errors) and len(item_errors) == 1:
+                    item_errors = item_errors[0]
+                errors.update({str(idx): item_errors})
```
sloria: I don't think there is a need to cast `idx` to a string. When passing `many=True` to a `Schema`, errors are indexed by integer values, so best to keep that consistent here.

marshmallow/fields.py
```diff
@@ -531,12 +531,26 @@ def _serialize(self, value, attr, obj):
...
 
+        result = []
+        errors = {}
+        for idx, each in enumerate(value):
+            try:
+                result.append(self.container.deserialize(each))
+            except ValidationError as e:
+                result.append(e.data)
+                item_errors = e.messages
+                if utils.is_collection(item_errors) and len(item_errors) == 1:
```
sloria: I'm not sure this is necessary. For consistency, why not keep the errors in a list, even if there is only one error?

sloria:
Looks good! Thanks

PR #383, opened 2016-01-21
Add `equal` parameter in validate.Length

daniloakamine:
I have a String field and I need to validate if it contains exactly 6 characters.
Instead of use **validate.Regexp**, I'd like to use the proper way **validate.Length**.

``` python
name = fields.Str(required=True,
                          validate=[validate.Length(min=6, max=6)])
```

It works, but I need to repeat the same value in two different parameters.
In addition, I receive a weird error message _"Length must be between 6 and 6"_. 
This message don't make sense for my case.

I'm suggesting to create a new optional parameter.

``` python
name = fields.Str(required=True,
                          validate=[validate.Length(equal=6)])
```

In that case, I will receive the message _"Length must be 6"_.
What do you think about this?

### DIFF
--- a/marshmallow/validate.py
+++ b/marshmallow/validate.py
@@ -219,6 +219,12 @@ class Length(Range):
     message_equal = 'Length must be {equal}.'
 
     def __init__(self, min=None, max=None, error=None, equal=None):
+        if equal is not None and any([min, max]):
+            raise ValueError(
+                'The `equal` parameter was provided, maximum or '
+                'minimum parameter must not be provided.'
+            )
+
         super(Length, self).__init__(min, max, error)
         self.equal = equal
 
--- a/marshmallow/validate.py
+++ b/marshmallow/validate.py
@@ -207,6 +207,8 @@ class Length(Range):
         will not be checked.
     :param int max: The maximum length. If not provided, maximum length
         will not be checked.
+    :param int equal: The exact length. If provided, maximum and minimum
+        length will not be checked.
     :param str error: Error message to raise in case of a validation error.
         Can be interpolated with `{input}`, `{min}` and `{max}`.
     """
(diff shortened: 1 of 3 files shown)


drgarcia1986:
:+1: :clap: :clap: :clap:

daniloakamine-luizalabs:
@sloria can you review for me, please?

daniloakamine:
@sloria can you review for me, please?

marshmallow/validate.py
```diff
@@ -214,10 +216,27 @@ class Length(Range):
     message_min = 'Shorter than minimum length {min}.'
     message_max = 'Longer than maximum length {max}.'
     message_all = 'Length must be between {min} and {max}.'
+    message_equal = 'Length must be {equal}.'
+
+    def __init__(self, min=None, max=None, error=None, equal=None):
+        super(Length, self).__init__(min, max, error)
+        self.equal = equal
```
sloria, line 229: It might make sense to raise a `ValueError` if equal is passed and either `min` and `max` is not `None` .

sloria:
Other than my one suggestion above, this looks good! Thanks for the PR

marshmallow/validate.py
```diff
@@ -214,10 +216,27 @@ class Length(Range):
     message_min = 'Shorter than minimum length {min}.'
     message_max = 'Longer than maximum length {max}.'
     message_all = 'Length must be between {min} and {max}.'
+    message_equal = 'Length must be {equal}.'
+
+    def __init__(self, min=None, max=None, error=None, equal=None):
```
sloria, line 221: Note to self: In marshmallow 3.0, we should reorder these parameters so that `equal` comes before `error`.

daniloakamine, line 221: Great! I wish I had done this, but I didn't want to break the backward compatibility in a first moment.

marshmallow/validate.py
daniloakamine, line 229: Done :+1:

sloria:
Thanks!

PR #392, opened 2016-02-11
Accept dict in messages errors

DanielBacci:
Hi,
I had a problem for return the fields with errors. My api should follow:

`field_name = [{"code": 4001, "message": "error desc" }]`

I like solution for required fields http://marshmallow.readthedocs.org/en/latest/quickstart.html,
 `city = fields.String(required={'message': 'City required', 'code': 400})`, but not working to validate :(.

My solution to problem is this PR, but accept suggestions.

Thanks :)

### DIFF
--- a/marshmallow/validate.py
+++ b/marshmallow/validate.py
@@ -198,7 +198,7 @@ class Range(Validator):
         return 'min={0!r}, max={1!r}'.format(self.min, self.max)
 
     def _format_error(self, value, message):
-        if not isinstance(self.error, dict):or self.error is None:
+        if not isinstance(self.error, dict) or self.error is None:
             return (self.error or message).format(
                 input=value, min=self.min, max=self.max)
         if 'message' in self.error:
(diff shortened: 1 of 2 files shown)


danielbacci-luizalabs:
cc @sloria @justanr

Can you review please ?

sloria:
Passing a dictionary or string as the `required` argument was removed from the API (https://marshmallow.readthedocs.org/en/latest/changelog.html#id14). I've just update the documentation in the Quickstart accordingly. 

In your code, you would do:

``` python
city = fields.String(required=True, error_messages={'required': {'message': 'City required', 'code': 400}})
```

Does that meet your use case?

danielbacci-luizalabs:
@sloria it works for required fields, but in my case i need pass a dictionary or string to validate like Length or Regexp ...

Example:

`zipcode = fields.Str(
        required=True,
        error_messages={"required": {"message": "message error", "code": "code error"}},
        validate=[
            validate.Length(max=8, error={"message": "message error", "code", "code error"})])`

sloria:
OK, it does make sense to allow the `error` parameter for validators to be a dictionary, though I'm unsure that this PR takes the right approach. The special-casing of the `message` name feels off to me.

sloria:
Closing this for now; feel free to reopen if you think there's still a need for this.

PR #400, opened 2016-02-21
New method validators

jmcarp:
This patch is a proposed fix for the failing
`test_method_validator_override` in `new-method-validators`. Instead of
storing validators in a list, use an `OrderedDict` keyed by method name,
so that overrides by subclasses clobber parent methods with the same
name.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -16,3 +16,4 @@ simplejson
 
 # Syntax checking
 flake8==2.4.1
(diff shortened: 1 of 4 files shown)


jmcarp:
This proposal fixes the failing tests from #368, so it should satisfy the requirements discussed there. Are there other edge cases this needs to handle?

cc @sloria @justanr

sloria:
Thanks for this. Further discussion on this feature should happen in #368.

PR #180, opened 2015-03-20
added support for dot notation only fields in dump method

rozenm:
1. Added only to dump (dumps) 
2. "only" support for dot notation syntax to control nested schema 
   fields 

@sloria This is not the final commit, I am pushing this in to get your first impression
all tests are passing.
My next step is to add more tests and fix documentation.

Thanks

Issue #151, opened 2015-02-17
Optional nesting (handling an embed parameter)


stephenfin:
I'm aware of how to nest items, as described [in the documentation](http://marshmallow.readthedocs.org/en/latest/nesting.html). Per those examples (with slight modifications) I am able to nest, say, a `Author` in a `Book`:

``` python
class AuthorSchema(Schema):
    class Meta:
        fields = ('id', 'name')

class BookSchema(Schema):
    author = fields.Nested(AuthorSchema, only='id')
    class Meta:
        fields = ('id', 'title', 'author')
```

This is all fine and dandy. However, say I was exposing these models over a REST API and I wanted to implement an `embed` paremeter, as described [here](http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#autoloading). From what I can see, I would need to either (a) define a huge number of alternative schemas (i.e. `BookSchema_AuthorID`, `BookSchema_AuthorName`, `BookSchema_AuthorIDName` etc.) or (b) perform some processing on the serialized output and strip out fields I don't want. Neither of these are ideal. Is there any alternative?

davidism:
I'd also find this very convenient.  To word it differently: There should be a way to exclude specific fields on nested (at any level) schemas.

rozenm:
We have a similar requirement,
here is an optional solution, if approved I'll try and put up a PR:
1.  Support specifying the list of fields to be serialized in a nested schema using the dot notation in
the "only" argument. 
2. Add "only" to dumps/dump methods that overrides the fields and self.only
3. Raise validation error if the fields or nested fields passed as "only" aren't a subset
of Meta class fields.

arturbekasov:
I've had a similar problem. The solution I've came up with is to use meta-classes and create schemas on the fly, like so:

``` python
# Function for creating schemas
def SchemaFactory(model, fields={}):
    dict = fields.copy()
    dict['Meta'] = type('Meta', (), {'model': model})

    return type(model.__name__ + 'Schema', (ma.ModelSchema,), dict)

# Sample usage
Schema = SchemaFactory(UnitTopic, {
  'topic': fields.Nested(TopicSchema),
  'unit': fields.Nested(UnitSchema)
})

schema = Schema(many=True)
schema.dump(unit_topics).data
```

So, you get to control what gets extended, by providing a fields.Nested entry to the factory. It might be a bit hacky, but seems to work.

luismartingil:
Would this be another hacky solution?

[...]

PR #405, opened 2016-03-02
Nested 'only' parameter

### LINKED ISSUE #402: Dynamically specify 'only' parameter for nested fields
I'm able to specify the `only` parameter when instantiating a schema (`MySchema(only=('field1', 'field2')`), however, this does not work for nested fields. I'd like to be able to have something like `MySchema(only=('field1', ('field2', ('subfield1', 'subfield3'))))` where _field2_ is a nested field of which only _subfield1_ and _subfield2_ should be serialized.

I'd be willing to implement it, however, first I wanna know if it has a chance to get into marshmallow. Secondly, it seems to not be a straight forward addition; instead of (or additionally to) during instantiation of the nested schema, the _only_ parameter needs to be passed when serializing. So I'd appreciate any hints of how to implement this properly.

The motivation behind this are APIs like [JSONAPI](http://jsonapi.org/) which allow to dynamically specify fields like so: `GET /articles?include=author&fields[articles]=title,body&fields[people]=name`. One could argue that resources are not really rendered in a nested fashion in JSONAPI responses. So lets assume, we use a (not yet implemented) `IncludeSchema` instead. That will, however, leave us with the same challenges of specifying the _only_ parameter dynamically.

Archelyst:
This allows to specify a nested `only` parameter both at schema instantiation time and during dump() like so:

``` python
class ChildSchema(Schema):
    foo = fields.Field()
    bar = fields.Field()
    baz = fields.Field()

class ParentSchema(Schema):
    bla = fields.Field()
    bli = fields.Field()
    blubb = fields.Nested(ChildSchema)

data = dict(bla=1, bli=2, blubb=dict(foo=42, bar=24, baz=242))

# either when instantiating
sch = ParentSchema(only=('bla', ('blubb', ('foo', 'bar'))))
result = sch.dump(data)

#or when dumping
sch = ParentSchema()
result = sch.dump(data, only=('bla', ('blubb', ('foo', 'bar'))))
```

It is fully backwards compatible. Fixes #402

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -364,6 +364,7 @@ class Nested(Field):
     def __init__(self, nested, default=missing_, exclude=tuple(), only=None, **kwargs):
         self.nested = nested
         self.only = only
+        self.only_explicit = None
         self.exclude = exclude
         self.many = kwargs.get('many', False)
         self.__schema = None  # Cached Schema instance
@@ -382,10 +383,11 @@ class Nested(Field):
             only = (self.only, )
         else:
             only = self.only
+        only = self.only_explicit or only
 
         # Inherit context from parent.
         context = getattr(self.parent, 'context', {})
-        if not self.__schema:
+        if not self.__schema or self.__schema.only != only:
             if isinstance(self.nested, SchemaABC):
                 self.__schema = self.nested
                 self.__schema.context.update(context)
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -318,7 +318,7 @@ class BaseSchema(base.SchemaABC):
         - ``index_errors``: If `True`, errors dictionaries will include the index
             of invalid items in a collection.
         - ``load_only``: Tuple or list of fields to exclude from serialized results.
-        - ``dump_only``: Tuple or list of fields to exclude from deserialization
+        - ``dump_only``: Tuple or list of fields to exclude from deserialization.
         """
         pass
 
@@ -471,8 +471,8 @@ class BaseSchema(base.SchemaABC):
         many = self.many if many is None else bool(many)
         if not many and utils.is_collection(obj) and not utils.is_keyed_tuple(obj):
             warnings.warn('Implicit collection handling is deprecated. Set '
-                            'many=True to serialize a collection.',
-                            category=DeprecationWarning)
+                          'many=True to serialize a collection.',
+                          category=DeprecationWarning)
 
         if many and utils.is_iterable_but_not_string(obj):
             obj = list(obj)
@@ -647,21 +647,37 @@ class BaseSchema(base.SchemaABC):
 
         return result, errors
 
+    def _get_only_fields(self):
+        """Get the fields that should only be serialized."""
+        if not self.only:
+            return None, {}
+        direct_fields = []
+        nested_fields = {}
+        for field in self.only:
+            if isinstance(field, (tuple, list)):
+                direct_fields.append(field[0])
+                nested_fields[field[0]] = field[1]
+            else:
+                direct_fields.append(field)
+        return direct_fields, nested_fields
+
     def _update_fields(self, obj=None, many=False):
         """Update fields based on the passed in object."""
-        if self.only:
+        only_direct, only_nested = self._get_only_fields()
+        if only_direct:
             # Return only fields specified in only option
             if self.opts.fields:
-                field_names = self.set_class(self.opts.fields) & self.set_class(self.only)
+                field_names = (self.set_class(self.opts.fields) &
+                               self.set_class(only_direct))
             else:
-                field_names = self.set_class(self.only)
+                field_names = self.set_class(only_direct)
         elif self.opts.fields:
             # Return fields specified in fields option
             field_names = self.set_class(self.opts.fields)
         elif self.opts.additional:
             # Return declared fields + additional fields
             field_names = (self.set_class(self.declared_fields.keys()) |
-                            self.set_class(self.opts.additional))
+                           self.set_class(self.opts.additional))
         else:
             field_names = self.set_class(self.declared_fields.keys())
 
@@ -670,6 +686,11 @@ class BaseSchema(base.SchemaABC):
         if excludes:
             field_names = field_names - excludes
         ret = self.__filter_fields(field_names, obj, many=many)
+        # Set nested only parameters overriding the only parameters declared on the fields
+        for field_name, field in [(fn, f) for fn, f in ret.items() if hasattr(f, 'only')]:
+            field.only_explicit = None
+            if field_name in only_nested:
+                field.only_explicit = only_nested[field_name]
         # Set parents
         self.__set_field_attrs(ret)
         self.fields = ret
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -669,6 +669,27 @@ def test_error_raised_if_additional_option_is_not_list():
                 additional = 'email'
 
 
+def test_nested_only():
+    class ChildSchema(Schema):
+        foo = fields.Field()
+        bar = fields.Field()
+        baz = fields.Field()
+    class ParentSchema(Schema):
+        bla = fields.Field()
+        bli = fields.Field()
+        blubb = fields.Nested(ChildSchema)
+    sch = ParentSchema(only=('bla', ('blubb', ('foo', 'bar'))))
+    data = dict(bla=1, bli=2, blubb=dict(foo=42, bar=24, baz=242))
+    result = sch.dump(data)
+    assert 'bla' in result.data
+    assert 'blubb' in result.data
+    assert 'bli' not in result.data
+    child = result.data['blubb']
+    assert 'foo' in child
+    assert 'bar' in child
+    assert 'baz' not in child
+
+
 def test_only_and_exclude():
     class MySchema(Schema):
         foo = fields.Field()


Archelyst:
Concerning the syntax: `only=('bla', ('blubb', ('foo', 'bar')))` means "only bla and blubb and of blubb only foo and bar". If anybody has a better idea, it's probably not hard to change it.

sloria:
Thanks for the PR! This is a nice feature. I will review this soon.

Archelyst:
Any thoughts about this one?

marshmallow/schema.py
```diff
@@ -452,7 +452,7 @@ def accessor(cls, func):
 
     ##### Serialization/Deserialization API #####
 
-    def dump(self, obj, many=None, update_fields=True, **kwargs):
+    def dump(self, obj, many=None, update_fields=True, only=None, **kwargs):
```
sloria: I'm not sure about adding `only` as a param here. I think it'd be good to limit the number of params to `dump` and `load`. Otherwise, users might expect all `Meta` options (e.g. `only`, `exclude`, `include`) to be parameters.

sloria:
How well does this approach generalize to `exclude`? This feature was requested in https://github.com/marshmallow-code/marshmallow/issues/151 .

marshmallow/schema.py
```diff
@@ -668,6 +686,11 @@ def _update_fields(self, obj=None, many=False):
         if excludes:
             field_names = field_names - excludes
         ret = self.__filter_fields(field_names, obj, many=many)
+        # Set nested only parameters overriding the only parameters declared on the fields
+        for field_name, field in [(fn, f) for fn, f in ret.items() if hasattr(f, 'only')]:
+            field.only_explicit = None
+            if only_nested and field_name in only_nested:
```
sloria: Minor: I think this can be simplified to `if field_name in only_nested:` , since `only_nested` will always be a `dict`.

sloria:
Initial pass finished. :+1: Sorry for the delay on my review; things have been busy at work.

sloria:
Also, I am bit concerned about the complexity this feature adds to the codebase, which makes me hesitant to merge this. I welcome feedback on whether this is a necessary feature.

marshmallow/schema.py
Archelyst: Ok, that's a good point. I'll see, if I can implement my use cases without this. What's the rationale for specifying `many` here? There are three and a half places where the schema can be configured (Meta, **init**, and dump/load); I don't see the logic yet, what goes where and why.

Archelyst:
> How well does this approach generalize to exclude? This feature was requested in #151 .

I think we could implement the same mechanism for `exclude`, however, from what I read in #151, a nested `only` parameter would suffice. It's just the inverse, it would probably be not too hard to convert any exclude to an only.

> Initial pass finished. :+1: Sorry for the delay on my review; things have been busy at work.

Thanks for the review, man. You've put so much work into this library from which I benefit greatly. No need to apologize :)

Archelyst:
> Also, I am bit concerned about the complexity this feature adds to the codebase, which makes me hesitant to merge this. I welcome feedback on whether this is a necessary feature.

Hm, you're right, the whole part the with figuring out which fields to render is a little tricky. However, the addition at hand is mostly encapsulated in one function. I think, "nested only" is worth the extra complexity.

You've probably seen my PR for marshmallow-jsonapi. IMO it would be big and increase adoption a lot if marshmallow would fully implement jsonapi (as much as serialization is concerned). One part of that is choosing the (nested) fields dynamically. You don't happen to come to Europe any time soon? Would be awesome to chat about how we could approach this. Maybe a little hackaton? :)

dscerri:
Thanks for all your hard work, this library has been of great benefit to me!

I agree that this adds some complexity to the codebase, however I think this feature would be very useful. My current solution involves having two parent schemas which only differ in the included fields in the nested schema, and this adds unnecessary complexity. 

An alternative dot notation was proposed in a previous PR (#180), which was maybe a little cleaner.

Archelyst:
It's probably not too hard to create a little helper that transforms any (dot) syntax into the easier to process list syntax. However, I don't really like the Syntax in PR #180, because adding several nested fields of one nested item is very verbose: `'first_name, last_name, address.street, address.zip_code, address.city'`. Could rather be `'first_name, last_name, address.fields(street, zip_code, city)'`, that's how facebook does it. Anyhow, we could add parsers for both.

Archelyst:
I removed the 'only' parameter from dump(). Now it doesn't look that scary any more :)

marshmallow/fields.py
```diff
@@ -499,14 +500,14 @@ def __init__(self, cls_or_instance, **kwargs):
         if isinstance(cls_or_instance, type):
             if not issubclass(cls_or_instance, FieldABC):
                 raise ValueError('The type of the list elements '
-                                           'must be a subclass of '
-                                           'marshmallow.base.FieldABC')
+                                 'must be a subclass of '
+                                 'marshmallow.base.FieldABC')
```
deckar01: Unnecessary formatting change.

Making formatting changes to code you didn't functionally modify makes a PR much more difficult to review.

I would recommend removing unnecessary formatting changes and squashing most of these commits since this PR has churned so much.

marshmallow/fields.py
```diff
@@ -499,14 +500,14 @@ def __init__(self, cls_or_instance, **kwargs):
...
             self.container = cls_or_instance()
         else:
             if not isinstance(cls_or_instance, FieldABC):
                 raise ValueError('The instances of the list '
-                                           'elements must be of type '
-                                           'marshmallow.base.FieldABC')
+                                 'elements must be of type '
+                                 'marshmallow.base.FieldABC')
```
deckar01: Unnecessary formatting change.

marshmallow/marshalling.py
```diff
@@ -118,9 +118,9 @@ def serialize(self, obj, fields_dict, many=False,
         if many and obj is not None:
             self._pending = True
             ret = [self.serialize(d, fields_dict, many=False,
-                                    dict_class=dict_class, accessor=accessor,
-                                    index=idx, index_errors=index_errors)
-                    for idx, d in enumerate(obj)]
+                                  dict_class=dict_class, accessor=accessor,
+                                  index=idx, index_errors=index_errors)
+                   for idx, d in enumerate(obj)]
```
deckar01: Unnecessary formatting change.

marshmallow/marshalling.py
```diff
@@ -233,9 +233,9 @@ def deserialize(self, data, fields_dict, many=False, partial=False,
         if many and data is not None:
             self._pending = True
             ret = [self.deserialize(d, fields_dict, many=False,
-                        partial=partial, dict_class=dict_class,
-                        index=idx, index_errors=index_errors)
-                    for idx, d in enumerate(data)]
+                       partial=partial, dict_class=dict_class,
+                       index=idx, index_errors=index_errors)
+                   for idx, d in enumerate(data)]
```
deckar01: Unnecessary formatting change.

marshmallow/schema.py
```diff
@@ -318,7 +318,7 @@ class Meta:
         - ``index_errors``: If `True`, errors dictionaries will include the index
             of invalid items in a collection.
         - ``load_only``: Tuple or list of fields to exclude from serialized results.
-        - ``dump_only``: Tuple or list of fields to exclude from deserialization
+        - ``dump_only``: Tuple or list of fields to exclude from deserialization.
```
deckar01: Unnecessary formatting change.

marshmallow/schema.py
```diff
@@ -471,8 +471,8 @@ def dump(self, obj, many=None, update_fields=True, **kwargs):
         many = self.many if many is None else bool(many)
         if not many and utils.is_collection(obj) and not utils.is_keyed_tuple(obj):
             warnings.warn('Implicit collection handling is deprecated. Set '
-                            'many=True to serialize a collection.',
-                            category=DeprecationWarning)
+                          'many=True to serialize a collection.',
+                          category=DeprecationWarning)
```
deckar01: Unnecessary formatting change.

marshmallow/fields.py
```diff
@@ -938,7 +938,6 @@ def _serialize(self, value, attr, obj):
             return value.isoformat()
         except AttributeError:
             self.fail('format', input=value)
-        return value
```
deckar01, line 941: Why was this removed? It seems unrelated to the feature.

This was returning `value` if an exception other than `AttributeError` occurred.

marshmallow/fields.py
```diff
@@ -915,7 +916,6 @@ def _deserialize(self, value, attr, data):
         """Deserialize an ISO8601-formatted time to a :class:`datetime.time` object."""
         if not value:   # falsy values are invalid
             self.fail('invalid')
-            raise err
```
deckar01, line 918: Why was this removed? It seems unrelated to the feature.

If `err` is not defined, then a separate issue should be created.

Archelyst, line 918: Yeah, this could be removed in an act of bureaucracy and properness :) I'm more in favor of the [Boy Scout rule](bureaucracy). But if it helps merging this PR, sure, I can undo all cleanup. (By the way, I was not looking for things to clean up, I just stumbled over them.)

marshmallow/fields.py
Archelyst: I'll definitely squash the commits once everything is settled here.

marshmallow/fields.py
deckar01, line 918: I can't merge this PR, but I would encourage keeping PRs as concise as possible for the sake of the reviewer.

It is my personal preference to keep unrelated changes out of a PR, although it is generally acceptable to include necissary formatting fixes in files you modified in a separate commit.
http://stackoverflow.com/questions/2214480/committing-when-changing-source-formatting

Removing unrelated lines of code without testing the behavior is probably not a good idea. Opening a separate PR would help keep unrelated changes from getting blocked on each other.

sloria, line 918: Good suggestions, @deckar01 . 

I definitely prefer to have PRs as concise and cohesive as possible. That said, I'm not a stickler on it, and I appreciate when people clean up the code. People are volunteering their valuable time to make these PRs, so I typically don't force people to do too much git "busywork".

sloria:
I think we should move forward with this PR, as this feature has been requested by a number of users by now.

There are a couple things to address before merging:

1) Syntax: `dotted.notation` or `('nested', ('tuple', 'notation'))`?

@Tim-Erwin makes a fair point about boilerplate here: https://github.com/marshmallow-code/marshmallow/pull/405#issuecomment-191287474 . That said, we already use dotted notation when accessing attributes, so using dotted notation for `only` might be more consistent.

``` python
city = fields.Str(attribute='address.city')
```

I'm still up in the air about this issue.

2) Nested `exclude` parameter

If we allow for this notation with `only`, users will expect the same notation to work with `exclude`, so this should be implemented, too.

marshmallow/fields.py
deckar01, line 918: I went ahead and removed comments about formatting nits that didn't have additional discussion.

Thanks for working on this PR @Tim-Erwin.

deckar01:
Although the dotted syntax is more verbose, it is very literal and therefore very readable. Nested tuple syntax representing a dictionary filter does not seem common and seems like it would be difficult to document clearly.

That said, I understand the desire to express a tree concisely and without duplication, so my proposal would be to bake the dotted syntax into the implementation and docs, then provide a utility method that flattens a nested tuple into its corresponding dotted property chains.

`flatPropertyChain('nested', ('tuple', ('notation')))`

would evaluate to

`('nested', 'tuple.notation')`

> Edit: I updated the example, because I had it wrong, which also made me realize it is a little confusing. It is not obvious that the nesting logic doesn't apply until the second level of tuples. There are also valid tuples that are invalid property chains. `('nested', ('tuple',))` is `('nested.tuple',)` or `('nested', 'tuple')`?

sloria:
That's a sound approach, @deckar01. I'm hesitant to support two ways of doing the same thing, though. One compromise would be to support the dotted syntax out-of-the-box, then add an example in the docs and/or examples/ folder using a helper to use the tuple syntax.

deckar01:
One concern that I have is the addition of `Nested.only_explicit`. I'm not completely sure why it is necessary to maintain a separate collection of only fields.

Here is my stab at an implementation: https://github.com/marshmallow-code/marshmallow/compare/dev...deckar01:dotted-nesting-support-402

Is there an issue with applying the `only` and `exclude` options to the nested schema directly?

Archelyst:
Ok, I reverted the commit that introduced the formatting and unrelated changes. I'll make a separate PR for that. Also I squashed everything into one commit.

Concerning the syntax, the nested tuple one seems cleaner to me. It's closer to what's actually needed (a tree) and easier to construct programmatically. Also, it seems more concise compared to what I've seen in your tests. I'd go the other way around and provide a utility to transform dotted notation to tuples. However, I wouldn't strongly insist on that. The dotted notation looks fine as well.

I don't get your statement

> It is not obvious that the nesting logic doesn't apply until the second level of tuples.

Could you give an example for that? The programming is straight forward and does not distinguish between first and following levels.

Archelyst:
> One concern that I have is the addition of Nested.only_explicit. I'm not completely sure why it is necessary to maintain a separate collection of only fields.

Well, `only` can be used during schema init as well as during `dump()`.

deckar01:
> I don't get your statement [about the ambiguity of nested tuples.]
> Could you give an example for that?

`('a', ('b',), 'c') -> ('a', 'b', 'c')` or invalid?

`('a', ('b', ()), 'c') -> ('a', 'c')` or invalid?

`('a', ('b', 'c')) -> ('a', 'b', 'c')` or invalid?

`('a', ('b', ('c',)), ('b', ('d',))) -> ('a', 'b.c', 'b.d')` but currently seems to overwrite to `('a', 'b.d')`.

Dotted strings don't allow representing the same field in more than one way, and invalid dotted strings error naturally with the existing field name validation logic.

> Well, `only` can be used during schema init as well as during `dump()`.

There is no `only` parameter for `Schema.dump()`.

Archelyst:
Ok, I get that there is a way to write invalid tuples.

> > Well, only can be used during schema init as well as during dump().
> 
> There is no only parameter for Schema.dump().

Sorry, that was bs. The reason for `only_explicit` is that you can specify `only` on `NestedField`s which should not be overwritten. Otherwise next time a schema is instantiated, not the `only` param specified on the `NestedField` will be active, but the last one specified by a nested `only` param during schema init.

deckar01:
`update_fields` does seem to suggest `dump` has mutation side effects that can be disabled. It might be a good idea to add a test for this behavior.

Archelyst:
Hm, do you think we should accept side effects in `dump`? That seems counterintuitive to me. `update_fields` is IMHO just an implementation detail (performance improvement for `NestedFields`) that's unfortunately visible from the outside.

deckar01:
[The commit that introduced `update_fields`](https://github.com/marshmallow-code/marshmallow/commit/d39790255d00a7520dbf450a67d91e5f34d698c9) seems to prevent the same schema's fields from being updated more than once. Rather than performance, could this be required to keep recursively nested schema from cycling? It was originally undocumented, but later received [documentation](https://github.com/marshmallow-code/marshmallow/commit/90d11fc4813f6604872c0e53933cf3f07c6e39f4).

Something to keep in mind is that the `Nested` field's underlying `__schema` object [is always a schema instance](https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/fields.py#L394) not a class, so you shouldn't have to worry about `dump()` mutating any classes, just schema instances that may have been reused to construct `Nested` fields.

Archelyst:
`update_fields` is used in the case of `many=True` where it is sufficient to only update the fields once and use that configuration for all items.

`__schema` is an instance, but it shares the fields (instances) with any other instances. So while it is fine to modify the schema instance, we shouldn't modify fields if it's avoidable. And it's avoidable quite easily, why do you object to that?

deckar01:
> we shouldn't modify fields if it's avoidable ... why do you object to that?

I suggest adding a test for this behavior. Technically you are still modifying the fields, but just adding a new property to avoid overriding `only`. 

> `update_fields` does seem to suggest `dump` has mutation side effects that can be disabled. It might be a good idea to add a test for this behavior.

Archelyst:
> I suggest adding a test for this behavior.

Which behavior and what should the test test?

> Technically you are still modifying the fields, but just adding a new property to avoid overriding only.

That's true, but at least one shouldn't override settings a user sets with an operation like `dump()`.

My favorite solution would be to distinguish between fields on a schema class and field instances on a schema instance.

deckar01:
I would test the purpose of `only_explicit` by dumping then asserting that fields reused on another schema have not mutated unexpectedly, and the other schema's dump is not polluted by the first.

It seems that if a `only_explicit` property is not overwritten due to a nested `only` in a second dump, the previous value of `only_explicit` will remain on the field and affect the second dump.

It seems like a lot of complexity has risen out of adding a property to a field that is later inspected by the marshaller to filter fields. I don't think it should be the responsibility of a field to "remember" that it should be dumped. Assigning state info to a persistent field property as a result of a schema operation seems like a poor design choice and `only_explicit` is repeating this pattern.

deckar01:
After looking deeper into it, modifying field instances is not an issue across schema instances, because the `Schema` constructor creates a [deep copy of the `declared_fields` class attribute](https://github.com/marshmallow-code/marshmallow/blob/4c2354c/marshmallow/schema.py#L329) for the instance. It still could be an issue for the field state leaking across multiple `dump()` calls with different arguments on the same schema instance.

Archelyst:
Thanks for analyzing this so thoroughly. I missed the deep copy, will make a test like you suggested as soon as I have some time. 

> It still could be an issue for the field state leaking across multiple dump() calls with different arguments on the same schema instance.

Do you have a specific scenario in mind? I'd include that in the tests then.

@sloria , do we have a decision on the syntax and whether the same should be possible with `exclude`? I'd really like to finish this soon.

deckar01:
> It still could be an issue for the field state leaking across multiple dump() calls with different arguments on the same schema instance.

Since `dump()` doesn't accept an `only` parameter, and field instances aren't shared between schema instances, there is no new field state that is created between `dump()` calls, so it appears that `only_explicit` is unnecessary, an `only` could be applied to the nested field directly.

It is a little misleading that `_update_fields()` is called when the schema is initialized (without a data object), then again in every `dump()` (with the data object it is serializing). Nearly everything that `_update_fields()` does during initialization is repeated unnecessarily during a `dump()` (mainly `__set_field_attrs()`).

That said I think this feature can safely be implemented without modifying `fields.py`. Since we should think of modifications to field instances in `_update_fields()` as initialization logic (which are never changed during a dump), so modifying `only` on a `Nested` field during `_update_fields()` semantically is the same as initializing it with the nested `only` fields supplied to the parent field.

Archelyst:
I agree. The `_update_fields` is a little odd, especially because it's really nothing that should be used form the outside. Well, that's not our concern here (could be addressed for v3) and it seems we can live without `only_explicit`.

deckar01:
Now that I am thinking about it in the context of initialization logic, it would probably be better to strip off the nested field names from `only` and apply them to the nested fields in a private method that is called directly by the `Schema` constructor.  This would avoid introducing additional complexity into the `_update_fields()` method and more explicitly indicate that the `only` parameter is used to override the construction (deep copy) of nested field instances.

Archelyst:
@sloria , let's move on with this. @deckar01 has a [PR with the dot notation](https://github.com/marshmallow-code/marshmallow/pull/468), while I don't agree with the approach (modifying `opts`), it's cleanly implemented and has support for `exclude`. If you wanna go that way, feel free to close this PR and then please merge his. I'd be happy to modify my code here to use the dot-notation and support `exclude` as well. It just would be really nice to come to an end here.

sloria:
I've gone ahead and merged #468 . Thanks so much for your work on this.

PR #408, opened 2016-03-07
Provide nested context in ctor of the nested schema

immerrr:
That way the context is made available to nested on_bind_field customizations.


sloria:
Good change. Thanks!

PR #411, opened 2016-03-08
removed unused attributes from schema base class

### LINKED ISSUE #410: remove root, parent, and name from SchemaABC
In the `SchemaABC` class there are three attributes which I don't see used anywhere:

```
class SchemaABC(object):
    """Abstract base class from which all Schemas inherit."""

    root = None
    parent = None
    name = None
```

Also they make it awkward when adding fields with the very common (even in the tutorials) names `parent` and `name` to a schema. If there is consensus that they can be removed, I can provide a PR.

Archelyst:
Fixes #410

### DIFF
--- a/marshmallow/base.py
+++ b/marshmallow/base.py
@@ -32,10 +32,6 @@ class FieldABC(object):
 class SchemaABC(object):
     """Abstract base class from which all Schemas inherit."""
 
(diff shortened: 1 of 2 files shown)


sloria:
Thanks!

PR #413, opened 2016-03-09
made the 'many' param of Nested fields composable with other fields

Archelyst:
Change how the `many` param of `Nested` fields is used in `__init__` so that the field can be composed together with other fields that have the same parameter. This would helpful for https://github.com/marshmallow-code/marshmallow-sqlalchemy/pull/60


sloria:
Thanks again!

Archelyst:
My pleasure, thanks for merging.

PR #414, opened 2016-03-10
Prefer `load_from` in nested errors about required fields

yumike:
Given this schema

```
class UserSchema(Schema):
    time_registered = fields.DateTime(load_from='timeRegistered', required=True)

class BlogSchema(Schema):
    user = fields.Nested(UserSchema, required=True)
```

loading `{}` and `{"user": {}}` raises errors with different field names:

```
>>> schema = BlogSchema()

>>> schema.load({}).errors
{'user': {'time_registered': [u'Missing data for required field.']}}  # Invalid

>>> schema.load({"user": {}}).errors
{'user': {'timeRegistered': [u'Missing data for required field.']}}  # Valid
```

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -454,17 +454,18 @@ class Nested(Field):
             for field_name, field in self.schema.fields.items():
                 if not field.required:
                     continue
+                error_field_name = field.load_from or field_name
                 if (
                     isinstance(field, Nested) and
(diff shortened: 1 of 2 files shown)


sloria:
Cherry-picked 528c36d into `2.6-line` then forward-ported to `dev`. Thanks for the fix!

PR #415, opened 2016-03-11
Deprecated getargspec

iiogmgo:
I fixed getargspec to signature because DeprecationWarning is occurred in Python 3.5.

### DIFF
--- a/marshmallow/utils.py
+++ b/marshmallow/utils.py
@@ -339,11 +339,11 @@ def get_func_args(func):
     `functools.partial` objects and class-based callables.
     """
     if isinstance(func, functools.partial):
-        return inspect.getargspec(func.func).args
+        return inspect.signature(func.func).args
     if inspect.isfunction(func) or inspect.ismethod(func):
-        return inspect.getargspec(func).args
+        return inspect.signature(func).args
     # Callable class
-    return inspect.getargspec(func.__call__).args
+    return inspect.signature(func.__call__).args
 
 
 def if_none(value, default):
--- a/marshmallow/utils.py
+++ b/marshmallow/utils.py
@@ -339,11 +339,11 @@ def get_func_args(func):
     `functools.partial` objects and class-based callables.
     """
     if isinstance(func, functools.partial):
-        return inspect.signature(func.func).args
+        return list(inspect.signature(func.func).parameters.keys())
     if inspect.isfunction(func) or inspect.ismethod(func):
-        return inspect.signature(func).args
+        return list(inspect.signature(func).parameters.keys())
     # Callable class
-    return inspect.signature(func.__call__).args
+    return list(inspect.signature(func.__call__).parameters.keys())
 
 
 def if_none(value, default):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -210,7 +210,7 @@ def test_from_iso_date(use_dateutil):
     assert_date_equal(result, d)
 
 def test_get_func_args():
-    def f1(self, foo, bar):
+    def f1(foo, bar):
         pass
 
     f2 = partial(f1, 'baz')
@@ -221,4 +221,4 @@ def test_get_func_args():
     f3 = F3()
 
     for func in [f1, f2, f3]:
-        assert utils.get_func_args(func) == ['self', 'foo', 'bar']
+        assert utils.get_func_args(func) == ['foo', 'bar']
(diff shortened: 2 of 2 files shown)


sloria:
Thanks for the PR. `signature` doesn't exist in Py2, so the builds are failing. Can you please fix this up to be py2/3-compatible?

sloria:
This also introduces a backwards-incompatible change in how method signatures are handled. Currently,  the following test passes:

``` python
    class F3(object):
        def __call__(self, foo, bar):
            pass
    f3 = F3()

    for func in [f1, f2, f3]:
        assert utils.get_func_args(func) == ['self', 'foo', 'bar']
```

Using `inspect.signature` will make this test fail.

Not sure what the best solution is. This will require more thought.

deckar01:
I looked into supporting Py2/3 compatibility. It is far from trivial.

> [Django developer group discussing the deprecation](https://groups.google.com/forum/#!topic/django-developers/NZskysjasx8/discussion)
> I ended up abstracting our usage of `inspect.getargspec()` into some utility methods in a django.utils.inspect module. Each method has a separate branch for Python 2 and Python 3. It's not ideal, but I think it's preferable to adding a short-lived dependency or vendored copy [...]

AFAIK `marshamallow` doesn't have any hard dependencies, so adding a back port like [`funcsigs`](https://github.com/testing-cabal/funcsigs), is probably not desirable.

I wanted to replicate the behavior of `getargspec()` with `signature()` to provide backwards compatibility, but I am afraid it is not possible due to the [documented behavior](https://www.python.org/dev/peps/pep-0362/#implementation) of `signature()` 

> [PEP 362 > Implementation](https://www.python.org/dev/peps/pep-0362/#implementation)
> If the object is a bound method, construct and return a new Signature object, with its first parameter (usually self or cls ) removed.

Although the `getmembers()` method allows getting the value of bound arguments like `self`, it does not expose the original name the parameter was given. This makes it impossible to replicate the behavior of `getargspec()` when a signature like `def __call__(me, arg1, arg2)` renames the `self` parameter.

Replicating the behavior of `signature()` with `getargspec()` should be possible, but the solution in the [implementation in the Django PR](https://github.com/django/django/pull/4846/files#diff-661c241427e347ab93c204317d4f68dcR37) seems over simplified. The `self` arg is not always present, which is a case they don't seem to consider.

Looking at the source of [funcsigs](https://github.com/testing-cabal/funcsigs/blob/master/funcsigs/__init__.py#L73) which replicates removing bound methods like `self`, reveals it is classifying the `kind` of the parameters.

``` py
kind = params[0].kind
if kind in (_POSITIONAL_OR_KEYWORD, _POSITIONAL_ONLY):
    # Drop first parameter:
    # '(p1, p2[, ...])' -> '(p2[, ...])'
    params = params[1:]
```

[funcsigs/__init__.py#L73](https://github.com/testing-cabal/funcsigs/blob/master/funcsigs/__init__.py#L73)

In summary, providing compatibility between Python 2 and 3 requires back porting a non-trivial amount of the `signature` implementation for Python 2.

deckar01:
I have a reasonably simple fallback for Python 2 working. I opened a [PR against iiogmgo/deprecated-getargspec](https://github.com/iiogmgo/marshmallow/pull/1/files).

This is basically the same strategy used in the Django PR, but with additional handling for bound methods.
http://stackoverflow.com/questions/53225/how-do-you-check-whether-a-python-method-is-bound-or-not

> Using inspect.signature will make this test fail.

It really should. Consider how this utility is being used currently.

https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/fields.py#L1203

That logic is currently invalid for bound methods, because including `self` in the arg count would make `_call_or_raise` think that the callable can handle a context for the second parameter, when in reality it can only be called with one parameter. There does not seem to be any tests where a `Function` field is constructed with a bound serializer or deserializer that accept a single unbound parameter, but they would currently fail.

I added a [test to the PR for providing a bound method](https://github.com/iiogmgo/marshmallow/pull/1/commits/518473586514e73197dca227df6b3057bf1062ce) with one unbound parameter as the serializer for a `Function` field. On dev it fails with the exception `No context available for Function field 'is_collab'`, but now that bound parameters are ignored it passes.

sloria:
Thanks @deckar01 for looking into this. I'm thinking that the best solution might be to change the behavior of `get_func_args` to match the behavior of `inspect.signature`, i.e. omit `self` for bound methods.

This would happen in marshmallow 3.0, since it's a breaking change.

deckar01:
I went ahead and opened #478 since merging my PR into this PR may not be a viable approach.

PR #416, opened 2016-03-14
Add support for heterogeneous schemas

### LINKED ISSUE #412: Support for object type hierarchies
Consider the following use case: you have a hierarchy of object types each of which can have unique attributes and you want to be able to serialize all them in a generic manner. E.g. you have type Event with subtypes representing particular type of event and having attributes specific to that type of event. I want to return a list of all events thus this list can contain events of different types. I would like to create a schema per event type (the same way I create a subclass per event type) and have a union-like Schema that will encode event type in some attribute on serialization and will use this field to pick particular schema on deserialization. This can also be extended to types that are do not have common predecessor.

``` python
class FooSchema(marshmallow.Schema):
    foo = marshmallow.fields.String(required=True)

class BarSchema(Schema):
    bar = marshmallow.fields.Integer(required=True)
```

What I would like to do is to have mixed list of objects which schemas are above and have them serialize uniformly.

``` python
MyUberSchema().dump([Foo(foo='hello'), Bar(bar=123)], many=True).data
# => [{'type': 'foo', 'foo': 'hello'}, {'type': 'bar', 'bar': 123}]
```

Is there an existing way to do that and/or is this a valid use case ?

maximkulkin:
Adds OneOfSchema class which is a special
schema that can multiplex between several
other schemas allowing to work with
heterogeneous collections (collections of
objects of different types).

### DIFF
--- a/marshmallow/__init__.py
+++ b/marshmallow/__init__.py
@@ -4,6 +4,7 @@ from __future__ import absolute_import
 from marshmallow.schema import (
     Schema,
     SchemaOpts,
+    OneOfSchema,
     MarshalResult,
     UnmarshalResult,
 )
@@ -19,6 +20,7 @@ __author__ = 'Steven Loria'
 __all__ = [
     'Schema',
     'SchemaOpts',
+    'OneOfSchema',
     'validates',
     'validates_schema',
     'pre_dump',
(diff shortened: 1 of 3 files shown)


sloria:
Thanks for the PR. I don't this this sort of custom schema is within the scope of marshmallow core. I suggest putting this in its own library, or perhaps sending a PR to [marshmallow-schema](https://github.com/Bachmann1234/marshmallow-polyfield)

PR #418, opened 2016-03-16
Pass through user ValidationError kwargs

russelldavies:
I noticed that in `ValidationError` there is a `kwargs` dictionary, which seems to be for user arguments. I thought I'd use it to add a `status_code` kwarg so that when I catch the exception in my code I could add a bit more semantic information to the response. However, because of how the schema loading works, it is discarded.

This patch fixes this but I'm not sure it's clean enough to get merged. Perhaps a more extensive refactoring is required.

### DIFF
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -35,6 +35,8 @@ class ErrorStore(object):
         self.error_field_names = []
         #: True while (de)serializing a collection
         self._pending = False
+        #: Dictionary of extra kwargs from user raised exception
+        self.kwargs = {}
 
     def reset_errors(self):
         self.errors = {}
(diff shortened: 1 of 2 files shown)


sloria:
Thanks for the PR. This is a known issue.

This issue is the likely the cause of https://github.com/sloria/webargs/issues/85.

I'll test this out when I get some time.

marshmallow/marshalling.py
```diff
@@ -181,6 +183,7 @@ def run_validator(self, validator_func, output,
                 raise ValidationError(self.default_schema_validation_error)
         except ValidationError as err:
             errors = self.get_errors(index=index)
+            self.kwargs = err.kwargs
```
sloria: Since multiple `ValidationErrors` may be raised during validation, it might make sense to do `self.kwargs.update(err.kwargs)`.

russelldavies: Good point, I've updated the pull request.

sloria:
A more complete implementation of this is in #428.

lafrech:
@russelldavies any thoughts on https://github.com/marshmallow-code/marshmallow/issues/996?

russelldavies:
@lafrech I haven't used this library for a long time now so can't comment.

PR #420, opened 2016-03-21
Option to treat empty strings as null values

cmanallen:
Empty strings are transformed to `None` during deserialization if the appropriate options are set.

**Changes:**
- Adds a schema option for interpreting ALL empty strings (`""`) as `None`.
- Adds a field option for interpreting field specific empty strings (`""`) as `None`.

**Reasoning:**
Clients will often send `""` when a field value has been deleted by the user or the user has tabbed through the form inputs without entering any information.  These inputs have the intention of being null but not the appropriate type.  Having an `empty_as_none` option allows us to use pre-existing marshmallow validators rather than validating after the fact.


sloria:
Thanks for the PR.

I'm going to hold off on merging this for now. You can achieve the desired behavior either by (1) having your client-side code encode empty strings as `null` before they are sent to marshmallow or (2) using a `pre_load` method to convert empty strings to `None`.

PR #422, opened 2016-03-26
Optimize cython

BrandonHoffman:
added slots to classes and made changes that will compile for cython. There were a few things that had to change worth noteing:

1) _creation_index was being used as both a class level and instance level variable which cause problems when adding slots. For this reason I just changed the instance level _creation_index to just be creation_index and switched the places referring to it to now refer to the new variable.

2) There was some code that said 'raise err' but there was no err defined. It appears like it would never get hit which was why it was not causing problems before. However with the cython compilation change this line made the code not compile. So that was removed. 

everything else was just required for the slots changes or compilation.

### DIFF
--- a/marshmallow/validate.py
+++ b/marshmallow/validate.py
@@ -17,6 +17,7 @@ class Validator(object):
         This class does not provide any behavior. It is only used to
         add a useful `__repr__` implementation for validators.
     """
+    __slots__ = ()
 
     def __repr__(self):
         args = self._repr_args()
@@ -44,6 +45,7 @@ class URL(Validator):
         ``ftp``, and ``ftps`` are allowed.
     """
 
+    __slots__ = ('relative', 'error', 'schemes')
     URL_REGEX = re.compile(
         r'^(?:[a-z0-9\.\-\+]*)://'  # scheme is validated separately
         r'(?:[^:@]+?:[^:@]*?@|)'  # basic auth
@@ -107,6 +109,7 @@ class Email(Validator):
         interpolated with `{input}`.
     """
 
+    __slots__ = ('error')
     USER_REGEX = re.compile(
         r"(^[-!#$%&'*+/=?^`{}|~\w]+(\.[-!#$%&'*+/=?^`{}|~\w]+)*$"  # dot-atom
         # quoted-string
@@ -171,6 +174,7 @@ class Range(Validator):
         Can be interpolated with `{input}`, `{min}` and `{max}`.
     """
 
+    __slots__ = ('min', 'max', 'error')
     message_min = 'Must be at least {min}.'
     message_max = 'Must be at most {max}.'
     message_all = 'Must be between {min} and {max}.'
@@ -213,6 +217,7 @@ class Length(Range):
         Can be interpolated with `{input}`, `{min}` and `{max}`.
     """
 
+    __slots__ = ('equal')
     message_min = 'Shorter than minimum length {min}.'
     message_max = 'Longer than maximum length {max}.'
     message_all = 'Length must be between {min} and {max}.'
@@ -263,6 +268,7 @@ class Equal(Validator):
         Can be interpolated with `{input}` and `{other}`.
     """
 
+    __slots__ = ('comparable', 'error')
     default_message = 'Must be equal to {other}.'
 
     def __init__(self, comparable, error=None):
@@ -292,6 +298,7 @@ class Regexp(Validator):
         Can be interpolated with `{input}` and `{regex}`.
     """
 
+    __slots__ = ('regex', 'error')
     default_message = 'String does not match expected pattern.'
 
     def __init__(self, regex, flags=0, error=None):
@@ -323,6 +330,7 @@ class Predicate(Validator):
     :param kwargs: Additional keyword arguments to pass to the method.
     """
 
+    __slots__ = ('method', 'error', 'kwargs')
     default_message = 'Invalid input.'
 
     def __init__(self, method, error=None, **kwargs):
@@ -353,6 +361,7 @@ class NoneOf(Validator):
         interpolated using `{input}` and `{values}`.
     """
 
+    __slots__ = ('iterable', 'values_text', 'error')
     default_message = 'Invalid input.'
 
     def __init__(self, iterable, error=None):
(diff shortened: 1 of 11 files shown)


sloria:
This is great! Thanks for the PR. I am on vacation out of the country. I will review this when I am back.

BrandonHoffman:
Sounds good, I'll keep an eye out for a response.

justanr:
To be curious, what benefits does the slots change bring other than memory efficiency? Is this something Cython uses?

BrandonHoffman:
Slots are not cython specifically. For the most part it is to save RAM space which can lead to much faster applications especially any long running application or application that uses a lot of data.

BrandonHoffman:
Just as a note these changes were made with the python hug framework in mind (https://github.com/timothycrosley/hug). We just recently added integrated support for marshmallow types. Being that it is a long running web framework we are putting a lot of effort in minimizing data usage and maximizing speed. Slots and cython compilation being two of the main optimizations.

sloria:
Is there a way we can test the Cython compilation on Travis?

BrandonHoffman:
I believe py26 - py35 in tox are all different Cython versions which should now be using the compiled versions.

BrandonHoffman:
Yah we should just be able to add cython as a dependency into a tox environment. if the cython library is available it would run using the cython compiled version. should be easy enough to add a new endpoint with that added dependency.

marshmallow/validate.py
```diff
@@ -107,6 +109,7 @@ class Email(Validator):
         interpolated with `{input}`.
     """
 
+    __slots__ = ('error')
```
ionelmc, line 112: This is missing a comma. Not sure if `__slots__` works with just a string.

ionelmc, line 112: Actually it's fine ([it's automatically converted to tuple](https://github.com/python/cpython/blob/2.7/Objects/typeobject.c#L2197)). Still a bit inconsistent style tho :-)

sloria:
@BrandonHoffman The travis builds don't currently run tox (although I am open to changing that). Could we just add it to the travis matrix like so?

``` yaml
# Config file for automatic testing at travis-ci.org

language: python
# http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/
sudo: false

env:
  - CYTHON=true
  - CYTHON=false

python:
  - "3.5"
  - "3.4"
  - "3.3"
  - "2.7"
  - "2.6"
  - "pypy"

before_install:
  - pip install -U pip

install:
  - pip install -U .[reco]
  - pip install -U -r dev-requirements.txt
  - if [[ $CYTHON == true ]]; then pip install cython

script: invoke test
```

sloria:
Would someone be willing to do benchmarks to see how much of a difference the `__slots__` change makes?

ionelmc:
Is there a well balanced schema to run benchmarks on?

sloria:
A schema that includes a combination of simple fields, `Nested` fields, and fields with validators should suffice.

BrandonHoffman:
Yah, that should work. As long as it gets installed via pip.

BrandonHoffman:
I did some benchmarking a while ago not specifically tied to marshmallow though. test was done on a mac running python 2.7 and python 3.3. If you want to do some more marshmallow specific benchmarking then hopefully this at least provides a good starting point. the python files are attached for my benchmarks (I had to change the extension from .py to .txt to attach them)

[test.txt](https://github.com/marshmallow-code/marshmallow/files/221188/test.txt)
[test_slots.txt](https://github.com/marshmallow-code/marshmallow/files/221187/test_slots.txt)

______________results______________
time python test.py && time python test_slots.py 
peak memory usage: 44199936

real    0m0.287s
user    0m0.177s
sys     0m0.051s
peak memory usage: 15745024

real    0m0.129s
user    0m0.110s
sys     0m0.016s

time python3.3 test.py && time python3.3 test_slots.py 

peak memory usage: 24846336

real    0m0.260s
user    0m0.159s
sys     0m0.041s
peak memory usage: 14241792

real    0m0.139s
user    0m0.122s
sys     0m0.014s

sloria:
Thanks @BrandonHoffman . That is certainly a good starting point.

So, the things that need to happen before this can get merged are
- [ ] Add Cython compilation to Travis builds
- [ ] Marshmallow benchmarks

Also, I'm on the fence on whether to consider this a breaking change. Using `__slots__` will prevent users from setting arbitrary attributes, which could break some users' code.

BrandonHoffman:
I have not had much time to look into these things much but I'll try to get to them this weekend. As far as the breaking change piece, there is an option to add **dict** to the slots tupple which allows for arbitrary properties to get set on the object still but does decrease some of the speed/memory improvements of slots. Looking through the classes I added slots to I cannot imagine why someone would be adding arbitrary properties. With that said though there is the option to add that **dict** to some or all of the classes if this is a big concern for you. I'll try to include benchmarks with and without the **dict** in my benchmarking. 

Also something that is worth noting is that any subclasses of a class with slots by default revert to the none slot behavior. In other words even without including **dict** in the base class any subclasses will be able to have arbitrary properties as normal.

sloria:
We actually ran into a case for setting an arbitrary attribute here: https://github.com/marshmallow-code/marshmallow-jsonapi/pull/40#discussion_r60829397 . It isn't necessary, and users probably should not set arbitrary attributes, but some might consider this a breaking change. I don't have a strong opinion on this. Any feedback is welcome.

I look forward to the benchmarks. Thanks @BrandonHoffman for your hard work on this.

Archelyst:
`__slots__` are helpful when used with classes that get instantiated a lot, I mean really if there are tons of instances. Which of the classes in marshmallow are instantiated that often? If there is any, I'd propose to only set `__slots__` on those specific ones. Benchmarks will show...

Tim-Erwin:
By the way: we can use slots and still assign arbitrary fields, if we add '__dict__' to the respective slots.

rowillia:
@BrandonHoffman @sloria https://github.com/marshmallow-code/marshmallow/pull/573 might be of interest to you.  I'm able to get a 10-20x speedup by codegenning the deserialize methods.

sloria:
Closing this for now. @BrandonHoffman Please feel free to reopen if you are still interested in this.

PR #425, opened 2016-03-31
Add function name in error message when @validates_schema raises a generic error

daniloakamine:
I was getting an error **'NoneType' object is not callable** in my schema.
I have lost a lot of time trying to find this error, after much time I find out that one of my functions decorated by @validates_schema was wrong...

This change could help to find out problems quickly.
What do you think?

~~PS: I'm not sure if `validator_func` will always have a value or it can be `None`.~~

### DIFF
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -10,6 +10,9 @@ and from primitive types.
 
 from __future__ import unicode_literals
 
+import inspect
+import functools
+
 from marshmallow.utils import is_collection, missing
 from marshmallow.compat import text_type, iteritems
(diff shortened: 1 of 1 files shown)


sloria:
I don't really like the idea of catching all exceptions raised by schema validators. What made your particular `TypeError` difficult to debug? Can you post the traceback here?

daniloakamine-luizalabs:
Let it go... It's really not necessary.
I will configure my logging better, thanks!

daniloakamine:
Let it go... It's really not necessary.
I will configure my logging better, thanks!

Issue #434, opened 2016-04-15
Decimal objects should respect as_string argument at either the validation or deserialization stages


davidthornton:
The `Decimal` object subclasses `Number` which establishes an `as_string` constructor parameter in order to serialize values as strings.

While this is a good idea and probably caters to a canonical use-case in serialization, there is the potential for unexpected behaviour in deserialization and/or validation (when using the `Decimal` object, specifically).

This is because the Python `decimal.Decimal()` method constructs values from integers, strings, floats, or tuples; and (per https://docs.python.org/2.7/library/decimal.html#quick-start-tutorial):

> Construction from an integer or a float performs an exact conversion of the value of that integer or float

From the Python documentation, even with `as_string` set, the current behaviour is to deserialize by passing directly to `decimal.Decimal()`, causing a situation like this:

```
>>> Decimal('3.14')
Decimal('3.14')
>>> Decimal(3.14)
Decimal('3.140000000000000124344978758017532527446746826171875')
```

This behaviour is at best arcane but at worst unexpected or wrong regardless of whether `as_string` is set.

I propose that when `as_string` _is_ set to true, either the `_deserialize` method implicitly casts the value to a string before passing to `decimal.Decimal()` or non-string values do not pass validation. Legacy behaviour would be preserved when `as_string` is false.

I prefer the implicit coercion to a string and can submit a small pull request to implement this. Failing validation is a bit more difficult because you have to differentiate between `str` and `unicode` - but I have presented it here as a potential solution to the problematic behaviour for comparison.

deckar01:
A test for this issue would help prevent this behavior from regressing.

```py
def test_decimal_field_deserialization_float(self):
    m1 = 3.14

    field = fields.Decimal()
    assert isinstance(field.deserialize(m1), decimal.Decimal)
    assert field.deserialize(m1) == decimal.Decimal('3.14')
```

PR #435, opened 2016-04-15
Always coerce values to string in Decimal object's deserialization method

davidthornton:
When `as_string` is set to true, the `_deserialize` method will implicitly cast the value to a string before passing to `decimal.Decimal()`.

This preserves legacy behaviour when `as_string` is false, and prevents unexpected `float` to `decimal.Decimal()` conversions when it is true.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -65,3 +65,4 @@ Contributors (chronological)
 - Mike Yumatov `@yumike <https://github.com/yumike>`_
 - Tim Mundt `@Tim-Erwin <https://github.com/Tim-Erwin>`_
 - Russell Davies `@russelldavies <https://github.com/russelldavies>`_
+- David Thornton `@davidthornton <https://github.com/davidthornton>`_
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -659,6 +659,16 @@ class Decimal(Number):
         a JSON library that can handle decimals, such as `simplejson`, or serialize
         to a string by passing ``as_string=True``.
 
+    .. warning::
+
+        If a JSON `float` value is passed to this field for deserialization it will
+        first be cast to its corresponding `string` value before being deserialized
+        to a `decimal.Decimal` object. The default `__str__` implementation of the
+        built-in Python `float` type may apply a destructive transformation upon
+        its input data and therefore cannot be relied upon to preserve precision.
+        To avoid this, you can instead pass a JSON `string` to be deserialized
+        directly.
+
     :param int places: How many decimal places to quantize the value. If `None`, does
         not quantize the value.
     :param rounding: How to round the value during quantize, for example
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -690,7 +690,7 @@ class Decimal(Number):
         if value is None:
             return None
 
-        num = decimal.Decimal(value)
+        num = decimal.Decimal(str(value))
 
         if self.allow_nan:
             if num.is_nan():
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -88,8 +88,9 @@ class TestFieldDeserialization:
         m1 = 12
         m2 = '12.355'
         m3 = decimal.Decimal(1)
-        m4 = 'abc'
-        m5 = [1, 2]
+        m4 = 3.14
+        m5 = 'abc'
+        m6 = [1, 2]
 
         field = fields.Decimal()
         assert isinstance(field.deserialize(m1), decimal.Decimal)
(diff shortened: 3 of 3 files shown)


sloria:
Thanks for the PR @davidthornton . I agree with the premise of this PR and the corresponding issue; I think it is correct to cast to a string before constructing the `Decimal`. `as_string` is only meant to affect serialization, however. I'm thinking we should always cast to string during deserialization, regardless of the value of `as_string`. 

Thoughts?

davidthornton:
The reason I didn't submit a pull request with unconditional deserialization was that I was not absolutely sure that there did not exist a valid reason to deserialize a float. I mean it definitely is super unintuitive, but is there a reason `decimal.Decimal()` accepts a float? Secondarily, is that a valid use case for HTTP transactions? I would tend against unnecessarily losing the generality of Marshmallow unless there is a good reason.

On a potentially unrelated note, as an uninitiated user I did miss the idea behind `as_string` as applying only at serialization time. However, upon going back through the documentation I can see this idea coming through strongly. Not breaking the trend with this idea may be a good enough reason to deserialize decimals independently of the value of `as_string`.

I'm very interested on your thoughts on the foregoing, and actively committed to amending the pull request to reflect the perfect semantics of this use case :)

sloria:
I see your point about preserving generality. In this case, though, I lean towards "Doing The Right Thing ™"--as you've suggested in #434 , the behavior when deserializing a float will be unexpected for the vast majority of users.

It's worth noting that django-rest-framework and colander both implement this behavior:
- https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/fields.py#L924
- https://github.com/Pylons/colander/blob/master/colander/__init__.py#L1366

deckar01:
A test for this issue would help prevent this behavior from regressing.

``` py
def test_decimal_field_deserialization_float(self):
    m1 = 3.14

    field = fields.Decimal()
    assert isinstance(field.deserialize(m1), decimal.Decimal)
    assert field.deserialize(m1).as_tuple() == (0, (3, 1, 4), -2)
```

davidthornton:
Hi guys, I've gone ahead and amended the pull request to prevent this regressive behaviour in the future.

As an aside, is it clear why this behaviour when dealing with floating point values is actually part of the Python `decimal.Decimal()` API in the first place?

deckar01:
> If value is a float, the binary floating point value is losslessly converted to its exact decimal equivalent. This conversion can often require 53 or more digits of precision. For example, Decimal(float('1.1')) converts to Decimal('1.100000000000000088817841970012523233890533447265625').

https://docs.python.org/2/library/decimal.html#decimal.Decimal

Since there is no way to distinguish between `1.1` and `1.100000000000000088817841970012523233890533447265625` the safest thing to do is preserve the precision.

It appears that a the `__str__` implementation for the builtin `float` type truncates the value to 12 significant digits.

``` py
str(1.10000000001)
# '1.10000000001'
str(1.100000000001)
# '1.1'
str(0.0000000000123456789012)
# '1.23456789012e-11'
str(0.00000000001234567890123)
# '1.23456789012e-11'
```

This seems like a potentially destructive transformation. Maybe it would be better to keep the existing behavior, document the existing behavior, and advise against passing decimal values as float literals.

On the other hand, since the built in float type can only guarantee about 12 digits of precision, it seems like a reasonable compromise to truncate the ambiguous precision.

The only use case I am concerned about is a schema that a developer may not have control over, but must deserialize with marshmallow. If a decimal value requires more that 12 digits of precision, but is supplied as a float literal, this truncation is almost guaranteed to introduce more error into the value than the error produced by the underlying binary representation.

This change seems to compromise accuracy of all values in favor of the perfection of an arbitrary subset of values.

A more surgical solution would be to inspect the original data provided to `loads()` in order to construct an exact `Decimal` from a string. There seems to be no hope for recovering `float` literals provided to `load()`, so preserving the existing precision seems like the correct behavior.

davidthornton:
I alluded to this by way of my previous comment:

> The reason I didn't submit a pull request with unconditional deserialization was that I was not absolutely sure that there did not exist a valid reason to deserialize a float.

However, what swayed me was the argument that two other libraries work this way explicitly - and it will be the canonical use case nearly all of the time. Another potentially compelling argument is that: absent explicit detail or documentation you'd expect a JSON float sent via HTTP to be interpreted as a string because all HTTP request bodies begin as strings (save for form data) and it would be much less expected to rely on extremely high precision floating point value coercion simply because the argument was sent unquoted as therefore as a "float".

To summarize I believe the best course of action is to defer to the package maintainer in this case.

If anyone thinks we should issue a caveat in the documentation just let me know and I'll insert one.

sloria:
I think casting to a string is the expected behavior and meets the 90% use case, so I think we can move forward with this PR. A caveat in the docs would definitely help. @davidthornton Please go ahead and add this to the PR.

sloria:
👍

Issue #110, opened 2014-12-21
Using dictionaries in ValidationError when using validate()


ksesong:
Calling ValidationError with a dictionary-type argument is not properly propagated to the resulting errors dictionary when using the `validate()` method from a `Schema instance`, despite the [documentation](http://marshmallow.readthedocs.org/en/dev/api_reference.html#marshmallow.exceptions.ValidationError).

The problem originates from the use of [`extend()`](https://github.com/sloria/marshmallow/blob/dev/marshmallow/fields.py#L391) when gathering the different errors. A solution is to wrap each dictionary within a list, i.e. change [how dictionaries are parsed](https://github.com/sloria/marshmallow/blob/dev/marshmallow/exceptions.py#L54), but I don't know how this would entail for the other methods.

Thanks!

sloria:
Yes, the current behavior may be incorrect when raising passing a dict to `ValidationError`. That said, passing a `dict` to a `ValidationError` was meant to be an in internal API, i.e. not used for user validators. What is your use case? What would expect the output to be?

``` python
def validator(val):
    raise ValidationError({'bar': 'err'})

class MySchema(Schema):
    foo = fields.Field(validate=validator)

s = MySchema()
s.validate({'foo': 42})   # what should this return?
# {'foo': ['err']}  ?
# OR
# [{'foo': ['err']}]  ?
```

ksesong:
Supposing I have a list of examples as an input (i.e. in use `schema.validate({'examples': ['random string', 'another random 2 string'...]})`), and I would like to know which elements are not validated: (here with a set of validators):

```
'errors': {
    'examples': [
        {
            'code': 'example_too_long',
            'message': 'Some of the examples are too long.'
            'indexes': [1, 4, 5]
        },
        {
            'code': 'example_containing_a_number',
            'message': 'Some of the examples contain invalid characters.'
            'indexes': [2, 5]
        },
    ]
} 
```

I thought I would raise the error directly with the dictionary, but I had to wrap it within a list.

ksesong:
Another remark which is still related to this issue: when registering some [schema-level validation functions](http://marshmallow.readthedocs.org/en/dev/extending.html#schema-level-validation), if I try to raise a ValidationError with a dict or a list, the argument is transformed to a string:

```
raise ValidationError([{'a':'bb'}], field='language')
```

gives:

```
"errors": {
    "language": [
      "[{'a': 'bb'}]"
    ]
}
```

[...]

Issue #73, opened 2014-12-04
Remove legacy API


sloria:
In version 2.0, the pre-1.0 legacy API will be completely removed from the codebase.

This includes:
- [x] Passing object to be serialized to `Schema` constructor
- [x] `data` and `errors` properties of `Schema`
- [x] The `error` param of `Fields` (still in question)
- [x] `Arbitrary`, `Fixed` and `Price` fields (remove in 2.0)
- [x] `Select` field (remove in 2.0)
- [x] Deprecated function validators. Alias validator classes?
- [x] `context` argument of Method fields? (in question)
- [x] `@Schema.preprocessor`, `@Schema.data_handler`, etc.
- [x] `MarshallingError` and `UnmarshallingError` (remove in 2.1)
- [x] `QuerySelect` and `QuerySelectList` (remove in 2.2)
- [x] `allow_none` and `required` string arguments (remove in 2.2)

EDIT: Updated checklist based on comments.

sloria:
The `Arbitrary`, `Fixed`, and `Price` fields may also be removed in favor of the `Decimal` field.

sloria:
Also, the deprecated function validators in the `validate` module.

philtay:
Also, deprecate/remove the `Select` field in favor of `OneOf` validator.

sloria:
Let's do this for the final release of 2.0.0 (rather than in 2.0.0-a) to ease the transition to the 2.0 line.

sloria:
Went ahead and removed `Schema.data` and `Schema.errors`, as it is has been deprecated since 1.0.

philtay:
Add the deprecated function validators in the checklist.

sloria:
Done.

PR #442, opened 2016-04-26
Validates schema field errors

### LINKED ISSUE #441: Reporting field errors from whole schema validator
Sometimes validating a field at a time is not enough and you want to validate a combination of fields. The fields could be deeply nested inside your structure. In that case you create a `validate_schema` validator and do all checks there. At the end you want to report errors not for the whole schema, but for particular fields. Documentation for ValidationError states that it accepts 

> String, list, or dictionary of error messages. If a dict, the keys will be field names and the values will be lists of messages

We can probably extrapolate last statement that dictionary values could contain not only list of messages, but also dictionary of nested field errors.

But in reality when you raise a ValidationError with a dictionary it is treated as just another error message (which happen to be a dictionary, not a string). Instead this error dictionary should be deeply merged with errors reported by individual fields.

I have an implementation for that BUT there is one test case (tests.test_decorators.TestValidatesSchemaDecorator.test_passing_original_data) which on it's own tests a different thing, but it showcases passing a dictionary to ValidationError that is not a field-to-error mapping, but rather complex error message data structure:

``` python
    raise ValidationError({'code': 'invalid_field'})
    ...
    assert '_schema' in errors                                                      
    assert len(errors['_schema']) == 1                                              
    assert errors['_schema'][0] == {'code': 'invalid_field'}
```

I believe, this usage is wrong (since documentation states that error messages should be strings, see quote above), the test should be altered to use simple string as an error message, library should be extended to support getting errors for multiple fields from whole-schema validator.

Here is a sample test of what I'm trying to achieve:

``` python
def test_allow_reporting_field_errors_in_schema_validator(self):
    class NestedSchema(Schema):
        baz = fields.Int(required=True)

    class MySchema(Schema):
        foo = fields.Int(required=True)
        bar = fields.Nested(NestedSchema, required=True)
        bam = fields.Int(required=True)

        @validates_schema(skip_on_field_errors=True)
        def consistency_validation(self, data):
            errors = {}
            if data['bar']['baz'] != data['foo']:
                errors['bar'] = {'baz': 'Non-matching value'}

            if data['bam'] > data['foo']:
                errors['bam'] = 'Value should be less than foo'

            if errors:
                raise ValidationError(errors)

    schema = MySchema()
    errors = schema.validate({'foo': 2, 'bar': {'baz': 5}, 'bam': 6})
    assert 'bar' in errors
    assert 'baz' in errors['bar']
    assert errors['bar']['baz'] == 'Non-matching value'
    assert 'bam' in errors
    assert errors['bam'] == 'Value should be less than foo'
```

What do you think ?

PS I understand that deep merging errors have corner cases and I already have an algorithm which IMO does that in a sane way.

maximkulkin:
Improves validation error reporting functionality enabling reporting field errors from whole-schema (`validates_schema`) validator.

Introduces new public function marshmallow.utils.merge_errors which allows deep merging of errors. It can be useful to simplify accumulating errors in users' validators.

Updates validation error documentation to clarify error messages format.

Fixes #441

### DIFF
--- a/tests/test_decorators.py
+++ b/tests/test_decorators.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import pytest
+from collections import namedtuple
 
 from marshmallow import (
     Schema,
@@ -432,6 +433,8 @@ class TestValidatesSchemaDecorator:
 
     def test_passing_original_data(self):
 
+        CustomError = namedtuple('CustomError', ['code'])
+
         class MySchema(Schema):
             foo = fields.Int()
             bar = fields.Int()
@@ -447,7 +450,7 @@ class TestValidatesSchemaDecorator:
                 def check(datum):
                     for key, val in datum.items():
                         if key not in self.fields:
-                            raise ValidationError({'code': 'invalid_field'})
+                            raise ValidationError(CustomError(code='invalid_field'))
                 if many:
                     for each in original_data:
                         check(each)
@@ -458,7 +461,7 @@ class TestValidatesSchemaDecorator:
         errors = schema.validate({'foo': 4, 'baz': 42})
         assert '_schema' in errors
         assert len(errors['_schema']) == 1
-        assert errors['_schema'][0] == {'code': 'invalid_field'}
+        assert errors['_schema'][0] == CustomError(code='invalid_field')
 
         errors = schema.validate({'foo': '4'})
         assert '_schema' in errors
@@ -469,7 +472,7 @@ class TestValidatesSchemaDecorator:
         errors = schema.validate([{'foo': 4, 'baz': 42}], many=True)
         assert '_schema' in errors
         assert len(errors['_schema']) == 1
-        assert errors['_schema'][0] == {'code': 'invalid_field'}
+        assert errors['_schema'][0] == CustomError(code='invalid_field')
 
     # https://github.com/marshmallow-code/marshmallow/issues/273
     def test_allow_arbitrary_field_names_in_error(self):
@@ -530,6 +533,37 @@ class TestValidatesSchemaDecorator:
         errors = schema.validate([{'foo': 3, 'bar': 'not an int'}], many=True)
         assert 'bar' in errors[0]
         assert '_schema' not in errors
+
+    def test_allow_reporting_field_errors_in_schema_validator(self):
+
+        class NestedSchema(Schema):
+            baz = fields.Int(required=True)
+
+        class MySchema(Schema):
+            foo = fields.Int(required=True)
+            bar = fields.Nested(NestedSchema, required=True)
+            bam = fields.Int(required=True)
+
+            @validates_schema(skip_on_field_errors=True)
+            def consistency_validation(self, data):
+                errors = {}
+                if data['bar']['baz'] != data['foo']:
+                    errors['bar'] = {'baz': 'Non-matching value'}
+
+                if data['bam'] > data['foo']:
+                    errors['bam'] = 'Value should be less than foo'
+
+                if errors:
+                    raise ValidationError(errors)
+
+        schema = MySchema()
+        errors = schema.validate({'foo': 2, 'bar': {'baz': 5}, 'bam': 6})
+        assert 'bar' in errors
+        assert 'baz' in errors['bar']
+        assert errors['bar']['baz'] == 'Non-matching value'
+        assert 'bam' in errors
+        assert errors['bam'] == 'Value should be less than foo'
+
 
 def test_decorator_error_handling():
     class ExampleSchema(Schema):
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -10,7 +10,7 @@ and from primitive types.
 
 from __future__ import unicode_literals
 
-from marshmallow.utils import SCHEMA, is_collection, missing
+from marshmallow.utils import SCHEMA, is_collection, merge_errors, missing
 from marshmallow.compat import iteritems
 from marshmallow.exceptions import (
     ValidationError,
@@ -182,29 +182,24 @@ class Unmarshaller(ErrorStore):
         except ValidationError as err:
             errors = self.get_errors(index=index)
             self.error_kwargs.update(err.kwargs)
+
             # Store or reraise errors
             if err.field_names:
-                field_names = err.field_names
-                field_objs = [fields_dict[each] if each in fields_dict else None
-                              for each in field_names]
+                all_errors = {}
+                for field_name in err.field_names:
+                    all_errors[field_name] = err.messages
+
+                self.error_field_names = err.field_names
+                self.error_fields = \
+                    [fields_dict[each] if each in fields_dict else None
+                     for each in err.field_names]
             else:
-                field_names = [SCHEMA]
-                field_objs = []
-            self.error_field_names = field_names
-            self.error_fields = field_objs
-            for field_name in field_names:
-                if isinstance(err.messages, (list, tuple)):
-                    # self.errors[field_name] may be a dict if schemas are nested
-                    if isinstance(errors.get(field_name), dict):
-                        errors[field_name].setdefault(
-                            SCHEMA, []
-                        ).extend(err.messages)
-                    else:
-                        errors.setdefault(field_name, []).extend(err.messages)
-                elif isinstance(err.messages, dict):
-                    errors.setdefault(field_name, []).append(err.messages)
-                else:
-                    errors.setdefault(field_name, []).append(text_type(err))
+                all_errors = err.messages
+
+                self.error_field_names = [SCHEMA]
+                self.error_fields = []
+
+            errors.update(merge_errors(errors, all_errors))
 
     def deserialize(self, data, fields_dict, many=False, partial=False,
             dict_class=dict, index_errors=True, index=None):
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -10,8 +10,8 @@ and from primitive types.
 
 from __future__ import unicode_literals
 
-from marshmallow.utils import is_collection, missing
-from marshmallow.compat import text_type, iteritems
+from marshmallow.utils import SCHEMA, is_collection, missing
+from marshmallow.compat import iteritems
 from marshmallow.exceptions import (
     ValidationError,
 )
@@ -160,9 +160,6 @@ class Marshaller(ErrorStore):
 
     # Make an instance callable
     __call__ = serialize
-
-# Key used for schema-level validation errors
-SCHEMA = '_schema'
 
 class Unmarshaller(ErrorStore):
     """Callable class responsible for deserializing data and storing errors.
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -20,7 +20,7 @@ from marshmallow.exceptions import ValidationError
 from marshmallow.orderedset import OrderedSet
 from marshmallow.decorators import (PRE_DUMP, POST_DUMP, PRE_LOAD, POST_LOAD,
                                     VALIDATES, VALIDATES_SCHEMA)
-from marshmallow.utils import missing
+from marshmallow.utils import missing, merge_errors
 
 
 #: Return type of :meth:`Schema.dump` including serialized data and errors
@@ -656,12 +656,12 @@ class BaseSchema(base.SchemaABC):
                 self._invoke_validators(pass_many=True, data=result, original_data=data, many=many,
                                         field_errors=field_errors)
             except ValidationError as err:
-                errors.update(err.messages)
+                errors = merge_errors(errors, err.messages)
             try:
                 self._invoke_validators(pass_many=False, data=result, original_data=data, many=many,
                                         field_errors=field_errors)
             except ValidationError as err:
-                errors.update(err.messages)
+                errors = merge_errors(errors, err.messages)
         # Run post processors
         if not errors and postprocess:
             try:
@@ -906,14 +906,14 @@ class BaseSchema(base.SchemaABC):
                                                   item, original_data, self.fields, many=many,
                                                   index=idx, pass_original=pass_original)
                     except ValidationError as err:
-                        errors.update(err.messages)
+                        errors = merge_errors(errors, err.messages)
             else:
                 try:
                     self._unmarshal.run_validator(validator,
                                               data, original_data, self.fields, many=many,
                                               pass_original=pass_original)
                 except ValidationError as err:
-                    errors.update(err.messages)
+                    errors = merge_errors(errors, err.messages)
         if errors:
             raise ValidationError(errors)
         return None
--- a/marshmallow/utils.py
+++ b/marshmallow/utils.py
@@ -14,7 +14,11 @@ from decimal import Decimal, ROUND_HALF_EVEN, Context, Inexact
 from email.utils import formatdate, parsedate
 from pprint import pprint as py_pprint
 
-from marshmallow.compat import OrderedDict, binary_type, text_type
+from marshmallow.compat import OrderedDict, binary_type, iteritems, text_type
+
+
+# Key used for schema-level validation errors
+SCHEMA = '_schema'
 
 
 dateutil_available = False
@@ -348,3 +352,58 @@ def get_func_args(func):
 
 def if_none(value, default):
     return value if value is not None else default
+
+
+def merge_errors(errors1, errors2):
+    """Deeply merges two error messages. Error messages can be
+    string, list of strings or dict of error messages (recursively).
+    Format is the same as accepted by :exc:`ValidationError`.
+    Returns new error messages.
+    """
+    if errors1 is None:
+        return errors2
+    elif errors2 is None:
+        return errors1
+
+    if isinstance(errors1, list):
+        if not errors1:
+            return errors2
+
+        if isinstance(errors2, list):
+            return errors1 + errors2
+        elif isinstance(errors2, dict):
+            return dict(
+                errors2,
+                **{SCHEMA: merge_errors(errors1, errors2.get(SCHEMA))}
+            )
+        else:
+            return errors1 + [errors2]
+    elif isinstance(errors1, dict):
+        if isinstance(errors2, list):
+            return dict(
+                errors1,
+                **{SCHEMA: merge_errors(errors1.get(SCHEMA), errors2)}
+            )
+        elif isinstance(errors2, dict):
+            errors = dict(errors1)
+            for k, v in iteritems(errors2):
+                if k in errors:
+                    errors[k] = merge_errors(errors[k], v)
+                else:
+                    errors[k] = v
+            return errors
+        else:
+            return dict(
+                errors1,
+                **{SCHEMA: merge_errors(errors1.get(SCHEMA), errors2)}
+            )
+    else:
+        if isinstance(errors2, list):
+            return [errors1] + errors2 if errors2 else errors1
+        elif isinstance(errors2, dict):
+            return dict(
+                errors2,
+                **{SCHEMA: merge_errors(errors1, errors2.get(SCHEMA))}
+            )
+        else:
+            return [errors1, errors2]
--- a/marshmallow/utils.py
+++ b/marshmallow/utils.py
@@ -14,7 +14,9 @@ from decimal import Decimal, ROUND_HALF_EVEN, Context, Inexact
 from email.utils import formatdate, parsedate
 from pprint import pprint as py_pprint
 
-from marshmallow.compat import OrderedDict, binary_type, iteritems, text_type
+from marshmallow.compat import OrderedDict, binary_type, iteritems, \
+    string_types, text_type
+from marshmallow.exceptions import ValidationError
 
 
 # Key used for schema-level validation errors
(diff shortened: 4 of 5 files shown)


maximkulkin:
Hey guys, any update on this ?

sloria:
Apologies for the delay. I am hesitant to merge this, as it adds significant complexity to the codebase. I would like more feedback on the #441 before merging this.

maximkulkin:
@sloria What parts you think are still complex? IMO having a well tested function that does heavy lifting of unifying error format actually simplifies things. The error merging function is exposed as public API so users can also benefit from it. As of existing library code, it became more compact than before.

tests/test_decorators.py
```diff
@@ -457,7 +457,7 @@ def check(datum):
         errors = schema.validate([{'foo': 4, 'baz': 42}], many=True)
         assert '_schema' in errors
         assert len(errors['_schema']) == 1
-        assert errors['_schema'][0] == {'code': 'invalid_field'}
+        assert errors['_schema'][0] == 'invalid_field'
```
deckar01: Why did these 3 errors change from dictionaries to strings?

deckar01:
It looks like this PR is adding about 60 lines of code to do what was being done in about 20 before. The increase in complexity seems to be that it is combining 2 collections of arbitrary type, instead of adding a collection of error messages to the `error` dictionary as before.

I would not oppose this as a public utility if there was a strong use case, but it seems like a public utility is a little out of scope for #441.

maximkulkin:
@deckar01 Good point on counting lines. Those additional 40 lines is for value added:
1. Even before Marshmallow supported different forms of errors (string, list of strings, dictionary of errors), but this functionality was rather limited (e.g. you couldn't report field errors from whole-schema validation function, although it seem to be possible by raising ValidationError with dictionary of errors. That didn't work).
2. The public function is to simplify building errors inside whole-schema validation function: you just build error for your particular field and merge_errors will handle the rest:

```
@marshmallow.validates_schema
def validate_all(self, data):
    errors = {}
    if data['foo']['bar'] >= data['baz']['bam']:
        errors = marshmallow.merge_errors(
            errors, {'foo': {'bar': 'Should be less than bam'}}
        )
    if data['foo']['quux'] >= data['baz']['bam']:
        errors = marshmallow.merge_errors(
            errors, {'foo': {'quux': 'Should be less than bam'}}
        )
    ...
    if errors:
        raise ValidationError(errors)
```

You could accumulate errors without that function by carefully coding which goes where and building all those dictionaries if you like, but it could be tricky. `merge_errors` is a convenient utility for this job.

deckar01:
- The [`ValidationError`](https://marshmallow.readthedocs.io/en/latest/api_reference.html#marshmallow.exceptions.ValidationError) seems to only support a single error, but supports associating the error with multiple fields.
- The [`@validate_schema`](https://marshmallow.readthedocs.io/en/latest/api_reference.html#marshmallow.decorators.validates_schema) decorator can only handle a single exception that gets raised.
- The `fields` parameter for [`ValidationError`](https://marshmallow.readthedocs.io/en/latest/api_reference.html#marshmallow.exceptions.ValidationError) probably accepts fields regardless of how deeply they are nested, but `field_names` has no way to express nested field names.

A common design pattern is for a validator to return a list of errors. Nested field names are starting to be expressed using [dotted nesting syntax](https://github.com/marshmallow-code/marshmallow/pull/405#issuecomment-220862224).

``` py
@marshmallow.validates_schema
def validate_all(self, data):
    errors = []
    if data['foo']['bar'] < data['baz']['bam']:
        errors.append(ValidationError('Should be less than baz.bam', ['foo.bar']))
    if data['foo']['quux'] < data['baz']['bam']:
        errors.append(ValidationError('Should be less than baz.bam', ['foo.quux']))
    ...
    return errors
```

maximkulkin:
@deckar01 

> The ValidationError seems to only support a single error, but supports associating the error with multiple fields.

If you look into documentation, it says, that first argument to ValidationError is `String, list, or dictionary of error messages. If a dict, the keys will be field names and the values will be lists of messages`. Obviously, it is by design to allow reporting errors for multiple fields.

> The @validate_schema decorator can only handle a single exception that gets raised.

Completely agree, couldn't have said better than this. Although I'm not sure why you have mentioned that in this topic's regard.

> The fields parameter for ValidationError probably accepts fields regardless of how deeply they are nested, but field_names has no way to express nested field names.

Actually, this whole design of ValidationError arguments (apart from the first one) is not clear to me. What is the use case for reporting the same error for multiple fields (which will be copied to all fields). Although in my PR I tried to maintain the same behavior, I think that those arguments are leftovers of some old error reporting design and are not actually used that much.

That code snippet that you've presented is not supported as of now as far as I know. Although I agree that common design pattern for validator is to return list of errors, this is not how it is done right now: value of validation function is not used anywhere, you need to raise ValidationError instead. Reporting errors as an exception has some advantages, e.g. you can mix deserialization with validation and stop deserialization when you encounter error you can not proceed with.

Dotted syntax seem to be the case for specifying attributes for "only" and "exclude" schema parameters, I haven't seen mention of plans to use it for error reporting. Moreover, having errors for particular sub-object grouped together is more convenient than filtering strings by prefixes. Compare 

```
{
  'foo.bar.baz': 'Error 1',
  'foo.bar.bam': 'Error 2',
  'foo.quux': 'Error 3',
}
```

to

```
{
  'foo': {
    'bar': {
      'baz': 'Error 1',
      'bam': 'Error 2',
    },
    'quux': 'Error 3',
  }
}
```

If you want to parse errors and display them in place with edit controls, the second variant is far easier to work with.

deckar01:
> Obviously, it is by design to allow reporting errors for multiple fields.
> 
> Actually, this whole design of ValidationError arguments (apart from the first one) is not clear to me. What is the use case for reporting the same error for multiple fields (which will be copied to all fields).

There don't seem to be any complex examples in the docs or the tests, but here is a possible use case for a single error on multiple fields:

``` py
class PriceSchema(Schema):
    amount = fields.Decimal()
    currency = fields.Str()

class CheapProductSchema(Schema):
    price = fields.Nested(PriceSchema)
    shipping = fields.Nested(PriceSchema)
    tax = fields.Nested(PriceSchema)

    @validates_schema
    def validate_product_is_cheap(self, data):
        if data['price']['amount'] + data['shipping']['amount'] + data['tax']['amount'] > 10:
            raise ValidationError('Product is not cheap', ['price', 'shipping', 'tax'])

    @validates_schema
    def validate_product_currencies_match(self, data):
        if len(set([data['price']['currency'], data['shipping']['currency'], data['tax']['currency']])) > 1:
            raise ValidationError('Product prices must use the same currency', ['price', 'shipping', 'tax'])
```

Notice how all 3 fields share responsibility for the error and the error could be fixed by changing any of the values.

> That code snippet that you've presented is not supported ...
> ... value of validation function is not used anywhere ...

It isn't currently, but I think it would be a developer friendly interface to support the use case you have proposed.

> I haven't seen mention of plans to use [dotted syntax] for error reporting.

I am proposing it. When [the discussion](https://github.com/marshmallow-code/marshmallow/pull/405#issuecomment-220862224) about how to represent nested field names came up for `only` and `exclude`, dotted syntax was preferred over a nested collection.

maximkulkin:
That's an interesting example of using those extra fields in ValidationError, but in the result you will have the same error reported multiple times (on all fields) which in some cases could be confusing for users (e.g. in your example incorrect currency in one price results errors reported for all prices). Also, in case of currency validation, it would be sometimes desirable to report that error for "currency" component of prices, not for price as a whole. Current implementation with fields does not allow that, although with this patch you can still do that if you like:

``` python
    @validates_schema
    def validate_product_currencies_match(self, data):
        errors = {}
        if data['price']['currency'] != data['shipping']['currency']:
            errors = merge_errors(errors, {
                'shipping': {'currency': 'Should be the same as in price'}
            })
        if data['price']['currency'] != data['tax']['currency']:
            errors = merge_errors(errors, {
                'tax': {'currency': 'Should be the same as in price'}
            })
        if errors:
            raise ValidationError(errors)
```

I agree that not using exceptions in validation has some pros, returning errors as a list of ValidationError instances looks awkward (because e.g. you still won't be able to report nested errors). Even in that case, it would be better to return string/list/dictionary of errors and in that case `merge_errors` utility will come in handy.

As of dotted syntax, I like that proposal and I wish you luck in settling it down, but it is completely orthogonal to this case.

deckar01:
@sloria The `merge_errors` utility is introduced to merge the validation error messages of unrelated errors together so they can be raised as single error after the entire schema has been validated. I interpreted the docs as `ValidationError` is meant to express one error involving potentially multiple fields. What do you think about using it for multiple errors? What do you think about using a nested dictionary to express nested fields?

maximkulkin:
@deckar01 You interpret documentation wrong, ValidationError was used to report nested errors long before. Here is code sample:

``` python

class FooSchema(Schema):
    bar = fields.Integer(required=True)

class BarSchema(Schema):
    foo = fields.Nested(FooSchema, required=True)

BarSchema(strict=True).validate({'foo': {'bar': 'abc'}})
# => marshmallow.exceptions.ValidationError: {'foo': {'bar': [u'Not a valid integer.']}}
```

You get ValidationError raised that contains multiple errors (including errors for nested fields).

Before this patch, errors for each particular field could be reported only from one place - marshmallow.Field subclass instance, and there was no need to merge them. But now each field can get errors from either field validator or schema validator(s), that's why `merge_errors` utility was added.

maximkulkin:
@sloria Looks like no more objections, right?

deckar01:
My remaining objections:
- `merge_errors()` does not belong in `utils`, because it is so tightly coupled with `ValidationError`.
- Requiring a utility function to build input for the `ValidationError` constructor is a poor interface.

> ValidationError was used to report nested errors

I will concede that `ValidationError` is in fact being used to represent multiple unrelated field errors to provide a single schema level validation error object. It would have been better if this behavior was given its own class like `SchemaError`, but it didn't so I think the use case is justified.

What do you think about moving the `merge_errors()` logic into the error class and exposing it as an instance method like `add_fields()`?

I am imagining an interface like:

``` py
@marshmallow.validates_schema
def validate_all(self, data):
    errors = ValidationError()
    if data['foo']['bar'] >= data['baz']['bam']:
        error.add_fields({'foo': {'bar': 'Should be less than bam'}})
    if data['foo']['quux'] >= data['baz']['bam']:
        error.add_fields({'foo': {'quux': 'Should be less than bam'}})
    ...
    if not errors.is_empty():
        raise errors
```

I think that would be easier to use as a developer and clearer to document.

maximkulkin:
I agree that some kind of error builder object would be nice (in our code we use such object). As of usage of ValidationError to be that builder - I'm not sure.

Also, in our builder we use a string attribute path to simplify reporting nested errors:

``` python
  errors_builder = ErrorsBuilder()
  errors_builder.add_error('foo/bar/baz', 'Some message')
  print errors_builder.errors
  # => {'foo': {'bar': {'baz': 'Some message'}}}
```

maximkulkin:
@sloria @deckar01 Looks like there is a problem with Travis CI running tests. Is there a known way to fix that?

deckar01:
It seems to have been caused by a breaking change in [`invoke`](https://github.com/pyinvoke/invoke) from v0.12.2 to v0.13.0. I opened an issue to see if they can provide some insight. https://github.com/pyinvoke/invoke/issues/362

A temporary fix would be to pin the version in [`dev-requirements.txt`](https://github.com/marshmallow-code/marshmallow/blob/dev/dev-requirements.txt#L2) to `invoke==0.12.2`.

maximkulkin:
@deckar01 I guess it's not important to have latest version of invoke for this project, so can you put a PR for that so everybody else can rebase?

deckar01:
Invoke fix PR: https://github.com/marshmallow-code/marshmallow/pull/475

maximkulkin:
Ok, I have addressed comments. Can we finish with this?

tests/test_decorators.py
```diff
@@ -518,3 +518,33 @@ def validate_many(self, data, many):
...
+                    errors['bar'] = {'baz': 'Non-matching value'}
+
+                if data['bam'] > data['foo']:
+                    errors['bam'] = 'Value should be less than foo'
+
+                if errors:
+                    raise ValidationError(errors)
+
+        schema = MySchema()
+        errors = schema.validate({'foo': 2, 'bar': {'baz': 5}, 'bam': 6})
```
deckar01, line 560: Before this PR `errors` would have been

``` py
{'_schema': [{'bam': 'Value should be less than foo', 'bar': {'baz': 'Non-matching value'}}]}
```

but now is

``` py
{'bam': 'Value should be less than foo', 'bar': {'baz': 'Non-matching value'}}
```

The `_schema` key is documented in the [Schema Level Validation](https://marshmallow.readthedocs.io/en/dev/extending.html#schema-level-validation) section of the Extending page.

Assigning schema validation errors to specific fields is documented in the [Storing Errors on Specific Fields](https://marshmallow.readthedocs.io/en/dev/extending.html#storing-errors-on-specific-fields) section.

maximkulkin, line 560: Exactly: before this PR, reported errors were not compliant to schema which says that inside lists you could have only strings (not dictionaries). This makes not much sense to report field errors inside "_schema". It looks inconsistent.
This PR fixes that. It still allows reporting errors for "_schema": just raise validation error with string or list of strings and they will be reported as "_schema" errors.

deckar01, line 560: I think you are confusing the documentation for the [`message` parameter of `ValidationError`](https://marshmallow.readthedocs.io/en/dev/api_reference.html#marshmallow.exceptions.ValidationError) with the documentation for the return value of [Schema.load()](https://marshmallow.readthedocs.io/en/dev/api_reference.html#marshmallow.Schema.load). The `errors` value returned by `load()` is a named tuple, and its value is documented in the sections on [Schema Level Validation](https://marshmallow.readthedocs.io/en/dev/extending.html#schema-level-validation) and [Storing Errors on Specific Fields](https://marshmallow.readthedocs.io/en/dev/extending.html#storing-errors-on-specific-fields).

> It looks inconsistent. This PR fixes that.

The "inconsistency" between the two interfaces is fully documented. This PR is a breaking change.

maximkulkin, line 560: I'm not confusing message parameter. What you're saying ("The `errors` value returned by `load()` is a named tuple") is incorrect. The _result_ of `load()` is named tuple out of which `errors` field is a dictionary of errors:

``` python
class FooSchema(marshmallow.Schema):
    bar = marshmallow.fields.String(required=True)

FooSchema().load({'bar': 123})
# => UnmarshalResult(data={}, errors={'bar': [u'Not a valid string.']})
```

So UnmarshalResult is your named tuple while `errors` is a dictionary of errors.

https://marshmallow.readthedocs.io/en/dev/extending.html#schema-level-validation describes what will happen if you report a string error inside function decorated with `validates_schema`. This stays the same with this PR. Storing errors on specific fields also works as before.

| The "inconsistency" between the two interfaces is fully documented. This PR is a breaking change.
Point me into documentation where it shows raising dicts in ValidationError and expecting it to not be treated as nested field errors.

deckar01, line 560: I think I see the issue. [You pointed out](https://github.com/marshmallow-code/marshmallow/issues/441#issuecomment-230115307) that the `messages` attribute says:

> if [it is] a dict, the keys will be field names and the values will be lists of messages.

This documentation is misleading, because it is seems to actually be referring to the way [`Unmarshaller.run_validator()` internally builds `errors`](https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/marshalling.py#L198) by copying the message for each field name, not how you have to build your `messages` dictionary or how it is going to transform your `messages` dictionary. I think this comment probably should have been added to `Schema.load()` considering the implementation.

The implementation of [`Unmarshaller.run_validator()`](https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/marshalling.py#L198) explicitly allows an arbitrary `message` dictionary argument to pass through as an "error message" value.

> Point me into documentation where it shows raising dicts in ValidationError and expecting it to not be treated as nested field errors.

Is it documented that passing a dictionary to the `message` argument of `ValidationError` is supposed to treat the keys as field names? I don't think that the `messages` attribute of `ValidationError` was actually meant to suggest that behavior.

maximkulkin, line 560: >  I don't think that the messages attribute of ValidationError was actually meant to suggest that behavior.

But I do. I don't understand why I need to persuade you to use this feature. If your use cases are fine without it, don't use it. My use cases require that. I do not see a reason to forbid doing that (reporting multiple different errors for different (nested) fields from whole-schema validator).

I haven't seen a better suggestions to do that. The proposed interface and the way resulting errors are reported look consistent.

deckar01, line 560: > I don't understand why I need to persuade you to use this feature.

If this were purely a new feature it would not be an issue, but you had to [modify existing test cases](https://github.com/marshmallow-code/marshmallow/pull/442/files#diff-8107272ea49a7ea4505985b05542ce00R449), which makes this PR backwards incompatible.

You are arguing that the documentation suggests there is a feature you want, but it is not implemented, so it is a bug that needs to be fixed. I am arguing that your interpretation of the documentation conflicts with the documentation for the `message` argument, the `ValidationError` implementation, and the `decorator` tests. One ambiguous footnote is not a strong argument for introducing breaking changes in v2.

> If your use cases are fine without it, don't use it.

The test cases you changed represent a use case that is not "fine without it". Those test cases are supported by a reasonable interpretation of the existing documentation.

> I haven't seen a better suggestions to [report] multiple different errors for different (nested) fields from whole-schema validator

You could break your whole-schema validator up into several schema validators, so that you can raise one `ValidationError` per error message. If you want to be able to assign error messages deeply nested fields, I would be interested in reviewing a PR that adds the dotted nesting syntax to the `fields` argument which would be a backwards compatible change.

maximkulkin, line 560: As I mentioned before (in [original issue description](issues/441)), the test case that I modified look like that feature (using non-string structure as error message) was accidental. The test itself aims to verify completely different aspect of the library (passing original data to validator). If there would be a special test that checked that it is possible to use non-string error messages than fine. But there is no such test case.

As of breaking validator into separate validators, this is not acceptable for me:
1. You still can not report errors for nested fields (only fields of top level object)
2. You can not report errors for list elements (since you can not make library re-run validator multiple times, reporting one error at a time).

maximkulkin:
Of course @deckar01! Why bother when you can just put labels the way you like!! EZ

deckar01:
@maximkulkin I apologize for for any frustration I have caused. I value your opinion and want to see marshmallow satisfy your use case.

I did some detective work :mag: and found that #110 explicitly [added support for arbitrary dictionaries](https://github.com/marshmallow-code/marshmallow/commit/3ca29d046d1aa2117f6b378f5c5a9143e69417ba) in `1.2.0`.

> Fix error handling when passing a `dict` or `list` to a `ValidationError` #110. Thanks @ksesong for reporting.
- [Allow dict to be passed as error in user validator](https://github.com/marshmallow-code/marshmallow/commit/e5e7520d93dc4981eb484dfa3ffab0155f6d05af)
- [Fix handling schema validation errors with dict or list](https://github.com/marshmallow-code/marshmallow/commit/3ca29d046d1aa2117f6b378f5c5a9143e69417ba)
1. These two commits contain explicit tests for the behavior in question.
2. Instead of directly testing the changes in `marshalling.py`, the test cases were added for `Schema.validator`.
3. When the `@validates_schema` decorator was introduced it did not tests this behavior.
4. When the deprecated `Schema.validator` API was removed in #73, the tests were deleted even though the behavior of `marshalling.py` never changed.

maximkulkin:
Ok, now #110 shows that people experienced problem with not being able to report multiple errors long ago (since 1.2.0), but haven't enough knowledge how to design it elegantly. The situation improved since then: with my #382 you can actually have errors for different items in list, so you do not need to implement custom schemas for error messages just to point on particular error location.

That proves the point that this feature is in demand.

deckar01:
I appreciate that you improved how marshmallow handles the specific use case presented in #110 with #382, but I can't advocate breaking a feature in v2 that has been present since v1.2.0.

This PR would force errors to be strings, but it is very useful for errors to be dictionaries that contain machine readable metadata about validation. Taking away this expressivity is a step backwards.

I would encourage you to close this PR and move the discussion to your original issue. I'm positive we can find an interface that satisfies your use case while preserving the functionality other developers rely on.

maximkulkin:
The problem with reporting non-string errors is that it makes resulting errors ambiguous: was at the message itself or just a nested field errors. This is result of solution sloppiness in 1.2.

If you want to report rich error messages, you can wrap them in your own types (other than lists and dicts) and you'll be fine:

``` python
import marshmallow
from collections import namedtuple

MyError = namedtuple('MyError', ['code', 'message'])

class FooSchema(marshmallow.Schema):
    foo = marshmallow.fields.String(required=True)

    @marshmallow.validates_schema
    def validate_schema(self, data):
        raise marshmallow.ValidationError({'foo': MyError(code=123, message='blah')})

print FooSchema().validate({})
# => {'foo': [u'Missing data for required field.', MyError(code=123, message='blah')]}
```

or do this

``` python
class FooSchema(marshmallow.Schema):
    foo = marshmallow.fields.String(required=True)

    @marshmallow.validates_schema
    def validate_schema(self, data):
        raise marshmallow.ValidationError(MyError(code=123, message='blah'))

print FooSchema().validate({})
# => {'foo': [u'Missing data for required field.'], u'_schema': [MyError(code=123, message='blah')]}
```

I propose moving this to next major release for making incompatible changes. I do not see why we need to keep accumulating mistakes instead of fixing them.

sloria:
Closing this for now, as it has become stale. May reopen this after we evaluate #441 and #581 further.

PR #444, opened 2016-04-28
Improve README badges

singingwolfboy:


singingwolfboy:
Ah, these are already addressed by 3a8640f934463909f95f0e9553366614dfacd9dd. Closing.

singingwolfboy:
@sloria: do you want the RST substitution references commit from this PR?

sloria:
I'm +0 on the substitution references for the badges. Sure, they make the plain rst a little nicer to look at, but when are people looking at the plain rst? Also, they add an extra step of navigation to find the links when editing them.

PR #447, opened 2016-05-02
conda recipe for marshmallow

ijstokes:
With a conda recipe it will make it easier for Anaconda users to leverage marshmallow.
But otherwise this provides a conda recipe that can build a conda package.

This was created specifically to support webargs for which marshmallow is a dependency.


sloria:
Thank you for the PR!

I'm not sure I want to maintain a conda recipe and add complexity to the release process at this time. Marshmallow is pure python with no hard dependencies, so installing with conda doesn't bring much benefit over installing with pip.

PR #449, opened 2016-05-03
Set child's parent before on_bind_field

immerrr:
This PR ensures that `field_obj.parent` and `field_obj.name` are set before `on_bind_field` hook is invoked.

Got bitten by this when trying to access `nested.schema` inside `on_bind_field` of the parent which triggered an early instantiation of the schema with parent still unset which meant it did not inherit the parent context (as implemented in #408).


sloria:
Thanks!

sloria:
Since this is the expected behavior and this is a minor change, I'll go ahead and backport this to the 2.7-line.

Issue #453, opened 2016-05-14
dump_only whitelist/inverse / load_only whilst including when dumping


nickw444:
I feel like this kind of touches on #375, but that didn't seem to align with the behaviour that I think many would find useful. Also highly related to #61 

It would be useful to use the meta class (or constructor) to be able to specify a set of fields that are whitelisted for deserialisation. If the field doesn't exist in the iterable, it cannot be deserialised. 

If a user of marshmallow had a large schema class, with many fields, but only wanted 1 of those fields to be deserializable/writable, from my understanding they could either:
- name every field they want as read only in `dump_only` attribute (lots of fields, prone to errors, ie missing a field)
- When loading data, specify `only=('fieldname',)` upon instantiating the schema. (prone to errors, forgetting to put this in when instantiating)

Am I missing something, or is this a shortcoming that could be addressed? 

Also to note, the naming of `load_only` feels like a misnomer without reading the docs in detail - If you specify a field in load_only, it will only be available for loading, not dumping - but taking another perspective to this could imply that it's an iterable of fields that are whitelisted to be loaded, but can still be dumped.

deckar01:
> ... the naming of load_only feels like a misnomer ...

I see what you mean with `load_only`. I assumed the wrong behavior until I looked at the docs.

> **load_only**: A list or tuple of fields to skip during serialization
> **dump_only**: A list or tuple of fields to skip during deserialization, read-only fields

A good way to illustrate why this is a strange interface is to create a schema with a field that is both `load_only` and `dump_only`.

```
class TestSchema(Schema):
    field1 = fields.Field()
    field2 = fields.Field()

    class Meta:
        load_only=('field1',)
        dump_only=('field1',)
```

`field1` will not be serialized nor deserialized in this example. `load_only` actually means "don't dump" and `dump_only` means "don't load".

|  | load | dump |
| --- | --- | --- |
| **True** |  |  |
| **False** | dump_only | load_only |

A clearer terminology might be "force" and "skip" which have clear connotations and no implicit behaviors.

|  | load | dump |
| --- | --- | --- |
| **True** | force_load | force_dump |
| **False** | skip_load | skip_dump |

The only ambiguity is priority, but I believe that in the rest of the library exclusion has priority over inclusion and instances have priority over classes.

nickw444:
I see exactly what you're saying, and I understand the functionality of using `load_only` and `dump_only`. 

[...]

PR #454, opened 2016-05-14
Add a dump_only whitelist attribute 'writable' - #453

nickw444:
This PR reflects the comments made in #453. If this were to become a feature, I feel that a better name, other than `writeable` would be needed, as this is still very ambiguous.


sloria:
Closing this for now. See https://github.com/marshmallow-code/marshmallow/issues/453#issuecomment-287653931 for a possible solution to your use case.

PR #458, opened 2016-05-21
Fix Length validator class inheritance

### LINKED ISSUE #455: Poor validator inheritance
The inheritance model for [`Range`](https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/validate.py#L159) and its subclass [`Length`](https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/validate.py#L201) does not seem semantically valid. This issue arose when [attempting to differentiate the two](https://github.com/marshmallow-code/apispec/pull/73) by duck typing the validators.

`Length` is not really a type of `Range`. `Range` validates numbers, and `Length` validates sequences and collections. This indicates that their domains are disjoint.

It would be more appropriate for the `Length` validator to have an attribute named something like `lengthRange` which is an instance of the `Range` validator.

This issue is equivalent to extending an `Engine` class to create a `Car` class. A `Car` has an `Engine`, but it is not a type of `Engine` itself.

deckar01:
Fixes #455.

Extend the `Length` validator directly from `Validator` instead of from the `Range` validator.
#455 suggested using an instance of the `Range` validator in the `Length` class, but it would have broken backwards compatibility with the way the parameters for custom error messages currently behave. `"{input} {min} {max}"` currently evaluates to `"foo 1 9"`, but would be `"3 1 9"` if `Range` formatted the message.

> Note: The class interface did not change, but the inheritance tree changed from
> `Length → Range → Validator → object` to `Length → Validator → object`.

### DIFF
--- a/marshmallow/validate.py
+++ b/marshmallow/validate.py
@@ -198,7 +198,7 @@ class Range(Validator):
         return value
 
 
-class Length(Range):
+class Length(Validator):
     """Validator which succeeds if the value passed to it has a
     length between a minimum and maximum. Uses len(), so it
     can work for strings, lists, or anything with length.
@@ -225,7 +225,9 @@ class Length(Range):
                 'minimum parameter must not be provided.'
             )
 
-        super(Length, self).__init__(min, max, error)
+        self.min = min
+        self.max = max
+        self.error = error
         self.equal = equal
 
     def _repr_args(self):


sloria:
Thanks for the PR. I'm not sure composing an instance of `Range` into `Length` is necessary; feels a bit overengineered. I don't feel strongly about either way.

Since this is a potentially breaking change, merging this into `3.0`.

PR #463, opened 2016-05-24
Support nested fields in "only" and "exclude" options

deckar01:
- [x] Implements #402 using dotted nesting notation for the `only` option.
- [x] Implements the same nesting notation for the `exclude` option according to [@sloria's recommendation](https://github.com/marshmallow-code/marshmallow/pull/405#issuecomment-220862224).
- [x] Tests
- [ ] Documentation

Punting [example of nested tuple syntax](https://github.com/marshmallow-code/marshmallow/pull/405#issuecomment-221029042) to a separate PR.

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -650,11 +650,13 @@ class BaseSchema(base.SchemaABC):
     def _update_fields(self, obj=None, many=False):
         """Update fields based on the passed in object."""
         if self.only:
+            # Include fields and the parents of nested fields.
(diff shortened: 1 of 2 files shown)


deckar01:
Closing for now since @Tim-Erwin raises a valid point that field instances can be reused, which may lead to unexpected behavior if `dump()` modifies a `Nested` field instance's `only` property.

https://github.com/marshmallow-code/marshmallow/pull/405#issuecomment-222597374

PR #468, opened 2016-06-05
Represent nested only and exclude options using dot delimiters

### LINKED ISSUE #402: Dynamically specify 'only' parameter for nested fields

deckar01:
Redesigned implementation of #463 based on the [discussion in #405](https://github.com/marshmallow-code/marshmallow/pull/405#issuecomment-223774115).

This fixes #402 by applying nested `only` and `exclude` arguments to the `Nested` field instances during the `Schema` construction and normalizing the `only` and `exclude` attributes of the schema to the flat field names previously expected by `Schema.__filter_fields()` and `Marshaller.serialize()`.

### DIFF
--- a/docs/nesting.rst
+++ b/docs/nesting.rst
@@ -82,6 +82,23 @@ You can explicitly specify which attributes of the nested objects you want to se
     #     'author': {'email': u'monty@python.org'}
     # }
 
+You can represent the attributes of deeply nested objects using dot delimiters.
+
+.. code-block:: python
+    :emphasize-lines: 5
+
+    class SiteSchema(Schema):
+        blog = fields.Nested(BlogSchema2)
+
+    schema = SiteSchema(only=['blog.author.email'])
+    result, errors = schema.dump(site)
+    pprint(result)
+    # {
+    #     'blog': {
+    #         'author': {'email': u'monty@python.org'}
+    #     }
+    # }
+
 .. note::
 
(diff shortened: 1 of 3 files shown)


sloria:
This is awesome. Thanks so much to you and @Tim-Erwin for your tremendous work on this.

deckar01:
Great to see this feature land! Can't wait to use it.

Thanks for reviewing @sloria.

PR #476, opened 2016-06-20
Update README.rst

icyc9:
Typo


icyc9:
pprint is a marshmallow method. my bad.

PR #478, opened 2016-06-22
Deprecated getargspec

deckar01:
> DeprecationWarning is occurring in Python 3.5

https://github.com/marshmallow-code/marshmallow/pull/415#issue-140108078

> I'm thinking that the best solution might be to change the behavior of get_func_args to match the behavior of inspect.signature, i.e. omit self for bound methods.
> 
> This would happen in marshmallow 3.0, since it's a breaking change.

https://github.com/marshmallow-code/marshmallow/pull/415#issuecomment-222388642
- Ignore bound arguments returned from `inspect.getargspec()` to match the behavior of `inspect.signature()`.
- Add a test demonstrating how using `get_func_args()` with `inspect.getargspec()` was broken when the serializer had a bound argument.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -210,7 +210,7 @@ def test_from_iso_date(use_dateutil):
     assert_date_equal(result, d)
 
 def test_get_func_args():
-    def f1(self, foo, bar):
+    def f1(foo, bar):
         pass
 
     f2 = partial(f1, 'baz')
@@ -221,4 +221,4 @@ def test_get_func_args():
     f3 = F3()
 
     for func in [f1, f2, f3]:
-        assert utils.get_func_args(func) == ['self', 'foo', 'bar']
+        assert utils.get_func_args(func) == ['foo', 'bar']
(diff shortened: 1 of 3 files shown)


tests/test_utils.py
```diff
@@ -123,42 +136,51 @@ def __marshallable__(self):
...
 
+
 def test_from_datestring():
     d = dt.datetime.now()
     rfc = utils.rfcformat(d)
     iso = d.isoformat()
     assert_date_equal(utils.from_datestring(rfc), d)
     assert_date_equal(utils.from_datestring(iso), d)
 
+
```
sloria: FWIW, we explicitly ignore the "2-line rule" in setup.cfg because it encourages unnecessary whitespace. This is fine, though.

sloria:
This is good. @deckar01 Could you please rebase this on top of the `3.0` branch and send a PR to that branch?

tests/test_utils.py
deckar01: I went ahead and dropped the pep 8 commits that added white space to the tests.

deckar01:
Rebased and opened against `3.0` in #479.

PR #479, opened 2016-06-23
Deprecated getargspec

deckar01:
Replaces #415 and #478.
- Ignore bound arguments returned from `inspect.getargspec()` to match the behavior of `inspect.signature()`.
- Add a test demonstrating how using `get_func_args()` with `inspect.getargspec()` was broken when the serializer had a bound argument.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,5 +1,5 @@
 # Task execution
-invoke
+invoke==0.12.2
 
 # Soft dependencies
 python-dateutil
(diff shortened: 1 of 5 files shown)


deckar01:
@sloria `3.0` doesn't have the pyinvoke fix yet. Should I cherry pick it into this PR or open a separate PR and rebase later?

deckar01:
I went ahead and opened #480 to bring the pyinvoke fixes to `3.0` and rebased this branch so the tests would work.

sloria:
Thanks!

PR #482, opened 2016-06-29
Change the get_value util signature to match built-ins

### LINKED ISSUE #341: Improve the consistence of the `utils.get_value()`
`utils.get_value()` is the entry-point for most of the fields accessor methods.

Its interface (aka. arguments order on a call) is different from a generic and common interface of a getter, be that for a `dict` or an `object`. This means that you can not overwrite this method with a simple assignment of `Field.get_attribute = dict.get` or `Field.get_attribute = getattr`. Instead you will need to create a new method that properly maps the arguments. This needs to be fixed... hopefully in 3.0.

This started here:
https://github.com/marshmallow-code/marshmallow/commit/7cba47481cbfb0f7e8b3fd36b303a419e236812a#commitcomment-14486377

Lets keep this task to remind us fixing it in the next major release.

deckar01:
Fixes #341 by changing the `get_value` utility's signature to match built-in python accessors.

### DIFF
--- a/docs/extending.rst
+++ b/docs/extending.rst
@@ -277,7 +277,7 @@ However, if you want to specify how values are accessed from an object, you can 
 
         # If we know we're only serializing dictionaries, we can
         # use dict.get for all input objects
-        def get_attribute(self, key, obj, default):
+        def get_attribute(self, obj, key, default):
             return obj.get(key, default)
 
         class Meta:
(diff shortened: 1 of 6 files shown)


sloria:
Thanks @deckar01 . The "Overriding how attributes are accessed" section of the docs will need to be updated to be consistent with this change. I can do this when I merge.

sloria:
Oh, scratch that. Didn't catch that you already changed that part of the docs. Good looks.

Issue #484, opened 2016-06-29
`Eqaul(False)` can't validate correctly


pktangyue:
```
online = fields.Bool(required=True, attribute='active', validate=Equal(False))
```

I add this field to my Schema to accept a `False` value, but when I pass a `False`, it will raise `{'online': [u'Invalid value.']}` error.

sloria:
Thanks @pktangyue for reporting. This seems to be a bug. I would gladly review and merge a PR for this!

PR #488, opened 2016-07-04
fix `Eqaul(False)` cant validate correctly

pktangyue:
#484

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -192,7 +192,7 @@ class Field(FieldABC):
         kwargs = {}
         for validator in self.validators:
             try:
-                if validator(value) is False and value is not False:
+                if validator(value) is False:
                     self.fail('validator_failed')
             except ValidationError as err:
                 kwargs.update(err.kwargs)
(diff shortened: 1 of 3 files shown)


deckar01:
It looks like this PR would cause a false positives for a custom validator that returns `False` instead of raising a `ValidationError`.

> Validation functions either return a boolean or raise a ValidationError.

https://marshmallow.readthedocs.io/en/latest/quickstart.html#validation

It looks like the built in validators should not be returning the value.

pktangyue:
@deckar01 Yes, I agree with you.
At the first, I change `__call__` method of `Eqaul` to:

``` python
def __call__(self, value):
  if value != self.comparable:
    raise ValidationError(self._format_error(value))
  return True
```

But I found this change can't pass the test, like `assert validate.Equal('a')('a') == 'a'`.

I couldn't find out why these built in validators return the value.

Maybe I could change all built in validators to return False/True, as well the tests?

deckar01:
I don't think that the validators need to return anything since they already raise an error.

> **either** return a boolean **or** raise a ValidationError. (emphasis added)

Raising an exception implicitly causes the test to fail, so no `assert` would be required.

Thanks for working on this @pktangyue.

sloria:
This bug is fixed in a simpler way in #527.

Issue #472, opened 2016-06-14
Allow errors to be created in post processors


dbertouille:
First a little bit of background in case there is a better way to do this. I'm working with marshmallow, sqlalchemy and marshmallow-sqlalchemy. I'd prefer to have validation done when creating the sqlalchemy models since that is where the fields are defined. The marshmallow schemas are simply using the ModelSchema from marshmallow-sqlalchemy to dynamically populate the fields.

Perhaps we could optionally pass a reference of the errors dictionary to post processors so that they could populate it with validation (or other) errors. 

Thoughts?

sloria:
Raising validation errors in a pre/post-processor is certainly a valid use case. I would certainly review and merge a PR for this.

PR #489, opened 2016-07-05
Catch validation errors raised during post processing.

dbertouille:
Catch validation errors raised during post processing.
Moved post processing before error handling.
Added a simple test.

Issue #472

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -67,3 +67,4 @@ Contributors (chronological)
 - Russell Davies `@russelldavies <https://github.com/russelldavies>`_
 - David Thornton `@davidthornton <https://github.com/davidthornton>`_
 - Vuong Hoang `@vuonghv <https://github.com/vuonghv>`_
+- David Bertouille `@dbertouille <https://github.com/dbertouille>`_
--- a/marshmallow/exceptions.py
+++ b/marshmallow/exceptions.py
@@ -40,6 +40,12 @@ class ValidationError(MarshmallowError):
         self.kwargs = kwargs
         MarshmallowError.__init__(self, message)
 
+    def normalized_messages(self, no_field_name="_schema"):
+        if isinstance(self.messages, dict):
+            return self.messages
+        if len(self.field_names) == 0:
+            return {no_field_name: self.messages}
+        return dict((name, self.messages) for name in self.field_names)
 
 class RegistryError(NameError):
     """Raised when an invalid operation is performed on the serializer
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -369,12 +369,6 @@ class BaseSchema(base.SchemaABC):
                     each.update(self.extra)
             else:
                 data.update(self.extra)
-        if self._marshal.errors:
-            # TODO: Remove self.__error_handler__ in a later release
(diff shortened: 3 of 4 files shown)


sloria:
@dbertouille Could you please update this with the latest dev branch?

dbertouille:
@sloria Ah my bad, done.

sloria:
It would make sense to handle ValidationErrors in any of the `pre_`/`post_`-decorated methods. @dbertouille Could you update this PR to implement that behavior?

dbertouille:
@sloria  The calls to pre/post dump/load will now catch ValidationError exceptions. Was there anything I missed?

sloria:
Thank you for working on this @dbertouille .

There are a few lingering, potential issues:
- What should the behavior be if a string is passed to ValidationError?

``` python

class MySchema(Schema):
    foo = fields.Field()

    @pre_load
    def pre_load(self, data):
        raise ValidationError('pre_load error')

sch = MySchema()
print(sch.load({'foo': 'green'}).errors)
```

Current output with this PR: `['pre_load error']`

I'm wondering if we should instead store this on the `_schema` key, such that the output would be `{'_schema': ['pre_load error']` . Though I'm not 100% sure on this.
- Passing a field name to `ValidationError` doesn't work

In schema validators, you can pass a field name or a list of field names to store the error on in the errors dict. See http://marshmallow.readthedocs.io/en/latest/extending.html#storing-errors-on-specific-fields . This doesn't currently work in this PR:

``` python
class MySchema(Schema):
    foo = fields.Field(validate=OneOf(['red', 'blue']))

    @pre_load
    def pre_load(self, data):
        raise ValidationError('pre_load error', 'foo')

sch = MySchema()
print(sch.load({'foo': 'green'}).errors)
```

Expected output: `{'foo': ['pre_load error']}`
Actual output: `['pre_load error']`

dbertouille:
I've had a chance to look at this again @sloria. I've updated the PR to handle those two cases.

I'm also not 100% sure that we should default the error messages key to '_schema' when nothing is supplied. The only other thing that comes to my mind is to add some other special keys. Something along the lines '_preprocess' or '_postprocess'.

sloria:
@dbertouille Thank you for updating the PR. Looks to be behaving as expected. 👍 

I think adding to `_schema` is a fine default, since errors raised without a field name will be schema-level validation failures. If the user wants to save on a different key, they can do so:

``` python
    @pre_load
    def pre_load(self, data):
        raise ValidationError('pre_load error', '_pre_load')
```

Or even:

``` python
class PreprocessingError(ValidationError):

    def __init__(self, message, *args, **kwargs):
        field_names = kwargs.pop('field_names', '_pre_load')
        super().__init__(message, field_names=field_names, *args, **kwargs)

class MySchema(Schema):
   #...

    @pre_load
    def pre_load(self, data):
        raise PreprocessingError('pre_load error')
```

henryfjordan:
I know this is really old, but I just ran into a subtle issue that broke some of my unit tests in a project relying on marshmallow:

This PR changes the behavior of `dump()` to now not run `@post_dump` methods if an error is encountered. While understandable (and probably more correct), this is a breaking change and probably should be noted in the changelog.

PR #498, opened 2016-07-15
Remove get_from_dict from extending docs.

johbo:
While experimenting with a custom way of accessing attributes I got trapped by the option `accessor` in the class `Meta`. It seems to me that this is not supported anymore.


sloria:
Ah, good catch. Thanks!

Issue #464, opened 2016-05-27
Time field wrong result with microseconds


Archelyst:
I have a schema with a `Time` field. When it serializes a value with microseconds, e.g. `2016-05-27T11:02:35.267899`, it results in  `2016-05-27T1`.

The reason are those lines:

```
if value.microsecond:
    return ret[:12]
```

https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/fields.py#L911

I have no idea what your idea was behind that, @sloria . `ret[:19]` would do it in my case, don't know if there might be other cases.

sloria:
Certainly looks like a bug. I'd gladly review and merge a PR fixing this.

PR #499, opened 2016-07-16
Fix _serialize() of Time field with microseconds

vuonghv:
When a Time object has microsecond value, the _serialize() method
returns the following string format: 'HH:mm:ss.SSSSSS'

Issue #464


sloria:
Ah, I forgot to create a support branch for the 2.9 release line. Sorry about that. I've gone ahead and cherry-picked this into `dev`. Thanks for the fix!

Issue #391, opened 2016-02-11
Field validator as method is buggy/inconsistent


martinstein:
Let's take the example from the docs at http://marshmallow.readthedocs.org/en/latest/quickstart.html#field-validators-as-methods

If you run this:

``` python
from marshmallow import Schema, fields, validates, ValidationError

class ItemSchema(Schema):
    quantity = fields.Integer()

    @validates('quantity')
    def validate_quantity(self, value):
        if value < 0:
            raise ValidationError('Quantity must be greater than 0.')
        if value > 30:
            raise ValidationError('Quantity must not be greater than 30.')

schema = ItemSchema()
result, errors = schema.load({'quantity': 40})
print("result:", result) 
print("errors:", errors)
```

you get the result:

```
result: {'quantity': 40}
errors: {'quantity': ['Quantity must not be greater than 30.']}
```

instead of the empty result `{}` (because validation failed).

Just for comparison, the behavior is correct for validation via lambda-function:

``` python
class ItemSchema2(Schema):
    quantity = fields.Integer(validate=lambda x: 0 <= x <= 30)

schema = ItemSchema2()
result, errors = schema.load({'quantity': 40})
print("result:", result)
print("errors:", errors)
```

This time you get the expected result:

```
result: {}
errors: {'quantity': ['Invalid value.']}
```

sloria:
Thanks for reporting @martinstein . I would certainly review and merge a PR addressing this inconsistency.

jmcarp:
It looks like this inconsistency might get fixed for free once #368 is merged and `validates` is deprecated.

vuonghv:
@sloria @jmcarp 
Do you think we really need a PR to fix this issue because `validates` is deprecated in the future (#368)?

sloria:
Even though `validates` will be deprecated after #368, I would gladly review and merge a PR fixing the inconsistency.

maximkulkin:
Isn't that if validation fails, the contents of result does not make sense and could just be garbage/inconsistent? Where can I find documentation on guarantees of what result will be in case of errors?

vuonghv:
I think that removing the invalid values from `result` will make getting all valid values easier by iterating `result`'s keys. We do not have to check if the key is in `errors` or not.

maximkulkin:
Valid values should be determined by absence of errors for that path. How will removing fields help detecting good values? Do you need ANOTHER validator to distinguish values that have all fields from values that do not have all fields?

[...]

PR #501, opened 2016-07-18
Fix the inconsistency of validates decorator

vuonghv:
Make the result of field validators as methods (through validates decorator) consistent with validations via validate callable (e.g function, lambda, or object with `__call__` defined).

Issue #391

### DIFF
--- a/tests/test_decorators.py
+++ b/tests/test_decorators.py
@@ -285,6 +285,18 @@ class TestValidatesDecorator:
         errors = schema.validate({})
         assert errors == {}
 
+        result, errors = schema.load({'foo': 41})
+        assert errors
+        assert result == {}
+
+        result, errors = schema.load([{'foo': 42}, {'foo': 43}], many=True)
+        assert len(result) == 2
+        assert result[0] == {'foo': 42}
+        assert result[1] == {}
+        assert 1 in errors
+        assert 'foo' in errors[1]
+        assert errors[1]['foo'] == ['The answer to life the universe and everything.']
+
     def test_field_not_present(self):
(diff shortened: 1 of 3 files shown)


tests/test_decorators.py
```diff
@@ -285,6 +285,18 @@ def test_validates_decorator(self):
         errors = schema.validate({})
         assert errors == {}
 
+        result, errors = schema.load({'foo': 41})
+        assert errors
+        assert result == {}
+
+        result, errors = schema.load([{'foo': 42}, {'foo': 43}], many=True)
+        assert len(result) == 2
+        assert result[0]['foo'] == 42
```
maximkulkin: Maybe this is more readable?

``` python
assert result[0] == {'foo': 42}
```

tests/test_decorators.py
```diff
@@ -285,6 +285,18 @@ def test_validates_decorator(self):
...
+        assert result[1] == {}
+        assert 'foo' in errors[1]
+        assert len(errors[1]['foo']) == 1
+        assert errors[1]['foo'][0] == 'The answer to life the universe and everything.'
```
maximkulkin:
``` python
assert 1 in errors
assert 'foo' in errors[1]
assert errors[1]['foo'] == ['The answer to life the universe and everything.']
```

maximkulkin:
What is the sacred benefit of this feature?

vuonghv:
@maximkulkin Thank you for your review, I think your above test code is more readable than mine.
And this PR is not a feature, if fixed the inconsistency in issue #391 .
You can take a look.

PR #518, opened 2016-08-25
Add a new validat'Or

Shir0kamii:
Hi,

I've written a new validator, named `Or` as you can guess by the title. :)

It takes an undetermined number of validator as arguments, and it will try to validate the given value with each validator until one does it with success.

I think it would be a good thing to merge it, since it allows to compose validator easily.

I hope you'll like it!

### DIFF
--- a/marshmallow/validate.py
+++ b/marshmallow/validate.py
@@ -465,3 +465,48 @@ class ContainsOnly(OneOf):
                 del choices[index]
 
         return value
+
+
+class Or(Validator):
+    """Validator which succeeds if ``value`` is validated by at least one of
+    ``validator_list`` elements
+
+    :param validator_list:
+    :param str error: Error message to raise in case of a validation error. Can be
+        interpolated with `{input}`.
+    """
+
+    default_message = "{input!r} can't be validated by any of the validators"
+
+    def __init__(self, *validator_list, error=None, **kwargs):
+        if any(not isinstance(validator, Validator) for validator in
+                validator_list):
+            raise ValueError("This validator accepts only validators as "
+                             "parameters")
+        self.validator_list = validator_list
+        self.error = error or self.default_message
+        super(Or, self).__init__(**kwargs)
+
+    def __call__(self, value):
+        for validator in self.validator_list:
+            try:
+                validator(value)
+            except ValidationError:
+                pass
+            else:
+                return value
+
+        # All validators raised an error
+        raise ValidationError(self._format_error(value))
+
+    def _repr_args(self):
+        return "validator_list={!r}".format(self.validator_list)
+
+    def _format_error(self, value):
+        return self.error.format(input=value)
+
+    def add_validator(self, validator):
+        if not isinstance(validator, Validator):
+            raise ValueError("This validator accepts only validators as "
+                             "parameters")
+        self.validator_list.append(validator)


maximkulkin:
"Just because I can" or there is a use case for that ?

I was thinking of creating a set of validators that is turing complete ("And", "Or", "Not", "If" and validators that can actually modify data), but then I recalled that it is just validators.

A lot of validations are done by declaring methods with `@validates(attr)` decorator. How does this `Or` validator compose with those?

marshmallow/validate.py
```diff
@@ -465,3 +465,48 @@ def __call__(self, value):
...
+    :param validator_list:
+    :param str error: Error message to raise in case of a validation error. Can be
+        interpolated with `{input}`.
+    """
+
+    default_message = "{input!r} can't be validated by any of the validators"
+
+    def __init__(self, *validator_list, error=None, **kwargs):
+        if any(not isinstance(validator, Validator) for validator in
+                validator_list):
```
maximkulkin, line 483: The visually better way to split into lines is to put expression at first line and "for .. " on second:

``` python
if not all(isinstance(validator, Validator)
           for validator in validator_list):
```

marshmallow/validate.py
```diff
@@ -465,3 +465,48 @@ def __call__(self, value):
...
+        # All validators raised an error
+        raise ValidationError(self._format_error(value))
+
+    def _repr_args(self):
+        return "validator_list={!r}".format(self.validator_list)
+
+    def _format_error(self, value):
+        return self.error.format(input=value)
+
+    def add_validator(self, validator):
```
maximkulkin, line 508: Why do you need this? Validators are mostly anonymous objects created during field declaration.
If there is a function to add more, why there is not functions to remove/clear validator list?

marshmallow/validate.py
justanr: Even better, allow it accept any callable that accepts the correct arguments.

Shir0kamii:
In my company, we use a custom decorator on views of our API. It uses `schema.load` on the body of the request. In the representation of our data, we uses a string which can be an empty string `Or` an url. It's surely not the best decision but it is shared across all our backends.

I can think of similar use cases people must have. For exemple, people could have a field 'contact' with an email `Or` a phone number. Not a great decision either, but in fact, a lot of people use one field for multiple purpose.

I have thought about the possibility of adding other logic operators, because it seemed fun to do, but indeed, I don't have use cases for other operators. How could people want a field to be an email `And` something else ?

If you have a better solution for my use case, I'll be happy to take it!

marshmallow/validate.py
Shir0kamii: Alright, I will do it soon!

marshmallow/validate.py
Shir0kamii: I admit I don't _need_ this method. I only thought that people _could_ want to add validators, but if they do it by accessing the list attribute, the validator check wouldn't be done. But as you highlight, I don't think a lot of people use schema in this way. Shall I remove it ?

justanr:
Your very specific usecase wouldn't need an `And` but I can think of quite a few that do. As well as `Not`. 

On the surface, `And` seems redundant, but it's actually useful when nested inside an `Or` -- "I want all if these _or_ all of these". 

And `Not` is most useful negating `Or`/`And` instances

Potentially, there's also the usecase for a `Xor` validator (only one of these may be true, but not all of them). 

I am wondering if this belongs in core marshmallow or in another package.

maximkulkin:
Guys, statement about other boolean validator combinators was a sarcasm, sorry for not putting corresponding label there.

Why wouldn't you just write a validator that matches either empty string or url ? Like

``` python
import marshmallow as m
import marshmallow.fields as mf
import marshmallow.validate as mv

def optional_url(s):
    s == '' or mv.URL()(s)

class MySchema(m.Schema):
    url = mf.String(validate=optional_url)
```

I do not think that it is useful outside of checking for empty string for simple cases. For more complex cases you can easily create custom validator like shown above.

Shir0kamii:
That's the "better solution" I was talking about! I didn't thought about the possibility of defining a simple function and using it as a validator, i feel stupid haha. Now I understand better why the validators use the `__call__` method.

I still think using an `Or` validator is cleaner but your solution is largely enough and beats the added complexity of maintaining new code by far.

Thank you for pointing out that solution!

sloria:
As @maximkulkin mentioned, any combining of validators should happen in a custom validator. That leaves things flexible without having to add any code to marshmallow core.

Issue #521, opened 2016-09-07
Dumps ignores strict for pre_dump/post_dump errors


tvuotila:
After #489 following code does not raise:

```
from marshmallow import Schema, pre_dump, ValidationError

class ExampleSchema(Schema):
    @pre_dump
    def pre_dump_error(self, item):
        errors = {
          'foo' : ['error',]
        }
        raise ValidationError(errors)

schema = ExampleSchema(strict=True)
schema.dumps(object())
```

PR #522, opened 2016-09-07
Fix strict ignored

tvuotila:
Refs marshmallow-code/marshmallow#521

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -483,6 +483,8 @@ class BaseSchema(base.SchemaABC):
         except ValidationError as error:
             errors = error.normalized_messages()
             result = None
+            if self.strict:
+                raise error
         else:
             errors = {}
 
@@ -520,6 +522,8 @@ class BaseSchema(base.SchemaABC):
                     original_data=obj)
             except ValidationError as error:
                 errors.update(error.normalized_messages())
+                if self.strict:
+                    raise error
         if errors:
             # TODO: Remove self.__error_handler__ in a later release
             error_handler = self.handle_error or self.__error_handler__
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -465,6 +465,7 @@ class BaseSchema(base.SchemaABC):
 
         .. versionadded:: 1.0.0
         """
+        errors = {}
         many = self.many if many is None else bool(many)
         if not many and utils.is_collection(obj) and not utils.is_keyed_tuple(obj):
             warnings.warn('Implicit collection handling is deprecated. Set '
@@ -485,8 +486,6 @@ class BaseSchema(base.SchemaABC):
             result = None
             if self.strict:
                 raise error
-        else:
-            errors = {}
 
         if not errors:
             if update_fields:
@@ -508,8 +507,6 @@ class BaseSchema(base.SchemaABC):
                 preresult = error.data
                 if self.strict:
                     raise error
-            else:
-                errors = {}
 
             result = self._postprocess(preresult, many, obj=obj)
 
@@ -521,7 +518,7 @@ class BaseSchema(base.SchemaABC):
                     many,
                     original_data=obj)
             except ValidationError as error:
-                errors.update(error.normalized_messages())
+                errors = error.normalized_messages()
                 if self.strict:
                     raise error
         if errors:
@@ -620,6 +617,7 @@ class BaseSchema(base.SchemaABC):
         :param bool postprocess: Whether to run post_load methods..
         :return: A tuple of the form (`data`, `errors`)
         """
+        errors = {}
         many = self.many if many is None else bool(many)
         if partial is None:
             partial = self.partial
@@ -632,8 +630,6 @@ class BaseSchema(base.SchemaABC):
         except ValidationError as err:
             errors = err.normalized_messages()
             result = None
-        else:
-            errors = {}
         if not errors:
             try:
                 result = self._unmarshal(
@@ -646,8 +642,6 @@ class BaseSchema(base.SchemaABC):
                 )
             except ValidationError as error:
                 result = error.data
-            else:
-                errors = {}
             self._invoke_field_validators(data=result, many=many)
             errors = self._unmarshal.errors
             field_errors = bool(errors)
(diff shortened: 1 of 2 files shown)


marshmallow/schema.py
```diff
@@ -525,6 +525,16 @@ def dump(self, obj, many=None, update_fields=True, **kwargs):
...
             if callable(error_handler):
                 error_handler(errors, obj)
+            exc = ValidationError(
+                errors,
+                field_names=self._marshal.error_field_names,
+                fields=self._marshal.error_fields,
+                data=result,
+                **self._marshal.error_kwargs
+            )
+            self.handle_error(exc, result)
```
maximkulkin: Here error handler is called second time (see above). Also, according to documentation it is a common practice to raise a different kind of error in error handler. So, by calling `handle_error` you're already violating strictness setting. I guess, you do not need to call `handle_error` at all if not strict.

tests/test_decorators.py
```diff
@@ -668,3 +668,77 @@ def make_item(foo, bar):
     assert '_schema' in errors
     assert len(errors['_schema']) == 1
     assert errors['_schema'][0] == 'preloadmsg1'
+    schema = ExampleSchema(strict=True)
```
maximkulkin, line 685: This is just one test nightmare. It would be better to split (at least yours) tests into individual tests making it obvious with test names what invariant you are testing. It will be easier to see guaranteed invariants and figure out what invariants still lack tests.

marshmallow/schema.py
tvuotila: > So, by calling handle_error you're already violating strictness setting. I guess, you do not need to call handle_error at all if not strict.

I have to disagree, documentation says:

> By default, Schema.dump() and Schema.load() will return validation errors as a dictionary (unless strict mode is enabled).

and

> You can specify a custom error-handling function for a Schema by overriding the handle_error method.

See [Handling errors](http://marshmallow.readthedocs.io/en/latest/extending.html#handling-errors).

So, regardless of strictness setting, we will have errors. If we have errors, they should be passed to the error handler.

I think that, it is the responsibility of error handler to handle errors correctly. It might comply with strictness setting or not.

> Here error handler is called second time

Of course, this is a mistake. After fixing this, [handling errors is similar to load method](https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/schema.py#L671-L684).

maximkulkin: Not correct. Yes, you will have errors. With `strict=True` you have to throw them as an exception (regardless of exception type), with `strict=False` you should not throw exception and just return errors. The thing is that if you handle errors with `self.handle_error` there is no guarantee that it won't throw exception. According to code, result of `self.handle_error` is always ignored. Although you might modify original error dictionary, this is not a common case. The common case is to throw an application-specific exception. Which violates strictness in some cases, because `self.strict` is not taken into account.

Anyways, now it makes even less sense: before it used either `self.handle_error` OR `self.__error_handler__`, now you call `self.__error_handler__`, and then `self.handle_error` without even checking if `self.handle_error` is callable.

tests/test_decorators.py
```diff
@@ -668,3 +668,41 @@ def make_item(foo, bar):
...
+        post_dump,
+    ]
+)
+def test_decorator_strict_error_handling_with_dump(decorator):
+    class ExampleSchema(Schema):
+        @decorator
+        def raise_value_error(self, item):
+            raise ValidationError({'foo': 'error'})
+
+    schema = ExampleSchema(strict=True)
```
maximkulkin, line 704: Did you check that it was broken before? Theses tests look like they would work without any modifications. What you really want to check is 
1. With `strict=True` it throws exception from pre/post_\* callbacks (that's how it worked before);
2. (more importantly) Without `strict=False` it does not raise exceptions BUT returns those errors as `errors` field in Marshmal/Unmarshal result.

tvuotila: > Did you check that it was broken before?

Yes

> Theses tests look like they would work without any modifications.

They **should** work. So, there is a bug. See marshmallow-code/marshmallow#521

> With strict=True it throws exception from pre/post_\* callbacks (that's how it worked before);

These tests check exactly that.

> (more importantly) Without strict=False it does not raise exceptions BUT returns those errors as errors field in Marshmal/Unmarshal result.

There are already tests for that https://github.com/marshmallow-code/marshmallow/blob/dev/tests/test_decorators.py#L534-L670

marshmallow/schema.py
tvuotila: > without even checking if self.handle_error is callable.

It is callable, I checked it https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/schema.py#L384
If it is overridden with non-callable, it is an error and "Errors should never pass silently." 🐍 

> now you call self.**error_handler**, and then self.handle_error

I do not see why anybody would have both of those. Keeping calls separate helps to rip out deprecated `__error_handler__` when it is time.

maximkulkin:
Ok, I just dived into the code and found out that it is very fishy. If you look at structure of `dump` method, it first calles pre_dump processors, then dump, then post_dump processors, then analysis of accumulated errors and running error handler. Thus error handler is never called if strict=True (except for pre_dump, which looks like a bug).

Although, dump and post_dump won't be called at all if previous stages have returned errors. And if those two stages return errors, exception will be immediately raised if strict=True. That means, that there is no point of analyzing strictness at last stage as you do, because the error was probably already raised. Except for pre_dump: this one does not raises exception. So, the proper way to fix it is to put `if self.strict: raise error` right after this line https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/schema.py#L483. No need to modify block where you are trying to fix it.

The whole implementation for error handler looks incorrect: it would be better to remove these lines:
https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/schema.py#L507-L508   and raise exception at the bottom of `dump` method.

As of removing deprecation, this should be done by a dedicated commit which removes it from everywhere, not just from a single place it is used. Also, that commit should only go to next major release since it breaks backward compatibility.

tvuotila:
I hope I understood you correctly. I had to add `if self.strict: raise error` to two places, in order to get my tests pass.

> The whole implementation for error handler looks incorrect: it would be better to remove these lines:
> https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/schema.py#L507-L508 and raise exception at the bottom of dump method.

I want the current PR to be a bug fix. It restores the functionality broken between versions 2.9.1 and 2.10.0. I even checked that my tests pass in 2.9.1.

I will create a new PR for "raise exception at the bottom of dump method".

PR #523, opened 2016-09-08
Fixed error in Specifying Serialization Keys quickstart example.

iakbar:
**The issue**
UserSchema has field **email**, but the data object has **CamelCasedEmail**.  The dump method will not find the email field in the data object and will only dump the name field. 

**The fix**
Renamed **CamelCasedEmail** to **email** in the data object


sloria:
Thanks again!

PR #527, opened 2016-09-11
Fix raw False value validation

### LINKED ISSUE #484: `Eqaul(False)` can't validate correctly

tuukkamustonen:
Fixes #484.

Validator's return value is now evaluated only when using "ad-hoc" callable
(lambda, function or method). With `Validator` instances, return value has no
meaning (instead, `ValidationError` are thrown) so set to ignore it when using
`Validator` sub-classes.

Would you like if-else block instead of using a temporary `r` variable?

Not sure if tests are in correct place?

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -15,6 +15,7 @@ from marshmallow.base import FieldABC, SchemaABC
 from marshmallow.utils import missing as missing_
 from marshmallow.compat import text_type, basestring
 from marshmallow.exceptions import ValidationError
+from marshmallow.validate import Validator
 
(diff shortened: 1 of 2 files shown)


tuukkamustonen:
Oh, just noticed there's also #488 open. Which one would be more suitable?

sloria:
This looks great. Simple and does the job. Thanks!

Issue #528, opened 2016-09-14
Errors from Nested fields are ignored


tvuotila:
Example:

```
from marshmallow import Schema, fields

class A(Schema):
    name = fields.Integer()

class B(Schema):
    user = fields.Nested(A)

# Has no errors
B().dump({'user': {'name': 'a'}})

# Has errors {'name': ['Not a valid integer.']}
A().dump({'name': 'a'})
```

PR #529, opened 2016-09-14
Fix nested errors missing from dump

tvuotila:
Refs marshmallow-code/marshmallow#528


maximkulkin:
Although it raises question of why do you ignore errors for nested fields in case of `only` argument is specified (e.g. error is in the nested field that is specified in `only` argument), but that feature is very strange and probably works only in simple cases compared to the issue here, so I'd rather say 👍 to this patch.

sloria:
Looks good. Thanks!

PR #530, opened 2016-09-14
Refactor dump exception

tvuotila:
PR had to be rebased to #522 because of conflicts.


sloria:
The behavior here looks correct. Thanks @tvuotila and thanks @maximkulkin for reviewing #522.

PR #532, opened 2016-09-20
Fix UUID fields not validating on serialize

pauljz:
UUIDs are currently only validated on deserialization/load. This validates them on serialize/dump as well. This is technically a breaking change, but the current behavior is probably undesirable in any case where it's being used.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -71,3 +71,4 @@ Contributors (chronological)
 - Alexandre Bonnetain `@Shir0kamii <https://github.com/Shir0kamii>`_
 - Tuukka Mustonen `@tuukkamustonen <https://github.com/tuukkamustonen>`_
 - Tero Vuotila `@tvuotila <https://github.com/tvuotila>`_
+- Paul Zumbrun `@pauljz <https://github.com/pauljz>`_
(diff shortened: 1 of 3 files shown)


maximkulkin (approved):
Although it's a strange thing to fail validation when you serialize data (meaning data that is originating from you does not hold invariants), but PR looks legit.

fuhrysteve:
FYI this change breaks compatibility with sqlalchemy-utils. It seems like a bug in sqlalchemy-utils where it has a `python_type` method rather than a class.

I have a pull request which should fix the issue:
https://github.com/kvesteri/sqlalchemy-utils/pull/248

http://docs.sqlalchemy.org/en/latest/core/type_api.html#sqlalchemy.types.TypeEngine.python_type

fuhrysteve:
Here's an excerpt of the traceback I get:

```python
marshmallow_sqlalchemy/convert.py", line 219, in _add_column_kwargs
    if not python_type or not issubclass(python_type, uuid.UUID):
TypeError: issubclass() arg 1 must be a class

(Pdb++) python_type
<bound method PhoneNumberType.python_type of PhoneNumberType(length=20)>
```

PR #534, opened 2016-09-23
Update Decimal field serialization to use fixed-point notation

gdub:
This avoids engineering-notation output, which occurs with default Decimal type string formatting in certain situations.

While Python does not have an issue round-tripping the engineering-notation strings, I imagine the fixed-point notation [[1]](https://docs.python.org/3/library/string.html#formatspec) more compatible with other languages/parsers.  FWIW, Django REST framework also uses fixed-point notation string output for Decimals [[2]](https://github.com/tomchristie/django-rest-framework/blob/3.4.7/rest_framework/fields.py#L1018).

Old behavior:

```
>>> fields.Decimal(as_string=True).serialize('x', {'x': Decimal('0.0000000')})
'0E-7'
```

New behavior:

```
>>> fields.Decimal(as_string=True).serialize('x', {'x': Decimal('0.0000000')})
'0.0000000'
```

### DIFF
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -238,6 +238,20 @@ class TestFieldSerialization:
         assert isinstance(m7s, decimal.Decimal)
         assert m7s.is_zero() and m7s.is_signed()
 
+        field = fields.Decimal(as_string=True, allow_nan=True)
+
+        m2s = field.serialize('m2', user)
+        assert isinstance(m2s, basestring)
+        assert m2s == user.m2
+
+        m5s = field.serialize('m5', user)
+        assert isinstance(m5s, basestring)
+        assert m5s == user.m5
+
+        m6s = field.serialize('m6', user)
+        assert isinstance(m6s, basestring)
+        assert m6s == user.m6
+
     def test_decimal_field_special_values_not_permitted(self, user):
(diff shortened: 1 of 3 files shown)


tests/test_serialization.py
```diff
@@ -266,6 +280,25 @@ def test_decimal_field_special_values_not_permitted(self, user):
...
         assert m7s.is_zero() and m7s.is_signed()
 
+    def test_decimal_field_fixed_point_representation(self, user):
+        """
+        Test we get fixed-point string representation for a Decimal number that would normally
+        output in engineering notation.
+        """
+        user.m1 = '0.00000000100000000'
+
+        field = fields.Decimal()
```
maximkulkin, line 290: How about caching serialization result instead of calling serialization twice?

``` python
result = fields.Decimal().serialize('m1', user)
assert isinstance(result, decimal.Decimal)
assert result == decimal.Decimal(user.m1)

result = fields.Decimal(as_string=True).serialize('m1', user)
assert isinstance(result, basestring)
assert result == user.m1
```

maximkulkin (approved):
LGTM 👍

Just make sure to squash commits together before you merge it.

sloria (approved):
Looks good. Thanks!

PR #536, opened 2016-09-29
Missing partial attribute in docstring

daniloakamine:
There's an attribute `partial` in `Schema.validate()`, but it's not present in documentation.
To fix this, I put there the same description from `Schema.load()`.


sloria:
Thanks!

PR #539, opened 2016-10-03
Fixes missing import in validates example

gotgenes:
Current example would break on validation

```
>>> item_schema = ItemSchema()
>>> item_schema.load({'quantity': 35})  # raises "NameError: name 'ValidationError' is not defined"
```


sloria:
Thanks!

Issue #540, opened 2016-10-11
Function fields break with Python 3 type annotations


martinstein:
Let's take the following simple test case:

``` python
from marshmallow import fields, Schema

def get_split_words(value: str):
    return value.split(';')

class TestSchema(Schema):
    name = fields.String(required=True)
    friends = fields.Function(serialize=None, deserialize=get_split_words)

data = {'name': 'Bruce Wayne', 'friends': 'Clark;Alfred;Robin'}
result = TestSchema().load(data)
```

That results in an error:

```
Traceback (most recent call last):
  File "mmtest.py", line 14, in <module>
    result = TestSchema().load(data)
  File "C:\tools\py35-venv\lib\site-packages\marshmallow\schema.py", line 542, in load
    result, errors = self._do_load(data, many, partial=partial, postprocess=True)
  File "C:\tools\py35-venv\lib\site-packages\marshmallow\schema.py", line 610, in _do_load
    index_errors=self.opts.index_errors,
  File "C:\tools\py35-venv\lib\site-packages\marshmallow\marshalling.py", line 294, in deserialize
    index=(index if index_errors else None)
  File "C:\tools\py35-venv\lib\site-packages\marshmallow\marshalling.py", line 67, in call_and_store
    value = getter_func(data)
  File "C:\tools\py35-venv\lib\site-packages\marshmallow\marshalling.py", line 287, in <lambda>
    data
  File "C:\tools\py35-venv\lib\site-packages\marshmallow\fields.py", line 263, in deserialize
    output = self._deserialize(value, attr, data)
  File "C:\tools\py35-venv\lib\site-packages\marshmallow\fields.py", line 1199, in _deserialize
    return self._call_or_raise(self.deserialize_func, value, attr)
  File "C:\tools\py35-venv\lib\site-packages\marshmallow\fields.py", line 1203, in _call_or_raise
    if len(utils.get_func_args(func)) > 1:
  File "C:\tools\py35-venv\lib\site-packages\marshmallow\utils.py", line 344, in get_func_args
    return inspect.getargspec(func).args
  File "C:\tools\Python35-32\lib\inspect.py", line 1045, in getargspec
    raise ValueError("Function has keyword-only arguments or annotations"
ValueError: Function has keyword-only arguments or annotations, use getfullargspec() API which can support them
```

[...]

PR #475, opened 2016-06-17
Fix compatibility with pyinvoke 0.13.0

deckar01:
Fixes #474

PR #548, opened 2016-11-10
Bugfix Issue 540 type annotations

sabinem:
[Fix for Issue 540](https://github.com/marshmallow-code/marshmallow/issues/540)

**Problem**
In Python 3 function parameters can have type annotations. Adding
annotations to a deserialization function used in schema caused a crash
on a utility function: `utils.get_func_args`. 

**Fix**
Since the inspect modul changed a lot from Python 2 to Python 3, I implemented a branching in
compat and a new version of the utility function for python 3. 
Both of these functions return now the arguments as tuples, the old function returned a list, but the new `inspect.signature` in Python3 returns a Ordered Dict, therefore I settled for a tuple as was stated in the docstring of the original function. 

**Open Points**
1. I tried to branch from `1.2-line`, but run into an error, like this here: [#475](https://github.com/marshmallow-code/marshmallow/pull/475), when I tried to set up my development environment, therefor I branched from `dev`instead.
2. I did not know how to set up a test case, since the test case with annotations is not valid python2 code and the python2 syntax check crashes on it.

**Feedback?**
This is my first meaningful open source bug fix. So please excuse, if I did something wrong and tell me how to improve.

### DIFF
--- a/marshmallow/compat.py
+++ b/marshmallow/compat.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 import sys
 import itertools
+import functools
+import inspect
 
 PY2 = int(sys.version_info[0]) == 2
 PY26 = PY2 and int(sys.version_info[1]) < 7
@@ -22,6 +24,13 @@ if PY2:
     else:
         from collections import OrderedDict
     OrderedDict = OrderedDict
+    def get_func_args(func):
+        if isinstance(func, functools.partial):
+            return list(inspect.getargspec(func.func).args)
+        if inspect.isfunction(func) or inspect.ismethod(func):
+            return list(inspect.getargspec(func).args)
+        if callable(func):
+            return list(inspect.getargspec(func.__call__).args)
 else:
     import urllib.parse
     urlparse = urllib.parse
@@ -36,6 +45,13 @@ else:
     zip_longest = itertools.zip_longest
     from collections import OrderedDict
     OrderedDict = OrderedDict
+    def get_func_args(func):
+        if isinstance(func, functools.partial):
+            return list(inspect.signature(func.func).parameters)
+        if inspect.isfunction(func):
+            return list(inspect.signature(func).parameters)
+        if callable(func) or inspect.ismethod(func):
+            return ['self'] + list(inspect.signature(func.__call__).parameters)
 
 
 # From six
--- a/marshmallow/utils.py
+++ b/marshmallow/utils.py
@@ -4,7 +4,6 @@ from __future__ import absolute_import, unicode_literals
 
 import collections
 import datetime
-import functools
 import inspect
 import json
 import time
@@ -15,6 +14,7 @@ from email.utils import formatdate, parsedate
 from pprint import pprint as py_pprint
 
 from marshmallow.compat import OrderedDict, binary_type, text_type
+from marshmallow.compat import get_func_args as compat_get_func_args
 
 
 dateutil_available = False
@@ -334,16 +334,10 @@ def callable_or_raise(obj):
     return obj
 
 
-def get_func_args(func):
-    """Given a callable, return a tuple of argument names. Handles
-    `functools.partial` objects and class-based callables.
-    """
-    if isinstance(func, functools.partial):
-        return inspect.getargspec(func.func).args
-    if inspect.isfunction(func) or inspect.ismethod(func):
-        return inspect.getargspec(func).args
-    # Callable class
-    return inspect.getargspec(func.__call__).args
+get_func_args = compat_get_func_args
+"""Given a callable, return a list of argument names.
+Handles `functools.partial` objects and callable objects.
+"""
 
 
 def if_none(value, default):
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -73,3 +73,4 @@ Contributors (chronological)
 - Tero Vuotila `@tvuotila <https://github.com/tvuotila>`_
 - Paul Zumbrun `@pauljz <https://github.com/pauljz>`_
 - Gary Wilson Jr. `@gdub <https://github.com/gdub>`_
+- Sabine Maennel `@sabinem <https://github.com/sabinem>`_
--- a/tasks.py
+++ b/tasks.py
@@ -21,6 +21,8 @@ def test(ctx, watch=False, last_failing=False):
         args.append('-f')
     if last_failing:
         args.append('--lf')
+    if int(sys.version_info[0]) < 3:
+        args.append('--ignore={0}'.format(os.path.join('tests', 'test_py3')))
     retcode = pytest.main(args)
     sys.exit(retcode)
 
--- a/tests/test_py3/test_utils.py
+++ b/tests/test_py3/test_utils.py
@@ -1,0 +1,14 @@
+from marshmallow import fields, Schema
+
+
+# Regression test for https://github.com/marshmallow-code/marshmallow/issues/540
+def test_function_field_using_type_annotation():
+    def get_split_words(value: str):
+        return value.split(';')
+
+    class MySchema(Schema):
+        friends = fields.Function(deserialize=get_split_words)
+
+    data = {'name': 'Bruce Wayne', 'friends': 'Clark;Alfred;Robin'}
+    result = MySchema().load(data)
+    assert result.data == {'friends': ['Clark', 'Alfred', 'Robin']}


marshmallow/compat.py
```diff
@@ -22,6 +22,7 @@
     else:
         from collections import OrderedDict
     OrderedDict = OrderedDict
+    from marshmallow.utils import get_func_args_2 as get_func_args
```
maximkulkin: I think these functions should be put here instead of just imported. Because otherwise they are both available through marshmallow.utils, although each one of them does not work properly in different versions of Python.

marshmallow/utils.py
```diff
@@ -334,16 +334,31 @@ def callable_or_raise(obj):
...
+    This is the orignial Python 2 version
     """
     if isinstance(func, functools.partial):
-        return inspect.getargspec(func.func).args
+        return tuple(inspect.getargspec(func.func).args)
     if inspect.isfunction(func) or inspect.ismethod(func):
-        return inspect.getargspec(func).args
-    # Callable class
-    return inspect.getargspec(func.__call__).args
+        return tuple(inspect.getargspec(func).args)
```
sloria: Any reason not to return a list to preserve backwards-compatibility?

marshmallow/utils.py
```diff
@@ -334,16 +334,31 @@ def callable_or_raise(obj):
     return obj
 
 
-def get_func_args(func):
```
sloria, line 337: Code that imports `get_func_args` from `marshmallow.utils` will break. We can preserve compatibility by adding

``` python
from marshmallow.compar import PY2

get_func_args = get_func_args2 if PY2 else get_func_args3
```

to this file.

sloria (changes requested):
Looking good! Just a few small changes to preserve backwards-compat.

Also, I've added a failing regression test for this change in https://github.com/marshmallow-code/marshmallow/tree/issue540 . Can you please rebase or merge that branch with this PR?

marshmallow/utils.py
sabinem: Okay.

marshmallow/utils.py
sabinem: No, you are right, it could be a list as well. The previous comment on the function said it returns a tuple, which was somewhat inconsistent, since it returned a list. 

```
"""Given a callable, return a tuple of argument names. Handles
'functools.partial' objects and class-based callables.
"""
```

So if it returns a list, I think the comment should be made consistent, otherwise this is fine as well.

marshmallow/utils.py
sabinem: I just tried my best to rebase and update the pull request. Hope it worked all out, otherwise please let me know.

marshmallow/utils.py
```diff
@@ -334,16 +333,34 @@ def callable_or_raise(obj):
...
-    return inspect.getargspec(func.__call__).args
+        return list(inspect.getargspec(func).args)
+    if callable(func):
+        return list(inspect.getargspec(func.__call__).args)
+
+
+def get_func_args3(func):
+    """Given a callable, return a list of argument names. Handles
+    `functools.partial` objects and class-based callables.
+    This is the Python 3 version where getargspec is depreciated and replaced by
```
sloria: Typo here: `depreciated` -> `deprecated`. FWIW, I don't think the docstring needs to talk about `getargspec` vs. `inspect.signature` because that is getting too far into implementation details.

Perhaps we should only document `get_func_args`, and leave the version-specific functions "private". Something like:

``` python
def _get_func_args2(func):
     # ... no docstring ...

def _get_func_args3(func):
    # ... no docstring ...

#: Given a callable, return a list of argument names.
#: Handles `functools.partial` objects and callable objects.
get_func_args = _get_func_args2 if PY2 else _get_func_args3
```

marshmallow/utils.py
sloria: Thanks. 

I was suggesting to rebase this branch onto the `issue540` branch, so that the test will be included with this branch/PR. Would you mind doing that?

marshmallow/utils.py
```diff
@@ -334,16 +333,34 @@ def callable_or_raise(obj):
     return obj
 
 
-def get_func_args(func):
-    """Given a callable, return a tuple of argument names. Handles
+def get_func_args2(func):
+    """Given a callable, return a list of argument names. Handles
```
sloria: Good catch on the error in the docstring!

sloria (changes requested):
Thanks for making those changes. Just a couple more nitpicky-changes to make, and this is good to merge!

marshmallow/utils.py
sabinem: okay.

marshmallow/utils.py
sabinem: Okay I think I misunderstood about the rebase. I did rebase but not with that branch, but with marshmallow/dev. I will try again. Do you have a tutorial at hand on how rebases easily?

lafrech: I generally don't use `rebase` and just go the easy way: merge his branch into yours.

```
git fetch upstream
git checkout issue_540_annotations_bug
git merge upstream/issue540
```

Rebase rewirtes the whole branch to make it look like it was pulled from `issue540` in the first place. This probably looks nicer once integrated in master (fewer merge commits). Some people care, some don't. I'm not sure about the policy, here, but it's never too late to squash all the commits together before integration if this is required when the implementation is validated, so if you're uncomfortable with this, merge is an option.

maximkulkin: @sabinem You need to take your new commits and put them on top of new branch. This is done with

```
git rebase <old_base> <your_branch> --onto <new_base>
```

e.g.

```
git rebase upstream/dev issue_540_annotations_bug --onto upstream/issue540
```

It really helps to understand how Git works.

@lafrech Best practice is to rebase: it allows fixing all merge conflicts so that CI/Github can do merges for you, simplifies per-commit reviewing and makes history nice.

marshmallow/compat.py
```diff
@@ -38,6 +38,7 @@
     OrderedDict = OrderedDict
 
 
+
```
maximkulkin, line 56: Looks like unnecessary change

marshmallow/utils.py
```diff
@@ -334,16 +333,27 @@ def callable_or_raise(obj):
...
+        return list(inspect.signature(func.func).parameters)
+    if inspect.isfunction(func):
+        return list(inspect.signature(func).parameters)
+    if callable(func) or inspect.ismethod(func):
+        return ['self'] + list(inspect.signature(func.__call__).parameters)
+
+
+# Given a callable, return a list of argument names.
+# Handles `functools.partial` objects and callable objects.
+get_func_args = get_func_args2 if PY2 else get_func_args3
```
maximkulkin: I still think that those functions belong to marshmallow/compat.py

maximkulkin (changes requested):

marshmallow/compat.py
sabinem: Sorry that extra blank line got in there by accident. Should I remove it? I was changing compat, but then sloria suggested to change only `utils.py`.

marshmallow/utils.py
maximkulkin: @sloria how about moving this version specific functions into compat.py instead of utils.py?

lafrech: Shouldn't compat.py stick to pure Python stuff, so that it can be imported anywhere without circular import issues? For instance, what if we need to import compat.py in utils.py some day?

maximkulkin: No circular dependencies should exist: compat.py should not import anything from utils.py (the reverse is ok). And, yes, those functions are pure python stuff.

I assumed all this is implied stuff is obvious.

lafrech: Sorry, I still had the first commit in mind, where the functions were imported from utils.py into compat.py. Taking [your comment here](https://github.com/marshmallow-code/marshmallow/pull/548/commits/c971b98bf074481200c54d198f7e7e18a5ea359b#r87485599) into account, it makes sense.

marshmallow/utils.py
sabinem: @maximkulkin: yes I think you are right: there are two points to be considered:
1. the branching or PY2 and PY3 should be done in compat.py rather then in utils.py
2. the upward compatibilty: it should be still possible to import get_func_arg from utils.

So I suggest the following solution: 
I put the two functions versions into compat and then I leave the simple relabeling of the function in utils, so that the function can be still imported from utils. I will try to implement this and send another pull request, okay?

marshmallow/utils.py
sabinem: I think I agree with maximkulkin and I programmed it and will send a pull request. I think putting the two functions into compat will be the cleanest solution, sind compat is about branching and compat may be imported by utils but not the other way around.

marshmallow/compat.py
```diff
@@ -1,10 +1,31 @@
...
 import sys
 import itertools
+import functools
+import inspect
 
 PY2 = int(sys.version_info[0]) == 2
 PY26 = PY2 and int(sys.version_info[1]) < 7
 
+
+def get_func_args2(func):
```
maximkulkin: Why not just this:

``` python
if PY2:
    import urlparse
    # ...
    def get_func_args(func):
        # code for Python 2
else:
    import urllib.parse
    # ...
    def get_func_args(func):
        # code for Python 3
```

sabinem: okay, sounds fine to me. And thanks for helping me out with how to do the rebase.

maximkulkin:
👍 LGTM

Just make sure to squash all your commits into one. You can do that by doing `git rebase --interactive upstream/dev` in your branch then in opened file leave your first commit ("Bugfix Issue 540 type annotations") having "pick" action and for all later commits change action to "fixup". Then save the file and exit. As a result all your commits will be merged (squashed) together into one commit.

marshmallow/utils.py
```diff
@@ -335,15 +335,9 @@ def callable_or_raise(obj):
 
 
 def get_func_args(func):
```
sloria: This can be simplified to:

``` python
get_func_args = compat_get_func_args
"""Given a callable, return a list of argument names.
Handles `functools.partial` objects and callable objects.
"""
```

marshmallow/fields.py
```diff
@@ -1244,7 +1244,7 @@ def _deserialize(self, value, attr, data):
         return value
 
     def _call_or_raise(self, func, value, attr):
-        if len(utils.get_func_args(func)) > 1:
+        if len(get_func_args(func)) > 1:
```
sloria: This change isn't really necessary, but it's not a big deal.

marshmallow/compat.py
```diff
@@ -22,6 +24,14 @@
...
         from collections import OrderedDict
     OrderedDict = OrderedDict
+    def get_func_args(func):
+        if isinstance(func, functools.partial):
+            return list(inspect.getargspec(func.func).args)
+        if inspect.isfunction(func) or inspect.ismethod(func):
+            return list(inspect.getargspec(func).args)
+        if callable(func):
+            return list(inspect.getargspec(func.__call__).args)
+        get_func_args = get_func_args2
```
sloria: This line should be removed.

sloria (changes requested):
Thanks again @sabinem . Just a couple more small changes and this is good to merge. Would you also add yourself to AUTHORS.rst please?

sabinem:
should I rebase once more to pull these two commits together, if everything is okay now?

sloria:
@sabinem Yes, one last rebase would be helpful.

sloria (approved):

sloria:
Looks good. Thanks @sabinem for the PR and thanks @lafrech and @maximkulkin for reviewing!

PR #552, opened 2016-11-22
Fix typo in decorator module code example. Fixes #551

### LINKED ISSUE #551: [Documentation] Syntax error
https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/decorators.py#L27

```python
 @post_dump(pass_many=True):
        def add_envelope(self, data, many):
            namespace = 'results' if many else 'result'
            return {namespace: data}
```

maximkulkin:


sloria:
Thanks!

PR #557, opened 2016-11-30
Automatically import `fields` module.

mindojo-victor:
`marshmallow/__init__.py` should do `from . import fields`. This would save some lines for the users. Any schema needs fields, there is no reason to not import the module automatically.

This would help in such case:
```
import marshmallow as mm

class UpgradeNowSchema(mm.Schema):
    plan = mm.fields.String(required=True)
```

Otherwise I need to:
```
import marshmallow as mm
import marshmallow.fields
```
which looks strange and ugly.

I would be even better to have fields in the main namespace:
```
plan = marshmallow.String(required=True)
```
then I would avoid shortening `marshmallow` -> `mm` or `ma`.

### DIFF
--- a/marshmallow/__init__.py
+++ b/marshmallow/__init__.py
@@ -7,6 +7,7 @@ from marshmallow.schema import (
     MarshalResult,
     UnmarshalResult,
 )
+from . import fields
 from marshmallow.decorators import (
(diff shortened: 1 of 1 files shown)


sloria:
Thanks!

PR #559, opened 2016-12-11
Better example of check_unknown_fields validator

### LINKED ISSUE #558: Better example for check_unknown_fields
The example validation function given in 'Validating Original Input Data' finds at most one unknown field at a time.  A more useful example would include all unknown fields in an error list.

altaurog:
Include all unknown fields in the error message.  The current example
reports at most one, even if more are present.  Fixes #558

### DIFF
--- a/docs/extending.rst
+++ b/docs/extending.rst
@@ -262,7 +262,7 @@ Validating Original Input Data
 Normally, unspecified field names are ignored by the validator. If you would like access to the original, raw input (e.g. to fail validation if an unknown field name is sent), add ``pass_original=True`` to your call to `validates_schema <marshmallow.decorators.validates_schema>`.
 
 .. code-block:: python
-    :emphasize-lines: 5
+    :emphasize-lines: 7
 
     from marshmallow import Schema, fields, validates_schema, ValidationError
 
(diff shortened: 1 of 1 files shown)


docs/extending.rst
```diff
@@ -270,13 +270,14 @@ Normally, unspecified field names are ignored by the validator. If you would lik
 
         @validates_schema(pass_original=True)
         def check_unknown_fields(self, data, original_data):
-            for key in original_data:
-                if key not in self.fields:
-                    raise ValidationError('Unknown field name {}'.format(key))
+            unknown = set(original_data) - set(self.fields)
+            if unknown:
+                errors = list(map("Unknown field '{}'".format, unknown))
+                raise marshmallow.ValidationError(errors)
```
maximkulkin: How about using ValidationError structure to point (machine-readable) to fields that have error?!
```python
raise marshmallow.ValidationError('Unknown field', list(unknown))
```
The reported result would be in a dictionary form like this:
```python
{'unknown_field1': 'Unknown field', 'unknown_field2': 'Unknown field'}
```

maximkulkin (changes requested):

docs/extending.rst
altaurog: Oh.  I see.  Yes, that makes sense.   `list` is superfluous then, as `set` is already iterable.

maximkulkin (approved):
LGTM 👍  Just make sure to squash all commits into one before merging

sloria:
Thanks!

PR #562, opened 2016-12-18
Fix redefinition of test_utils.test_get_value()

nelfin:
This redefinition was preventing py.test from collecting and executing
three test cases of `utils.get_value`


sloria:
Good catch. Thanks!

PR #563, opened 2016-12-19
Reset user-defined error kwargs on schema dump/load

jbasko:
Fixes issue when user-defined kwargs on ValidationError aren't reset and kwargs for each new ValidationError are a merge of all previous kwargs.


jbasko:
Did I create the PR incorrectly? Why are those other changes showing up?

PR #564, opened 2016-12-19
Error kwargs reset on schema on dump/load

jbasko:
Fixes issue when user-defined kwargs on ValidationError aren't reset and kwargs for each new ValidationError are a merge of all previous kwargs.


sloria:
Good catch. Thanks for the patch!

jbasko:
Thanks for merging and releasing this quickly!

Issue #550, opened 2016-11-18
Allow strict=False to override strict setting in Meta options class.


douglas-treadwell:
I prefer ```strict=True``` by default (as someone else requested in another issue), so I was setting it in the Meta class to True... but then I found I could not set it to False in cases when I wanted it to be False.

Workaround for now... implemented a BaseSchema with overridden strict kwarg like this:

```python
class BaseSchema(Schema):
    def __init__(self, strict=True, **kwargs):
        super(BaseSchema, self).__init__(strict=strict, **kwargs)
```

sloria:
Fixed by #569

PR #569, opened 2017-01-06
strict value is overidable, in all case. (especially,  __init__(strict=False), and Meta.strict=True)

podhmo:
Current marshmallow's implementation is not allow to overwrite strict value by `strict=False`, when strict value is set by Meta.

```python
class S(Schema):
    class Meta:
        strict = True

S(strict=False).strict  # => True
```
this pull request is the solution for #550

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -2048,3 +2048,44 @@ class TestLoadOnly:
         assert 'str_dump_only' not in result.data
         assert 'str_load_only' in result.data
         assert 'str_regular' in result.data
+
+
+class TestStrictDefault:
+
+    class SchemaTrueByMeta(Schema):
+        class Meta:
+            strict = True
+
+    class SchemaFalseByMeta(Schema):
+        class Meta:
+            strict = False
+
+    class SchemaWithoutMeta(Schema):
+        pass
+
+    def test_default(self):
+        assert self.SchemaWithoutMeta().strict is False
+
+    def test_meta_true(self):
+        assert self.SchemaTrueByMeta().strict is True
+
+    def test_meta_false(self):
+        assert self.SchemaFalseByMeta().strict is False
+
+    def test_default_init_true(self):
+        assert self.SchemaWithoutMeta(strict=True).strict is True
+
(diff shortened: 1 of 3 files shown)


podhmo:
It is a bit strange that default value of strict is `None`.  If so, this is better, maybe.

```python
delegated = object()

    .. snip
    def __init__(self, extra=None, only=(), exclude=(), prefix='', strict=delegated,
                 many=False, context=None, load_only=(), dump_only=(),
                 partial=False):
```

tests/test_schema.py
```diff
@@ -2048,3 +2048,38 @@ def test_dump_only(self, schema, data):
...
+        class Meta:
+            strict = False
+
+    class SchemaWithoutMeta(Schema):
+        pass
+
+    def test_default(self):
+        assert self.SchemaWithoutMeta().strict is False
+
+    def test__meta_true(self):
```
maximkulkin: 1. I think it is safe and more common not to use double underscores in test names.
2. You miss two tests when you use SchemaWithoutMeta and override with init argument (both True and False).

maximkulkin (changes requested):
I think there is no need for custom object as default value implementation and using None instead is OK. See my inline comments though.

podhmo:
Oh, yes I missed, adding two tests. And renaming test name (honestly, this is something like my individual habits,  separating test method's name by double underscore for readability. But this is not strong will).

maximkulkin:
👍 LGTM

Just make sure to squash your 3 commits about tests into one before merging, their commit descriptions are not informative and it is completely unnecessary to have all this small fixes in master.

podhmo:
sure. rebased.

sloria:
Thanks @podhmo for the PR and thanks @maximkulkin for reviewing. @podhmo : would you mind adding yourself to `AUTHORS.rst`? Once that's done, this is good to merge.

I consider this a bugfix, so I'll backport this to the 2.10-line branch once this is merged.

podhmo:
added.

sloria:
Thanks!

PR #572, opened 2017-01-18
Allow deeply nested load and dump only

YuriHeupa:
This allows deeply nested `load_only` and `dump_only` for `Nested` fields, it takes declarations from parent and normalizes it passing to the Nested schema defined.

For example:
```python
class ChildSchema(Schema):
    foo = fields.String()
    bar = fields.String()

class ParentSchema(Schema):
    childs = NestedDumpOnly('ChildSchema')

ParentSchema(dump_only=('childs.foo', ))
# The Above will mark the ChildSchema.foo field as dump_only=True
```

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -408,21 +408,33 @@ class Nested(Field):
             elif isinstance(self.nested, type) and \
                     issubclass(self.nested, SchemaABC):
                 self.__schema = self.nested(many=self.many,
-                        only=only, exclude=self.exclude, context=context)
+                        only=only, exclude=self.exclude, context=context,
+                        load_only=self._nested_normalized_option('load_only'),
+                        dump_only=self._nested_normalized_option('dump_only'))
             elif isinstance(self.nested, basestring):
                 if self.nested == _RECURSIVE_NESTED:
                     parent_class = self.parent.__class__
                     self.__schema = parent_class(many=self.many, only=only,
-                            exclude=self.exclude, context=context)
+                            exclude=self.exclude, context=context,
+                            load_only=self._nested_normalized_option('load_only'),
+                            dump_only=self._nested_normalized_option('dump_only'))
                 else:
                     schema_class = class_registry.get_class(self.nested)
                     self.__schema = schema_class(many=self.many,
-                            only=only, exclude=self.exclude, context=context)
+                            only=only, exclude=self.exclude, context=context,
+                            load_only=self._nested_normalized_option('load_only'),
+                            dump_only=self._nested_normalized_option('dump_only'))
             else:
                 raise ValueError('Nested fields must be passed a '
                                  'Schema, not {0}.'.format(self.nested.__class__))
         self.__schema.ordered = getattr(self.parent, 'ordered', False)
         return self.__schema
+
+    def _nested_normalized_option(self, option_name):
+        nested_field = '%s.' % self.name
+        return [field.split(nested_field, 1)[1]
+                for field in getattr(self.parent, option_name, set())
+                if field.startswith(nested_field)]
 
     def _serialize(self, nested_obj, attr, obj):
         # Load up the schema first. This allows a RegistryError to be raised
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -77,3 +77,4 @@ Contributors (chronological)
 - Victor Varvaryuk `@mindojo-victor <https://github.com/mindojo-victor>`_
 - Jāzeps Baško `@jbasko <https://github.com/jbasko>`_
 - `@podhmo <https://github.com/podhmo>`_
+- Yuri Heupa `@YuriHeupa <https://github.com/YuriHeupa>`_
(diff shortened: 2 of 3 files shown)


marshmallow/fields.py
```diff
@@ -401,23 +401,36 @@ def schema(self):
 
         # Inherit context from parent.
         context = getattr(self.parent, 'context', {})
+
+        # Normalize nested load_only and dump_only from parent sets
+        nested_field = '%s.' % self.name
+        dump_only = [field.split(nested_field, 1)[1]
+                     for field in getattr(self.parent, 'dump_only', set(()))
+                     if nested_field in field]
```
maximkulkin: Shouldn't this be `if field.startswith(nested_field)` ?

marshmallow/fields.py
```diff
@@ -401,23 +401,36 @@ def schema(self):
 
         # Inherit context from parent.
         context = getattr(self.parent, 'context', {})
+
+        # Normalize nested load_only and dump_only from parent sets
+        nested_field = '%s.' % self.name
```
maximkulkin: This implementation has performance impact of calculating this every time schema is requested although schema itself is being cached. I suggest moving this into two functions and call them whenever you create schema class.

maximkulkin (changes requested):

marshmallow/fields.py
YuriHeupa: Make sense, fixed.

marshmallow/fields.py
YuriHeupa: Indeed. I've made changes based on your suggestion, please let me know if it makes sense to you.

marshmallow/fields.py
```diff
@@ -401,29 +401,42 @@ def schema(self):
...
                 raise ValueError('Nested fields must be passed a '
                                  'Schema, not {0}.'.format(self.nested.__class__))
         self.__schema.ordered = getattr(self.parent, 'ordered', False)
         return self.__schema
 
+    def _nested_normalized_option(self, option_name):
+        nested_field = '%s.' % self.name
+        return [field.split(nested_field, 1)[1]
+                for field in getattr(self.parent, option_name, set(()))
+                if nested_field in field.startswith(nested_field)]
```
maximkulkin: This thing is incorrect

maximkulkin (changes requested):

marshmallow/fields.py
YuriHeupa: Fixed.

marshmallow/fields.py
```diff
@@ -408,22 +408,34 @@ def schema(self):
...
             else:
                 raise ValueError('Nested fields must be passed a '
                                  'Schema, not {0}.'.format(self.nested.__class__))
         self.__schema.ordered = getattr(self.parent, 'ordered', False)
         return self.__schema
 
+    def _nested_normalized_option(self, option_name):
+        nested_field = '%s.' % self.name
+        return [field.split(nested_field, 1)[1]
+                for field in getattr(self.parent, option_name, set(()))
```
maximkulkin: Although I do not understand why there is so much parens in `set()` call, I guess it's minor.

maximkulkin (approved):

marshmallow/fields.py
YuriHeupa: 🤔  They were unnecessary. removed!

YuriHeupa:
Is there any plan or objection for merging this?

sloria:
This looks like a good feature. Thanks for the PR and thanks @maximkulkin for reviewing. I'll get to reviewing and merging this soon, once I have a moment of free time.

sloria (approved):
Looks great. Thanks again!

avish:
Uhm, this change seems to be breaking list-of-nested fields (i.e. a `Nested` field whose `parent` is a `List` field rather than a `Schema`, which means its `load_only` and `dump_only` properties contain boolean values rather than sets): 

```

class ChildSchema(Schema):
    name = fields.Str()

class ParentSchema(Schema):
    children = fields.List(fields.Nested(ChildSchema))

def test_list_of_nested():
    data = {'children': [{'name': 'Child'}]}
    ParentSchema().dump(data)

```

Gives: 
```
    def _nested_normalized_option(self, option_name):
        nested_field = '%s.' % self.name
        return [field.split(nested_field, 1)[1]
>               for field in getattr(self.parent, option_name, set())
                if field.startswith(nested_field)]
E       TypeError: 'bool' object is not iterable
```

YuriHeupa:
@avish Thanks for reporting this, I've submitted a fix here #575

touilleMan:
I've encountered another cornercase where this feature break (see https://travis-ci.org/Scille/umongo/jobs/195873309#L493)

```python
    @property
    def root(self):
        """Reference to the `Schema` that this field belongs to even if it is buried in a `List`."""
        ret = self
        while hasattr(ret, 'parent') and ret.parent:
            ret = ret.parent
        return ret
```

Basically if a nested field has no parent (not linked yet to a schema), then `field.root` will return itself... who's `load_only`/`dump_only` fields are boolean !

I think `root` implementation is broken : it should not return itself but `None` if the field has no schema.
Once this behavior fixed `_nested_normalized_option` should work properly :smile:

YuriHeupa:
@touilleMan I'm not sure if that's the correct approach for fixing this issue, but I've submitted a fix here #578

PR #573, opened 2017-01-19
Optimize dumping objects by code generating functions to dump objects.

rowillia:
After profiling some of our services running Marshmallow, we noticed that it was
consuming a ton of CPU.  This approach is similar to that of namedtuple in that
we generate python code for dumping objects.  We then fall back to the original,
slower path should anything go wrong.

In the example provided in performance/benchmark.py we see a 16x speedup.  For
some of our internal schemas, this is closer to 20x.

```
Benchmark Result:
	Original Time: 2382.36 usec/dump
	Optimized Time: 145.83 usec/dump
	Speed up: 16.34x
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,4 +65,4 @@ venv
 
 # Other
 .directory
-
+*.pprof
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -76,3 +76,4 @@ Contributors (chronological)
 - Sabine Maennel `@sabinem <https://github.com/sabinem>`_
 - Victor Varvaryuk `@mindojo-victor <https://github.com/mindojo-victor>`_
 - Jāzeps Baško `@jbasko <https://github.com/jbasko>`_
+- `@podhmo <https://github.com/podhmo>`_
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,13 @@
 Changelog
 ---------
 
+2.11.1 (2017-01-08)
++++++++++++++++++++
+
+Bug fixes:
+
+- Allow ``strict`` class Meta option to be overriden by constructor (:issue:`550`). Thanks :user:`douglas-treadwell` for reporting and thanks :user:`podhmo` for the PR.
+
 2.11.0 (2017-01-08)
 +++++++++++++++++++
 
--- a/marshmallow/__init__.py
+++ b/marshmallow/__init__.py
@@ -14,7 +14,7 @@ from marshmallow.decorators import (
 from marshmallow.utils import pprint, missing
 from marshmallow.exceptions import ValidationError
 
-__version__ = '2.11.0'
+__version__ = '2.11.1'
 __author__ = 'Steven Loria'
 
 __all__ = [
--- a/marshmallow/compat.py
+++ b/marshmallow/compat.py
@@ -4,8 +4,11 @@ import itertools
 import functools
 import inspect
 
+import operator
+
 PY2 = int(sys.version_info[0]) == 2
 PY26 = PY2 and int(sys.version_info[1]) < 7
+PY34 = sys.version_info[0:2] >= (3, 4)
 
 if PY2:
     import urlparse
@@ -19,6 +22,9 @@ if PY2:
     itervalues = lambda d: d.itervalues()
     iteritems = lambda d: d.iteritems()
     zip_longest = itertools.izip_longest
+    def is_overridden(instance_func, class_func):
+        return instance_func.__func__ is not class_func.__func__
+    reload = reload
     if PY26:
         from .ordereddict import OrderedDict
     else:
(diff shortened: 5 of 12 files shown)


maximkulkin:
I would really want not to have this kind of code in the library. Apart from missing some features, it's completely unmaintainable.

I would suggest to author to try a different serialization library if he feels like Marshmallow being a bottleneck. There is a [performance comparison page](https://voidfiles.github.io/python-serialization-benchmark/) of different (sometimes not so sophisticated) serialization frameworks.

rowillia:
@maximkulkin Thanks for the feedback.  I'd be very interested in learning more about which features are missing, I've structured the test suite to run both in `optimized` and `unoptimized` mode and it passes in both.  It's great that Marshmallow has such a comprehensive test suite as this is generally able to ensure that the implementations are in sync.  If there's a feature missing I'm happy to add tests for it.  Additionally, if you have any other ideas on how to improve maintainability I'd be happy to take a look.  I fully realize this code needs more comments and plan on adding them in, but I'd also be happy to write up a wiki page with examples and a full explanation. 

This approach effectively generates an optimistic fast path, relying on the code around it for everything else.  The generated code simply unrolls grabbing properties off of the object being serialized and storing them in the result dictionary (See https://github.com/marshmallow-code/marshmallow/pull/573/files#diff-e1bc6e780c4b78fd3a4a85bf7b8ae0acR2099 for an example of the generated code).  Should there be a validation error, an exception, or even a missing field without a default we fall back to the original code.  The intention here is to optimize the common case and duplicate as little code as possible.  There is still some duplication (checking for `_CHECK_ATTRIBUTE`, generating a key based on a path, etc...) but it is intentionally kept to a minimum.

For the most part, the generated code calls directly back into the backing field's `_serialize` method.  There is an affordance for a Field to specify how wo inline it's serialization/validation which gets leveraged by the simpler fields like `String` or `Number` to speed things up even further.  Without these optimizations, it's still a big win (on the order of 7x - 8x).

It's quite common for popular libraries in the critical path to maintain a series of speedups that require code duplication and implementing the same thing twice (markupsafe, sqlalchemy, simplejson, etc...).  While it does of course come with some cost, generally the speedups are worth it.  If this were less than a 5x speedup I likely would not have pursued it because of the maintenance burden.

sloria:
Thanks @rowillia for the PR. That is an impressive speedup.

I am also concerned about the maintainability of this change. I wonder if there is some way this code could exist outside of marshmallow. I would be much more open to adding the necessary hooks to allow for an optimized Schema subclass in a separate library. Thoughts?

rowillia:
Thanks for taking a look @sloria !  I realize this is quite the PR, but the perf gains are pretty massive.  It allows users to have the great API marshmallow provides with speeds close to what they would be if they hand-rolled their validation.

There's some perf improvements here that are unrelated to this change, but only yielded meaningful improvements once the parsing code was jit'd (e.g. 100µs wouldn't be worth it with the current implementation, but when jitted it is).  We could potentially pull those into a separate PR.  It seems safest to have some of those behind a flag

If we were to take the separate API approach, the API I'd want is something like this:

```python
from marshmallow_jit import jit

class Foo(Schema):
   ....

foo_schema = jit(Foo())
```

In this world, we'd need 2 hooks.  One for invoking the jitted method instead of `Marshaller` and falling back to `self._marshal` if needed (https://github.com/marshmallow-code/marshmallow/pull/573/files#diff-035365c88325ed9779fcb020f7bcb35aR660) and one for recompiling the generated methods on `update_fields` - https://github.com/marshmallow-code/marshmallow/pull/573/files#diff-035365c88325ed9779fcb020f7bcb35aR933

In the future there may need to be other hooks, next would likely be deserialize and perhaps a fast-path for schemas without processors.

maximkulkin:
> In the future there may need to be other hooks, next would likely be deserialize and perhaps a fast-path for schemas without processors.

I've done some benchmarking on my own project and it looks like the biggest slowdown there is the fact that there are too many calls to small function during the process, not the extra checks that are made. So, if you can just inline all functions into one large function (keeping all checks and all functionality), you can get a pretty significant boost to overall performance.

As a side note, I've checked other projects @rowillia mentioned and I couldn't find any trace of similar technique (which is generating code on the fly) there.

rowillia:
@maximkulkin I was referring to the practice of having code duplication between what are optional speedups and the regular implementation.  For example - https://github.com/pallets/markupsafe/tree/master/markupsafe implements `_native.py` and `_speedups.c` which effectively duplicate code for the purposes of optionally speeding things up.

maximkulkin:
@rowillia Agreed. Although this one is a bit different. And I wonder if any popular framework does code generation to speed up existing implementation or if it even considered a good practice.

rowillia:
@maximkulkin Beyond namedtuple in the standard lib there's the popular package `attrs` - https://github.com/hynek/attrs/blob/1481227583f2cf1566dc2836d3861ffe092b8e33/src/attr/_make.py#L140

maximkulkin:
1. Those use cases are different.
2. It's actually an example of bad usage of code generation, because all that functionality could be implemented without code generation at all in a more compact way:
```python
def _make_attr_tuple_class(cls_name, attr_names):
    """
    Create a tuple subclass to hold `Attribute`s for an `attrs` class.
    The subclass is a bare tuple with properties for names.
    class MyClassAttributes(tuple):
        __slots__ = ()
        x = property(itemgetter(0))
    """
    return type(
        "{}Attributes".format(cls_name),
        (tuple,),
        dict({'__slots__': tuple()}, **{
            attr_name: property((lambda x: itemgetter(x))(idx))
            for idx, attr_name in enumerate(attr_names)
        }
    )
```

rowillia:
@maximkulkin Sure, I don't want to get into the particulars as to whether or not another library that does codegen should have avoid it in one case.  In general, that whole library relies on codegen to achieve the same speed you would get with hand rolled classes which is fairly analogous to what we're doing here.  We're providing the user with a great, simple API and achieving a similar speed to the hand written code through codegen.

sloria:
Closing this for now, as it has become stale. @rowillia Feel free to reopen if there is further interest in this. As I mentioned in my previous comment, I am open to adding the necessary hooks to make it possible to put these optimizations in a separate library.

PR #574, opened 2017-01-20
Ability to pass schemes into fields.Url

mosquito:


marshmallow/fields.py
```diff
@@ -1106,13 +1106,18 @@ class Url(ValidatedField, String):
     """
     default_error_messages = {'invalid': 'Not a valid URL.'}
 
-    def __init__(self, relative=False, **kwargs):
+    def __init__(self, relative=False, schemes=None, **kwargs):
         String.__init__(self, **kwargs)
+
+        if schemes is None:
```
sloria: These two lines should be unnecessary; `validate.URL` accepts `schemes=None`, in which case the `default_schemes` are used.

marshmallow/__init__.py
```diff
@@ -14,7 +14,7 @@
 from marshmallow.utils import pprint, missing
 from marshmallow.exceptions import ValidationError
 
-__version__ = '2.11.1'
+__version__ = '2.11.2'
```
sloria: No need to bump the version; I will do this before I release. 

This is technically a feature rather than a bugfix, so it will go in 2.12.0. Would you mind reverting this change?

sloria (changes requested):
Thanks for the PR. Just a few comments to address. Also, could you please add yourself to AUTHORS.rst? 

Once you make those changes, feel free to squash your commits into a single commit if you know how to. If not, I can do it when I merge.

mosquito:
Done :-D

sloria (approved):

sloria:
Thanks!

PR #577, opened 2017-01-25
Make a series of small optmizations for dumping objects

rowillia:
These optimizations are fairly small, a 5-10% win on the benchmark, but when
jitting serializers in https://github.com/marshmallow-code/marshmallow/pull/573
these changes wind up being a fairly big when the serializer is compiled.


sloria:
Looks good. Thanks!

PR #578, opened 2017-01-28
Fix Nested field root method

YuriHeupa:
Makes `root` method in Nested field not return itself but `None` if the `ret` field isn't a schema

This is a fix for the issue being reported here: https://github.com/marshmallow-code/marshmallow/pull/572#issuecomment-275800288


sloria:
Thanks!

PR #581, opened 2017-02-03
Allow schema validators to set specific error messages for multiple fields at once.

cpoppema:
I've mentioned this shortly in https://github.com/marshmallow-code/marshmallow/issues/441, but here comes the pull request with an example of how it works:

```
>>> from collections import defaultdict
>>> 
>>> from marshmallow import fields, Schema, validates_schema
>>> from marshmallow.validate import ValidationError
>>> 
>>> 
>>> class MySchema(Schema):
...     country = fields.Str()
...     zip_code = fields.Str()
...     area = fields.Int()
...     
...     @validates_schema
...     def validate_all(self, data):
...         errors = defaultdict(list)
...         
...         country = data.get('country')
...         zip_code = data.get('zip_code')
...         area = data.get('area')
...         
...         if country:
...             if country == 'us':
...                 if zip_code in ['', None]:
...                     errors['zip_code'].append('Missing zip_code missing for country us')
...                 if area:
...                     errors['area'].append('area is not supported for country us')
...             else:
...                 if zip_code:
...                     errors['zip_code'].append('zip_code is not supported for country %s' % country)
...                 if area in ['', None]:
...                     errors['area'].append('Missing area missing for country %s' % country)
...         
...         if errors:
...             raise ValidationError(dict(errors), field_names=errors.keys())
>>> MySchema(many=True).load([{
...     'country': 'us',
...     'area': 1,
... }]).errors
```

Before this pull request `errors` looks like:
```
{0: {
    'zip_code': [
        {
            'zip_code': ['Missing zip_code missing for country us'],
            'area': ['area is not supported for country us']
        }
    ],
    'area': [
        {
            'zip_code': ['Missing zip_code missing for country us'],
            'area': ['area is not supported for country us']
        }
    ]
}}
```

With the proposed changes `errors` looks like:
```
{0: {
    'zip_code': ['Missing zip_code missing for country us'],
    'area': ['area is not supported for country us']
}}
```

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -79,3 +79,4 @@ Contributors (chronological)
 - `@podhmo <https://github.com/podhmo>`_
 - Dmitry Orlov `@mosquito <https://github.com/mosquito>`_
 - Yuri Heupa `@YuriHeupa <https://github.com/YuriHeupa>`_
+- WeAreSpindle `<https://github.com/wearespindle>`_
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -206,7 +206,10 @@ class Unmarshaller(ErrorStore):
                     else:
                         errors.setdefault(field_name, []).extend(err.messages)
                 elif isinstance(err.messages, dict):
-                    errors.setdefault(field_name, []).append(err.messages)
+                    if field_name in err.messages:
+                        errors.setdefault(field_name, []).extend(err.messages.get(field_name))
+                    else:
+                        errors.setdefault(field_name, []).append(err.messages)
                 else:
                     errors.setdefault(field_name, []).append(text_type(err))
 


maximkulkin:
@cpoppema Just to let you know, raising ValidationError with a dictionary of different field errors is considered "a breaking change" since someone in the past asked for ability to raise errors represented as dicts. There is a [PR for that](https://github.com/marshmallow-code/marshmallow/pull/442) already, hanging there for almost a year.

cpoppema:
I have seen that PR for #441 and it was breaking because it fails existing tests and those tests required change to make them succeed again, right ?  As mentioned here: https://github.com/marshmallow-code/marshmallow/pull/442#r69383349
> but you had to modify existing test cases, which makes this PR backwards incompatible.

In my PR all tests succeed without modifications, so how does that makes this backwards incompatible exactly? Yes the (as far as I know) undocumented interface to ValidationError changes the output error dict _only when_ you pass a dict _and_ a list of field names. If you pass a list of field names and a message or a list of messages, these errors are still duplicated across those fields in the output error dict.

maximkulkin:
Thing is that although the no test fail, this patch does change semantics of ValidationError arguments (the same way the original PR does, although it also requires to specify fields names which make it even more ridiculous and convoluted). Accompanied by abusing lack of good test coverage this PR does indeed pass all tests (the `fields_names` argument in ValidationError was supposed to mean that THE SAME error applies to all arguments listed. This PR changes that to "messages dictionary contains errors for specified field names" which does not play well if you want to report the same error, which is a dictionary, to multiple fields). So, it's backward incompatible. And if to choose between two backward incompatible PRs, I would choose the first one: it's more complete, exposes useful helper functions/classes for building errors, IT ADDS TESTS.

sloria:
Thanks @cpoppema  for the PR. I'm not going to merge this in right now because I think there is more discussion to be had about the proper API. As @maximkulkin points out, having to pass `field_names` along with the messages dict is a bit kludgy. 

That said, I am still open to adding functionality to address #441--I would love to have this in 3.0. I'm hoping we can arrive at a simple API for this in the near future.

PR #582, opened 2017-02-06
Add `strict` parameter for fields.Method

dongweiming:
I had create a schema like this:

```


class TestSchema(Schema):
     avatar_url = fields.Method('get_avatar_url')

     def get_avatar_url(self, obj):
          path = obj['path']
          if not path.starts_with('/'):
               path = '/' + path
          return path


s = TestSchema()
print(s.dump({'path': '/some/path'}))
```

the result is `MarshalResult(data={}, errors={})`. because AttributeError be ignored. but error is caused by other place(In this case, path has no attribute 'starts_with') rather than `obj`

i think this is  very puzzled, should add a `strict` parameter for developers

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -1172,22 +1172,29 @@ class Method(Field):
     :param str deserialize: Optional name of the Schema method for deserializing
         a value The method must take a single argument ``value``, which is the
         value to deserialize.
+    :param bool strict: If `True` and catch AttributeError will reraise it during
+        serialization, otherwise AttributeError will be ignored.
 
     .. versionchanged:: 2.0.0
         Removed optional ``context`` parameter on methods. Use ``self.context`` instead.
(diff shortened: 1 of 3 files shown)


maximkulkin:
I wonder why AttributeError should be ignored in the first place. Obviously, if you write some method in your schema to get values from serialized objects and you're getting AttributeError, you're doing it wrong and "let it crash". If you still want to serialize something, you can put `try/catch` clause **INSIDE** your method.

sloria:
Yes, the `AttributeError` is a wart from marshmallow's early days that accidentally made it into 2.0. I'm leaning towards removing the exception handling in marshmallow 3.0 and cutting an early release.

sloria:
Closing this for now, as this is implemented in the 3.x line. Use the implementation from there if you want stricter (de)serialization.

PR #590, opened 2017-03-03
Actually allow setting Boolean truthy and falsy via kwargs

zwack:
Contrary to the docs `truthy` and `falsy` couldnt be set through kwargs.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -773,6 +773,8 @@ class Decimal(Number):
 class Boolean(Field):
     """A boolean field.
 
+    :param set truthy: see below
+    :param set falsy: see below
     :param kwargs: The same keyword arguments that :class:`Field` receives.
     """
     #: Values that will (de)serialize to `True`. If an empty set, any non-falsy
@@ -784,6 +786,14 @@ class Boolean(Field):
     default_error_messages = {
         'invalid': 'Not a valid boolean.'
     }
+
+    def __init__(self, truthy=None, falsy=None, **kwargs):
+        super(Boolean, self).__init__(**kwargs)
+
+        if truthy is not None:
+            self.truthy = truthy
+        if falsy is not None:
+            self.falsy = falsy
 
     def _serialize(self, value, attr, obj):
         if value is None:
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -773,14 +773,13 @@ class Decimal(Number):
 class Boolean(Field):
     """A boolean field.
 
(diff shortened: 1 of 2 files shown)


marshmallow/fields.py
```diff
@@ -785,6 +787,14 @@ class Boolean(Field):
         'invalid': 'Not a valid boolean.'
     }
 
+    def __init__(self, truthy=None, falsy=None, **kwargs):
+        super(Boolean, self).__init__(**kwargs)
+
+        if truthy is not None:
+            self.truthy = truthy
```
maximkulkin: Explicitly wrap this with a `set()` maybe?

marshmallow/fields.py
```diff
@@ -773,6 +773,8 @@ def _to_string(self, value):
 class Boolean(Field):
     """A boolean field.
 
+    :param set truthy: see below
```
maximkulkin: That's a bad way to write argument descriptions: when you generate documentation, there will be no "below" available.

Please move below descriptions here and replace those with something like "Default values for constructor arguments".

maximkulkin (changes requested):

sloria (changes requested):
I agree with @maximkulkin 's comments. Can you please also add tests for this change?

When those are addressed, this is good to merge.

zwack:
Ok, made changes as requested.

tests/test_deserialization.py
```diff
@@ -284,18 +283,29 @@ class MyBoolean(fields.Boolean):
     ])
     def test_boolean_field_deserialization_with_custom_truthy_values_invalid(
             self, in_val):
-        class MyBoolean(fields.Boolean):
-            truthy = set(['yep'])
-        field = MyBoolean()
+        field = fields.Boolean(truthy=('yep',))
```
maximkulkin: Can you please leave existing tests unchanged and add new ones separately, because existing functionality is part of public interface (and thus need to be kept functioning).

maximkulkin (changes requested):

tests/test_deserialization.py
sloria: I agree. @zwack Once the above change is implemented, this should be good to go.

sloria (changes requested):

zwack:
Old tests have been restored now. Also merged with upstream as default values have changed meanwhile.

sloria (approved):

sloria:
Thanks for updating @zwack .

PR #591, opened 2017-03-04
Minor code refactoring

frol:


frol:
```
RuntimeError: dictionary changed size during iteration
```

Hmm, that is really odd. I am taking a deeper look now.

frol:
I believe the code became simpler and more efficient (I have eliminated two list allocations (one for `list(iteritems)` and another for `sorted()`), and also replaced the `getter` (function call on every iteration) with the value that is already available as `field_value`).

sloria:
Looks good. Thanks @frol !

PR #592, opened 2017-03-04
Fix ordered schema support for inherited fields

frol:
@sloria I am sorry, I am completely lost here. I have a biggish project in which I use ordered schemas with Marshmallow-SQLAlchemy, and I have noticed that fields on some of the schemas end up being resorted on every restart... My debug lead me to this change, but I cannot create a test for it :(


sloria:
Hm, I would expect [this test](https://github.com/marshmallow-code/marshmallow/blob/82c327ee38b32a3ba0d50b1b70c4fe267c091517/tests/test_options.py#L84) to fail without this change, but this change looks correct nonetheless. I'll backport this to the 2.x-line branch and release a hotfix.

PR #593, opened 2017-03-05
Inherit ordered schemas when Meta is defined

frol:


frol:
I have reimplemented this feature in a cleaner way. Originally, I patched the `meta` instance right in the `__new__`.

sloria:
We should not be messing with Python's inheritance behavior any more than necessary. `class Meta` can be inherited just like a normal class.

```python

class Parent(Schema):
    class Meta:
        ordered = True


class Child(Schema):
    class Meta(Parent.Meta):
        pass

assert Child().opts.ordered is True
```

frol:
Without this change we have inconsistent behavior as `ordered` is inferred from base classes in `__new__`, so I believe that either you drop that logic (that would be sad) or complete it with this change.

frol:
I am talking about these lines: https://github.com/marshmallow-code/marshmallow/blob/596015ad86676c74a11879c88a2aa69e91b0d40c/marshmallow/schema.py#L84-L97. You infer the `ordered` option there and since you already treat the schema as ordered, it is absolutely fine to help the SchemaOpts to set the `ordered` value correctly.

This behavior was introduced 2 years ago in https://github.com/marshmallow-code/marshmallow/commit/927704be32a999b5c65edaa8ee6f76e58fe6890a to close #162.

BTW, if the Meta is inherited, we won't break anything as `ordered` is checked on the Meta: https://github.com/marshmallow-code/marshmallow/blob/596015ad86676c74a11879c88a2aa69e91b0d40c/marshmallow/schema.py#L83

sloria:
OK, thanks for your input. I will take a closer look at this when I have the time.

sloria:
@frol OK, this looks like a good change. Since this is a bug fix, would you mind resending the PR to the `2.x-line` branch?

frol:
Thank you for your review! I agree it is should be in 2.x-line. I will do this shortly.

frol:
Rebased and changed the target branch.

sloria:
Thanks for the quick turnaround!

PR #599, opened 2017-03-10
Implement `ordered` option inheritance without breaking the API of 2.x line

### LINKED ISSUE #597: TypeError - unexpected keyword argument 'ordered'
When running our tests with marshmallow-2.13.2 and marshmallow-jsonapi-0.10.2, we encountered this stack trace:

```
tests/fa/test_updater.py:3: in <module>
    from fa import updater
src/fa/__init__.py:15: in <module>
    from game_process import instance as instance
src/fa/game_process.py:8: in <module>
    import util
src/util/__init__.py:718: in <module>
    from crash import CrashDialog
src/util/crash.py:11: in <module>
    from faf.api import BugReportSchema
/usr/lib64/python2.7/site-packages/faf/api/__init__.py:4: in <module>
    from faf.api.leaderboard_schema import LeaderboardSchema
/usr/lib64/python2.7/site-packages/faf/api/leaderboard_schema.py:1: in <module>
    from marshmallow_jsonapi import Schema, fields
/usr/lib64/python2.7/site-packages/marshmallow_jsonapi/__init__.py:2: in <module>
    from .schema import Schema, SchemaOpts
/usr/lib64/python2.7/site-packages/marshmallow_jsonapi/schema.py:31: in <module>
    class Schema(ma.Schema):
/usr/lib64/python2.7/site-packages/marshmallow/schema.py:106: in __new__
    klass.opts = klass.OPTIONS_CLASS(meta, ordered=ordered)
E   TypeError: Error when calling the metaclass bases
E       __init__() got an unexpected keyword argument 'ordered'
```

This might be related to #593.

frol:
*Fixes #597*

Reimplemented 3ffb74d71b5e647ddc3793f464e271a2eeb98bad for 2.x-line to avoid API breakage.

I have added a test to ensure that the API don't break.

I think this implementation shouldn't land to 3.x branch because it feels a bit cumbersome.

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -103,7 +103,9 @@ class SchemaMeta(type):
         meta = getattr(klass, 'Meta')
         # Set klass.opts in __new__ rather than __init__ so that it is accessible in
         # get_declared_fields
-        klass.opts = klass.OPTIONS_CLASS(meta, ordered=ordered)
+        klass.opts = klass.OPTIONS_CLASS(meta)
+        # Pass the inherited `ordered` into opts
+        klass.opts.ordered = ordered
         # Add fields specifid in the `include` class Meta option
         cls_fields += list(klass.opts.include.items())
 
@@ -180,7 +182,7 @@ class SchemaMeta(type):
 class SchemaOpts(object):
     """class Meta options for the :class:`Schema`. Defines defaults."""
 
-    def __init__(self, meta, ordered=False):
+    def __init__(self, meta):
         self.fields = getattr(meta, 'fields', ())
         if not isinstance(self.fields, (list, tuple)):
             raise ValueError("`fields` option must be a list or tuple.")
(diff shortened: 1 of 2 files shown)


sloria:
Looks good. Can you please resend this to `2.x-line`?

sloria:
Also, I'm lukewarm about keeping the `ordered` parameter in 3.x. What do you find cumbersome about this change?

frol:
Rebased.

frol:
I think that the original implementation of the inheritance is better because it leaves the control of the value up to the SchemaOpt whereas this implementation overrides/patches the value.

sloria:
I think I see your point now. Patching `ordered` after the options class is instantiated means that a custom `SchemaOpts` wouldn't be able to force the value of `ordered`.

```python
from marshmallow import SchemaOpts, Schema

class CustomOpts(SchemaOpts):

    def __init__(self, meta):
        super().__init__(meta)
        self.ordered = False


class MySchema(Schema):
    OPTIONS_CLASS = CustomOpts

    class Meta:
        ordered = True

assert MySchema().opts.ordered is False  # AssertionError
```

I just didn't like having a special-cased option in the constructor, but I think it's fine as long as we have a documented migration path.

PR #604, opened 2017-03-18
Allow duplicates in ContainsOnly

### LINKED ISSUE #603: validate.ContainsOnly raises error if input contains duplicate, valid entries
Reported in https://github.com/marshmallow-code/marshmallow/issues/516#issuecomment-287561530:

> As pointed out by @maximkulkin 
> 
>     validate.ContainsOnly([1, 2, 3])([1, 2, 3, 3])
> 
> raises `ValidationError`.
> 
> It seems intended (explicit `del choices[index]` in the code) but is not explicit in the docs, and in fact, I don't get the rationale.
> 
> The docs says 
> 
> > succeeds if ``value`` is a sequence and each element in the sequence is also in the sequence passed as ``choices``

This behavior is unintended. Duplicate, valid entries should be considered valid.

lafrech:
Fixes #603 

From the tests, it is clear that this behavior was intentional, but I really think duplicates should be accepted. Someone wanting the old behavior can do it in a custom validator.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -83,3 +83,4 @@ Contributors (chronological)
 - Roy Williams `@rowillia <https://github.com/rowillia>`_
 - Vlad Frolov `@frol <https://github.com/frol>`_
 - Erling Børresen `@erlingbo <https://github.com/erlingbo>`_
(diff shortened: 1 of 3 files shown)


sloria:
Thanks!

PR #605, opened 2017-03-21
Check for valid iso8601-format datetime strings

### LINKED ISSUE #556: Rationale behind fields.DateTime().deserialize('1916') returning datetime(1916, 11, 30, 0, 0)?
I just noticed that when I deserialize a date expressed as a 'YY' or 'YY-mm' string, I get a datetime with month defaulting to 11 and day defaulting to 30 (except for february).

```py
f = fields.DateTime()
f.deserialize('1916')
# datetime.datetime(1916, 11, 30, 0, 0)
f.deserialize('1916-2')
# datetime.datetime(1916, 2, 29, 0, 0)
f.deserialize('1916-3')
# datetime.datetime(1916, 3, 30, 0, 0)
```

Is this intended?

Shouldn't this be either rejected or default to a documented value (could be january 1st)?

To prevent this seemingly non-deterministic behavior (it seems reproducible, but I don't get the rationale), as a workaround, it is possible to set a format explicitly, which rejects such incomplete inputs, but this might not always be desired (be loose on inputs / strict on outputs).

```py
f = fields.DateTime(format='%Y-%m-%d')
f.deserialize('1916-05-04')
# datetime.datetime(1916, 5, 4, 0, 0)
f.deserialize('1916')
# marshmallow.exceptions.ValidationError: Not a valid datetime.
```

### LINKED ISSUE #566: Is there a specific way to format datetime in json to conform to?
Is there a specific way to format datetime in json?

Some of my json dates come out as "Tue, 22 Sep 2015 10:17:00 GMT" and some come out as "2015-02-03T15:40:36.078358+00:00"

data in database all look the same. I'm not sure how to get rid of the GMT way of formatting...

sloria:
Fixes #556

### DIFF
--- a/marshmallow/utils.py
+++ b/marshmallow/utils.py
@@ -6,6 +6,7 @@ import collections
 import datetime
 import inspect
 import json
+import re
 import time
 import types
 from calendar import timegm
@@ -210,6 +211,14 @@ def rfcformat(dt, localtime=False):
     else:
         return local_rfcformat(dt)
 
+# From Django
+_iso8601_re = re.compile(
+    r'(?P<year>\d{4})-(?P<month>\d{1,2})-(?P<day>\d{1,2})'
+    r'[T ](?P<hour>\d{1,2}):(?P<minute>\d{1,2})'
+    r'(?::(?P<second>\d{1,2})(?:\.(?P<microsecond>\d{1,6})\d{0,6})?)?'
+    r'(?P<tzinfo>Z|[+-]\d{2}(?::?\d{2})?)?$'
+)
+
 
 def isoformat(dt, localtime=False, *args, **kwargs):
     """Return the ISO8601-formatted UTC representation of a datetime object.
@@ -254,6 +263,8 @@ def from_iso(datestring, use_dateutil=True):
 
     Use dateutil's parser if possible and return a timezone-aware datetime.
     """
+    if not _iso8601_re.match(datestring):
+        raise ValueError('Not a valid ISO8601-formatted string')
     # Use dateutil's parser if possible
     if dateutil_available and use_dateutil:
         return parser.parse(datestring)
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -310,6 +310,11 @@ class TestFieldDeserialization:
         msg = 'Not a valid datetime.'.format(in_value)
         assert msg in str(excinfo)
 
+    def test_datetime_passed_year_is_invalid(self):
+        field = fields.DateTime()
+        with pytest.raises(ValidationError):
+            field.deserialize('1916')
+
     def test_custom_date_format_datetime_field_deserialization(self):
 
         dtime = dt.datetime.now()
@@ -325,10 +330,7 @@ class TestFieldDeserialization:
         assert_datetime_equal(field.deserialize(datestring), dtime)
 
         field = fields.DateTime()
-        if utils.dateutil_available:
-            assert_datetime_equal(field.deserialize(datestring), dtime)
-        else:
-            assert msg in str(excinfo)
+        assert msg in str(excinfo)
 
     @pytest.mark.parametrize('fmt', ['rfc', 'rfc822'])
     def test_rfc_datetime_field_deserialization(self, fmt):


marshmallow/utils.py
```diff
@@ -210,6 +211,14 @@ def rfcformat(dt, localtime=False):
     else:
         return local_rfcformat(dt)
 
+# From Django
+_iso8601_re = re.compile(
+    r'(?P<year>\d{4})-(?P<month>\d{1,2})-(?P<day>\d{1,2})'
+    r'[T ](?P<hour>\d{1,2}):(?P<minute>\d{1,2})'
+    r'(?::(?P<second>\d{1,2})(?:\.(?P<microsecond>\d{1,6})\d{0,6})?)?'
```
Timothee, line 218: As far as I can tell, ISO-8601 requires numbers to be zero-padded, so `\d{1,2}` should be `\d{2}` in the three lines above.

marshmallow/utils.py
```diff
@@ -210,6 +211,14 @@ def rfcformat(dt, localtime=False):
     else:
         return local_rfcformat(dt)
 
+# From Django
+_iso8601_re = re.compile(
+    r'(?P<year>\d{4})-(?P<month>\d{1,2})-(?P<day>\d{1,2})'
```
Timothee, line 216: Wikipedia writes:

> Representations can be done in one of two formats – a basic format with a minimal number of separators or an extended format with separators added to enhance human readability.[5][6] The standard notes that "The basic format should be avoided in plain text

Not sure if that's something that should be supported here or not.

marshmallow/utils.py
```diff
@@ -254,6 +263,8 @@ def from_iso(datestring, use_dateutil=True):
 
     Use dateutil's parser if possible and return a timezone-aware datetime.
     """
+    if not _iso8601_re.match(datestring):
+        raise ValueError('Not a valid ISO8601-formatted string')
```
Timothee, line 267: I'd be tempted to say "Not a valid ISO8601-formatted **datetime** string" as `2017-04-13` is a valid ISO-8601 string, just not one with time. (for that matter `2016` is also valid, though you'd be asking for trouble 😄 )

tests/test_deserialization.py
```diff
@@ -310,6 +310,11 @@ def test_invalid_datetime_deserialization(self, in_value):
         msg = 'Not a valid datetime.'.format(in_value)
         assert msg in str(excinfo)
 
+    def test_datetime_passed_year_is_invalid(self):
+        field = fields.DateTime()
+        with pytest.raises(ValidationError):
+            field.deserialize('1916')
```
Timothee, line 316: I'd add a test case for a something like `2017-04-13`, as some kind of documentation, as they're common as ISO 8601 strings, but not valid in the context of `fields.DateTime`. (also, the initial issue with `1916` was how the current date was used to fill in the blanks; but with `2017-04-13`, it didn't raise such red flags)

Timothee (commented):
Just a few comments.

For the record, this change was "breaking" for us. To be clear, we were mistakenly using `DateTime` where we should have been using `Date`; so we were certainly at fault. The parsing still worked fine when we were sending strings like `2017-04-13`.

As `marshmallow` updated automatically to that version (as part of `flask-apispec`) when rebuilding our container, we were naturally puzzled at first. 😄 

I'm not arguing that this was not "technically not breaking", but wanted to give an example where the resulting behavior was still correct, predictable and not unreasonable, even when misusing the API.

I guess I'm just saying that when making something less forgiving, it can be better to err on the side of caution and consider the change breaking even if it's technically not. Just my two cents 😉

marshmallow/utils.py
sloria: Good suggestion. I will change the message.

marshmallow/utils.py
sloria: This regex was taken from Django, so I am hesitant to change it unless it is changed there as well.

tests/test_deserialization.py
sloria: Also a good suggestion.

sloria:
@Timothee Thank you very much for the feedback. I'm sorry for any inconvenience this change may have caused you. This was, however, a bug fix--DateTime was accepting invalid input, and this fixes that.

marshmallow/utils.py
Timothee: That makes sense.

Timothee:
> This was, however, a bug fix--DateTime was accepting invalid input, and this fixes that.

Certainly understood 😄  Cheers!

PR #609, opened 2017-04-02
When validators called when many elements are provided, but pass_many is False, all elements from the original_data are still passed to the validator

thehuz:
In _invoke_validators, where many data elements exist, but pass_many is not True, there is a for loop to call the validator on each element. However, for that case, while the call to the validator only provides the specific data element, all of the original_data is sent (instead of the respective original_data element). Change is to only provide the respective element for the original_data to the call to the validator

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -63,6 +63,18 @@ def test_dump_with_strict_mode_raises_error(SchemaClass):
     assert type(exc.messages) == dict
     assert exc.messages == {'email': ['Not a valid email address.']}
 
+def test_invoke_validators_with_many():
+    class MySchema(Schema):
+        email = fields.Email()
+
+        @validates_schema(pass_original=True)
+        def validate_schema_with_original(self, data, original_data):
+            assert not isinstance(original_data, list)
+
+    myschema = MySchema()
+    myschema.load([ { 'name': 'Joe', 'email': 'joe@null.null' }, { 'name': 'Jill', 'email': 'jill@null.null'} ], many=True)
+
+
 def test_dump_resets_errors():
     class MySchema(Schema):
         email = fields.Email()
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -72,7 +72,8 @@ def test_invoke_validators_with_many():
             assert not isinstance(original_data, list)
 
     myschema = MySchema()
-    myschema.load([ { 'name': 'Joe', 'email': 'joe@null.null' }, { 'name': 'Jill', 'email': 'jill@null.null'} ], many=True)
+    myschema.load([{'name': 'Joe', 'email': 'joe@null.null'},
+        {'name': 'Jill', 'email': 'jill@null.null'}], many=True)
 
 
 def test_dump_resets_errors():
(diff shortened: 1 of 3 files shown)


sloria (changes requested):
@thehuz Thanks for the PR! I think this is the desired behavior. Could you please add a test for this change?

thehuz:
Hello Sloria. I added a test case to test_schema to cover this fix. Please let me know if there is an appropriate place in the file the test function would fit best and I can move it there (currently it is placed randomly near the top). Also, feel free to let me know if the test case function name should follow a certain structure/format.

tests/test_schema.py
```diff
@@ -63,6 +63,19 @@ def test_dump_with_strict_mode_raises_error(SchemaClass):
     assert type(exc.messages) == dict
     assert exc.messages == {'email': ['Not a valid email address.']}
 
+def test_invoke_validators_with_many():
```
sloria: Can you move this to `tests.test_decorators.TestPassOriginal`?

thehuz: I have moved it. Let me know if it looks good or if there are other changes you would like implemented.

sloria:
After giving this a second thought, I'm not sure this is the correct thing to do. It makes the behavior of `validates_schema(pass_original=True)` inconsistent with the other decorators. I think if you want to validate individual items, you should pass `pass_many=True` and validate each item if `many=True`.

thehuz:
Hello Sloria. 

The expectation is that by leaving `pass_many=False`, you can write a validator function that is agnostic of whether many items are sent. Then, as expected, you receive as the `data` parameter, the specific element that was processed; however, for the `original_data` parameter, you receive either the specific one original element (if only one element was processed -- i.e. `many=False` by the code calling the Schema method) or the list of all the elements (if a list of elements was processed -- i.e. `many=True` by the code calling the Schema method). This leads to the following challenges:

1) When `pass_many` is `False`, the validator function is presumably written expecting that many elements won’t be passed -- i.e. the code will be written to process an element at a time (even if `many=True` was used by the calling code). However, while this is true for the processed `data` element, the `original_data` parameter will either be a single element OR a list of elements (depending on whether `many=False` or `many=True` was used by the calling code). This increases the code complexity for the validator functions, as the code will need to handle the `original_data` parameter being either a single element OR a list of all the elements.

2) It isn't possible to identify the original data element in the list from the `original_data` parameter that corresponds to the single [processed] element passed as the `data` parameter. (i.e. there is no index value passed to identify the element in the list from the `original_data` parameter that corresponds to the passed single element `data` parameter)


An alternate solution, is to pass an `original_data_idx` parameter to the validator function. i.e. instead of passing the specific original data element that corresponds with the specific processed data element passed, when `many=True` from the calling code, just provide the index into the list of original data elements that corresponds to the specific process data element passed. This would resolve the 2nd challenge. There will still be code complexity with having to check if an [optional] `original_data_idx` value is passed to the validator function, and only then dereferencing the `original_data` parameter to get the specific original data element.

To your point that other decorators currently operate the same way, this alternate solution also supports backwards compatibility, as the additional optional argument `original_data_idx=None` (which is only supplied if multiple elements in a list are passed as the original_data parameter) can safely be ignored by existing code.

Thoughts?

sloria:
This behavior was implemented in https://github.com/marshmallow-code/marshmallow/pull/744 . Thanks again for your work on this.

PR #618, opened 2017-04-28
Fix a custom return in filter_fields method.

YuriHeupa:
Makes filter_fields method only returns fields pairs specified by `field_names`.

This will avoid schema fields to be overwritten with already excluded fields.


sloria:
Looks correct. Thanks @YuriHeupa !

PR #619, opened 2017-04-30
Fix inaccessible python-dateutil deserialization logic by adding a configuration parameter to DateTime

mchalek:
The existing implementation of `fields.DateTime` contains an [inaccessible code path](https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/fields.py#L959-L967) that appears to have intended fall back to `python-dateutil` for deserialization in cases when a `dateformat` is not defined.  It even contains a warning message suggesting that `python-dateutil` be installed, but the logging of this message is also inaccessible.

The code is inaccessible because it only runs if `self.dateformat` is `None`, but `self.dateformat` is always re-assigned to `self.DEFAULT_FORMAT` in an [earlier line](https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/fields.py#L947).

The changes I made here enable format-free deserialization via `python-dateutil`, requiring the user to "opt-in" by setting a parameter when constructing the `DateTime` field, e.g.:
```python
class MySchema(Schema):
    dt = DateTime(auto_deserialize_format=True)
```
I considered implementing this as a fallback, rather than an opt-in parameter, as the existing code seemed to be written.  However, it is clear that this could have unintended consequences: for example, one of the unit tests requires that a year-only input date be rejected, while `dateutil` will parse a year-only input using *today's date* as the source of default values for all missing fields, which is quite unintuitive (IMO).

Also, as another minor change, the reassignment of `self.dateformat` described above is both unnecessary and unexpected, as a typical user would probably not assume that `deserialize()` or `serialize()` methods would mutate class variables.  While it seems unlikely that this would ever have caused a problem, it still should probably be avoided, so I have changed both `serialize()` and `deserialize()` to eliminate this pattern.

I added a unit test to confirm that the code does correctly parse non-standard date formats when so configured (and not otherwise).  All other existing unit tests continue to pass.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -917,12 +917,17 @@ class DateTime(Field):
         'format': '"{input}" cannot be formatted as a datetime.',
     }
 
-    def __init__(self, format=None, **kwargs):
+    def __init__(self, format=None, auto_deserialize_format=False, **kwargs):
         super(DateTime, self).__init__(**kwargs)
         # Allow this to be None. It may be set later in the ``_serialize``
         # or ``_desrialize`` methods This allows a Schema to dynamically set the
         # dateformat, e.g. from a Meta option
         self.dateformat = format
+
+        if auto_deserialize_format and not utils.dateutil_available:
+            raise RuntimeError('auto_deserialize_format option requires the python-dateutil library')
+
+        self.auto_deserialize_format = auto_deserialize_format
 
     def _add_to_schema(self, field_name, schema):
         super(DateTime, self)._add_to_schema(field_name, schema)
@@ -964,17 +969,10 @@ class DateTime(Field):
             except (TypeError, ValueError):
                 raise self.fail('invalid')
 
-        if self.dateformat:
-            return deserialize_with_format(self.dateformat)
-        elif utils.dateutil_available:
+        if self.auto_deserialize_format:
             return deserialize_with_dateutil()
-        elif self.DEFAULT_FORMAT:
-            return deserialize_with_format(self.DEFAULT_FORMAT)
         else:
-            warnings.warn('It is recommended that you install python-dateutil '
-                          'for improved datetime deserialization.')
-            raise self.fail('invalid')
-
+            return deserialize_with_format(self.dateformat or self.DEFAULT_FORMAT)
 
 class LocalDateTime(DateTime):
     """A formatted datetime string in localized time, relative to UTC.
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -931,29 +931,29 @@ class DateTime(Field):
     def _serialize(self, value, attr, obj):
         if value is None:
             return None
-        self.dateformat = self.dateformat or self.DEFAULT_FORMAT
-        format_func = self.DATEFORMAT_SERIALIZATION_FUNCS.get(self.dateformat, None)
+        dateformat = self.dateformat or self.DEFAULT_FORMAT
+        format_func = self.DATEFORMAT_SERIALIZATION_FUNCS.get(dateformat, None)
         if format_func:
             try:
                 return format_func(value, localtime=self.localtime)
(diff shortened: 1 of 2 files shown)


znatty22:
Hey @mchalek, 

I'm also running into the same issue and wondering if you're still actively working on this PR?

Thanks!

mchalek:
hi @znatty22 to be honest I totally forgot about this.  It looks like the CI errors come from an unused variable, I should be able to push a commit to fix those.

But otherwise I never heard anything back from the maintainers, I don't know whether this is something they would consider or what the process would be to complete the merge.  Do you have any idea?

znatty22:
No worries, and that would be great if you could make that last commit! Hmm not sure if you've perused the contributing guidelines [here](https://github.com/marshmallow-code/marshmallow/blob/dev/CONTRIBUTING.rst)

But it looks like you might need to open an issue and reference it here and then maybe you need to submit this PR against the latest maintenance release branch since its a bug fix. Looks like that branch would be 
[2.x-line](https://github.com/marshmallow-code/marshmallow/tree/2.x-line)

mchalek:
ok @znatty22 thanks for the links.  I fixed the formatting errors.  I'll try to remember to follow through with the rest of the steps either this weekend or next week.

mchalek:
As a follow-up to #758 , I thought I should show how the code handles the example from that issue after this PR is merged:

By default, nothing changes:
```python
>>> import marshmallow as ma
>>> f = ma.fields.DateTime()
>>> f.deserialize('2018/03/26')
  File "marshmallow/fields.py", line 261, in deserialize
    output = self._deserialize(value, attr, data)
  File "marshmallow/fields.py", line 929, in _deserialize
    raise self.fail('invalid')
  File "marshmallow/fields.py", line 216, in fail
    raise ValidationError(msg)
marshmallow.exceptions.ValidationError: Not a valid datetime.
```

However, if the `DateTime` field is created with the `auto_deserialize_format` argument, then `dateutil` will be used:
```python
>>> f = ma.fields.DateTime(auto_deserialize_format=True)
>>> f.deserialize('2018/03/26')
datetime.datetime(2018, 3, 26, 0, 0)
```

lafrech:
Maybe the parameter could be `strict=False` to match `Integer(strict=False)` expressing that we're loose on the input.

mchalek:
hi @lafrech ,

sure, that's a good suggestion, I can make that change.  Are you able to review and/or merge this commit once it's done?  Should I contact any of the other contributors to help get this pushed through?

thanks,

Kevin

lafrech:
Ultimately, @sloria is the one who decides. If I were you, I'd wait for his feedback before I put too much effort on this.

mchalek:
Great, thanks @lafrech .

sloria:
Sorry it has taken so long for me to review this! Thanks @mchalek for your work.

I wonder if @deckar01 's suggestion [here](https://github.com/marshmallow-code/marshmallow/issues/758#issuecomment-378603115) is the way to go:

> What if we registered dateutil in DATEFORMAT_DESERIALIZATION_FUNCS instead?

The code would be much more explicit about using dateutil, rather than trying to fall back on it.

mchalek:
@sloria sure, whatever solution you think is best.  At this point, well over a year into the PR, It would be nice to have the issue be fixed in some way, but it really does not matter to me whether that's my original way or some other way, lol.  I'm glad to have raised the issue but I will probably not personally submit the reimplementation using the `DATEFORMAT_DESERIALIZATION_FUNCS` unless this issue comes up to bite me again sometime.

sloria:
@mchalek OK, no problem.

@lafrech @deckar01 Thoughts on this? I lean towards adding `'dateutil'` as a dateformat option. Should we get rid of automatic usage of dateutil if it's installed?

lafrech:
Closed (for now) by https://github.com/marshmallow-code/marshmallow/pull/962.

PR #620, opened 2017-05-01
Proposal: add support for field-level pre/post-processing

altaurog:
This is a feature I've been meaning to suggest since we used marshmallow in 
a small project back in December.  I am not a long time marshmallow user, 
and I haven't fully explored all its features or dug deep into the 
"marshmallow way."  I certainly haven't read any dev discussions regarding 
what features to include, but it seems to me this one could simplify 
several things.

I really like the ease `fields.Function` provides for defining custom field 
behavior. I'd love to see an equally easy way to build on top of all the 
functionality packed into the existing type fields.  To leverage the type 
conversion/checking they provide, I would have to subclass:

```
>>> f = fields.Int()
>>> f.deserialize('2')
2
>>> f.deserialize('2.1')
...
>>> f = fields.Function(deserialize=lambda x: x + 1)
>>> f.deserialize(2)
3
>>> f.deserialize(2.1)
3.1
>>> f.deserialize('2')
...
TypeError: Can't convert 'int' object to str implicitly
```

With pre- and post-processing functions on the base field class, this 
becomes much easier:

```
>>> f = fields.Int(post_deserialize=lambda x: x + 1)
>>> f.deserialize('2')
3
>>> f.deserialize('2.1')
...
```

One real-world case for this I came across was in handling optional search 
paramaters for a GET /entries api endpoint.  Each entry has a timestamp, 
and we needed to implement a `limit_days` parameter to get only entries 
from the last _n_ days.  `fields.Function(deserialize=timedelta)` doesn't 
cut it, because we really want all that `fields.Int()` validation first.  
So the options are either subclass `fields.Int` or process the data in a 
`post_load` function:

```
class EntrySearch(Schema):
    limit_days = fields.Int(missing=None)

    @post_load
    def post_process(self, data):
        if data['limit_days'] is not None:
            data['limit_days'] = timedelta(max(1, data['limit_days']))
```

A `post_deserialize` function on the field makes this much more 
straightforward:

```
>>> class EntrySearch(Schema):
...   limit_days = fields.Int(missing=None,
...                           post_deserialize=lambda d: timedelta(max(1, d)))
... 
>>> s = EntrySearch()
>>> s.load({})
UnmarshalResult(data={'limit_days': None}, errors={})
>>> s.load({'limit_days': '3'})
UnmarshalResult(data={'limit_days': datetime.timedelta(3)}, errors={})
```

Another real-world case I came across was handling different field formats, 
such as comma-separated list:

```
>>> f = fields.List(fields.Int(),
>>>                 pre_deserialize=lambda v: v.split(','),
>>>                 post_serialize=lambda v: ','.join(map(str, v)))
>>> f.deserialize('1,2,3')
[1, 2, 3]
>>> f.serialize('data', {'data': [2, 4, 6]})
'2,4,6'
```

(Incidentially, I would have thought `List(Int(as_string=True))` would 
serialize a list of ints to a list of str, but it doesn't)

Another real-world case I came across was a situation where a certain 
framework was parsing data sometimes as a list and sometimes as a scalar.

```
def force_list(val)
    if hasattr(v, 'join'):
        return v
    return [v]

class EntrySchema(Schema):
    ref_ids = fields.List(fields.Int(), pre_deserialize=force_list)
```

and another:

```
    id = fields.UUID(pre_deserialize=decrypt, post_serialize=encrypt)
```

and so forth.  Schema-level `pre_load`, `post_load`, `pre_dump` and 
`post_dump` still have applicability, just like schema-level validation in 
contradistinction to field-level validation.

So I hope you'll agree that this is a worthwhile feature to support.

I believe with this change, some of the examples could be simplified, but I 
only glanced at them.  If you like, I could update the examples to use 
these where sensible.

### DIFF
--- a/docs/custom_fields.rst
+++ b/docs/custom_fields.rst
@@ -4,6 +4,8 @@
 Custom Fields
 =============
 
+A lot of functionality can be added to the existing field classes on Instantiation using :ref:`field-level pre- and post-processing <field_pre_post>` functions.  For more flexibility in creating new functionality, marshmallow supports custom field classes.
+
 There are three ways to create a custom-formatted field for a `Schema`:
 
 - Create a custom :class:`Field <marshmallow.fields.Field>` class
--- a/docs/extending.rst
+++ b/docs/extending.rst
@@ -7,7 +7,9 @@ Extending Schemas
 Pre-processing and Post-processing Methods
 ------------------------------------------
 
-Data pre-processing and post-processing methods can be registered using the `pre_load <marshmallow.decorators.pre_load>`, `post_load <marshmallow.decorators.post_load>`, `pre_dump <marshmallow.decorators.pre_dump>`, and `post_dump <marshmallow.decorators.post_dump>` decorators.
+:ref:`Field-level pre-processing and post-processing <field_pre_post>` functions can cover many use cases in a very straightforward manner, but sometimes it may be necessary to apply pre- or post-processing to several fields or even the input collection as a whole.
+
+Schema-level pre-processing and post-processing methods can be registered using the `pre_load <marshmallow.decorators.pre_load>`, `post_load <marshmallow.decorators.post_load>`, `pre_dump <marshmallow.decorators.pre_dump>`, and `post_dump <marshmallow.decorators.post_dump>` decorators.
 
 
 .. code-block:: python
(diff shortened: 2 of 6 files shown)


sloria:
Closing this for now since it has become quite stale, and we don't really have the bandwidth to address this at the moment. Thanks anyway for the suggestion. Feel free to open up an issue for discussion if you're still interested in this feature.

PR #621, opened 2017-05-02
Record invalid type errors per index with many=True

tuukkamustonen:
Earlier, errors about invalid types (such as giving primitive for dict) got recorded on `_schema` level for list elements, rather than for each index separately.

So for example, first block in the added test used to produce:

```python
{
    '_schema': ['Invalid input type.', 'Invalid input type.']}
    0: {},
    1: {}
}
```

With this change it now produces:

```python
{
    0: {'_schema': ['Invalid input type.']}
    1: {'_schema': ['Invalid input type.']}
}
```

I don't know marshmallow internals well, and I am quite unsure this. But if this fix makes sense, we can iterate it into a proper form! :)

I used slightly different style to do asserts. I can replace strings by references to `Field.default_error_messages`, if that's better. Or remove value comparisons altogether. I just think this slightly more verbose form gives reader a better understanding of what is actually in `result.errors`.

### DIFF
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -256,14 +256,19 @@ class Unmarshaller(ErrorStore):
                 try:
                     raw_value = data.get(attr_name, missing)
                 except AttributeError:  # Input data is not a dict
-                    errors = self.get_errors(index=index)
                     msg = field_obj.error_messages['type'].format(
                         input=data, input_type=data.__class__.__name__
                     )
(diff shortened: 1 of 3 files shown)


tuukkamustonen:
Also this, what do you think @sloria? Candidate for 3.0.0?

tuukkamustonen:
This seems to be fixed (at least in `3.0.0b13`).

Issue #616, opened 2017-04-28
Record errors for invalid types passed to Schema.load()


tuukkamustonen:
With `2.13.5`:

```python
In [1]: from marshmallow import Schema, fields

In [2]: Schema().load({})
Out[2]: UnmarshalResult(data={}, errors={})

In [3]: Schema().load([])
Out[3]: UnmarshalResult(data={}, errors={})

In [4]: Schema().load(123)
Out[4]: UnmarshalResult(data={}, errors={})
```

1) I assume this doesn't produce errors, because there are no fields in the schema. However, as schema itself kinda represents `dict` type (=requires dict to parse values from), would it make sense to record error for `[]` and `123` etc.?

Then with `many=True`:

```python
In [5]: Schema(many=True).load({})
Out[5]: UnmarshalResult(data=[], errors={})

In [6]: Schema(many=True).load([])
Out[6]: UnmarshalResult(data=[], errors={})

In [7]: Schema(many=True).load(123)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-8ad8cdeda05d> in <module>()
----> 1 Schema(many=True).load(123)

/home/musttu/Code/virtualenvs/pastry/local/lib/python2.7/site-packages/marshmallow/schema.pyc in load(self, data, many, partial)
    578         .. versionadded:: 1.0.0
    579         """
--> 580         result, errors = self._do_load(data, many, partial=partial, postprocess=True)
    581         return UnmarshalResult(data=result, errors=errors)
    582 

/home/musttu/Code/virtualenvs/pastry/local/lib/python2.7/site-packages/marshmallow/schema.pyc in _do_load(self, data, many, partial, postprocess)
    658                     partial=partial,
    659                     dict_class=self.dict_class,
--> 660                     index_errors=self.opts.index_errors,
    661                 )
    662             except ValidationError as error:

/home/musttu/Code/virtualenvs/pastry/local/lib/python2.7/site-packages/marshmallow/marshalling.pyc in deserialize(self, data, fields_dict, many, partial, dict_class, index_errors, index)
    237                         partial=partial, dict_class=dict_class,
    238                         index=idx, index_errors=index_errors)
--> 239                     for idx, d in enumerate(data)]
    240 
    241             self._pending = False

TypeError: 'int' object is not iterable
```

2) I would kinda expect `{}` to record error (because it's not list). And int value (non-iterable value) should be gracefully caught into errors, rather than raising exception?

[...]

PR #622, opened 2017-05-02
Accept only lists and sets when many=True

tuukkamustonen:
It doesn't make sense to accept any iterable, because list of primitives cannot be defined anyhow (on `Schema` level).

Accepting any iterable (such as some "string") produces weird outcome with no good. It's better to fail fast and record error on invalid type. I assume there is some reasoning why any iterable is accepted, even though library doesn't support list of non-dicts?

This also fixes `TypeError`, that is thrown when non-iterable value is given to `load()` with `many=True` (see https://github.com/marshmallow-code/marshmallow/issues/616).

### DIFF
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -10,6 +10,7 @@ and from primitive types.
 
 from __future__ import unicode_literals
 
+from marshmallow.fields import Field
 from marshmallow.utils import is_collection, missing, set_value
 from marshmallow.compat import text_type, iteritems
 from marshmallow.exceptions import (
@@ -231,7 +232,11 @@ class Unmarshaller(ErrorStore):
         # Reset errors if not deserializing a collection
         if not self._pending:
             self.reset_errors()
-        if many and data is not None:
+        if many and data is not None and not isinstance(data, (list, set)):
+            self.error_field_names.append(SCHEMA)
+            self.errors.setdefault(SCHEMA, []).append(Field.default_error_messages['type'])
+            ret = data
+        elif many and data is not None:
             self._pending = True
             ret = [self.deserialize(d, fields_dict, many=False,
                         partial=partial, dict_class=dict_class,
@@ -247,7 +252,7 @@ class Unmarshaller(ErrorStore):
                     data=ret,
                 )
             return ret
-        if data is not None:
+        elif data is not None:
             partial_is_collection = is_collection(partial)
             ret = dict_class()
             for attr_name, field_obj in iteritems(fields_dict):
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -14,6 +14,7 @@ from marshmallow import (
     validates, validates_schema
 )
 from marshmallow.exceptions import ValidationError
+from marshmallow.fields import Field
 
 from tests.base import *  # noqa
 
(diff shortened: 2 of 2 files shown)


tuukkamustonen:
I wonder if there's need to support iterators and generators, still.

edelooff:
Accepting only lists and sets (or subclasses thereof) instead of any generic iterable (generator or otherwise) seems like an overly blunt way of solving the bad behavior of one iterable type (the string).

An explicit test and failure for string-like types makes sense. But for any other type, if it supports iteration, `many=True` should not suddenly raise an error because it's not on a whitelist.

tuukkamustonen:
@edelooff String is one example. It works, but produces weird output into `errors` dict. Then there are non-iterables, which seem to raise `TypeError`. After you tackle those cases, isn't it just list/sets (and I guess iterables/generates) that are left behind?

edelooff:
Doing an explicit `isinstance` check for string-like types makes sense. The rest of it could probably be solved by wrapping the relevant code in a `try` block and catching the inevitable `TypeError` in case of a non-iterable value (on the assumption that any `TypeError` in nested a schema does not bubble up, which I'm fairly sure holds true)

tuukkamustonen:
I would rather fail fast (=check type) than use try-except, but whatever works and follows project style. I can try that.

Any pointers to what should be passed as `ValidationError(data=X)`? I am not sure what is done with it, if anything (I assume it's for debug purposes only).

edelooff:
Both approached fail about as fast, but the try/except is more accepting in what will work, which seems more in keeping with general Python style. As far as the project and its style goes, I'm just a user with an opinion.

As for the `ValidationError`, I've no idea really.

tuukkamustonen:
@sloria Seeing that marshmallow 3.0.0 is on its way, care to consider this to be merged?

I haven't touched the PR, so feedback by @edelooff is yet to be applied, but what do you think about the idea?

rgant:
For marshmallow-jsonapi I need to customize the Schema to add this sort of checking for a list.
https://github.com/rgant/saas-api-boilerplate/blob/master/ourmarshmallow/schema.py#L145

It would be nice to have this folded into the core marshmallow. I could then look for those errors and then raise the appropriate validation errors for JSONAPI.

marshmallow/marshalling.py
```diff
@@ -231,7 +232,11 @@ def deserialize(self, data, fields_dict, many=False, partial=False,
         # Reset errors if not deserializing a collection
         if not self._pending:
             self.reset_errors()
-        if many and data is not None:
+        if many and data is not None and not isinstance(data, (list, set)):
```
rgant, line 235: What about using marshmallow.utils.is_collection instead of this direct call to isinstance?

tuukkamustonen: Yeah, that would be perfect.

sloria:
Sorry it's taken so long to get to this.

I understand the problem this is trying to solve, but I'm concerned that this solution won't account for all the collection types that might be passed to a schema. Using `is_iterable_but_not_string` might not be enough; it essentially just checks if an object inherits from `collections.Iterable`. There are many ORMs out there with custom collection types, e.g., querysets, and I would hate to break use cases with those.

tuukkamustonen:
Thanks for looking into it, we know you're busy.

I agree - PR as is would be too restrictive, though I just don't see use case for accepting `str` with `many=True`.

> Using is_iterable_but_not_string might not be enough; it essentially just checks if an object inherits from collections.Iterable

It's been a while, but does the code check anything at the moment? Could we check & raise strings, and let everything else pass as currently (I assume it fails somewhere else then).

Or would you recommend to rather explicitly check & disallow strings?

> There are many ORMs out there with custom collection types, e.g., querysets, and I would hate to break use cases with those.

That's a reasonable concern. But aren't they iterators or generators anyway...? `is_iterable_but_not_string` appears to check for both.

---

My use case here: use `many=True` in BATCH `POST/PATCH` endpoint in HTTP API, which accepts either a JSON object or a list of objects, but I want strings to be rendered as invalid input.

tuukkamustonen:
Something like this was implemented in #954.

PR #625, opened 2017-05-11
Make UUIDs valid if input is string representation or 16 byte rep

JeffBerger:
My database contains UUIDs in a 16 byte representation and not the string representation. The python UUID library does not automatically detect the difference if passed as the first argument, byte representations must be passed into the `bytes` kwarg of the constructor.

This PR is to allow marshmallow to accept both the normal string representation and the byte representation.

In python 2.x bytes is the same type as string so for a valid string representation this will pass, so an additional check is made on the length being 12.  For python 3.x a string representation will be a different type so the first statement will catch it.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -623,7 +623,10 @@ class UUID(String):
         if isinstance(value, uuid.UUID):
             return value
         try:
-            return uuid.UUID(value)
+            if isinstance(value, bytes) or len(value) == 16:
+                return uuid.UUID(bytes=value)
+            else:
+                return uuid.UUID(value)
         except (ValueError, AttributeError):
             self.fail('invalid_uuid')
 
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -623,8 +623,11 @@ class UUID(String):
         if isinstance(value, uuid.UUID):
             return value
         try:
-            return uuid.UUID(value)
-        except (ValueError, AttributeError):
+            if isinstance(value, bytes) and len(value) == 16:
+                return uuid.UUID(bytes=value)
+            else:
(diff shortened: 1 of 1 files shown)


sloria:
Thanks @JeffBerger for this PR. I *think* this is the correct behavior. I looked into what djangorestframework does, and it looks like they only pass `hex` (https://github.com/encode/django-rest-framework/blob/1e9e1a5bfebaa7bb1ceb5dcb546b7bbd83c209ba/rest_framework/fields.py#L836-L837). I'm unsure if this was an oversight, but the discussion [here](https://github.com/encode/django-rest-framework/issues/3679) suggests that the authors were at least aware of the `bytes` argument. I'd like to hear a response on my comment [here](https://github.com/encode/django-rest-framework/issues/3679#issuecomment-302589556) before proceeding with this.

JeffBerger:
Sounds great - thanks for getting back to me on this, I'll start watching the other thread as well.

JeffBerger:
@sloria - I am just checking in about getting this merged.  I looked at the reference and it seemed that it was given the ok in a comment in October on the django-rest-framework thread

sloria:
@JeffBerger The DRF thread has been silent, but I think we can move forward with this as I think it's the correct behavior. Would you be up for adding a test for this?

Also, don't forget to add yourself to AUTHORS.rst.

Thanks!

JeffBerger:
Sure I'll make some tests, I'll see if I can get to it before this holiday break but if not I'll get back to you in the new year.  Cheers!

sloria:
Just checking in-- @JeffBerger do you still plan to add tests to this? If not, just let us know so that someone else can pick it up.

sloria:
I picked this up in #816.

PR #576, opened 2017-01-25
Add benchmark to Marshmallow

rowillia:
Pull out the benchmarking code from  https://github.com/marshmallow-code/marshmallow/pull/573 into a seperate PR.

PR #628, opened 2017-05-16
Add benchmark into Marshmallow

rowillia:
This rebases and re-adds https://github.com/marshmallow-code/marshmallow/pull/576

### DIFF
--- a/performance/benchmark.py
+++ b/performance/benchmark.py
@@ -1,0 +1,130 @@
+"""Simple benchmark for Marshmallow serialization of a moderately complex object.
+
+Uses the `timeit` module to benchmark serializing an object through Marshmallow.
+"""
+
+from __future__ import print_function, unicode_literals, division
+
+import argparse
+import cProfile
+import gc
+import timeit
+import time
+
+from marshmallow import Schema, fields, ValidationError, pre_load
+
+
+# Custom validator
+def must_not_be_blank(data):
+    if not data:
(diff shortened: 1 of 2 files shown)


tasks.py
```diff
@@ -32,6 +33,13 @@ def flake(ctx):
     ctx.run('flake8 .', echo=True)
 
 @task
+def benchmark(ctx):
+    """Run fast benchmark on codebase."""
+    # Explicitly shell out to get more consistent results by creating a new process
+    # every time, for example running out of process ensures a pristine class_registry.
+    ctx.run('python performance/benchmark.py --iterations=100 --repeat=3', echo=True)
```
rowillia, line 40: @sloria You had concerns about shelling out here.  I added a comment as to why running benchmarks out of band helps.

performance/benchmark.py
```diff
@@ -0,0 +1,130 @@
...
+        profile.enable()
+
+    gc.collect()
+    best = min(timeit.repeat(lambda: quotes_schema.dump(quotes),
+                             'gc.enable()',
+                             number=iterations,
+                             repeat=repeat))
+    if profile:
+        profile.disable()
+        profile.dump_stats('marshmallow.pprof')
```
sloria, line 99: This file should be gitignored.

sloria: Made this change when I merged.

PR #629, opened 2017-05-16
Implement hooks for a JIT to integrate into marshmallow.

rowillia:
This is a revival of https://github.com/marshmallow-code/marshmallow/pull/573

Unfortunately other personal/professional priorities prempted me from focusing on this for the past few months, but I'd like to revisit open sourcing this again.

When we deployed this to production we saw ~20x improvement in serialization time which, in turn, dropped our response times in half.

I totally understand the apprehension on the original PR due to the complexity this introduces, so I've created a follow on PR that integrates the hooks I need into marshmallow, leaving everything else relatively unchanged.

We're still able to get the same performance gains, and I'm able to iterate on my JIT library without having to pester you with PRs :D.  If we can get this integration working we'll have an awesome story about how Marshmallow allows engineers to have the amazing API of Marshmallow without having to sacrifice performance.

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -2,12 +2,15 @@
 """The :class:`Schema` class, including its metaclass and options (class Meta)."""
 from __future__ import absolute_import, unicode_literals
 
+from abc import ABCMeta, abstractmethod, abstractproperty
 from collections import defaultdict, Mapping
 import copy
 import datetime as dt
 import decimal
 import inspect
+import importlib
 import json
+import os
 import uuid
 import warnings
 from collections import namedtuple, OrderedDict
@@ -19,7 +22,32 @@ from marshmallow.exceptions import ValidationError
 from marshmallow.orderedset import OrderedSet
 from marshmallow.decorators import (PRE_DUMP, POST_DUMP, PRE_LOAD, POST_LOAD,
                                     VALIDATES, VALIDATES_SCHEMA)
-from marshmallow.utils import missing
+from marshmallow.utils import missing, suppress
+
+DEFAULT_JIT_ENVIRONMENT_VARIABLE = 'MARSHMALLOW_SCHEMA_DEFAULT_JIT'
+DEFAULT_JIT = missing
+
+def get_default_jit():
+    """Allows overriding the default JIT to use from the environment.
+
+    This is useful for running tests as well as rolling out a Marshmallow JIT to
+    every schema in a process.
+    """
+    global DEFAULT_JIT
+    if DEFAULT_JIT != missing:
+        return DEFAULT_JIT
+
+    DEFAULT_JIT = None
+
+    with suppress(Exception):
+        default_jit_path = os.getenv(DEFAULT_JIT_ENVIRONMENT_VARIABLE)
+
+        if default_jit_path:
+            parts = default_jit_path.split('.')
+            module_name = '.'.join(parts[0:-1])
+            class_name = parts[-1]
+            module = importlib.import_module(module_name)
+            DEFAULT_JIT = getattr(module, class_name, None)
 
 
 #: Return type of :meth:`Schema.dump` including serialized data and errors
@@ -174,6 +202,37 @@ class SchemaMeta(type):
                 self.__processors__[tag].append(attr_name)
 
 
+class SchemaJit(with_metaclass(ABCMeta)):
+    """Abstract Base Class for implementing a jit for Marshmallow.
+
+    A SchemaJit should create methods to marshal or unmarshal an object for a
+    given schema.  These methods should, given an object, optimistically perform
+    marshalling or unmarshalling.  If they fail, Marshmallow will fallback to the existing
+    reflection-based methods.
+    """
+    @abstractmethod
+    def __init__(self, schema):
+        pass
+
+    @abstractproperty
+    def jitted_marshal_method(self):
+        """Returns the jitted marshal method.
+
+        :return: A method that accepts an object and returns the marshalling result or None
+            if there is no applicable marshal method.
+        """
+        pass
+
+    @abstractproperty
+    def jitted_unmarshal_method(self):
+        """Returns the jitted unmarshal method.
+
+        :return: A method that accepts an object and returns the unmarshalling result or None
+            if there is no applicable unmarshal method.
+        """
+        pass
+
+
 class SchemaOpts(object):
     """class Meta options for the :class:`Schema`. Defines defaults."""
 
@@ -206,6 +265,7 @@ class SchemaOpts(object):
         self.include = getattr(meta, 'include', {})
         self.load_only = getattr(meta, 'load_only', ())
         self.dump_only = getattr(meta, 'dump_only', ())
+        self.jit_options = getattr(meta, 'jit_options', {})
 
 
 class BaseSchema(base.SchemaABC):
@@ -345,12 +405,26 @@ class BaseSchema(base.SchemaABC):
         self.context = context or {}
         self._normalize_nested_options()
         self._types_seen = set()
+        self._jit_class = get_default_jit()
+        self._jit_instance = None
+        self._current_name_to_type = None
         self._update_fields(many=many)
 
     def __repr__(self):
         return '<{ClassName}(many={self.many}, strict={self.strict})>'.format(
             ClassName=self.__class__.__name__, self=self
         )
+
+    @property
+    def jit(self):
+        return self._jit_class
+
+    @jit.setter
+    def jit(self, value):
+        self._jit_instance = None
+        self._current_name_to_type = None
+        self._jit_class = value
+        self._update_fields(many=self.many)
 
     @property
     def dict_class(self):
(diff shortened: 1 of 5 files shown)


sloria:
Thanks for reviving this. I am glad to hear you're developing the JIT code as a separate library; this is certainly an improvement.

I am still concerned about the added complexity here--the added hooks are very specific to JIT compilation. Can we implement more general-purpose hooks that will allow you to add JIT compilation? Also, I think it would be more intuitive to factor the JIT logic into a Schema class, rather than having to set a `jit` attribute on a schema.

```python
class AuthorSchema(JITSchema):
    ...
```

Then perhaps the JIT could be customized as a class Meta option:

```python
class AuthorSchema(JITSchema):
     class Meta:
         jit = SuperFastJIT
```

rowillia:
Thanks for the feedback @sloria.  The main hooks that are needed are fairly jit-specific.  We need the hook to recompile code upon fields changing (either from the constructor or from field types changing across invocations), we need the hooks to attempt the serialization/deserialization first before falling back to the methods in `marshmallow.marshalling`, and we need the hook for `Nested` fields to use the same JIT as their parent.

I refactored some of the code to create less code duplications and documented these single points.

RE: JitSchma inheriting from Schema - I thought about that at length but ultimately decided against it for a couple of reasons.  First, it makes it substantially easier to test the JIT without having to monkey patch `Schema` and second it makes the deployment model simpilier...we install the JIT library and set the environment variable override and we're good to go!

rowillia:
Ping @sloria !  Let me know if there's anything else I can do here or if there's any other ideas you've got as to what the right integration points would be.  I'm happy to VC to talk through any concerns as well whenever would work for you, feel free to reach out over email and we'll work a time out!

sloria:
> The main hooks that are needed are fairly jit-specific. 

Hm, ok, without having looked to deeply into the implementation, I guess I was hoping the names of the hooks could at least be more general. I'll have to think more on this 😔 

> RE: JitSchma inheriting from Schema - I thought about that at length but ultimately decided against it for a couple of reasons. First, it makes it substantially easier to test the JIT without having to monkey patch Schema and second it makes the deployment model simpilier...we install the JIT library and set the environment variable override and we're good to go!

You can still achieve these benefits even with a Schema subclass.

```python
BaseSchema = JITSchema if os.environ.get('USE_JIT') else Schema
    
class MySchema(BaseSchema):
    #...
```

rowillia:
> You can still achieve these benefits even with a Schema subclass.

@sloria That's definitely doable, but if we wanted to run the Marshmallow test suite as part of testing the JIT we'd have to go through and codemod Marshmallow itself before running the tests.  Additionally this makes adoption of a Jit harder as before it can be used users would have to go through and refactor their Schema code.

If that's the only point of contention I'm happy to remove the ability to override the default Jit and do some `sed` magic in our library.

marshmallow/schema.py
```diff
@@ -19,7 +22,31 @@
 from marshmallow.orderedset import OrderedSet
 from marshmallow.decorators import (PRE_DUMP, POST_DUMP, PRE_LOAD, POST_LOAD,
                                     VALIDATES, VALIDATES_SCHEMA)
-from marshmallow.utils import missing
+from marshmallow.utils import missing, suppress
+
+DEFAULT_JIT_ENVIRONMENT_VARIABLE = 'MARSHMALLOW_SCHEMA_DEFAULT_JIT'
```
sloria, line 27: I don't think this configuration parsing belongs in marshmallow. As I suggested in my previous comment, users can do this easily in their application code

```python
BaseSchema = JITSchema if os.environ.get('USE_JIT') else Schema
JITClass = import_object(os.environ.get('MARSHMALLOW_SCHEMA_DEFAULT_JIT'))
```

marshmallow/schema.py
```diff
@@ -174,6 +201,37 @@ def _resolve_processors(self):
                 self.__processors__[tag].append(attr_name)
 
 
+class SchemaJit(with_metaclass(ABCMeta)):
```
sloria, line 204: Couldn't this go in your external library?

Ideally, all JIT-related code would go in the library, i.e. `JITSchema`, `JITSchemaOpts`, `SchemaJIT`..

I imagine it would look like:

```python
import os
from toastedmarshmallow import Schema as JITSchema, DefaultJIT

BaseSchema = JITSchema if os.environ.get('MARSHMALLOW_USE_JIT') else Schema

class MySchema(BaseSchema):
    class Meta:
        jit = DefaultJIT()
```

marshmallow/schema.py
```diff
@@ -353,6 +418,17 @@ def __repr__(self):
         )
 
     @property
+    def jit(self):
```
sloria, line 421: This property would go on my proposed `Schema` subclass in the external library.

sloria (changes requested):
In general, I'd like to see more of JIT-related code in the external library. I am willing to add more hooks to  `marshmallow.Schema`  to minimize code duplication.

marshmallow/schema.py
rowillia: This has to be part of Marshmallow itself as it sets up the contract for dealing with the external library, otherwise the external library has to duplicate huge swaths of this file to inject itself.  Doing this via hooks and a contract creates a much looser coupling with Marshmallow and the external library.

The other main issue with the subclass approach is that it doesn't allow some of the other great libraries in the Marshmallow ecosystem to benefit from this.  For example there's no way for me to compose a `flask_marshmallow.Schema` with a `toastedmarshmallow.JitSchema` without effectively doubling the class hierarchy in all of those other frameworks.

s0undt3ch:
Looking forward on this getting in.

s0undt3ch:
Have we lost traction on getting this in?

s0undt3ch:
@rowillia hope you didn't lost interest in convincing @sloria on getting this in. BTW, conflicts.

sloria:
Update: I'm still not completely convinced that the JIT-related code can't be factored out of marshmallow core, but honestly, I haven't dug deep into this PR nor toasted-marshmallow.

I'm not opposed to adding more hooks to support toasted-marshmallow's use case. I'm just really hesitant to add JIT-related code since it's so specific.

Right now, we're focused on shipping marshmallow 3, which is why I haven't been able to get to this PR. Apologies for the delay. I'll revisit this once we get more of the high-priority marshmallow 3 issues and PRs out the door.

taion:
Looking at this code a bit, it seems almost like all the hooks could be implemented by wrapping the marshaller and unmarshaller with a custom class that implements the Jit hooks.

@rowillia Am I missing something here? If not, would you mind if I took a stab at doing this?

deckar01:
I would like to open an issue for improving serialization performance and compare JIT with some alternatives. I think the complexity of this proposal warrants a comprehensive justification.

taion:
I'm spiking something to let toasted-marshmallow inject itself without requiring any explicit hooks in Marshmallow itself.

I think the codegen idea is a pretty good idea... saves a lot of hitting extra functions and such.

taion:
See https://github.com/lyft/toasted-marshmallow/pull/16.

A quick test suggests that performance is comparable what it is with this implementation. Unless I'm missing something there, then it's not actually necessary to inject any hooks here to get the desired behavior.

s0undt3ch:
I believe the hooks were a way to have your own schemas JIT compiled and thus faster, but also, any other schema, from a 3rd-party library.
Your approach only enables JIT on Your own schemas... Which seems like the path of least surprise, but I also get the benefits of, because my library says so, have any 3rd party lib also use JIT...

taion:
That may be, but in that case I believe the approach above still permits monkeypatching the exported `Schema` from Marshmallow to have the desired semantics.

taion:
@deckar01 Did you get a chance to look into those performance issues? The speedup from toasted-marshmallow is pretty significant. The current release isn't quite drop-in compatible, though, so I'm coming on needing to make a choice as to what to do. I can help you look as well, especially if there are any particular culprits you've identified.

nikoladsp:
Hi all,
any update on this?

sloria:
@nikoladsp We're still focused on marshmallow 3 at the moment, so my comment above still applies:

> I'm not opposed to adding more hooks to support toasted-marshmallow's use case. I'm just really hesitant to add JIT-related code since it's so specific.
> Right now, we're focused on shipping marshmallow 3, which is why I haven't been able to get to this PR. Apologies for the delay. I'll revisit this once we get more of the high-priority marshmallow 3 issues and PRs out the door.

That said, we're significantly closer to a 3.0 release. In the meantime, anyone should feel free to do an analysis of alternatives, like the ones suggested by @taion and @deckar01 above.

taion:
With some refactoring, I think https://github.com/lyft/toasted-marshmallow/pull/16 is the way to go here. Unless you're somehow using a bunch of third-party schemas over which you have no control, the approach in that PR allows injecting the JIT hooks with no changes here, at least as of the time it was written.

I think we may need to do a small internal refactoring of the construction of the `Marshaller` objects with some subsequent code changes, but it'd be a ~10-LoC change, rather than something that adds very customized hooks.

pzybe:
commenting to show interest, we're looking to reduce our validation bottlenecks this would be very appreciated.

sloria:
Closing, as this has gone quite stale and activity on toasted-marshmallow has been pretty quiet. 

FWIW, marshmallow 3's serialization performance is significantly better than marshmallow 2 (not toasted fast, but acceptable for most use cases).

I'm still open to exposing new hooks that would make injecting a JIT possible, but the approach will have to be different from the one here.

PR #632, opened 2017-05-19
Enum field

lafrech:
Our models often use string fields where the value must be in an `Enum`. We use a `String` field and add a `OneOf` validator, so that the deserialized string is validated against the `Enum`.

This fields allows to deserialize as an `Enum` instance rather than as a string.

I also added an `as_string` parameter, like in `Number` field. Some may want to serialize as `Enum` and only use the validation feature. It defaults to `False`, like in `Number` field.

This does not add a dependency to `enum`. It is not imported in the code, and corresponding tests are skipped if it can't be imported. When using Python < 3.4, the backported version can be used.

This is not critical to us, but I thought it might help.

Feedback welcome.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -4,6 +4,7 @@ invoke>=0.13.0
 # Soft dependencies
 python-dateutil
 pytz
+enum34 ; python_version < '3.4'
 
 # Distribution
 wheel
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -1312,6 +1312,24 @@ class Constant(Field):
         return self.constant
 
 
+class Enum(Field):
+
+    default_error_messages = {'invalid': 'Not a valid choice.'}
+
+    def __init__(self, enum, *args, **kwargs):
+        super(Enum, self).__init__(*args, **kwargs)
+        self._enum = enum
+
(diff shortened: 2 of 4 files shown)


justanr:
The last time this came up, I ended up writing [Marshmallow-Enum](https://github.com/justanr/marshmallow_enum).

lafrech:
Thanks @justanr, I didn't see that. In fact I didn't even search before proposing this.

The differences between our implementations are:

- Mine offers to serialize either as string or as enum (in which case, the field only does validation, nothing more).

- Yours offers to serialize as value. I thought of this but dismissed the idea as values can be anything, not only integers, so they could need a non-trivial serialization themselves and things could get complicated.

@sloria, would you find any interest in such a field?

I added a commit to make enum a dev requirement, so there is still no install dependency on enum, but this ensures it is always tested.

sloria:
I am hesitant to merge this since marshmallow-enum exists as a stable and maintained 3rd-party package. @lafrech If there are limitations you see in marshmallow-enum, send a PR to that package for now. @justanr If you are ever in need of maintenance help for marshmallow-enum, let us know here; I and others would be glad to help out.

justanr:
Thanks, I'll be sure to shout if it's needed.

PR #635, opened 2017-05-24
Allow Number fields to prohibit boolean values.

### LINKED ISSUE #623: Don't accept True/False as Int()
In python, `True`/`False` are tracked by an int value, but I think marshmallow should not accept `True` or `False` as `Int()` field. Or at least it should be configurable.

rowillia:
After a bunch of research, I decided to _explicitly_ special case booleans as opposed
to adding a flag for only allowing numbers to be serialized.  I came across code in
our repositories that, for example, overload `__int__` or `__float__` that would expect
to be able to be serialized.  I agree, however, that allowing booleans to be serialized
as numbers is likely incorrect.

This fixes https://github.com/marshmallow-code/marshmallow/issues/623

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -639,8 +639,6 @@ class Number(Field):
     """Base class for number fields.
 
     :param bool as_string: If True, format the serialized value as a string.
-    :param bool prohibit_boolean: If True, attempting to serialize a boolean
-        will result in a ValueError instead of coercing the value to a number.
     :param kwargs: The same keyword arguments that :class:`Field` receives.
     """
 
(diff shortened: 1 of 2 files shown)


marshmallow/fields.py
```diff
@@ -639,6 +639,8 @@ class Number(Field):
     """Base class for number fields.
 
     :param bool as_string: If True, format the serialized value as a string.
+    :param bool prohibit_boolean: If True, attempting to serialize a boolean
```
sloria: I don't think we need to add this as a parameter. `Number` should only accept numbers as input; booleans should be considered invalid.

sloria (changes requested):

sloria:
Thanks for the PR! Just one suggestion above.

rowillia:
@sloria Done!

sloria (approved):

sloria:
Thanks!

PR #654, opened 2017-07-05
Improvement: Schema tags: added "order" option

vgavro:
For now, if you're using mixins or subclassing marshmallow-sqlalchemy (which is using @post_load for model creation), tags would be invoked in alphabetically order (because they're relying on dir() accessing attribute, which is not obvious anyway, and better be documented?). I think it would be better to pass verbose "order" argument, to make more complex mixins and schema subclassing. Please consider this PR, if you think it's ok I'll add tests and changes in documentation. Thank you.
Use-cases: data changes in @post_load and @pre_load, on which other mixins or subclasses may be relying.

### DIFF
--- a/marshmallow/decorators.py
+++ b/marshmallow/decorators.py
@@ -69,7 +69,8 @@ def validates(field_name):
     return tag_processor(VALIDATES, None, False, field_name=field_name)
 
 
-def validates_schema(fn=None, pass_many=False, pass_original=False, skip_on_field_errors=True):
+def validates_schema(fn=None, pass_many=False, pass_original=False, skip_on_field_errors=True,
+                     order=0):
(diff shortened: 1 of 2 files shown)


vgavro:
note that this PR is backward-compatible: so if you're relying on alphabetic order of method name it will remain with default sort value.

sloria:
Thanks for the PR! This is very similar to #600 . I think "order" may be more intuitive than "priority". That said, I'm still not sure whether this is necessary. See the [discussion on #600](https://github.com/marshmallow-code/marshmallow/pull/600#issuecomment-286306434): most use-cases should be covered by using a single `pre/post` method that calls other methods in order.

sloria:
Closing this for now. Further discussion of this should happen in #600

PR #665, opened 2017-07-31
fix #664: validate.URL: allow hostnames with no dots

### LINKED ISSUE #664: validate.URL rejects hostnames with no dots
My current use case:
I use Docker and I need to set some URL fields for the containers to communicate together. The URL fields look like `http://third-party/test`, where `third-party` is the hostname of one of the containers.

Currently, `validate.URL` rejects this URL as invalid, because there is no dot in the hostname `third-party`.

Other use cases include:
* a local network in the same subdomain, where hosts are queried using only the name of the machine, not the FQDN, having configured the DNS resolver to search `example.com` automatically, for example `webclient.example.com` querying `webserver` instead of `webserver.example.com`
* a stricly-local network, where there are no public TLD available, and all machines are name `machine1`, `machine2`, etc. very similarly to the Docker internal network

sduthil:
reason: a hostname with no dots is valid, and useful in some cases,
e.g. a host in the same subdomain than the client, a shorter alias to
a longer hostname, a local network without TLD

### DIFF
--- a/marshmallow/validate.py
+++ b/marshmallow/validate.py
@@ -47,7 +47,7 @@ class URL(Validator):
     URL_REGEX = re.compile(
         r'^(?:[a-z0-9\.\-\+]*)://'  # scheme is validated separately
         r'(?:[^:@]+?:[^:@]*?@|)'  # basic auth
-        r'(?:(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+'
+        r'(?:(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.?)+'
         r'(?:[A-Z]{2,6}\.?|[A-Z0-9-]{2,}\.?)|'  # domain...
         r'localhost|'  # localhost...
         r'\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}|'  # ...or ipv4
--- a/tests/test_validate.py
+++ b/tests/test_validate.py
@@ -23,6 +23,7 @@ from marshmallow import validate, ValidationError
     'http://123.45.67.8/',
     'http://2001:db8::ff00:42:8329',
     'http://www.example.com:8000/foo',
+    'http://my-local-neighbor',
 ])
 def test_url_absolute_valid(valid_url):
     validator = validate.URL(relative=False)


sloria:
Thanks for the PR @sduthil . 

Both Django's and WTForms' URL validators disallow URLs with no periods, which makes me hesitant to merge this. Your use cases sound valid, though, so I'll keep this open for discussion for the time being.

sduthil:
From what I read, WTForms allows dotless URL, with a specific `require_tld=False` option (see the [interface](https://github.com/wtforms/wtforms/blob/master/wtforms/validators.py#L381,L388) and the [implementation](https://github.com/wtforms/wtforms/blob/master/wtforms/validators.py#L518,L528))

About Django, there is a [opened feature request](https://code.djangoproject.com/ticket/25418) and another quite old [ticket closed](https://code.djangoproject.com/ticket/9202).

There's also a [recent discussion](https://groups.google.com/forum/#!topic/django-developers/ASBJ0ge2KYo/discussion) about email validation, where some comments are relevant to URLs, some aren't.

sloria:
Thanks @sduthil for doing that research.

It appears from both those Django tickets that there is still an open discussion on whether to make the validator more permissive:

https://code.djangoproject.com/ticket/25418#comment:2

>I believe the idea of URLValidator is to recognize URLs that usually work without some special DNS setup. I feel like this proposal has come up before and been rejected. If so, we should document the restriction (and how to lift it in your own validation) to try to prevent it from being proposed again and again.

https://code.djangoproject.com/ticket/25418#comment:8

> Tim, the fact that you've had a duplicate from me suggests that the regex is wrong. ​https://tools.ietf.org/html/rfc3986#section-3.2.2 does not specify the need for either a domain or a tld.
That's the technicality, the practicality is that most intranet urls are left in short form, without a domain or tld. These are not a "special" DNS setup, and I can't see a reason to simplify the regex.

https://code.djangoproject.com/ticket/9202#comment:1

>I'm not sure we want to be so permissive for URLField when a RegexField with a custom regex does the same thing. The point of the URLField is to validate "common" URLs; for the vast majority of users http://django/ is a typo.

https://code.djangoproject.com/ticket/9202#comment:4

> So this regexp, while technically correct, isn't appropriate to validate the contents of URLField; it's too permissive.

So I'm still hesitant to merge this. I would be more likely to merge something along the lines of WTForms' approach, i.e. strict validation by default, with an option to be more permissive.

sduthil:
I agree that the WTForms approach is better, and backward compatible. I'll open another PR when I have some time.

Thanks for the discussion, I learned a few things in the process :)

Issue #231, opened 2015-06-29
Make fields deserialization more strict ?


touilleMan:
Working with the library, it appears to me that the deserialization module has a lack of strictness from my point of view:

``` python
class DocSchema(Schema):
    str = fields.String()
   bool = fields.Boolean()

# Ok, None is not implicitly acceptable
DocSchema().load({'str': None, 'bool': None})
# UnmarshalResult(data={}, errors={'str': ['Field may not be null.'], 'bool': ['Field may not be null.']})

# But it's not the case for numbers
DocSchema().load({'str': 42, 'bool': 42})
# UnmarshalResult(data={'str': '42', 'bool': True}, errors={})

# Or for any complex structure (list, dict)
DocSchema().load({'str': {}, 'bool': {}})
# UnmarshalResult(data={'str': '{}', 'bool': False}, errors={})
```

Is there a way to enforce real strict validation ? Passing a dict as argument of a string field without getting any error sounds like a silent fail for me...

zachguo:
:+1: I'm having same issues.

density:
I think you can fix this by creating your own `Field` subclass and implementing a stricter `_deserialize` method. But I agree that the Fields should not be so forgiving. Would love to hear @sloria's view on this -- happy to help implement it if it's what the people want :)

sloria:
I agree with the premise of this issue.

For frame of reference, I wrote a script to compare the outputs of marshmallow, colander, and django-rest-framework:

``` python
DATA = {'foo': 42, 'bar': 24}

### marshmallow ###

from marshmallow import Schema, fields

class MySchema(Schema):
    foo = fields.Str()
    bar = fields.Bool()

data, errors = MySchema().load(DATA)

print('--- marshmallow ---')
print(errors)

### colander ###

import colander as col

class MySchema(col.MappingSchema):
    foo = col.SchemaNode(col.String())
    bar = col.SchemaNode(col.Boolean())

print('--- colander ---')
try:
    MySchema().deserialize(DATA)
except col.Invalid as err:
    print(err.asdict())
except:
    pass

### DRF ###

from django.conf import settings
settings.configure(
    INSTALLED_APPS=(
        'rest_framework',
    )
)
import django
django.setup()

from rest_framework import serializers as ser


class MySer(ser.Serializer):

    foo = ser.CharField()
    bar = ser.BooleanField()


ser = MySer(data=DATA)
ser.is_valid()
print('--- DRF ---')
print(ser.errors)
```

[...]

PR #667, opened 2017-08-10
Add strict option to fields.Integer

yoichi:
I want a feature to get validation error on float and Decimal value for fields.Integer instead of rounding it to integer.
Compare with https://github.com/marshmallow-code/marshmallow/issues/231, we already have tests expecting current rounding behaviour, so I've added a constructor option 'strict' and keep the default behavior.

# before the change

```

DATA = {'foo': '3.0', 'bar': 3.14, 'baz': 3.0}

class MySchema(Schema):
    foo = fields.Integer(required=True)
    bar = fields.Integer(required=True)
    baz = fields.Integer(required=True)

s = MySchema()
print('--- Marshmallow ---')
print(s.validate(DATA))


from django.conf import settings
settings.configure(INSTALLED_APPS=['rest_framework'])
import django
django.setup()

from rest_framework import serializers as ser

class MySer(ser.Serializer):
    foo = ser.IntegerField()
    bar = ser.IntegerField()
    baz = ser.IntegerField()

ser = MySer(data=DATA)
ser.is_valid()
print('--- DRF ---')
print(ser.errors)
```

```
--- Marshmallow ---
{'foo': ['Not a valid integer.']}
--- DRF ---
{'bar': ['A valid integer is required.']}
```

# after the change

```

DATA = {'foo': '3.0', 'bar': 3.14, 'baz': 3.0}

class MySchema(Schema):
    foo = fields.Integer(required=True, strict=True)
    bar = fields.Integer(required=True, strict=True)
    baz = fields.Integer(required=True, strict=True)

s = MySchema()
print('--- Marshmallow ---')
print(s.validate(DATA))
```

```
--- Marshmallow ---
{'baz': ['Not a valid integer.'], 'bar': ['Not a valid integer.'], 'foo': ['Not a valid integer.']}
```

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -86,3 +86,4 @@ Contributors (chronological)
 - Jérôme Lafréchoux  `@lafrech <https://github.com/lafrech>`_
 - Roy Williams `@rowillia <https://github.com/rowillia>`_
 - `@dradetsky <https://github.com/dradetsky>`_
+- Yoichi NAKAYAMA `@yoichi <https://github.com/yoichi>`_
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import, unicode_literals
 
 import collections
 import datetime as dt
+import numbers
 import uuid
 import warnings
 import decimal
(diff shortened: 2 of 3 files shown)


sloria:
Thanks! Seems like a good feature. @yoichi Can you please add yourself to AUTHORS.rst?

PR #668, opened 2017-08-10
Custom get item serialization bug fix

MichalKononenko:
I noticed that getting the object prototype for ``Nested`` schemas naively assumes that calling ``__getitem__`` with ``0`` will succeed. To that end, I've written a test to replicate the bug, and a small solution for it.


MichalKononenko:
Root was wrong, my bad!

PR #674, opened 2017-08-24
fix #664: validate.URL: add option to allow hostnames with no dots

### LINKED ISSUE #664: validate.URL rejects hostnames with no dots

sduthil:
reason: a hostname with no dots is valid, and useful in some cases,
e.g. a host in the same subdomain than the client, a shorter alias to
a longer hostname, a local network without TLD

This is a followup on #665.

### DIFF
--- a/marshmallow/validate.py
+++ b/marshmallow/validate.py
@@ -42,6 +42,7 @@ class URL(Validator):
         Can be interpolated with `{input}`.
     :param set schemes: Valid schemes. By default, ``http``, ``https``,
         ``ftp``, and ``ftps`` are allowed.
+    :param bool require_tld: Whether to reject non-FQDN hostnames
     """
 
     class RegexMemoizer(object):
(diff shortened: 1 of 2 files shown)


sloria:
Thanks!

PR #678, opened 2017-09-19
Add support for pre-deserialization validators (inspectors).

monotaro-yasushi-masuda:
This PR will add support of "raw input inspectors".
Inspectors are validators called before Field._deserialize() so that "raw input" values are checked.

Input validation before _deserialize() is necessary to have chance to reject malformed input before _deserialize() unexpectedly corerce it to "safe" value for validators.
 
For example, if you want to reject "incorrect width of digits" or prevent 1.23 (float, not string "1.23") for Integer field, inspector would be right place to implement it.

```
# this fails as expected
>>> Integer().deserialize('12.3')
  ...

# this should fail in particular case (but no help with current implementation)
>>> Integer().deserialize(12.3)
12

# Better if this can reject 12 (no padding) or 000012 (too many digits)
>>> Integer().deserialize("00012")
12
```

This patch should not change current validator behavior. If you want add raw input validation to your existing code, you may write inspector just as like as validator and specify it in inspector parameter (it allows both single inspector and list of inspectors).

```
>>> def only_digit(value):
...     if not str(value).isdigit():
...             raise ValidationError('Input contains non-digits')
...
>>> from marshmallow.validate import Length  # existing validators can be used as inspector
>>> Integer(inspect=[Length(5, 5), only_digit]).deserialize('0.123')  # FAIL: contains non-digit
  File "marshmallow/fields.py", line 283, in deserialize
    self._run_validators(value, self.inspectors, 'inspector_failed')
  File "marshmallow/fields.py", line 224, in _run_validators
    raise ValidationError(errors, **kwargs)
marshmallow.exceptions.ValidationError: ['Input contains non-digits']
>>> Integer(inspect=[Length(5, 5), only_digit]).deserialize('00123')  # SUCCESS: five digits
123
>>> Integer(inspect=[Length(5, 5), only_digit]).deserialize('123')  # FAIL: too few digits
  File "marshmallow/fields.py", line 283, in deserialize
    self._run_validators(value, self.inspectors, 'inspector_failed')
  File "marshmallow/fields.py", line 224, in _run_validators
    raise ValidationError(errors, **kwargs)
marshmallow.exceptions.ValidationError: [u'Length must be between 5 and 5.']
>>> Integer(inspect=[Length(5, 5), only_digit]).deserialize('000123')  # FAIL: to many digits
  File "marshmallow/fields.py", line 283, in deserialize
    self._run_validators(value, self.inspectors, 'inspector_failed')
  File "marshmallow/fields.py", line 224, in _run_validators
    raise ValidationError(errors, **kwargs)
marshmallow.exceptions.ValidationError: [u'Length must be between 5 and 5.']
>>>
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ python:
 
 before_install:
   - pip install -U pip
+  - pip install -U six
 
 install:
   - pip install -U .[reco]
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -70,8 +70,12 @@ class Field(FieldABC):
         be checked if the field's name is not found on the input dictionary. If checked,
         it will return this parameter on error.
     :param str dump_to: Field name to use as a key when serializing.
+    :param callable inspect: Validator or collection of validators that are called
+        before deserialization. Validator takes a field's raw input value as
+        its only parameter and returns a boolean.
+        If it returns `False`, an :exc:`ValidationError` is raised.
     :param callable validate: Validator or collection of validators that are called
-        during deserialization. Validator takes a field's input value as
+        immediately after deserialization. Validator takes a field's input value as
         its only parameter and returns a boolean.
         If it returns `False`, an :exc:`ValidationError` is raised.
     :param required: Raise a :exc:`ValidationError` if the field value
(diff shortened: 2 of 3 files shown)


sloria:
This adds complexity to the API that I'm not comfortable adding at this point. Can you use `@pre_load` to meet your use case?

monotaro-yasushi-masuda:
Sorry for late respond..

@pre_load can be used to reject values before de-serialization, but it requires repeating-myself in defining Schema :(
I hope pre-deserialization validators are loose-coupled with Field() so that you can reuse existing Field classes and popular data inspection tasks (checking input data length, encoding charset or type etc.)
My two cents it would be better to leave @pre_load for Schema-specific pre-loading task and to have some pre-deserialize validation hook (with whatever appropreate name; I'm not confident 'inspect' is good naming ;)

sloria:
I'm sorry, but I'm not comfortable adding this right now--I'm really trying to constrain the fields API. For now, I recommend sticking with `pre_load` methods, and perhaps you can open an RFC issue with your use case(s) so that others can discuss it.

PR #679, opened 2017-09-24
Use Changelog date instead of build date

bmwiedemann:
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good.

Without this patch, when building packages (e.g. for openSUSE Linux) in different years, results differed slightly:
```diff
     <div class="footer">
     -      &copy; 2017 <a href="http://stevenloria.com">Steven Loria</a>. 
     +      &copy; 2018 <a href="http://stevenloria.com">Steven Loria</a>.
```


monotaro-yasushi-masuda:
Filed https://github.com/marshmallow-code/marshmallow/issues/681 for build failure issue.

sloria:
Thanks!

PR #689, opened 2017-10-19
Add python 3.6 to tox

Fedalto:
I noticed that python 3.6 is supported, travis is running tests against it, but was missing in `tox.ini`.


sloria:
Thanks!

PR #692, opened 2017-10-21
add blank for convention

jun0jang:
this is based flake8.
Annotations also require spaces.


sloria:
We explicitly ignore that rule in our flake8 config. Thanks anyway, though!

PR #694, opened 2017-10-27
Changes from_iso to from_iso_datetime

sklarsa:
Since the iso regex used in `utils.from_iso` does not handle iso-formatted dates (`%Y-%m-%d`), I renamed this method to make its functionality clearer.


sloria:
I'm fine with the name change, but I think we should deprecate--rather than remove-- the old function.

```python
def from_iso(...):
    warnings.warn('from_iso is deprecated. Use from_iso_datetime instead.', UserWarning)
    return from_iso_datetime(...)
```

sloria:
Thanks!

Issue #634, opened 2017-05-24
Broken passing of context inside fields.List(fields.Nested(SomeSchema))


YuppY:
If schema is declared like this:
```python
class ChildSchema(Schema):
    foo = fields.Method('dump_foo')
    def dump_foo(self, obj):
        print 'context is', self.context
  
class ParentSchema(Schema):
    child = fields.List(fields.Nested(ChildSchema))
```
Then `context` is saved on first call and ignored on further calls:
```python
>>> ParentSchema(context={'bar': 1}).dump({'child': [{}]})
context is {'bar': 1}
MarshalResult(...)
>>> ParentSchema(context={'baz': 2}).dump({'child': [{}]})
context is {'bar': 1}
MarshalResult(...)
>>> ParentSchema().dump({'child': [{}]})
context is {'bar': 1}
MarshalResult(...)
```

It is because `copy.deepcopy(self._declared_fields)` doesn't copy `List.container` field and schema with `context` is cached inside `Nested` field and saved inside `_declared_fields`.

justanr:
Any reason to do it like that instead of declaring the nested field to be many? This sounds like a bug to me but on the other hand, is it a use case that Marshmallow should support?

jssuzanne:
Hi, I have the same issue, In my mind the solution is to allow the breaking of the cache in Nested field. because the context is propagated only when the making of the cache

jssuzanne:
Travis sound good for the PR #696

alexeysofin:
@jssuzanne #696 completely breaks nested schemas, because when self.withcache == False,self.__schema will not be initialized at all. Am I getting something wrong? :)

jssuzanne:
You may close this PR, I created my own nested field without cache for my need.

PR #696, opened 2017-11-02
#634, allow to deactivate the cache of the nested field

jssuzanne:


sloria:
Can you explain the use case for this change?

jssuzanne:
I need to propagate the context at each call (dump, load, validate), if the context change the cached schema is wrong, In my case I does not want this cache

erwinyusrizal:
I agree with @jssuzanne at least it really helps in development

s0undt3ch:
I seem to hitting this too and I also mess with context...

s0undt3ch:
Cache based on context repr instead? I'm fine with just disabling cache.

deckar01:
I think the correct solution for this use case is to construct a new schema instance each time you need to use a new context. The context should probably only be set via the schema constructor argument. If it's not safe to mutate the context, it should probably be a private property. Fixing the example in the documentation might make this more intuitive:

http://marshmallow.readthedocs.io/en/latest/custom_fields.html#adding-context-to-method-and-function-fields

```py
freddy = User('Freddie Mercury', 'fred@queen.com')
bicycle_blog = Blog('Bicycle Blog', author=freddy)
schema1 = UserSchema(context={'blog': bicycle_blog})
result1 = schema.dump(freddy)

tony = User('Tony Hawk', 'tony@proskater.com')
skateboard_blog = Blog('Skateboard Blog', author=tony)
schema2 = UserSchema(context={'blog': skateboard_blog})
result2 = schema.dump(tony)
```

deckar01:
I just noticed this was for #634. That is definitely a bug. We should probably construct a schema instance for nested fields to avoid mutating the class properties. We should not add a feature flag to hack around a broken feature though.

marshmallow/fields.py
```diff
@@ -395,7 +400,7 @@ def schema(self):
         .. versionchanged:: 1.0.0
             Renamed from `serializer` to `schema`
         """
-        if not self.__schema:
+        if not (self.__schema and self.withcache):
```
alexeysofin, line 403: This pull-request completely breaks nested schemas, because when self.withcache == False,self.__schema will not be initialized at all. Am I getting something wrong? :)

teleport1995: Why?
 If self.withcache == False -> (self.__schema and False) == False -> if not False == if True

deckar01: The logic is fine.

|   | withcache | not withcache |
|---|---|---|
| **no __schema** | True 👈 | True 👈 |
| **__schema** | False | True |

Negating grouped expressions can be hard to read though. It is better to expand the expression so that when read aloud it makes sense.

"Make a schema if there is no schema or the cache is disabled" -> `if not self.__schema or not self.withcache`

---

That said, I'm still not a fan of making this a configurable option.

sloria:
I'm with @deckar01 on avoiding a feature flag here. I don't this case justifies increasing the API surface. We'll have to give more thought on how we can properly propagate context to nested fields. In the meantime, you can work around the issue using a custom field. Now that `_schema` is [no longer mangled](https://github.com/marshmallow-code/marshmallow/pull/1319), you can do something like:

```python


class NestedUncached(fields.Nested):
    def _serialize(self, *args, **kwargs):
        self._schema = None
        super()._serialize(*args, **kwargs)

    def _deserialize(self, *args, **kwargs):
        self._schema = None
        super()._deserialize(*args, **kwargs)
```

Issue #650, opened 2017-06-23
String fields should return an error when given non-UTF-8


dan-blanchard:
Currently, if a user specifies a `bytes` value for a `String` field that is not UTF-8, marshmallow will raise a `UnicodeDecodeError`. A more sensible thing to expect from a validation library like marshmallow would be for it to return an error message about this not being a valid UTF-8 `String`.

cwisecarver:
@sloria Any thoughts on this? I can make a PR if it will help.

sloria:
@dan-blanchard @cwisecarver Sorry for the delayed response; this one fell through the cracks. Yes, I think the proposed behavior is correct. I would certainly review and merge a PR for this.

sloria:
This is fixed in 0.15.0.

Chris2048:
Hi. This fix wraps the ensure_text_type call in the _deserialize method with a try loop; might this also be needed in the _serialize method?

PR #699, opened 2017-11-29
Fix bug: field that is not UTF-8 will raise a UnicodeDecodeError

4lissonsilveira:
- Related: #650


sloria:
Thanks for the PR! Would you mind adding a test for your change?

4lissonsilveira:
Sure, I will.
I've thought about something like this:
```
    def test_field_deserialization_with_non_utf8_value(self):
        non_utf8_char = '\xc8'
        field = fields.String()
        # This exception only happens in Python version <= 2.7
        if isinstance(non_utf8_char, bytes):
                field.deserialize(non_utf8_char)
            assert excinfo.value.args[0] == 'Not a valid utf-8 string.'
```
Do you have a better approach for tests like this?

sloria:
@4lissonsilveira I think that is a fine approach.

PR #700, opened 2017-12-08
Structured Dictionaries

### LINKED ISSUE #483: How to create a Schema containing a dict of nested Schema?
Hi. I've been digging around and couldn't find the answer to this.

Say I've got a model like this:

``` python
class AlbumSchema(Schema):
    year = fields.Int()

class ArtistSchema(Schema):
    name = fields.Str()
    albums = ...
```

I want `albums` to be a `dict` of `AlbumSchema`, so that `ArtistSchema` serializes as

``` python
{ 'albums': { 'Hunky Dory': {'year': 1971},
              'The Man Who Sold the World': {'year': 1970}},
  'name': 'David Bowie'}
```

Naively, I would expect syntaxes like this to work:

``` python
fields.List(Schema)
fields.Dict(Schema)
```

or maybe

``` python
fields.List(fields.Nested(Schema))
fields.Dict(fields.Nested(Schema))
```

Serializing a `list` of `Schema` can be achieved through `Nested(Schema, many=True)`, which I find less intuitive, and I don't know about a `dict` of `Schema`.

Is there any way to do it? Or a good reason _not_ to do it?

(Question also [asked on SO](http://stackoverflow.com/questions/38048775/marshmallow-dict-of-nested-schema).)

deckar01:
Fixes #483

Add support for structured dictionaries by providing `values` and `keys` arguments to the `Dict` field's constructor. This mirrors the `List` field's ability to validate it's items.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -1125,12 +1125,19 @@ class TimeDelta(Field):
 
 
 class Dict(Field):
-    """A dict field. Supports dicts and dict-like objects.
+    """A dict field. Supports dicts and dict-like objects. Optionally composed
+    with another `Field` class or instance.
+
+    Example: ::
+
+        numbers = fields.Dict(values=fields.Float())
+
+    :param Field values: A field class or instance for dict values.
+    :param kwargs: The same keyword arguments that :class:`Field` receives.
 
     .. note::
-        This field is only appropriate when the structure of
-        nested data is not known. For structured data, use
-        `Nested`.
+        When the structure of nested data is not known, you may omit the
+        schema argument to prevent content validation.
 
     .. versionadded:: 2.1.0
     """
@@ -1139,11 +1146,54 @@ class Dict(Field):
         'invalid': 'Not a valid mapping type.'
     }
 
-    def _deserialize(self, value, attr, data):
-        if isinstance(value, collections.Mapping):
-            return value
+    def __init__(self, values=None, **kwargs):
+        super(Dict, self).__init__(**kwargs)
+        if values is None:
+            self.value_container = None
+        elif isinstance(values, type):
+            if not issubclass(values, FieldABC):
+                raise ValueError('The type of the dict elements '
+                                 'must be a subclass of '
+                                 'marshmallow.base.FieldABC')
+            self.value_container = values()
         else:
+            if not isinstance(values, FieldABC):
+                raise ValueError('The instances of the dict '
+                                 'elements must be of type '
+                                 'marshmallow.base.FieldABC')
+            self.value_container = values
+
+    def _serialize(self, value, attr, obj):
+        if value is None:
+            return None
+        if not self.value_container and not self.key_container:
+            return value
+        if isinstance(value, collections.Mapping):
+            return {
+                key: self.container._serialize(item, attr, obj)
+                for key, item in value.items()
+            }
+        self.fail('invalid')
+
+    def _deserialize(self, value, attr, data):
+        if not isinstance(value, collections.Mapping):
             self.fail('invalid')
+        if not self.value_container and not self.key_container:
+            return value
+
+        result = {}
+        errors = {}
+        for key, item in value.items():
+            try:
+                result.update({key: self.container.deserialize(item)})
+            except ValidationError as e:
+                result.update({key: e.messages})
+                errors.update({key: e.messages})
+
+        if errors:
+            raise ValidationError(errors, data=result)
+
+        return result
 
 
 class ValidatedField(Field):
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -1130,14 +1130,15 @@ class Dict(Field):
 
     Example: ::
 
-        numbers = fields.Dict(values=fields.Float())
+        numbers = fields.Dict(values=fields.Float(), keys=fields.Str())
 
     :param Field values: A field class or instance for dict values.
+    :param Field keys: A field class or instance for dict keys.
     :param kwargs: The same keyword arguments that :class:`Field` receives.
 
     .. note::
         When the structure of nested data is not known, you may omit the
-        schema argument to prevent content validation.
+        `values` and `keys` arguments to prevent content validation.
 
     .. versionadded:: 2.1.0
     """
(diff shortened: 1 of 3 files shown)


marshmallow/fields.py
```diff
@@ -1139,11 +1147,89 @@ class Dict(Field):
...
-    def _deserialize(self, value, attr, data):
-        if isinstance(value, collections.Mapping):
-            return value
+    def __init__(self, values=None, keys=None, **kwargs):
+        super(Dict, self).__init__(**kwargs)
+        if values is None:
+            self.value_container = None
+        elif isinstance(values, type):
+            if not issubclass(values, FieldABC):
+                raise ValueError('The type of the dict elements '
```
sloria: It might be more direct and more correct to say `"'values' must be a subclass of ...."`

tests/test_deserialization.py
```diff
@@ -542,6 +542,36 @@ def test_dict_field_deserialization(self):
...
+            field.deserialize({1: "bar"})
+        assert excinfo.value.args[0] == {1: ['Invalid key: Not a valid string.']}
+
+    def test_structured_dict_key_value_deserialization(self):
+        field = fields.Dict(keys=fields.Str, values=fields.Decimal)
+        assert field.deserialize({"foo": 1}) == {"foo": decimal.Decimal('1')}
+        with pytest.raises(ValidationError) as excinfo:
+            field.deserialize({1: "bar"})
+        assert excinfo.value.args[0] == {1: [
+            'Invalid key: Not a valid string.',
```
sloria: What does the error data look like if a key has more than one error? It would be worth adding a test that uses a field with a validator. Something like

```python
fields.Dict(keys=fields.Str(validate=[validate.Length(min=2)]))
```

marshmallow/fields.py
```diff
@@ -1139,11 +1147,89 @@ class Dict(Field):
...
+        result = {}
+        errors = {}
+        items = list(value.values())
+        keys = list(value.keys())
+        if self.key_container:
+            for idx, key in enumerate(keys):
+                try:
+                    keys[idx] = self.key_container.deserialize(key)
+                except ValidationError as e:
+                    key_messages = ['Invalid key: ' + message for message in e.messages]
```
sloria: Minor: use `str.format` rather than concatenation.

marshmallow/fields.py
```diff
@@ -1125,12 +1125,20 @@ def _deserialize(self, value, attr, data):
...
+    :param Field values: A field class or instance for dict values.
+    :param Field keys: A field class or instance for dict keys.
+    :param kwargs: The same keyword arguments that :class:`Field` receives.
 
     .. note::
-        This field is only appropriate when the structure of
-        nested data is not known. For structured data, use
-        `Nested`.
+        When the structure of nested data is not known, you may omit the
+        schema argument to prevent content validation.
```
sloria: It would be more accurate to say: "you may omit the ``values`` and ``keys`` arguments to prevent content valiation."

sloria (changes requested):
Looking really good on first pass! Just a few nits for now.

deckar01:
@sloria Thanks for reviewing. Those issues should be addressed.

marshmallow/fields.py
```diff
@@ -1139,11 +1147,91 @@ class Dict(Field):
...
+        errors = {}
+        items = list(value.values())
+        keys = list(value.keys())
+        if self.key_container:
+            for idx, key in enumerate(keys):
+                try:
+                    keys[idx] = self.key_container.deserialize(key)
+                except ValidationError as e:
+                    key_messages = [
+                        'Invalid key: {}'.format(message)
```
sloria: Just a thought: it *might* be worth exposing a way to customize these error messages, e.g. through a method hook, but...YAGNI might also apply here.

sloria (approved):
I'd still like to review this a little more closely, but I don't see any obvious blockers at the moment. 👍

marshmallow/fields.py
```diff
@@ -1139,11 +1147,91 @@ class Dict(Field):
...
+        if values is None:
+            self.value_container = None
+        elif isinstance(values, type):
+            if not issubclass(values, FieldABC):
+                raise ValueError('"values" must be a subclass of '
+                                 'marshmallow.base.FieldABC')
+            self.value_container = values()
+        else:
+            if not isinstance(values, FieldABC):
+                raise ValueError('"values" must be of type '
```
lxop, line 1161: "must be of type" -> "must be a subclass of"
Same thing with the keys check below

lxop: Actually, scratch that. It would be good if the message matched that for the List class though - refer to "The type of the values" and "The value instances" respectively

marshmallow/fields.py
```diff
@@ -1139,11 +1147,91 @@ class Dict(Field):
...
+                                 'marshmallow.base.FieldABC')
+            self.key_container = keys
+
+    def _serialize(self, value, attr, obj):
+        if value is None:
+            return None
+        if not self.value_container and not self.key_container:
+            return value
+        if isinstance(value, collections.Mapping):
+            items = value.values()
```
lxop: 'items' has another meaning in the context of a dictionary, perhaps 'values' would be a better variable name here

marshmallow/fields.py
```diff
@@ -1139,11 +1147,91 @@ class Dict(Field):
...
+        if self.key_container:
+            for idx, key in enumerate(keys):
+                try:
+                    keys[idx] = self.key_container.deserialize(key)
+                except ValidationError as e:
+                    key_messages = [
+                        'Invalid key: {}'.format(message)
+                        for message in e.messages
+                    ]
+                    result.update({key: list(key_messages)})
```
lxop: Why not just `result[key] = list(key_messages)`?

lxop: Also, shouldn't the `result` dict take `e.data`?

lxop: In fact, nothing should be written to `result` - that is overwritten at the end anyway by `result = dict(zip(keys, items))`. Shouldn't this be `keys[idx] = e.data`? Otherwise that zip will fail (or worse - with one bad key and one bad value independently, the zip will match up incorrect keys to values)

galuszkak:
This looks awesome, I was looking for something like that when for example Your keys are uuid's and they have nested objects!

Awesome work @deckar01 ! @sloria will be this merged?

marshmallow/fields.py
```diff
@@ -1125,12 +1125,20 @@ def _deserialize(self, value, attr, data):
...
+    :param Field values: A field class or instance for dict values.
+    :param Field keys: A field class or instance for dict keys.
+    :param kwargs: The same keyword arguments that :class:`Field` receives.
 
     .. note::
-        This field is only appropriate when the structure of
-        nested data is not known. For structured data, use
-        `Nested`.
+        When the structure of nested data is not known, you may omit the
+        `values` and `keys` arguments to prevent content validation.
```
aldanor, line 1141: Maybe (for clarity), description of `values` parameter could mention something like "or None, if values are not meant to be validated"; same for keys?

marshmallow/fields.py
```diff
@@ -1125,12 +1125,20 @@ def _deserialize(self, value, attr, data):
...
+
+    :param Field values: A field class or instance for dict values.
+    :param Field keys: A field class or instance for dict keys.
+    :param kwargs: The same keyword arguments that :class:`Field` receives.
 
     .. note::
-        This field is only appropriate when the structure of
-        nested data is not known. For structured data, use
-        `Nested`.
+        When the structure of nested data is not known, you may omit the
```
aldanor, line 1140: "you may omit the `values` and `keys` arguments" -- you can actually omit just either one of them as well?

aldanor:
I've left a few very minor comments re: the docstring, but this looks quite awesome as is; plans to merge this soon?

sloria:
Yes, I think this is good to go!

PR #707, opened 2017-12-26
Fix URL validator for ipv6 (brackets required always)

harlov:
By RFC "Format for Literal IPv6 Addresses in URL's" (https://tools.ietf.org/html/rfc2732) , square brackets always requried in ipv6 URI.


harlov:
PR resolve #708

sloria:
Thanks!

PR #709, opened 2017-12-28
Optionally pass field instance to validators

### LINKED ISSUE #289: Pass schema's context to validation functions
I am thinking about a way to pass context to validation function.

Currently validation function expect exactly one argument, value of a field. Maybe `Field._validate` method can check with inspect.getargspec and if function expect two arguments, send schema's context as second argument?

I know that context is available in validation methods declared with `@validates` decorator, but I want to reuse validation function in different schemas with different fields.

s0undt3ch:
Fixes https://github.com/marshmallow-code/marshmallow/issues/289

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -193,7 +193,11 @@ class Field(FieldABC):
         kwargs = {}
         for validator in self.validators:
             try:
-                r = validator(value)
+                try:
+                    r = validator(value, field=self)
+                except TypeError:
+                    # This validator does not accept the field instance
+                    r = validator(value)
                 if not isinstance(validator, Validator) and r is False:
                     self.fail('validator_failed')
             except ValidationError as err:
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -3,6 +3,7 @@ import pytest
 
 from marshmallow import fields, Schema, ValidationError
 from marshmallow.marshalling import missing
+from marshmallow.validate import Validator
 
 from tests.base import ALL_FIELDS, User
 
@@ -185,3 +186,41 @@ class TestErrorMessages:
 
         assert 'doesntexist' in excinfo.value.args[0]
         assert 'MyField' in excinfo.value.args[0]
+
+
+class TestFieldValidationPassingFieldToValidators:
+
+    def test_validator_receives_field_instance(self, user):
+
+        class ValidatorWithoutFieldPassingSupport(Validator):
+            error = 'No Error'
+
+            def __call__(self, value):
+                if value not in ('male', 'female'):
+                    raise ValidationError("Gender '{0}' is not valid".format(value))
+
+        class ValidatorWithFieldPassingSupport(Validator):
+            def __init__(self):
+                self.error = 'Validation failed'
+
+            def __call__(self, value, field=None):
+                if field is None:
+                    raise ValidationError('No field instance was passed')
+                if value not in field.context['allowed_names']:
+                    raise ValidationError(
+                        "Name '{0}' was not found in: {1}".format(
+                            value, ', '.join(field.context['allowed_names']))
+                    )
+
+        class MySchema(Schema):
+            name = fields.String(validate=ValidatorWithFieldPassingSupport())
+            gender = fields.String(validate=ValidatorWithoutFieldPassingSupport())
+
+        schema = MySchema()
+        schema.context['allowed_names'] = ('Python',)
+        result, errors = schema.load({'name': 'Monty', 'gender': 'unknown'})
+        assert result == {}
+        assert 'name' in errors
+        assert 'gender' in errors
+        assert errors['name'] == ["Name 'Monty' was not found in: Python"]
+        assert errors['gender'] == ["Gender 'unknown' is not valid"]


marshmallow/fields.py
```diff
@@ -193,7 +193,11 @@ def _validate(self, value):
         kwargs = {}
         for validator in self.validators:
             try:
-                r = validator(value)
+                try:
+                    r = validator(value, field=self)
```
sloria, line 197: `self` should probably passed as a positional argument rather than a keyword argument. That way, users wouldn't be required to name the second argument "field".

sloria:
I'm a bit worried about the performance overhead of automagically allowing for a second parameter (though I haven't done any benchmarking).

As suggested in https://github.com/marshmallow-code/marshmallow/issues/289#issuecomment-145008637 , we could have a decorator that specifies that the field should be passed. Thoughts?

marshmallow/fields.py
s0undt3ch: Users don't have to name the argument, but if they do, it should be field because that's what's being passed in. Unless you have a better name.

s0undt3ch:
I will try to get some benchmarks going. I'd prefer to just handle it by catching the exception(inspect is not even an option), than forcing the use of yet another decorator(this is a validator not a schema, ie, this would be the fisr4 use of decorators in validators?).
The idea behind catching the exception is to support old validators which don't accept the second argument, perhaps even deprecate this usage by throwing a deprecation warning and just move to always passing `field` as the second argument in 2 or 3 major versions?

marshmallow/fields.py
sloria: Constraining the name of the argument seems unnecessary. Different projects might have different naming conventions. For example, I've used `field_obj` to disambiguate field names from field instances.

sloria:
@s0undt3ch You make a fair point. I would definitely be willing to go the exception-handling approach if you can verify that it doesn't affect performance.

s0undt3ch:
Take this example script:
```python
def no_arg_func():
    pass


def arg_func(name):
    pass


def named_arg_func(name=None):
    pass


def accept_name(func):
    func.__accept_name__ = True
    return func


@accept_name
def decorated_func(name=None):
    pass


def try_except_no_arg_func(func):
    try:
        func()
    except TypeError:
        # Won't happen
        pass


def try_except_args_func(func):
    try:
        func(True)
    except TypeError:
        func()


def try_except_named_arg_func(func):
    try:
        func(name=True)
    except TypeError:
        func()


def try_decorated_func(func):
    try:
        func.__accept_name__
        func(name=True)
    except AttributeError:
        func()


def try_decorated_func_has_attr(func):
    if hasattr(func, '.__accept_name__'):
        func(name=True)
    else:
        func()


def main():
    import timeit
    # Cold start
    t1 = timeit.timeit('try_except_no_arg_func(no_arg_func)', setup='from __main__ import try_except_no_arg_func, no_arg_func')
    t2 = timeit.timeit('try_except_args_func(arg_func)', setup='from __main__ import try_except_args_func, arg_func')
    t3 = timeit.timeit('try_except_named_arg_func(named_arg_func)', setup='from __main__ import try_except_named_arg_func, named_arg_func')
    t4 = timeit.timeit('try_decorated_func(no_arg_func)', setup='from __main__ import try_decorated_func, no_arg_func')
    t5 = timeit.timeit('try_decorated_func(decorated_func)', setup='from __main__ import try_decorated_func, decorated_func')
    t6 = timeit.timeit('try_decorated_func_has_attr(no_arg_func)', setup='from __main__ import try_decorated_func_has_attr, no_arg_func')
    t7 = timeit.timeit('try_decorated_func_has_attr(decorated_func)', setup='from __main__ import try_decorated_func_has_attr, decorated_func')
    # Re-run
    t1 = timeit.timeit('try_except_no_arg_func(no_arg_func)', setup='from __main__ import try_except_no_arg_func, no_arg_func')
    t2 = timeit.timeit('try_except_args_func(arg_func)', setup='from __main__ import try_except_args_func, arg_func')
    t3 = timeit.timeit('try_except_named_arg_func(named_arg_func)', setup='from __main__ import try_except_named_arg_func, named_arg_func')
    t4 = timeit.timeit('try_decorated_func(no_arg_func)', setup='from __main__ import try_decorated_func, no_arg_func')
    t5 = timeit.timeit('try_decorated_func(decorated_func)', setup='from __main__ import try_decorated_func, decorated_func')
    t6 = timeit.timeit('try_decorated_func_has_attr(no_arg_func)', setup='from __main__ import try_decorated_func_has_attr, no_arg_func')
    t7 = timeit.timeit('try_decorated_func_has_attr(decorated_func)', setup='from __main__ import try_decorated_func_has_attr, decorated_func')

    print('                    Old behavior:', t1)
    print('           Try/Execpt using args:', t2)
    print('Try/Except using named arguments:', t3)
    print('              Non decorated func:', t4)
    print('                  Decorated func:', t5)
    print('     Non decorated func(hasattr):', t6)
    print('         Decorated func(hasattr):', t7)


if __name__ == "__main__":
    main()
```
Here's the results:
```
                    Old behavior: 0.13157559500541538
           Try/Execpt using args: 0.1553857090184465
Try/Except using named arguments: 0.14342665893491358
              Non decorated func: 0.5568735370179638
                  Decorated func: 0.195724172051996
     Non decorated func(hasattr): 0.4652692979434505
         Decorated func(hasattr): 0.4673313000239432
```

s0undt3ch:
Polite ping.

sloria:
Thanks for your patience. I'm currently on vacation out of the country. Will review this when I get back

s0undt3ch:
:+1:

sloria:
Your benchmark uses tests the faster code branches--not the slow ones:

```python
t2 = timeit.timeit('try_except_args_func(arg_func)', setup='from __main__ import try_except_args_func, arg_func')
```

Since you're passing `arg_func` (rather than `no_arg_func`), we never get to the `except`. 

I'm still concerned how this change will affect performance when validators do *not* pass the field as an argument.

s0undt3ch:
```python
def no_arg_func():
    pass


def arg_func(name):
    pass


def named_arg_func(name=None):
    pass


def accept_name(func):
    func.__accept_name__ = True
    return func


@accept_name
def decorated_func(name=None):
    pass


def try_except_no_arg_func(func):
    try:
        func()
    except TypeError:
        # Won't happen
        pass


def try_except_args_func(func):
    try:
        func(True)
    except TypeError:
        func()


def try_except_named_arg_func(func):
    try:
        func(name=True)
    except TypeError:
        func()


def try_decorated_func(func):
    try:
        func.__accept_name__
        func(name=True)
    except AttributeError:
        func()


def try_decorated_func_has_attr(func):
    if hasattr(func, '.__accept_name__'):
        func(name=True)
    else:
        func()


def main():
    import timeit
    # Cold start
    t1 = timeit.timeit('try_except_no_arg_func(no_arg_func)', setup='from __main__ import try_except_no_arg_func, no_arg_func')
    t2 = timeit.timeit('try_except_args_func(arg_func)', setup='from __main__ import try_except_args_func, arg_func')
    t21 = timeit.timeit('try_except_args_func(no_arg_func)', setup='from __main__ import try_except_args_func, no_arg_func')
    t3 = timeit.timeit('try_except_named_arg_func(named_arg_func)', setup='from __main__ import try_except_named_arg_func, named_arg_func')
    t31 = timeit.timeit('try_except_named_arg_func(no_arg_func)', setup='from __main__ import try_except_named_arg_func, no_arg_func')
    t4 = timeit.timeit('try_decorated_func(no_arg_func)', setup='from __main__ import try_decorated_func, no_arg_func')
    t5 = timeit.timeit('try_decorated_func(decorated_func)', setup='from __main__ import try_decorated_func, decorated_func')
    t6 = timeit.timeit('try_decorated_func_has_attr(no_arg_func)', setup='from __main__ import try_decorated_func_has_attr, no_arg_func')
    t7 = timeit.timeit('try_decorated_func_has_attr(decorated_func)', setup='from __main__ import try_decorated_func_has_attr, decorated_func')
    # Re-run
    t1 = timeit.timeit('try_except_no_arg_func(no_arg_func)', setup='from __main__ import try_except_no_arg_func, no_arg_func')
    t2 = timeit.timeit('try_except_args_func(arg_func)', setup='from __main__ import try_except_args_func, arg_func')
    t21 = timeit.timeit('try_except_args_func(no_arg_func)', setup='from __main__ import try_except_args_func, no_arg_func')
    t3 = timeit.timeit('try_except_named_arg_func(named_arg_func)', setup='from __main__ import try_except_named_arg_func, named_arg_func')
    t31 = timeit.timeit('try_except_named_arg_func(no_arg_func)', setup='from __main__ import try_except_named_arg_func, no_arg_func')
    t4 = timeit.timeit('try_decorated_func(no_arg_func)', setup='from __main__ import try_decorated_func, no_arg_func')
    t5 = timeit.timeit('try_decorated_func(decorated_func)', setup='from __main__ import try_decorated_func, decorated_func')
    t6 = timeit.timeit('try_decorated_func_has_attr(no_arg_func)', setup='from __main__ import try_decorated_func_has_attr, no_arg_func')
    t7 = timeit.timeit('try_decorated_func_has_attr(decorated_func)', setup='from __main__ import try_decorated_func_has_attr, decorated_func')

    print('                                             Old behavior:', t1)
    print('                                    Try/Execpt using args:', t2)
    print('           Try/Execpt using args(without supporting args):', t21)
    print('                         Try/Except using named arguments:', t3)
    print('Try/Except using named arguments(without supporting args):', t31)
    print('                                       Non decorated func:', t4)
    print('                                           Decorated func:', t5)
    print('                              Non decorated func(hasattr):', t6)
    print('                                  Decorated func(hasattr):', t7)


if __name__ == "__main__":
    main()
```
Here's the results:

```
                                             Old behavior: 0.15172904799692333
                                    Try/Execpt using args: 0.16716513900610153
           Try/Execpt using args(without supporting args): 1.1487227189936675
                         Try/Except using named arguments: 0.16583949900814332
Try/Except using named arguments(without supporting args): 0.6354002379957819
                                       Non decorated func: 0.6881285959971137
                                           Decorated func: 0.20219569199252874
                              Non decorated func(hasattr): 0.5365560189966345
                                  Decorated func(hasattr): 0.5888571029936429
```
Is this what you were after?

sloria:
Looks like there's a significant performance penalty if we hit the `except` block, which was my concern. 

```
Try/Execpt using args: 0.16716513900610153
Try/Execpt using args(without supporting args): 1.1487227189936675
```

s0undt3ch:
Yes, expected, and we should use named keyword arguments:
```
Try/Except using named arguments(without supporting args): 0.6354002379957819
```
Which has similar performance penalty to using a decorator:
```
 Non decorated func: 0.6881285959971137 Decorated func: 0.20219569199252874
```

And using a decorator, you never get to deprecate.

sloria:
Closing this, as it's gone stale, and I still don't think the try-except here is quite right. Besides the performance penalty if we hit the `except`, it will swallow `TypeError`s raised by the validator.

PR #710, opened 2018-01-02
Nested field: store data and valid_data in ValidationError on dump/load

lafrech:
I think this was forgotten when implementing afb12308.

Without this change, `valid_data` contains wrong data when loading a nested field with errors.


sloria:
Good catch! Thanks for the fix.

Issue #598, opened 2017-03-10
RFC: Return deserialized data when strict=True


sloria:
## Status quo

`load` currently returns a namedtuple (`UnmarshalResult`) containing both the deserialized data and error messages.

```python
from marshmallow import Schema, fields, ValidationError

class UserSchema(Schema):
    id = fields.Int(required=True)

    class Meta:
        strict = True


schema = UserSchema()
try:
    result = schema.load({'id': 42})
except ValidationError as e:
    print(e.messages)
assert result.data['id'] == 42
```

## Proposed API

When `strict=True`, return only the deserialized data. Messages can already be accessed from `ValidationError.messages`.

```python
from marshmallow import Schema, fields, ValidationError

class UserSchema(Schema):
    id = fields.Int(required=True)

    class Meta:
        strict = True


schema = UserSchema()
try:
    data = schema.load({'id': 42})
except ValidationError as e:
    print(e.messages)
assert data['id'] == 42
```

douglas-treadwell:
I am in favor of strict=True by default, but I think the API for load() returning different KINDS of data depending on configuration might be confusing.

douglas-treadwell:
I think it's better if methods always return the same "kind" or "shape" of data.  However, here's an alternative...

What about a second method such as .loaddata() which only retrieves the data and ignores the errors?  This way there's a shorthand for people who don't care about the errors.  Obviously also the corresponding .loadsdata().  I suppose this isn't much shorter though than .load().data... maybe your current API here is plenty good!

maximkulkin:
Can people who don't care about errors just write `schema.load(payload).data` instead?

maximkulkin:
@douglas-treadwell I vote not to do that. Once Marshmallow goes this route, there are plenty of option combinations to introduce as a new methods: `loaddata()`, `loadmany()`, `loadstrictmany()`...

@sloria I would rather opt to deprecate `strict` option at all and just do validation error exceptions all the time. Then `load()` could return just data in an nonconfusing way

douglas-treadwell:
@maximkulkin I agree and withdraw my suggestion about loaddata().  It's perfectly fine to do load().data.

sloria:
I like @maximkulkin 's suggestion. So we would

1. Implement #377 and make strict mode the default
2. Modify `load` and `dump` return (de)serialized data rather than a (Un)MarshalResult (as in the proposed API above
[...]

Issue #377, opened 2016-01-14
Strict Schema Validation should be enabled by default


MichalKononenko:
To me, it goes against the zen of Python that validation fails silently when the `strict` parameter is not set to `True`. Why is it false by default? Shouldn't it be `True` by default?

sloria:
This was also suggested in https://github.com/keleshev/ask-me-about-api-design/issues/1 .

> Also, I would make schemas always strict=True. I don't think that returning a tuple result, error is Pythonic. And there's no disadvantage of raising exception, as long as you provide a structured error together with exception message.

My response on that thread, which I still stand by:

> Perhaps. My concern with raising an exception by default is that the user would then have to learn the interface of the exception (error.json, in your example) in order to get at the error messages. With a tuple, they just receive two dicts.

Also, this is obviously a breaking change. That said, I could be convinced that this is a good idea to include in 3.0. Opening this up for discussion.

immerrr:
FWIW, I think it's a good idea. Having a common base schema class providing `class Meta: strict = True` is probably the first thing I do when introducing marshmallow.

douglas-treadwell:
I'm at a company that just started replacing reqparse from Flask-Restful with Marshmallow.  I agree that strict=True seems like a better default.  Most things in Python that don't go as expected raise errors, rather than returning a tuple with error messages.

douglas-treadwell:
Those of you who agree about this may find the code snippet in this issue useful:

https://github.com/marshmallow-code/marshmallow/issues/550

As mentioned in that issue, one drawback of using `class Meta: strict = True` is it cannot be overridden later with `SomeSchema(strict=False)`.

maximkulkin:
In our project we just use our own base class for all schemas:

``` python
class Schema(marshmallow.Schema):
    def __init__(self, strict=True, **kwargs):
        super(Schema, self).__init__(strict=strict, **kwargs)
```

douglas-treadwell:
Right, the same thing I posted in issue #550, although I named the extended schema a bit differently.

maximkulkin:
@douglas-treadwell Yes, sorry, I missed that. My take on that is: don't look into Meta class magic and just use old proved OOP approaches.

douglas-treadwell:
Agreed Max.  Also, I'm happy to see we took the same approach, and thanks for sharing it.  I always like to double check that a workaround seems reasonable to others.

sloria:
[...]

PR #711, opened 2018-01-02
Remove strict parameter and return [de|]serialiazed data or raise ValidationError

lafrech:
I started to work on https://github.com/marshmallow-code/marshmallow/issues/377 (_Strict Schema Validation should be enabled by default_).

This deprecates `strict`. A `ValidationError` is raised every time an error is encountered.

Note that this does not implement https://github.com/marshmallow-code/marshmallow/issues/598 (_Return deserialized data when strict=True_), and dump/load still returns data and errors.

The error in the tests is due to https://github.com/marshmallow-code/marshmallow/pull/710. I'll rebase this branch when this issue is fixed in `dev` branch.

Feedback welcome.

### DIFF
--- a/docs/upgrading.rst
+++ b/docs/upgrading.rst
@@ -17,6 +17,35 @@ The marshmallow 3.x series supports Python 2.7, 3.4, 3.5, and 3.6.
 
 Python 2.6 and 3.3 are no longer supported.
 
+
+``strict`` in Schema is deprecated
+**********************************
+
+`Schema().load` and `Schema().dump` don't return a (data, errors) duple anymore.
+
+Only data is returned. If invalid data are passed in, a :exc:`ValidationError <marshmallow.exceptions.ValidationError>` is raised. The dictionary of validation errors is accessible from the `ValidationError.messages <marshmallow.exceptions.ValidationError.messages>` attribute, along with the valid data (values deserialized without error).
+
+.. code-block:: python
+
+    from marshmallow import Schema, ValidationError
+
+    # 2.x
+    user = User(name="Monty", email="monty@python.org")
+    schema = UserSchema()
+    data, errors = schema.dump(user)
+
+    # 3.x
+    user = User(name="Monty", email="monty@python.org")
+    schema = UserSchema()
+    try:
+        data = schema.dump(user)
+    except ValidationError as err:
+        errors = err.messages
+        valid_data = err.valid_data
+
+:meth:`Schema.validate` now always returns a dictionary of validation errors (same as 2.x with ``strict=False``).
+
+
 Overriding ``get_attribute``
 ****************************
 
@@ -80,7 +109,7 @@ The `Method <marshmallow.fields.Method>` and `Function <marshmallow.fields.Funct
     # In 2.x, the following would pass without errors
     # In 3.x, and AttributeError would be raised
     result = schema.dump(None)
-    result.data  # => {}
+    result  # => {}
 
 
     # 3.x
@@ -95,7 +124,7 @@ The `Method <marshmallow.fields.Method>` and `Function <marshmallow.fields.Funct
 
     schema = ShapeSchema()
     result = schema.dump(None)
-    result.data  # => {}
+    result  # => {}
 
 Adding additional data to serialized output
 *******************************************
@@ -113,7 +142,7 @@ Use a `post_dump <marshmallow.decorators.post_dump>` to add additional data on s
         y = fields.Int()
 
     schema = MySchema(extra={'z': 123})
-    schema.dump({'x': 1, 'y': 2}).data
+    schema.dump({'x': 1, 'y': 2})
     # => {'z': 123, 'y': 2, 'x': 1}
 
     # 3.x
@@ -127,7 +156,7 @@ Use a `post_dump <marshmallow.decorators.post_dump>` to add additional data on s
             return output
 
     schema = MySchema()
-    schema.dump({'x': 1, 'y': 2}).data
+    schema.dump({'x': 1, 'y': 2})
     # => {'z': 123, 'y': 2, 'x': 1}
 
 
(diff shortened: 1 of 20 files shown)


lafrech:
@sloria, any idea why the tests fail for P2.7 and Pypy?

It fails with invoke but I can't reproduce when calling pytest directly in a P2.7 virtualenv.

I copied the mechanism used for json module deprecation.

lafrech:
I'm tackling #598 in the same branch as those two issues are related.

Feedback welcome.

The docs should be updated.

Not sure about `Schema.validate`. Currently it raises a `ValidationError` if an error occurs and returns `None` otherwise. Maybe we could let it always return a `dict` of errors (empty or not). That could be more practical from the caller point of view. I don't know.

sloria:
Thanks so much for tackling this @lafrech ! This is looking great.

The failure in Python 2 is due to an existing bug in pytest: https://github.com/pytest-dev/pytest/issues/2917 . I was able to work around it by upgrading pytest to the latest version and using `pytest.deprecated_call()`:

```python
def test_strict_is_deprecated():
    with pytest.deprecated_call():
        class StrictUserSchema(Schema):
            name = fields.String()

            class Meta:
                strict = False

    class UserSchema(Schema):
        name = fields.String()

    with pytest.deprecated_call():
        UserSchema(strict=True)
```

I just enabled pyup.io so that we can automate dependency upgrades. Once I have the initial update passing on Travis, pytest will be updated. Feel free to just do the update in this branch though if it's blocking you.

sloria:
OK, pytest is now updated on `dev`. Rebasing with `dev` should fix the tests.

lafrech:
Branch rebased and workaround applied.

docs/upgrading.rst
```diff
@@ -17,6 +17,35 @@ The marshmallow 3.x series supports Python 2.7, 3.4, 3.5, and 3.6.
...
+    # 3.x
+    user = User(name="Monty", email="monty@python.org")
+    schema = UserSchema()
+    try:
+        data = schema.dump(user)
+    except ValidationError as err:
+        errors = err.messages
+        valid_data = err.valid_data
+
+:meth:`Schema.validate` now always returns a dictionary of validation errors (same as before with ``strict=True``).
```
sloria: Wouldn't it be the same as before with `strict=False`?

lafrech: Oops. You're right. Let me fix this now.

sloria:
This is fantastic. Special thanks for updating the docs!

This doesn't deprecate `strict` so much as remove its functionality altogether. Are we OK with this? I'm all for this cleaner API, but it's a significant, breaking change.

sloria:
By the way, I've given you write access to this repo. Thanks again for all your hard work.

sloria:
Also, the code in the `examples/` directory will also need to be updated for this change.

lafrech:
> Also, the code in the examples/ directory will also need to be updated for this change.

Done.

> This doesn't deprecate strict so much as remove its functionality altogether. Are we OK with this? I'm all for this cleaner API, but it's a significant, breaking change.

Yes, I totally realize that. You're right about the fact that it is not a "deprecation". Since this is a major release, let's remove `strict` completely. I think I left it at first because I thought it would avoid breaking apps that use `strict=True` in all their `Schema`s, but those apps are broken anyway by the removal of `[Un|]MarhalResult` so it didn't make much sense. I pushed a new commit removing the parameter and the deprecation warning.

> By the way, I've given you write access to this repo. Thanks again for all your hard work.

Thanks. I don't think I'll be pushing stuff in `dev` branch, but now I'll be able to set labels and milestones on issues.

sloria:
Fine by me!

lafrech:
Great! Thanks for reviewing and merging.

> 3.0.0b7 (unreleased)
> 
> Features:
> 
>     Backwards-incompatible: Schemas are now strict by default (:issue:`377`).

Actually, they are not only strict by default, they can only be strict, the option has been removed.

sloria:
Fair point; I've tweaked the wording a bit.

marshmallow/fields.py
```diff
@@ -445,15 +445,17 @@ def _serialize(self, nested_obj, attr, obj):
...
-        if isinstance(self.only, basestring):  # self.only is a field name
-            if self.many:
-                return utils.pluck(ret, key=self.only)
-            else:
-                return ret[self.only]
-        if errors:
-            raise ValidationError(errors, data=obj, valid_data=ret)
+        except ValidationError as exc:
+            raise ValidationError(exc.messages, data=obj, valid_data=exc.valid_data)
+        finally:
```
deckar01, line 453: This causes a `NameError` to be raised instead of the `ValueError` when using the string version of `only`, because `ret` will always be undefined if an exception is raised. This shouldn't have been wrapped in a `finally` block. 

```py
try:
    raise ValueError()
    a = 1
except ValueError as e:
    raise ValueError()
finally:
    print(a)
# Traceback (most recent call last):
#   line 7, in <module>
# NameError: name 'a' is not defined
```

The should get fixed as part of #800.

lafrech: Good point @deckar01. I wonder how I missed this.

Should I do something about it or just wait for #800?

deckar01: I have a fix in a WIP PR. https://github.com/marshmallow-code/marshmallow/pull/823

mfow:
> @lafrech Since this is a major release, let's remove strict completely

@sloria, I think you should reconsider keeping the strict field, although I agree it should be the default.

Consider the case of a third party dependency that adds an additional field to a response from an API call. It's common practice to add additional json fields without considering this to be a breaking change.

Alternatively, consider two microservices that communicate over json. By forcing "strict", we make it difficult to deploy or rollback these microservices independently.

Even the [JSON Schema spec](http://json-schema.org/latest/json-schema-core.html#rfc.section.6.4) itself allows for additional fields: 

> Implementations MAY define additional keywords to JSON Schema

lafrech:
`strict` is not about additional fields (for this, see EXCLUDE, IGNORE, RAISE).

It is about raising an exception in case of error vs. returning a data + errors tuple.

PR #712, opened 2018-01-02
Don't validate Email and URL fields on serialization

### LINKED ISSUE #608: Email and URL fields should not validate on serialization
All fields except for `Email` and `URL` validate only upon deserialization. `Email` and `URL` fields should be modified to be consistent with the other fields.

lafrech:
Closes #608.

`email` was used to generate dump errors in tests not directly related to `Email` field. I replaced it with `age` except when using `age` broke the test (because the schema also has an `is_old` Method field depending on `age`), in which case I used `balance`.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -1228,18 +1228,7 @@ class Dict(Field):
         return result
 
 
-class ValidatedField(Field):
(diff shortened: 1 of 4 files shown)


sloria:
Thanks!

Issue #378, opened 2016-01-14
missing and default values should be specified in deserialized form


chadrik:
Marshmallow expects `missing` and `default` values to be specified in a pre-serialized form, which is inconvenient and unintuitive.

Here is a schema which I would expect to work:

``` python
import datetime
import uuid

class TestSchema(Schema):
    id = fields.UUID(
        missing=uuid.uuid1)
    ts = fields.DateTime(
        format='rfc',
        missing=datetime.datetime.now)
```

However, it does not:

``` python
schema = TestSchema()
pprint.pprint(tuple(schema.load({})))
({}, {'id': [u'Not a valid UUID.'], 'ts': [u'Not a valid datetime.']})
```

Instead, I have to convert my `default` and `missing` values to primitive types, which undermines the purpose of a library like marshmallow that already knows how to do this.  For data types like datetimes, which have many possible serialized formats, I have to keep the formatting of my `default` and `missing` values in sync with my field properties, which is fragile and convoluted:

``` python
from email.Utils import formatdate

class TestSchema(Schema):
    id = fields.UUID(
        missing=lambda: str(uuid.uuid1()))
    ts = fields.DateTime(
        format='rfc',
        missing=formatdate)
```

In the case of `default` values, marshmallow does not attempt to validate them, so no error is raised, but the resulting document is not deserializable.

immerrr:
I have hit this working with comma-delimited parameters in webargs.

costrouc:
I had this exact same error. I find it unintuitive that `missing=datetime.now` does not work but instead requires `missing=lambda: str(datetime.now())`

sloria:
I think the OP makes sense. I would gladly review and merge a PR for this.

lafrech:
I just sent a PR to achieve this.

@chadrik, @costrouc, @sloria, feedback welcome.

Kareeeeem:
That workaround works but then context is not there. I get the idea for the change but the defaulting nested schemas just became a hell of a lot more convoluted.

PR #714, opened 2018-01-02
Prevent using field name when using load_from

lafrech:
I'm afraid there is an issue with `load_from`.

In `Unmarshaller.deserialize`, it is only checked if `raw_value` is `missing`.

```python
    if raw_value is missing and field_obj.load_from:
```

So if incoming data has a key that accidentally matches the field name, the corresponding value is used. 

Although the tests explicitly test this, I don't think this is intended, so I assume the tests are wrong too.

This PR fixes that and fixes the tests.

I'm working on `dev` branch, but I'm now realizing this also affects `2.x`. Should I base this PR on `2.x`?

I believe this is a bugfix but not a breaking change. It only breaks scenarios when someone would write

```python
    years = fields.Integer(attribute='age', load_from='Years')
```

and deserialize data with a `years` (not `Years`!) field.

### DIFF
--- a/docs/upgrading.rst
+++ b/docs/upgrading.rst
@@ -300,6 +300,13 @@ deserialization. They do not validate on serialization. This makes them
 more consistent with the other fields and improves serialization
 performance.
 
+
+Field name is not looked for when ``load_from`` is specified
+************************************************************
+
+When ``load_from`` is specified on a field, only ``load_from`` is looked for in the input data, the field name is not looked for anymore.
+
+
 Upgrading to 2.3
 ++++++++++++++++
 
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -65,9 +65,7 @@ class Field(FieldABC):
         input value is missing. May be a value or a callable.
     :param str attribute: The name of the attribute to get the value from. If
         `None`, assumes the attribute has the same name as the field.
-    :param str load_from: Additional key to look for when deserializing. Will only
-        be checked if the field's name is not found on the input dictionary. If checked,
-        it will return this parameter on error.
+    :param str load_from: Key to look for when deserializing.
     :param str dump_to: Field name to use as a key when serializing.
     :param callable validate: Validator or collection of validators that are called
         during deserialization. Validator takes a field's input value as
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -254,8 +254,11 @@ class Unmarshaller(ErrorStore):
             for attr_name, field_obj in iteritems(fields_dict):
                 if field_obj.dump_only:
                     continue
+                field_name = attr_name
+                if field_obj.load_from:
+                    field_name = field_obj.load_from
                 try:
-                    raw_value = data.get(attr_name, missing)
+                    raw_value = data.get(field_name, missing)
                 except AttributeError:  # Input data is not a dict
                     errors = self.get_errors(index=index)
                     msg = field_obj.error_messages['type'].format(
@@ -267,10 +270,6 @@ class Unmarshaller(ErrorStore):
                     errors.setdefault(SCHEMA, []).append(msg)
                     # Input data type is incorrect, so we can bail out early
                     break
-                field_name = attr_name
-                if raw_value is missing and field_obj.load_from:
-                    field_name = field_obj.load_from
-                    raw_value = data.get(field_obj.load_from, missing)
                 if raw_value is missing:
                     # Ignore missing field if we're allowed to.
                     if (
@@ -283,11 +282,7 @@ class Unmarshaller(ErrorStore):
                 if raw_value is missing and not field_obj.required:
                     continue
 
-                getter = lambda val: field_obj.deserialize(
-                    val,
-                    field_obj.load_from or attr_name,
-                    data
-                )
+                getter = lambda val: field_obj.deserialize(val, field_name, data)
                 value = self.call_and_store(
                     getter_func=getter,
                     data=raw_value,
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -1071,17 +1071,17 @@ class TestSchemaDeserialization:
         data = {
             'Name': 'Mick',
             'UserName': 'foobar.com',
-            'years': 'abc'
+            'Years': 'abc'
         }
         result, errors = AliasingUserSerializer().load(data)
         assert errors['UserName'] == [u'Not a valid email address.']
-        assert errors['years'] == [u'Not a valid integer.']
+        assert errors['Years'] == [u'Not a valid integer.']
 
     def test_deserialize_with_load_from_param(self):
         class AliasingUserSerializer(Schema):
             name = fields.String(load_from='Name')
             username = fields.Email(attribute='email', load_from='UserName')
-            years = fields.Integer(attribute='age', load_from='Years')
+            years = fields.Integer(load_from='Years')
         data = {
             'Name': 'Mick',
             'UserName': 'foo@bar.com',
@@ -1090,7 +1090,7 @@ class TestSchemaDeserialization:
         result, errors = AliasingUserSerializer().load(data)
         assert result['name'] == 'Mick'
         assert result['email'] == 'foo@bar.com'
-        assert result['age'] == 42
+        assert 'years' not in result
 
     def test_deserialize_with_dump_only_param(self):
         class AliasingUserSerializer(Schema):
--- a/tests/test_marshalling.py
+++ b/tests/test_marshalling.py
@@ -247,12 +247,12 @@ class TestUnmarshaller:
         fields_dict = {
             'name': fields.String(load_from='Name'),
             'username': fields.Email(attribute='email', load_from='UserName'),
-            'years': fields.Integer(attribute='age', load_from='Years')
+            'years': fields.Integer(load_from='Years')
         }
         result = unmarshal.deserialize(data, fields_dict)
         assert result['name'] == 'Mick'
         assert result['email'] == 'foo@bar.com'
-        assert result['age'] == 42
+        assert 'years' not in result
 
     def test_deserialize_fields_with_dump_only_param(self, unmarshal):
         data = {


sloria:
The current behavior is expected. The docs say

> load_from (str) – Additional key to look for when deserializing. Will only be checked if the field’s name is not found on the input dictionary. 

So this is backwards-incompatible. I would be willing to change the behavior, though I'd like to understand the rationale for the original behavior. @hakjoon Do you remember why we decided to check the field name before checking `load_from` in [your PR](https://github.com/marshmallow-code/marshmallow/pull/126)?

lafrech:
Oh, right. It looked intentional in the code but I didn't see the rationale at all, so I thought it might be a mistake.

Note that current behavior does not allow swapping fields likes this:

```python
field_a = fields.Integer(load_from='field_b')
field_b = fields.Integer(load_from='field_a')
```
Not really a big deal as this would be a bit strange.

However, I don't like the fact that the deserializer accepts a value with a name from the model that is not exposed in the API. The API only exposes the `load_from` string. The field name is internal, implementation detail.

lafrech:
Good news is since it is a breaking change, it is 3.0, so I don't need to rebase to 2.x.

hakjoon:
My original thought was that It would allow you to use the same schema to load a foreign API with PascalCase and an API created with this schema with proper snake_case. If we didn’t check `attr_name` and you loaded a dataset, dumped it you could no longer load the dumped dataset using the same schema. 

I don’t know if this is worth supporting, clearly it can lead to confusion. In practice I’ve found it less confusing to just create different schemas when I have weird needs like this. Keeping straight how `load_from`, `dump_to` and `attribute` will interplay can get a little hairy. 

(In fact I would love the ability to specify a load only or dump only schema so it would only work one way.)

lafrech:
Thanks @hakjoon for the clarification.

I've always been confused by `load_from`, `dump_to` and `attribute`, thinking they must overlap, somehow. Now, I think I get it.

In the general case, dump/load is symmetric and if you have a model field that is exposed with another name, you use the API name in the schema, and specify the model name as `attribute`.

```python
         class MySchema(Schema):
             ApiName = fields.String(attribute='model_name')
```

`dump_to` and `load_from` introduce asymmetry, allowing to specify a different key for dump and load:

```python
         class MySchema(Schema):
             model_name = fields.String(load_from='api_load_name', dump_to='api_dump_name')
```

If you wanted to reproduce use case 1 using these, you would need to specify both `load_from` and `dump_to` and make sure they match:

```python
         class MySchema(Schema):
             model_name = fields.String(load_from='api_name', dump_to='api_name')
```

As you're saying, this may be a bit confusing. There might be use cases where asymmetry is desirable. I can't tell.

The use case I'm familiar with is an API exposing an object model, documented with apispec. In this case, it is probably a better idea to use two different schemas. In fact, AFAIU, apispec itself doesn't know what to do when `dump_to != dump_load` so it gives up and exposes the field name (see https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/marshmallow/swagger.py#L95 and https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/marshmallow/swagger.py#L407).

I don't have a strong stand against `load_from` / `dump_to` (I just don't use them) and maybe my confusion came from not reading the docs thoroughly. Of course, if there is consensus that they provide a feature that is best achieved another way (using different dump/load schemas), then I guess we might as well simplify things and remove them.

I stumbled upon this while trying to address https://github.com/marshmallow-code/marshmallow/issues/378 and figured I should solve this before introducing new changes in `Unmarshaller.deserialize`.

hakjoon:
one thing that I forgot to mention was that when `load_from` was added `dump_to` didn’t exist so I needed the weird fallback behavior with `attr_name`. Now that `dump_to` exists that fallback is probably no longer necessary and I think your suggested change is probably clearer behavior. 

I might be in an unusual use case in that I use serilaizers a lot for consuming external APIs so `load_from` is super useful, but I have moved to making serializers that are only meant for be used for loading in these circumstances instead of trying to make something work both ways. 

I feel like I tried using `attribute` to solve this like you demonstrated but it didn't work for some reason. I can't really remember why but you'll notice the original issue #125 was around exposing attribute for serialization and deserialization.  

My main usecase is to be able to do transforms like this. 

```
// input
{
    'ApiKey': 2,
    'weird-key': 3
}
// output
{
   'apikey': 2,
   'weird_key':3
}
```

lafrech:
I checked your example in #125 and I don't see what you couldn't do with `attribute`.

```python
class UserSchema(Schema):
    name = fields.String(deserialize_from="Name")
    email_addr = fields.String(attribute="email", deserialize_from="EmailAddress")
```

could be written

```python
class UserSchema(Schema):
    Name = fields.String(attribute='name')
    EmailAddress = fields.String(attribute='email')
```

However, `attribute` doesn't solve your example above:

```python
class MySchema(Schema):
    ApiKey = fields.String(attribute='apikey')
    weird-key = fields.String(attribute='weird_key')  # syntax error
```

This could be solved on user side with meta-programming, but it sucks.

Maybe this is the reason `attribute` didn't work for you.

A symmetric to `attribute` would allow this (let's call it `key`):

```python
class MySchema(Schema):
    apikey = fields.String(key='apikey')
    weird_key = fields.String(key='weird-key')
```
and in fact, if it existed, it would cover all that `attribute` does so `attribute` itself wouldn't be useful, would it?

This would allow dump/load to/from any API with keys that are invalid python keywords on a symmetric fashion (load key = dump key) and with unique keys (only one key accepted for a field).

What `load_from` / `dump_to` brings on top of this is

- asymmetry
- different keys allowed

If this is correct, then it boils down to whether we want asymmetry (from what we said, different schemas sounds better) and whether we want to allow multiple keys for a field (this sounds wrong).

Should we implement `attribute`'s symmetric parameter? How should it be named? Should we keep `attribute`, then?

hakjoon:
Regarding,
```python
class UserSchema(Schema):
    Name = fields.String(attribute='name')
    EmailAddress = fields.String(attribute='email')
```
I don't think it ever occurred to me to use `attribute` like this :)

The proposed use of `key` makes a lot of sense to me.  Honestly that was all I needed when I proposed this as I wanted to move a bunch of incoming API mappers from BPMappers, which had that exact feature, to Marshmallow because it supported a bunch of features I had bolted on top of BPMappers.

I think all the current options can be a little overwhelming, so I am all for simplifying them.  For example I can't tell at a glance what would happen here.
```python
class MySchema(Schema):
    email_addr = fields.String(attribute="email", load_from="EmailAddress", dump_to='the_email')
```
Depending on what Steven wants to support regarding symmetric vs asymmetric options, your proposal would work for my needs.

sloria:
There shouldn't be a need for `key`. Just use `load_from` to specify the key in the input. The name of the field will be the output key.

```python


class MySchema(Schema):
    apikey = fields.Int(load_from='ApiKey')
    weird_key = fields.Int(load_from='weird-key')

    class Meta:
        strict = True


in_data = {
    'ApiKey': 2,
    'weird-key': 3
}
out_data = {
    'apikey': 2,
    'weird_key': 3
}

sch = MySchema()
deserialized = sch.load(in_data).data
assert sch.dump(deserialized).data == out_data
```

lafrech:
Sure, but if you want this to be symmetrical, you need to pass both `load_from` and `dump_to`, right?

```python
class MySchema(Schema):
    apikey = fields.Int(load_from='ApiKey', dump_only='ApiKey')
    weird_key = fields.Int(load_from='weird-key', dump_only='weird-key')
```

Our point is that we could simplify things by disallowing asymmetry and use a single argument (let's call it `key` for now):

```python
class MySchema(Schema):
    apikey = fields.Int(key='ApiKey')
    weird_key = fields.Int(key='weird-key')
```
And if we do that, then there is no point in `attribute` anymore (except backward compatibility).

timc13:
it would be even cooler if we could specify a `key_func` in Meta that takes the field name as an argument.

My use case is that I'm camelCasing everything coming in to & going out of the API layer.  Right now, I'm doing some monkeypatching like this:

```
def to_camel_case(snake_str):
    components = snake_str.split('_')
    return components[0] + ''.join(x.title() for x in components[1:])


def camel_case_schema_all_fields():
    for cls_ in BaseSchema.__subclasses__():
        for field_name, field_value in cls_._declared_fields.items():
            field_value.dump_to = to_camel_case(field_name)
            field_value.load_from = to_camel_case(field_name)


camel_case_schema_all_fields()
```


it would be awesome if I could just do this:
```
class BaseSchema(Schema):
    class Meta:
        key_func = to_camel_case
```

sloria:
Ah, I see now. Thanks for the clarification @lafrech . I like the idea of having just one parameter for specifying both the input and output key, but I'm a bit nervous about what use cases we might break. Sure, having `load_from` and `dump_to` is more typing, but it allows the greatest control.

If you want that feature, you could do this:

```python
class BaseSchema(Schema):

        if field_obj.metadata.get('key'):
            key = field_obj.metadata['key']
            field_obj.load_from = key
            field_obj.dump_to = key
```

So I think for now, we'll move this PR forward then create a RFC issue about the `key` parameter. Sound good @lafrech and @hakjoon ?


@timc13 Your use case is also addressed with `on_bind_field`:

```python
class BaseSchema(Schema):

        if self.opts.key_func:
            key = self.opts.key_func(field_name)
            field_obj.load_from = key
            field_obj.dump_to = key
```

lafrech:
OK for the RFC issue about the new `key` parameter (we can probably find a better name). The proposal would be:

- Remove `load_from` and `dump_to`
- Remove `attribute`
- Add a new parameter as described above

We can't really split this in two, because if we remove `load_from` and `dump_to` without adding the new parameter, we break the `weird-key` use case. However, we could remove `load_from` and `dump_to`, add the new parameter and keep `attribute`. But I don't see the point of keeping `attribute` as it brings nothing more than what the new parameter would provide.

Setting `load_from` and `dump_to` at once can also be done in a `BaseSchema`. However, my point here is not only to simplify things for the user but also to simplify the code here and in related libs (webargs, apicspec,...) and avoid undefined cases. Like @hakjoon says, there can be combinations that are not easily grasped at first sight. The case where `load_from` and `dump_to` are different is a bit broken in apispec IIUC.

I think @timc13's proposal is not directly related. It could be a nice feature to have but it can be added afterwards in a non-breaking way and, as you say, this can be done in a `BaseSchema` in the meantime.

Regarding this PR, it sounds like it could be merged as current behaviour is calling for trouble and a better way to achieve the same thing would be to create two schemas.

lafrech:
Rebased.

CI error is a PyPi connection timeout. I don't have the credentials to relaunch.

sloria:
Thanks @lafrech . Would you mind updating the documentation for `load_from` to reflect the new behavior?

lafrech:
Done.

sloria:
Thanks! Will merge this over the weekend.

PR #719, opened 2018-01-06
Add code of conduct

sloria:
Add Code of conduct based on Trio's COC, which is based on the
WeAllJS COC, which is excellent. This COC will apply to all
marshmallow-code projects.

Attn people with commit bits, @deckar01 @lafrech : For now, I will take lead on enforcement when necessary. I encourage you to also take part as you see fit.


deckar01:
It's nice to have an explicit guideline to reference. Just reading that code of conduct has made me start to think a little deeper about my comments.

Issue #721, opened 2018-01-11
Add support for data URI scheme


jakubczaplicki:
Currently [data URI's](https://en.wikipedia.org/wiki/Data_URI_scheme) are not supported by `fields.Url` validation. Marshmallow should allow to pass a valid `data:` URI.
E.g. ```data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg==```

sloria:
See https://github.com/marshmallow-code/marshmallow/pull/722#issuecomment-357327474

atmo:
Hello, @sloria ! I understand that you don't want to include data `URI`s support to `fields.Url` (as `URL` is only a subset of all possible `URI`s, see [spec](https://tools.ietf.org/html/rfc3986#section-1.1.3) or [wikipedia](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#URLs_and_URNs)). How would you feel about defining a separate field for `URI`? I can see how `URL` might then be inherited from `URI` as `URI` will hold most of the common functionality.

PR #722, opened 2018-01-11
Add support for data URI scheme

jakubczaplicki:
Fix for #721 
Adds support for data URI's scheme. The change consist of adding a regex variant in `_regex_generator()`


sloria:
I'm going to have to pass on this one. Data URIs are not URLs and therefore should not be validated using the `URL` validator.

I recommend that you use a custom validator for Data URIs.

Issue #717, opened 2018-01-05
RFC: replace load_from and dump_to with a single parameter


lafrech:
See discussion in https://github.com/marshmallow-code/marshmallow/pull/714.

In the general case, dump/load is symmetric and if you have a model field that is exposed with another name, you use the API name in the schema, and specify the model name as `attribute`.

```python
         class MySchema(Schema):
             ApiName = fields.String(attribute='model_name')
```

`dump_to` and `load_from` introduce asymmetry, allowing to specify a different key for dump and load:

```python
         class MySchema(Schema):
             model_name = fields.String(load_from='api_load_name', dump_to='api_dump_name')
```

If you want to reproduce use case 1 using these, you need to specify both `load_from` and `dump_to` and make sure they match:

```python
         class MySchema(Schema):
             model_name = fields.String(load_from='api_name', dump_to='api_name')
```

The flexibility brought by having both `load_from` and `dump_to` comes at a price: complexity for the user but also in the code (marshmallow and related libs like webargs, apispec...), potential undefined cases for some weird combinations...

Assuming symmetry is not required, a single parameter should be enough. There is however a limitation with `attribute`. It can't be used for APIs where the keys are invalid Python variable names:

```python
class MySchema(Schema):
    ApiKey = fields.String(attribute='apikey')
    weird-key = fields.String(attribute='weird_key')  # syntax error
```

Currently, `load_from` and `dump_to` can be used for this as shown above.

The proposal here is to introduce a new parameter to unite them all. Let's call it `key` for now, short of a better name:

```python
class MySchema(Schema):
    apikey = fields.Int(key='ApiKey')
    weird_key = fields.Int(key='weird-key')
```

If we do that, then there is no point in `attribute` anymore (except backward compatibility). So we'd remove `attribute`, `dump_to` and `load_from`.

Any objection to this?

Any suggestion for a better name?

tinproject:
I was reading the discussion on PR #174. Just to add some context and see if I understand well (the doc is not always very clear):

We have a `Python Object` that by means of an `Schema` instance we serialize (`dump`) into a `Serialized Representation`, that we can take the backwards process and deserialize (`load`) into a `Python Object`.

A `Schema` have `Fields` that knows how the data can be de/serialized. That `Fields` can take an `attribute` from a `Python Object` and serialize it into a `key/name` in a map of a `Serialized Representation`, and back.


Some comments answering the questions:

[...]

PR #724, opened 2018-01-18
Merge load_from and dump_to into data_key

lafrech:
Working on https://github.com/marshmallow-code/marshmallow/issues/717.

- [x] Replace `load_from` with `key_xxx`
- [x] Replace `dump_to` with `key_xxx`
- [x] The name of the parameter is still to be determined, so I used `key_xxx` for easy find and replace.
- [x] Update docs

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -371,50 +371,35 @@ By default, `Schemas` will marshal the object attributes that are identical to t
     #  'date_created': '2014-08-17T14:58:57.600623+00:00'}
 
 
-Specifying Deserialization Keys
--------------------------------
+Specifying Serializarion/Deserialization Keys
+---------------------------------------------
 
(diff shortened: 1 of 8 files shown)


taion:
I'm a fan of this change.

Is the plan on `attribute` to drop it in a future PR?

What do you think of `data_key` as a name? The signatures on `Schema` are:

```python
    def dump(self, obj, many=None, update_fields=True, **kwargs):
    def load(self, data, many=None, partial=None):
```

So `data` seems like an okay name here to flag the serialized object.

lafrech:
Thank you, @taion, for your positive feedback.

> Is the plan on attribute to drop it in a future PR?

I think attribute would stay (see https://github.com/marshmallow-code/marshmallow/issues/717).

> What do you think of data_key as a name?

Why not. I can't think of a single word (no underscore) name. Maybe @sloria or anyone will come up with something.

Let's try to keep the discussion in #717.

lafrech:
Done. @sloria , I think this is good to go.

sloria:
Looks good to me.

Issue #511, opened 2016-08-17
schema load None


Windfarer:
If I load None to schema, it will not return validate error

Windfarer:
Is it a feature or a bug?

maximkulkin:
Although this behavior is debatable, it is still easy to check on your side because it only affects top-level objects, for nested objects it is handled by fields.Nested field type.

jasonab:
I agree that this is an error, especially if required fields are set. If I set required=True, I expect that any structure coming out of marshmallow has that field in it, so I should not need to check for its existence.

sloria:
I would be open to changing the behavior for deserializing `None`. I think either

1. Deserialize `None` as an empty dict. Required fields would raise an error.
2. Raise a schema-level error. The errors dictionary would be something like `{'_schema': 'No data provided'}`

Thoughts?

Windfarer:
I think there would be two level validation. 
Maybe a schema level ```ignore_none```or ```allow_none``` parameter could be introduced.
If we init a schema with this parameter, the loading will do or skip this schema level validation. And if we allow None to go into the field level validation, then the field will check the require fields and raise field level exception. 

Just my subjective opinion, for yours' discussion.

lafrech:
Wouldn't it be better if `dump` acted symmetrical so that `dump(load(stuff))` would always return `stuff`?

Currently, this is what happens:

    Test.dump(Test().load(None).data).data -> {}
    Test.load(Test().dump(None).data).data -> {}

I'm not sure whether `None` should [de|]serialize as `None` or raise en `Exception`, but empty `dict` is not equivalent to `None`. `None` has a special meaning.

CausticYarn:
I vote for option 1, return an empty dict, with an error thrown for missing required fields. Any non-required fields would return a KeyError on access, just as if they weren't provided.

douglas-treadwell:
I would be okay with option 1 as long as there is a way to opt-in for passing None to raise an exception.

lafrech:
I think I like option 2 better. More explicit. And it doesn't break symmetry (`dump(load(stuff)) = stuff`, see my comment above). But I can live with both options.

Anybody can provide a use case for `None` [|de]serializing as `{}`?

@sloria, I guess this should be labeled _backwards_incompat_.

lafrech:
I just opened a PR with an implementation of option 2 (raise `ValidationError`).

[...]

PR #725, opened 2018-01-18
Schema.load(None) raises ValidationError

lafrech:
Proposal for #511.


sloria:
This looks like desirable behavior. LGTM!

Issue #406, opened 2016-03-06
Nesting field should validate nested schema instance


frol:
Consider the following test cases:

``` Python
class Inner(Schema):
    foo = fields.Field()
    bar = fields.Field()
    baz = fields.Field()

def test_nested_schema_instance_only_and_exclude():
    class Outer(Schema):
        inner = fields.Nested(Inner(), only=('foo', 'bar'), exclude=('bar', ))

    sch = Outer()
    data = dict(inner=dict(foo=42, bar=24, baz=242))
    result = sch.dump(data)
    assert 'foo' in result.data['inner']
    assert 'bar' not in result.data['inner']

def test_nested_schema_class_only_and_exclude():
    class Outer(Schema):
        inner = fields.Nested(Inner, only=('foo', 'bar'), exclude=('bar', ))

    sch = Outer()
    data = dict(inner=dict(foo=42, bar=24, baz=242))
    result = sch.dump(data)
    assert 'foo' in result.data['inner']
    assert 'bar' not in result.data['inner']
```

The only difference is `Inner()` vs `Inner`. The first test case fails as `bar` field will be included even though it is passed to the exclusion list of the nesting field.

Thus, [when `SchemaABC` instance is passed](https://github.com/marshmallow-code/marshmallow/blob/78551b3d583394f75c4a82823247bdec7da87047/marshmallow/fields.py#L387) we should validate `many`, `only`, `exclude` settings and raise an error if something is conflicting.

P.S. The second test case is taken from [marshmallow/tests/test_schema.py](https://github.com/marshmallow-code/marshmallow/blob/78551b3d583394f75c4a82823247bdec7da87047/tests/test_schema.py#L694).

sloria:
I think validating `many`, `only`, and `exclude` is a good idea--it will certainly prevent incorrect usages.

remeika:
According to [the docs](https://github.com/marshmallow-code/marshmallow/blob/966685e61b976e9006051658626de474be8bac82/marshmallow/fields.py#L355), this is (now) a known limitation:

> When passing a Schema instance as the first argument, the instance’s exclude, only, and many attributes will be respected.
> 
> Therefore, when passing the exclude, only, or many arguments to fields.Nested, you should pass a Schema class (not an instance) as the first argument.

Is this a satisfactory state of affairs? It seems to me the following behaviors would be preferable, in descending order of preference:

1. Both Schema classes and instances have the same behavior when passed to `fields.Nested`
[...]

PR #731, opened 2018-01-30
Treatment for many and only parameters

4lissonsilveira:
- A small treatment to avoid conflict with only and exclude params
has been created.
- An If has been created to verify if many parameter is boolean.

Related: #406

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -386,6 +386,22 @@ class Nested(Field):
         self.many = kwargs.get('many', False)
         self.__schema = None  # Cached Schema instance
         self.__updated_fields = False
+
+        if not isinstance(self.many, bool):
+            raise ValueError('Many parameter must be boolean')
+
+        if only:
+            if isinstance(only, basestring):
+                only = [only]
+
+            # Catch fields used at the same time
+            # into only and exclude parameters
+            fields = [item for item in only if item in list(exclude)]
+            if fields:
+                raise ValueError(
+                    'Fields {0} cannot be at exclude and only at the same time'.format(
+                        fields))
+
         super(Nested, self).__init__(default=default, **kwargs)
 
     @property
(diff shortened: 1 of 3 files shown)


4lissonsilveira:
@sloria do you think that is a good approach? Could you(or another contributor) give me any suggestions for improvements?

Thanks!

marshmallow/fields.py
```diff
@@ -386,6 +386,22 @@ def __init__(self, nested, default=missing_, exclude=tuple(), only=None, **kwarg
         self.many = kwargs.get('many', False)
         self.__schema = None  # Cached Schema instance
         self.__updated_fields = False
+
+        if not isinstance(self.many, bool):
+            raise ValueError('Many parameter must be boolean')
```
lafrech, line 391: `many` does not have to be a boolean. You can pass anything, as long as it evaluates to a either truthy value if you expect a collection or a falsy value if you expect a single element.

marshmallow/fields.py
```diff
@@ -386,6 +386,22 @@ def __init__(self, nested, default=missing_, exclude=tuple(), only=None, **kwarg
...
+
+        if only:
+            if isinstance(only, basestring):
+                only = [only]
+
+            # Catch fields used at the same time
+            # into only and exclude parameters
+            fields = [item for item in only if item in list(exclude)]
```
lafrech, line 399: You could use set intersections to do that.

lafrech:
I'm not sure we're willing to add this kind of validation.

- You may want to pass anything as `many`. It does not have to be a boolean. It will evaluate to `True` or `False`.

- I believe `only` has priority over `exclude` (or conversely, I don't remember). Raising an exception when a field is in both lists may break some workflows. However, I agree that relying on this may be a bit scary, and in some cases, it could hide an error. And the fact that I can't remember which one takes priority and I can't find it quickly in the docs makes me think it could be sensible to prevent it. I suppose the assumption is "Since it can be solved by an explicit priority and it creates no issue, why prevent it? The user should be aware or what he is doing." Same applies to `many`.

deckar01:
These seem like good warnings for new users, but would likely be burdensome errors for existing users.

sloria:
Closing this in favor of #826

PR #734, opened 2018-02-07
Rework error message collection in structured Dict

### LINKED ISSUE #730: Dict(values=Nested()) doesn't collect complete validation errors
Using the new structured Dict support from #483 / #700, the error handling/collection logic seems unable to provide access to the original/detailed failures due to how it aggregates multiple errors.

* marshmallow.__version__ = '3.0.0b6'
* Python 2.7.14

Simple test is something like:

```python
from marshmallow import Schema, pprint, fields

class InnerSchema(Schema):
    inner = fields.String(validate=OneOf(['a','b'], error='inner validation fail on {input}'))

class OuterSchema(Schema):
    outer = fields.Dict(values=fields.Nested(InnerSchema))

foo = {
    'outer': {
        'x': {'inner': 'a'},
    }
}
foo2 = {
    'outer': {
        'x': {'inner': 'xxx'},
    }
}

# This works (no validation failures)
outer_sch = OuterSchema()
ret = outer_sch.load(foo)
pprint(ret)
# >>> UnmarshalResult(data={'outer': {'x': {'inner': u'a'}}}, errors={})

# Fails to validate Inner.inner
outer_sch2 = OuterSchema()
ret2 = outer_sch.load(foo2)
pprint(ret2)
# >>> UnmarshalResult(data={}, errors={'outer': {'x': [u'Invalid value: inner']}})
# Note: the custom error message on outer.x.inner is not present.

# validation + custom message is working as expected on the inner String field.
inner_sch = InnerSchema()
inner2 = inner_sch.load(foo2['outer']['x'])
pprint(inner2)
# >>> UnmarshalResult(data={}, errors={'inner': ['inner validation fail on xxx']})
```

I don't think it's possible to extract the actual failure message (`'inner validation fail on xxx`) with any of hte existing error handling overrides, because it's getting dropped (I think) here: 
https://github.com/marshmallow-code/marshmallow/blob/3.0.0b6/marshmallow/fields.py#L1219

(since `e.messages` is a dict but the loop is only taking its keys).

Not sure exactly what an appropriate result would be. Any access to the inner validation structures would be fine for what I need.

lafrech:
I believe this fixes #730.

@shabble, @deckar01, would you mind reviewing?

Thanks.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -1195,7 +1195,7 @@ class Dict(Field):
         if not self.value_container and not self.key_container:
             return value
 
-        errors = {}
+        errors = collections.defaultdict(dict)
         values = list(value.values())
         keys = list(value.keys())
         if self.key_container:
@@ -1203,10 +1203,7 @@ class Dict(Field):
                 try:
                     keys[idx] = self.key_container.deserialize(key)
                 except ValidationError as e:
-                    errors[key] = [
-                        'Invalid key: {}'.format(message)
-                        for message in e.messages
-                    ]
+                    errors[key]['key'] = e.messages
         if self.value_container:
             for idx, item in enumerate(values):
                 try:
@@ -1214,12 +1211,7 @@ class Dict(Field):
                 except ValidationError as e:
                     values[idx] = e.data
                     key = keys[idx]
-                    if key not in errors:
-                        errors[key] = []
-                    errors[key].extend([
-                        'Invalid value: {}'.format(message)
-                        for message in e.messages
-                    ])
+                    errors[key]['value'] = e.messages
         result = dict(zip(keys, values))
 
         if errors:
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -1305,26 +1305,6 @@ class TestSchemaDeserialization:
         errors = MySchema(partial=True).validate({'foo': 3}, partial=('bar', 'baz'))
         assert errors == {}
 
-    # regression test for https://github.com/marshmallow-code/marshmallow/issues/730
-    def test_structured_dict_nested_value_deserialization(self):
-        class Child(Schema):
-            email = fields.Email()
-
-        class Parent(Schema):
-            users = fields.Dict(
-                keys=fields.String(),
-                values=fields.Nested(Child))
-
-        with pytest.raises(ValidationError) as excinfo:
-            Parent().load({
-                'users': {('whatever', ): {'email': 'mick-stones-com', }, }})
-        assert excinfo.value.args[0] == {
-            'users': {
-                ('whatever', ): {
-                    'key': ['Not a valid string.'],
-                    'value': {'email': ['Not a valid email address.'], },
-                }}}
-
 
 validators_gen = (func for func in [lambda x: x <= 24, lambda x: 18 <= x])
 
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -543,12 +543,14 @@ class TestFieldDeserialization:
         assert excinfo.value.args[0] == 'Not a valid mapping type.'
 
     def test_structured_dict_value_deserialization(self):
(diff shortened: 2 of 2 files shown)


marshmallow/fields.py
```diff
@@ -1195,31 +1195,23 @@ def _deserialize(self, value, attr, data):
...
         if self.key_container:
             for idx, key in enumerate(keys):
                 try:
                     keys[idx] = self.key_container.deserialize(key)
                 except ValidationError as e:
-                    errors[key] = [
-                        'Invalid key: {}'.format(message)
-                        for message in e.messages
-                    ]
+                    errors[key]['key'] = e.messages
```
deckar01, line 1206: ~~I don't think changing the error structure is necessary to fix this bug. From what I can tell, the dict deserialization error structure I implemented is closer to the list deserialization structure than what was proposed in https://github.com/marshmallow-code/marshmallow/issues/730#issuecomment-361362161.~~

I actually like how using a dictionary simplifies the implementation by avoiding array extension and text prefixes, ~~but I propose moving changes to the error structure to a separate issue/PR~~.

marshmallow/fields.py
```diff
@@ -1195,31 +1195,23 @@ def _deserialize(self, value, attr, data):
...
                 try:
                     values[idx] = self.value_container.deserialize(item)
                 except ValidationError as e:
                     values[idx] = e.data
                     key = keys[idx]
-                    if key not in errors:
-                        errors[key] = []
-                    errors[key].extend([
-                        'Invalid value: {}'.format(message)
-                        for message in e.messages
```
deckar01, line 1221: Er, wait. The `messages` property of a `ValidationError` can be a "string, list, or dictionary of error messages".

https://marshmallow.readthedocs.io/en/latest/api_reference.html#marshmallow.exceptions.ValidationError.messages

Ya, I was assuming it was a list. We can't safely merge the value and key errors, so making it a dict to separate them makes sense.

marshmallow/fields.py
deckar01: I see why it is needed now. Thanks for digging into this @lafrech.

deckar01: It might be helpful to provide a more technical insight about this fix in the PR description for subsequent reviewers. It wasn't clear from the conversation in the issue why this was a necessary.

tests/test_deserialization.py
```diff
@@ -1304,6 +1303,60 @@ class MySchema(Schema):
...
+            })
+
+        assert excinfo.value.args[0] == {
+            'authors': {
+                'Mick': {'value': {
+                    'email': ['Not a valid email address.'],
+                    'birthdate': ['Not a valid datetime.']}},
+                ('whatever', ): {
+                    'key': ['Not a valid string.'],
+                    'value': {'birthdate': ['Not a valid datetime.']}}
```
deckar01: This test seems really big. Is it possible to boil the repro case down further? Any `ValidationError` that returns a dict for `messages` should cover the bug.

lafrech: You're right. This test is more a demo than a test. I just shortened it. I kept a success case and a failure. Do you think only the failure is relevant to the test?

deckar01: It looks like we only need to test the error from of custom validators. I was referring to the size of the data structure not the number of cases. Sorry for not being clear.

If this can be reproduced without using a nested schema it would be easier to follow. A schema with a single field that has a custom validator that either produces an error string or error dict should cover the bug.

I am OK with the tests as they are now since they match the reported issue. Any additional work would be icing on the cake. Thanks for working on this.

lafrech:
Thanks @deckar01 for the review. I think I managed to get it right and minimal.

deckar01:
Looks good to me. :+1:

lafrech:
@sloria, I think we're done with this.

sloria:
Awesome!

Issue #743, opened 2018-02-23
Example of validating original input data fails on `many=True`


stj:
https://marshmallow.readthedocs.io/en/latest/extending.html#validating-original-input-data

Given a schema
```python
import marshmallow
import marshmallow.fields

class X(marshmallow.Schema):
    text = marshmallow.fields.String()

    @marshmallow.validates_schema(pass_original=True)
    def check_unknown_fields(self, data, original_data):
        unknown = set(original_data) - set(self.fields)
        if unknown:
            raise marshmallow.ValidationError('Unknown field', unknown)

class Y(marshmallow.Schema):
  nested = marshmallow.fields.Nested(nested=X, many=True)
```

When sending extra fields for validation the `set` fails with a TypeError

```python
schema = Y()
schema.load({"nested": [{"fail": "fast", "text": "text"}]})
```

Result is:
`TypeError: unhashable type: 'dict'`

Changing `marschmallow/schema.py` line 841 from
```python
                        self._unmarshal.run_validator(validator,
                                                  item, original_data, self.fields, many=many,
                                                  index=idx, pass_original=pass_original)
```
to
```python
                        self._unmarshal.run_validator(validator,
                                                  item, original_data[idx], self.fields, many=many,
                                                  index=idx, pass_original=pass_original)
```
fixed the issue.

stj:
marshmallow 3.0.0b7
python 3.6.3

lafrech:
Thanks for the report. I had a quick look and I think you're right about the cause of this.

Your fix looks correct. It could also be written like this but I'm not sure it would be clearer:

```python
for idx, (item, original_item) in enumerate(zip(data, original_data)):
    try:
        self._unmarshal.run_validator(validator,
                                      item, original_item, self.fields, many=many,
                                      index=idx, pass_original=pass_original)
```

Would you like to send a PR and see if this breaks any test? And maybe add a test case for this?

Thanks.

stj:
Did a quick checkout of the repository and run the test suite with the change. No test failed.

Can try to create a PR with a test on Monday.

sloria:
Thanks for reporting and sending a PR @stj . This is the same as #315 ; let's continue discussion there.

PR #744, opened 2018-02-24
:bug: Fix pass_original for validate_schema on nested many

### LINKED ISSUE #315: post_dump is passing a list of objects as original object
Hi,

I think post_dump with pass_original=True should pass the original object related to the data serialized and not a list of objects which this object belongs to.

``` python
from marshmallow import fields, post_dump, Schema

class DeviceSchema(Schema):
    id = fields.String()

    @post_dump(pass_original=True)
    def __post_dump(self, data, obj):
        print(obj)  # <-- this is a list

devices = [dict(id=1), dict(id=2)]
DeviceSchema().dump(devices, many=True)
```

In the above example, the parameter `obj` is a list of devices rather than the device object itself.

What do you think?

stj:
Resolve #315 Pass the correct `original_data` for decorated `validate_schema` method
when dealing with a `Nested(many=True)` field.

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -838,9 +838,9 @@ class BaseSchema(base.SchemaABC):
             if many and not pass_many:
                 for idx, item in enumerate(data):
                     try:
-                        self._unmarshal.run_validator(validator,
-                                                  item, original_data, self.fields, many=many,
-                                                  index=idx, pass_original=pass_original)
+                        self._unmarshal.run_validator(
+                            validator, item, original_data[idx], self.fields,
+                            many=many, index=idx, pass_original=pass_original)
                     except ValidationError as err:
                         errors.update(err.messages)
             else:
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -836,11 +836,11 @@ class BaseSchema(base.SchemaABC):
             if pass_many:
                 validator = functools.partial(validator, many=many)
             if many and not pass_many:
-                for idx, item in enumerate(data):
+                for idx, (item, orig) in enumerate(zip(data, original_data)):
                     try:
                         self._unmarshal.run_validator(
-                            validator, item, original_data[idx], self.fields,
-                            many=many, index=idx, pass_original=pass_original)
+                            validator, item, orig, self.fields, many=many,
+                            index=idx, pass_original=pass_original)
                     except ValidationError as err:
                         errors.update(err.messages)
             else:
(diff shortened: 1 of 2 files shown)


sloria:
The desired behavior is:

> I think the intuitive behavior is for pass_original to return the corresponding original object instance unless pass_many is explicitly [true].

Can you add a test for passing both `pass_many=True` and `pass_original=True`? In this case, the original collection should be passed.

EDIT: Fix typo.

marshmallow/schema.py
```diff
@@ -838,9 +838,9 @@ def _invoke_validators(self, pass_many, data, original_data, many, field_errors=
             if many and not pass_many:
                 for idx, item in enumerate(data):
                     try:
-                        self._unmarshal.run_validator(validator,
-                                                  item, original_data, self.fields, many=many,
-                                                  index=idx, pass_original=pass_original)
+                        self._unmarshal.run_validator(
+                            validator, item, original_data[idx], self.fields,
```
sloria: I'm concerned that this will break for collections that don't support indexing (e.g. QuerySets in Django). I think @lafrech 's suggestion [here](https://github.com/marshmallow-code/marshmallow/issues/743#issuecomment-368141351) might address that.

stj:
The default for `pass_many` is `False`. Does that mean the PR #744 is superfluous?

Can we clarify what is expected to be send to the validator method when?

This is what I currently observe when changing the decorator kwargs in the test:
* ```
  @validates_schema(pass_many=True, pass_original=True)
  def validate_schema(self, data, original_data, many): 
    assert many is True
    assert original_data == [{'foo': 1, 'bar': 2}]
    assert data == [{'foo': 1}]```
* ```
  @validates_schema(pass_many=False, pass_original=True)
  def validate_schema(self, data, original_data):
     assert original_data == [{'foo': 1, 'bar': 2}]
     assert data == {'foo': 1}```

Currently `pass_many` only changes `data` but not `original_data`. Is this the expected behavior?

sloria:
Sorry, I had a typo in my comment above--I've fixed it.

From the discussion in #315, it seems the desired behavior is that both `data` and `original_data` change according to the value of `pass_many`.

stj:
Updated the PR with a fixup commit and edited the issue reference.

sloria:
Thanks!

sloria:
@stj Before this is released, the same behavior should also be implemented for `post_dump` and `post_load`. Would you be willing to help with that?

stj:
@sloria Not familiar with `post_dump` nor `post_load` but will take a look at it today.

sloria:
@stj  Here are the docs for them: https://marshmallow.readthedocs.io/en/latest/extending.html#pre-processing-and-post-processing-methods . They take similar parameters to validates_schema. 

Thanks so much!

stj:
@sloria have a look at #750 . Not super happy about the tests but they cover the change.

PR #750, opened 2018-03-06
Fix pass_original on nested many

### LINKED ISSUE #315: post_dump is passing a list of objects as original object

stj:
Pass the correct `original_data` for decorated `post_load` and
`post_dump` methods when dealing with a `Nested(many=True)` field.

Resolve #315

### DIFF
--- a/marshmallow/decorators.py
+++ b/marshmallow/decorators.py
@@ -107,6 +107,9 @@ def post_dump(fn=None, pass_many=False, pass_original=False):
     By default, receives a single object at a time, transparently handling the ``many``
     argument passed to the Schema. If ``pass_many=True``, the raw data
     (which may be a collection) and the value for ``many`` is passed.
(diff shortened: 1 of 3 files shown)


sloria:
Behavior looks correct, and it's cleanly implemented. Thanks!

PR #752, opened 2018-03-12
Marshmallow 3.x - serializer dump method directly return dictionary, …

magul:
…there's no reason to ask for data (and no attribute with that name)


lafrech:
Thanks. I missed those lines in https://github.com/marshmallow-code/marshmallow/commit/a19497b16cf89cf3d2f89166dc4bfad2dcd87f20.

Issue #727, opened 2018-01-24
Incorrect Integer validation for Floats


petargyurov:
When supplying a float value to a schema whose field is registered as `Integer`, Marshmallow will implicitly cast the given float to an integer, without raising errors, even when attempting to force strictness.

e.g.: a given value of 0.15 will be silently cast to 0, of type `int`.

This behaviour comes from the definition of the constructor for the Integer field. Line 707 in `fields.py`:

```
# override Number
def __init__(self, strict=False, **kwargs):
    self.strict = strict
    super(Integer, self).__init__(**kwargs)
```

`strict=False` is the culprit. 

This comes into effect when we format the value:

```
# override Number
def _format_num(self, value):
    if self.strict and isinstance(value, numbers.Number):  # this condition will never be satisfied!
        if not isinstance(value, numbers.Integral):
            self.fail('invalid')
    return super(Integer, self)._format_num(value)
```

Cannot seem to stop this by setting the strictness via the schema object either.
Expected output would be: `ValidationError: {'your_value': [u'Not a valid integer.']`}

Thoughts?

deckar01:
There has been some discussion about this in https://github.com/marshmallow-code/marshmallow/pull/755.

I agree that casting a numerical type is a pretty weak form of validation. I suspect it was implemented this way to handle numbers that were encoded as strings. The rest of the API is strict by default, with flags like `allow_none` to make fields less strict.

Here are some changes that would make number fields more consistent:

- Get rid of the `strict` flag
- Make the number fields strict by default
- Enforce strict integer validation
- Add an `allow_string` argument to control casting

deckar01:
The current strict flag confused some users and prompted a stackoverflow question. See #794.

lafrech:
> Add an `allow_string` argument to control casting

Or use existing `as_string` flag: let it act symmetrically, serialize to string, deserialize from string.

Maybe even raise a `ValidationError` if `as_string` is `True` and the data came as a number.

jerry-le:
In 2.15.4 version, I still don't know how to disable `fields.Integer()` not auto casting the input from float to int.
@deckar01

PR #755, opened 2018-03-20
fix: change how strict works on fields.Int() so it's really strict about types

galuszkak:


galuszkak:
This changes how strict work on Integer field. 
I think this may concern @yoichi who did #667 and @sloria.

sahilkataria:
This definitely fixes the issue of being able to validate integers better. It's currently broken since it doesn't differentiate between a "1" and a 1.

lafrech:
This behaviour looks good to me.

However, I would do the test in `test_deserialization.py`, in `test_strict_integer_field_deserialization` or another test function close to it.

(I would put those tests in `test_fields.py` but since the other `strict` tests are in `test_deserialization.py`, let's put all of them together. There must be a good reason.)

Regarding the deserialization vs. serialization, the rationale is that we want to validate what gets in our application, so we validate on deserialization.

IIUC, the fact that the `_format_num` function, which is called on both serialization and deserialization, does validation is more an accident that a design principle. We don't really want to validate our outputs as we assume our own code is safe. Therefore, I wouldn't test serialization, as we might remove this validation in the future without breaking the API contract.

There are places where validation is done on serialization, but the trend is to remove them (see Email and URL fields here: https://github.com/marshmallow-code/marshmallow/commit/3c649b92af52e1559b3b1cf52cf3b0aedb3c756c).

Just my humble opinion.

galuszkak:
@lafrech done, Is this better ? I didn't catch that strict for integer was in different test suite. I agree that it maybe make sense to change it to `test_fields.py` but for now I just want to fix issue with string as an input treated as strict int.

lafrech:
> @lafrech done, Is this better ? 

Thanks. Looks good to me.

> I agree that it maybe make sense to change it to test_fields.py but for now I just want to fix issue with string as an input treated as strict int.

Right. I was just thinking out loud. There's probably a good reason it is here and not in `test_fields.py` anyway.

yoichi:
I think this change does not match the design policy of marshmallow.
Deserialization should process strings without rejecting them.

I also think the type validation is useful feature to be used
with serialization format which support types (json, yaml, ...).
I think it should be handled in pre-deserialization process
as proposed in https://github.com/marshmallow-code/marshmallow/pull/678 (it was rejected though)

galuszkak:
@yoichi so that is what we have problem right now, because we don't want string to be processed because JSON has number type, and if it's string, we want to reject it, as this is wrong type for us. I would argue that if someone is using for example hexadimecal he may want to have use string, but this isn't our case, and for not there is no easy way to do this kind of validation. I was under impression that `strict` was designed that is really strict and it's rejecting all non integer types (including string). Becuase You can process string with `strict=False`. Maybe beside arguing on `strict` purpose we may need new param that will indicate that strings, won't be processed.

galuszkak:
> Deserialization should process strings without rejecting them.

Isn't that purpose of `strict` to not make it happen? I'm kinda surprised by this statement.

@yoichi so that is our main use case, because JSON/YAML distinguish Number and String as 2 different data types, we don't want to have `'1'` be the same as `1`. We need very strict validation. 

We may argue if this should be done through `strict` param but if this isn't way to do it, I would just change it to different param name, that would include strings (maybe `allow_string` or `reject_string`. Especially if someone wants to convert from string he can do it with `strict=False`. 

What do You think about it?

galuszkak:
@yoichi ping.  I'm happy to move this forward, as we would like to have this feature added or bug fixed (based on what action we will take on this change) :). What do You think about proposal shown above?

yoichi:
Different param name sounds good.
I think it is better to deal with not only Integer but Number and its all derived classes.

deckar01:
Number field deserialization is permissive by default. Anything that can be cast to the type is considered valid (except for bools...). I don't think the `strict` flag was a necessary addition to the API. It could have been implemented with a custom validator/field and so can this. Strict is going to be interpreted differently depending on the use case. If it is cumbersome to create a custom validator/field I think we should improve that instead of hard coding a special use case behind a feature flag.

christian-storm:
For what it's worth I ended up creating my own, e.g.,

```
class StrictInteger(fields.Integer):
    ''' Override fields.Number string to num coercion'''
    def _format_num(self, value):
        """Check instance of int  or raise a :exc:`ValidationError` if an error occurs."""
        if value is None:
            return None
        if not isinstance(value, self.num_type):
            raise ValidationError("Must be type int")
        return value
```

sloria:
>  I don't think the strict flag was a necessary addition to the API. It could have been implemented with a custom validator/field and so can this. Strict is going to be interpreted differently depending on the use case. If it is cumbersome to create a custom validator/field I think we should improve that instead of hard coding a special use case behind a feature flag.

This is a valid point. Validation is the responsibility of the validators. 

@lafrech @deckar01 How should we move this forward? Should we keep `strict`?

lafrech:
Validators validate additional constraints (range of values, multiple-choices,...) but fields should be able to validate data type by default. A `DateTime` field fails if fed an integer, for instance.

This case is complicated because `Integer` can deserialize a string or a number (int or float) and some want to be loose on inputs, others want to be strict, and some want to be loose on the string issue not on the float issue and conversely. We can't do that with a single keyword. I wouldn't use `strict` to toggle both string and float strictness.

I think the general philosophy is to be loose on inputs so it is fine to cast string to number. And it is acceptable to decide rejecting numbers expressed as strings should be done via a validator. (This validator could be provided in the code. It doesn't have to be written by each user.) The default won't round-trip, but that's what happens when being loose on inputs.

Regarding floats being cast to int, it is a bit different. It is fair to say that an integer and an integer expressed as a string hold the same meaning and the difference is related to the transport channel. However, an int and a float are different things (different types). It seems wrong to rely on the fact that Python will cast floats to int automatically and silently. From an API perspective, this Python behavior is internal detail, sort of.

Overall, on second thought, I would say

- Reject this PR and propose another solution to this. Maybe a "number not as string"  (name suggestions welcome) validator that could be used by all number fields. (Could it be more generic than just numbers? I guess not.)

- Perhaps keep `strict` and make it default to `True`. It could be renamed to be more explicit. Like `autocast=False` (name suggestions welcome).

Just thoughts. No definitive opinion. Possibly biased by my usage (web API with JSON inputs).

lafrech:
@deckar01:
>  It could have been implemented with a custom validator/field and so can this. 

@lafrech:
> Overall, on second thought, I would say
> 
>  - Reject this PR and propose another solution to this. Maybe a "number not as string" (name suggestions welcome) validator that could be used by all number fields. (Could it be more generic than just numbers? I guess not.)

Except validation happens after deserialization. This is at the core of marshmallow's logic.

At this stage, it is too late to check the type of the input. So I don't see how to do that with a validator.

So I don't see many alternatives:

- add an opt-in `reject_strings` (name suggestions welcome) feature flag
- accept strings and let the user write a subclass

deckar01:
Should the string and number formats be two different fields? `DateTime` and `TimeStamp`?

lafrech:
Hmmm, I'm not sure I get your point. Maybe I got you confused by talking about `DateTime`.

This issue is about number fields (`Int`, `Float`) and the fact that they may come as numbers or strings. Are you suggesting we provide two fields for each, like `Int` and `IntAsString`? I'd prefer an `as_string` feature flag.

lafrech:
Actually, `as_string` already exists and defines how to serialize. It is not used on deserialization (although it could as I suggested in related #727). It does not define whether or not to be loose on inputs.

deckar01:
I'm wondering when the complexity of a field grows enough that it needs to be split into two fields. Currently, a type like `int` has one field and we try to support all common serialization and deserialization formats through field options. In this case the complexity is pretty low so adding an extra argument or two is probably acceptable, but it might be useful to consider all the use cases:

 - `Number | String -> Integer` (default behavior)
 - `String | Number -> Integer` (`as_string` prefers string serialization)
 - `Number -> Integer` (strict behavior)
 - `String -> Integer` (string only format is currently not supported)

One way to model this is to decompose the base cases into strict fields (like `IntegerString` and `IntegerNumber`) that specialize in strict conversion. That would allow us to compose them into a more complex field with opinionated defaults (like `Integer`) without needing additional options to reduce it to one of the strict fields.

lafrech:
Which of those fields would allow loose input validation (accept any form of representation)?

deckar01:
In that example `Integer` would accept a number or a string (like it does currently) and under the hood it would be using `IntegerString` and `IntegerNumber`.

lafrech:
@deckar01:
> One way to model this is to decompose the base cases into strict fields (like IntegerString and IntegerNumber) that specialize in strict conversion. That would allow us to compose them into a more complex field with opinionated defaults (like Integer) without needing additional options to reduce it to one of the strict fields.

Looks like an important rework. Shouldn't we postpone this to after 3.0?

deckar01:
This bug fix looks good to me as is.

In the future it would be nice to have a more explicit API for declaring the accepted types. It's nice to have a convenient API for lenient parsing, but the field name should probably advertise that behavior so users don't get surprised.

lafrech:
I don't mind. We can merge this now so that strings are rejected when `strict` is `True`, and consider the fields refactor after 3.0 is out. @deckar01, would you like to open an issue about the refactor?

PR #756, opened 2018-03-21
Pass missing and default in deserialized form

### LINKED ISSUE #378: missing and default values should be specified in deserialized form

lafrech:
Closes #378

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -481,6 +481,25 @@ In the context of a web API, the ``dump_only`` and ``load_only`` parameters are 
         created_at = fields.DateTime(dump_only=True)
 
 
+Specify default Serialization/Deserialization values
+----------------------------------------------------
+
+Default values can be provided to a :class:`Field <fields.Field>` for both serialization and deserialization.
+
+`missing` is used for deserialization if the field is not found in the input data. Likewise, `default` is used for serialization if the input value is missing.
+
+.. code-block:: python
+
+    class UserSchema(Schema):
+        id = fields.UUID(missing=uuid.uuid1)
+        birthdate = fields.DateTime(default=dt.datetime(2017, 9, 29))
+
+    UserSchema().load({})
+    # {'id': UUID('337d946c-32cd-11e8-b475-0022192ed31b')}
+    UserSchema().dump({})
+    # {'birthdate': '2017-09-29T00:00:00+00:00'}
+
+
 Next Steps
 ----------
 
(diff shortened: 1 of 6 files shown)


lafrech:
TODO: update docs.

sloria:
Behavior looks good. This should be good to merge after docs are updated.

lafrech:
Rebased, added a note in "upgrading" and presented the feature in "quickstart".

docs/quickstart.rst
```diff
@@ -481,6 +481,25 @@ In the context of a web API, the ``dump_only`` and ``load_only`` parameters are
         created_at = fields.DateTime(dump_only=True)
 
 
+Specify default Serialization/Deserialization values
```
sloria, line 484: "Default" here should be capitalized. I'll fix this.

lafrech: Oh. I'm not sure what the rule is. Capitalize everything except common "little" words like "and", "as" "the", "a",...?

sloria: Correct. Headings are title cased.

taion:
This creates an interesting breaking change in behavior when both `required` and `missing` are specified. Before this change, `required` would essentially have no effect if `missing` was specified, as the default would get applied before field-level deserialization.

After this change, `missing` now has no effect if `required` is specified, as the `required` check is made before the `missing` logic is applied.

I think it would make more sense to restore the old behavior – as with `missing` implicitly overriding `allow_none`, it should also override `required`.

Issue #749, opened 2018-03-05
require_tld option of validate.URL is missing for fields.URL


DenisKuplyakov:
See subj. There is no way to pass `require_tld` to underlying validate when using `fields.URL`.

deckar01:
The validator was implemented in #674. It looks like there was originally some hesitation to add it to the field in #665 and it was closed for backwards compatibility reasons. I don't see any reason why it would be backwards incompatible. It should be safe to add another parameter before the kwargs.

We might be able to add it to v2. We can add it to v3. You will probably need to copy the url field implementation to a custom field with the extra argument supplied for now.

http://marshmallow.readthedocs.io/en/stable/custom_fields.html#custom-fields

lafrech:
Should `Email` have a `require_tld` parameter as well?

sloria:
Closed by #765 . @lafrech Let's defer on adding `require_tld` to Email for now, since there insn't a corresponding param for `validate.Email`.

PR #765, opened 2018-04-07
 Add require_tld argument to the Url field

mahenzon:
I saw the issue #749, but I cannot understand the problem out there with adding an optional argument `require_tld` to the Url field. It's already there in the URL validator initialiser so if we add it to the Url field's initialiser it will be only more convenient.

This fix helped me so I hope it can help other people

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -327,9 +327,11 @@ class Field(FieldABC):
             ret = ret.parent
         return ret if isinstance(ret, SchemaABC) else None
 
+
 class Raw(Field):
     """Field that applies no formatting or validation."""
     pass
+
 
 class Nested(Field):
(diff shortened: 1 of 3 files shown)


lafrech:
Looks good to me. Mind adding a regression test?

mahenzon:
@lafrech of course I don’t mind if you add any tests :)

sloria:
@surik00 He meant to suggest that you add a test for this. Would you like to do this?

mahenzon:
Sorry, @sloria, my bad. I've created some tests: https://github.com/marshmallow-code/marshmallow/pull/765/commits/f257ed61a48ea8fd0b6e94aa060fed50fdeaa85c

Is this what @lafrech asked for?

sloria:
@surik00 Yes, thanks for adding the tests.

sloria (approved):

PR #769, opened 2018-04-11
Remove unused **kwargs argument from BaseSchema.dump()

### LINKED ISSUE #729: dump(obj, many=None, update_fields=True, **kwargs) throws error when **kwargs included
I noticed the **kwargs in the argument list fo dump(). Here is the API documentation:

    dump(obj, many=None, update_fields=True, **kwargs)
        Serialize an object to native Python data types according to this Schema’s fields.

        Parameters:	
          obj – The object to serialize.
          many (bool) – Whether to serialize obj as a collection. If `None`, the value for self.many is used.
          update_fields (bool) – Whether to update the schema’s field classes. Typically set to `True`, 
            but may be `False` when serializing a homogenous collection. This parameter is used by `fields.Nested` to avoid multiple updates.
          Returns: A tuple of the form (`data`, `errors`)
          Return type: `MarshalResult, a collections.namedtuple`

The documentation doesn't mention what **kwargs are used for, and I thought that I might be able to get some data in to my @post_dump function by using it. So I started to play around with a simple example:

    dump(my_obj, many=None, update_fields=True, txn=1)

I received the following error:

    File "/home/apeters/git/donate_0/venv/lib/python2.7/site-packages/marshmallow/schema.py", line 513, in dump
        **kwargs
    TypeError: serialize() got an unexpected keyword argument 'txn'

Looking at schema.py at line 513:

    preresult = self._marshal(
                    processed_obj,
                    self.fields,
                    many=many,
                    # TODO: Remove self.__accessor__ in a later release
                    accessor=self.get_attribute or self.__accessor__,
                    dict_class=self.dict_class,
                    index_errors=self.opts.index_errors,
                    **kwargs
                )

Which takes us to marshalling.py. At line 96 we find the serialize() function that was throwing the error:

```
def serialize(self, obj, fields_dict, many=False,
                  accessor=None, dict_class=dict, index_errors=True, index=None)
```

I find that the serialize() does not have a **kwargs argument. Putting one in removes the error.

Two questions:

    - What is the purpose of **kwargs in dump()?
    - Should there be an extra argument in the serialize() function?

lafrech:
Closes #729

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -372,7 +372,7 @@ class BaseSchema(base.SchemaABC):
 
     ##### Serialization/Deserialization API #####
 
-    def dump(self, obj, many=None, update_fields=True, **kwargs):
+    def dump(self, obj, many=None, update_fields=True):
         """Serialize an object to native Python data types according to this
         Schema's fields.
 
@@ -424,8 +424,7 @@ class BaseSchema(base.SchemaABC):
                     many=many,
                     accessor=self.get_attribute,
                     dict_class=self.dict_class,
-                    index_errors=self.opts.index_errors,
-                    **kwargs
+                    index_errors=self.opts.index_errors
                 )
             except ValidationError as error:
                 errors = self._marshal.errors


sloria:
Thanks!

PR #771, opened 2018-04-12
Docs: BaseSchema docstring improvements

### LINKED ISSUE #535: How is deserialization affected by only and partial?
I was hoping to get some clarification on the differences between `only` and `partial` with respect to deserialization. (This question is similar to #509 which addresses serialization using the same fields.) Based on the docs and the response to that issue, I interpreted it to mean that `only` applies to serialization and `partial` applies to deserialization. However, in testing it looks like `only` also affects deserialization. What am I missing?

``` python

class MySchema(Schema):
    foo = fields.Field(required=True)
    bar = fields.Field(required=True)

only_schema = MySchema(only=('foo', ))
partial_schema = MySchema(partial=True)

print(only_schema.load({'foo': 42, 'bar': 24}))
print(only_schema.load({'foo': 42}))
print(partial_schema.load({'foo': 42, 'bar': 24}))
print(partial_schema.load({'foo': 42}))
```

Output:

``` python
UnmarshalResult(data={u'foo': 42}, errors={})
UnmarshalResult(data={u'foo': 42}, errors={})
UnmarshalResult(data={'foo': 42, 'bar': 24}, errors={})
UnmarshalResult(data={'foo': 42}, errors={})
```

I expected deserialization using `only_schema` to result in an error similar to the example below. 

``` python
MySchema().load({'foo':42})
```

Output:

``` python
UnmarshalResult(data={'foo': 42}, errors={'bar': [u'Missing data for required field.']})
```

Ultimately, I'm trying to get away with using the same schema for (de)serialization, but maybe that's not the ideal way to use marshmallow.

lafrech:
I think this closes #535.

I'm not 100% satisfied with this but I think it is better than the original.

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -231,19 +231,20 @@ class BaseSchema(base.SchemaABC):
         data, errors = schema.dump(album)
         data  # {'release_date': '1968-12-06', 'title': 'Beggars Banquet'}
 
-    :param tuple only: A list or tuple of fields to serialize. If `None`, all
-        fields will be serialized. Nested fields can be represented with dot delimiters.
-    :param tuple exclude: A list or tuple of fields to exclude from the
-        serialized result. Nested fields can be represented with dot delimiters.
+    :param tuple|list only: Whitelist of fields to select when instantiating the Schema.
+        If empty, all fields are used.
+        Nested fields can be represented with dot delimiters.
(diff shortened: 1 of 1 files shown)


lafrech:
Let's solve https://github.com/marshmallow-code/marshmallow/issues/772 first.

lafrech:
Reworked in https://github.com/marshmallow-code/marshmallow/pull/777.

PR #777, opened 2018-04-19
Fix #772 Empty Only Treated as None

### LINKED ISSUE #535: How is deserialization affected by only and partial?

### LINKED ISSUE #772: Empty Only Treated as None
When the `only` parameter is an empty list/set, it causes all the fields to be de/serialized like `None`. The implementation of 2.x and 3.0 are not in compliance with their respective docs.

> **only** (_tuple_) – A list or tuple of fields to serialize. If **None**, all fields will be serialized.

```py
class TestSchema(Schema):
    foo = fields.Field()
sch = TestSchema(only=())
data = dict(foo='bar')
result = sch.dump(data)
assert 'foo' not in result

# assert 'foo' not in {'foo': 'bar'}
```

This could create a security vulnerability if an application was dynamically generating the field set based on security role. A filter that was meant to hide everything would inadvertently show everything.

lafrech:
Implementation of the fix described in https://github.com/marshmallow-code/marshmallow/issues/772#issuecomment-380975300.

I also reworded the docstring a bit (this is a rework of https://github.com/marshmallow-code/marshmallow/pull/771 which hopefully should close https://github.com/marshmallow-code/marshmallow/issues/535). I'm open to suggestions for a better wording.

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -310,7 +310,7 @@ class BaseSchema(base.SchemaABC):
         """
         pass
 
-    def __init__(self, only=(), exclude=(), prefix='', many=False,
+    def __init__(self, only=None, exclude=(), prefix='', many=False,
                  context=None, load_only=(), dump_only=(), partial=False):
         # copy declared fields from metaclass
         self.declared_fields = copy.deepcopy(self._declared_fields)
@@ -624,7 +624,7 @@ class BaseSchema(base.SchemaABC):
 
     def _normalize_nested_options(self):
         """Apply then flatten nested schema options"""
-        if self.only:
+        if self.only is not None:
             # Apply the only option to nested fields.
             self.__apply_nested_option('only', self.only)
             # Remove the child field names from the only option.
@@ -657,7 +657,7 @@ class BaseSchema(base.SchemaABC):
 
     def _update_fields(self, obj=None, many=False):
         """Update fields based on the passed in object."""
-        if self.only:
+        if self.only is not None:
             # Return only fields specified in only option
             if self.opts.fields:
                 field_names = self.set_class(self.opts.fields) & self.set_class(self.only)
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -2,7 +2,7 @@
 """The :class:`Schema` class, including its metaclass and options (class Meta)."""
 from __future__ import absolute_import, unicode_literals
 
-from collections import defaultdict, Mapping
+from collections import defaultdict, Mapping, OrderedDict
 import copy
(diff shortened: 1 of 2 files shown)


deckar01:
Looks good to me. Thanks @lafrech!

@sloria Should this be released in the next patch of 2.x?

lafrech:
I don't know if we have a rule to express the `:param` types in the docstrings.

Since the text said "tuple or list" I figured I'd use `tuple|list` as type and remove this from the sentence.

But I might as well be listing all collection types... Maybe tuple was enough and the user can imagine a list will fit as well.

Later in the file, only tuple is specified:

```python
    def validate(self, data, many=None, partial=None):
        """Validate `data` against the schema, returning a dictionary of
        validation errors.
        [...]
        :param bool|tuple partial: Whether to ignore missing fields. If `None`,
            the value for `self.partial` is used. If its value is an iterable,
```

No big deal anyway. Just trying to get it right.

sloria:
Yes, this should be sent to the 2.x-line branch. 

> Since the text said "tuple or list" I figured I'd use tuple|list as type and remove this from the sentence.

Good idea.

@lafrech Would you mind resending this to `2.x-line`?

sloria:
Done in #782

Issue #783, opened 2018-04-25
Could I use a single schema instance on concurrent request?


yupeng0921:
For example, I have an AlbumSchema class, then I create a global variable 'g_schema' from it:
g_schema = AlbumSchema()
Could I use the same g_schema on the concurrent http requests? Is it safe?

lafrech:
Funny you ask because that very same day we've been facing issues that look quite like concurrency issues.

We use marshmallow + webargs and we pass webargs's `use_args` a `Schema` instance, so the instance is shared over all requests.

Some requests receive an error message that obviously corresponds to another request that occurred concurrently.

This does not occur when passing a schema class.

We're investigating this. The root cause could be the fact that the `Unmarshaller` is [created at init time](https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/schema.py#L332):

```python
        #: Callable unmarshalling object
        self._unmarshal = marshalling.Unmarshaller()
```

(same issue with `self._marshal` and perhaps other attributes)

The `Unmarshaller` instance is then shared among all requests and called several times during the loading process, which does not seem thread safe.

I've checked in webargs code and it is meant to receive either a class or an instance, so this looks like a bug.

If this is really the case, then it is a serious show-stopper.

I'm not sure how to isolate/reproduce/fix/test. Help very much welcome.

And if I'm wrong, well... great, what a relief, sorry for the noise.

lafrech:
From a quick analysis, I think the consequences are "limited" to spurious error messages.

Take the following with care, I have no certainty at this point:

- When all goes fine, data is not mixed up between two concurrent requests.

- When an error occurs, another concurrent requests can receive the error message meant for the other request.

- A wrong request can not pass due to another concurrent passing request.

So you'd have false positives, but no false negatives.

This would mitigate the issue a bit.

lafrech:
It should be noted that this happens only if two concurrent requests occur and one of them fails validation. In the case of a loaded server, as long as all requests are valid, there is no problem.

Also, we managed to see and reproduce this because we're using marshmallow to parse pretty fat inputs. In the general case of normal size inputs, the race condition is much less likely to occur.

taion:
Looking at the code, the fix in ~~#799~~ #785 is not quite sufficient. `Schema._update_fields` has some amount of shared state, and can cause issues with stomping on other updates if field types need to be inferred.

deckar01:
[...]

PR #785, opened 2018-04-26
Don't instantiate Marshaller and Unmarshaller in Schema.__init__()

lafrech:
Sharing those structures between concurrent dump/load raises concurrency issues (see https://github.com/marshmallow-code/marshmallow/issues/783).

-----------------------------------------------------------------

My colleague @davidfrederique identified the root cause for our concurrency issues. The `Marshaller` and `Unmarshaller` instances are instance attribute so they are shared among all requests sharing the same `Schema` instance.

This patch instantiates `Marshaller`/`Unmarshaller` on dump / load operations and ensures they are not reused for another dump/load.

Using this solves the problem we identified in our application.

I don't know how to test this with pytest. At least, no existing test is broken.

### DIFF
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -38,12 +38,6 @@ class ErrorStore(object):
         #: Dictionary of extra kwargs from user raised exception
         self.error_kwargs = {}
 
-    def reset_errors(self):
-        self.errors = {}
-        self.error_field_names = []
-        self.error_fields = []
-        self.error_kwargs = {}
-
     def get_errors(self, index=None):
         if index is not None:
             errors = self.errors.get(index, {})
@@ -113,9 +107,6 @@ class Marshaller(ErrorStore):
         .. versionchanged:: 1.0.0
             Renamed from ``marshal``.
         """
-        # Reset errors dict if not serializing a collection
-        if not self._pending:
-            self.reset_errors()
         if many and obj is not None:
             self._pending = True
             ret = [self.serialize(d, fields_dict, many=False,
@@ -228,9 +219,6 @@ class Unmarshaller(ErrorStore):
             serializing a collection, otherwise `None`.
         :return: A dictionary of the deserialized data.
         """
-        # Reset errors if not deserializing a collection
-        if not self._pending:
-            self.reset_errors()
         if many and data is not None:
             self._pending = True
             ret = [self.deserialize(d, fields_dict, many=False,
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -344,12 +344,6 @@ class BaseSchema(base.SchemaABC):
         self.partial = partial
         #: Dictionary mapping field_names -> :class:`Field` objects
         self.fields = self.dict_class()
-        #: Callable marshalling object
-        self._marshal = marshalling.Marshaller(
-            prefix=self.prefix
-        )
-        #: Callable unmarshalling object
-        self._unmarshal = marshalling.Unmarshaller()
         if extra:
             warnings.warn(
                 'The `extra` argument is deprecated. Use a post_dump '
@@ -470,6 +464,8 @@ class BaseSchema(base.SchemaABC):
 
         .. versionadded:: 1.0.0
         """
+        # Callable marshalling object
+        marshal = marshalling.Marshaller(prefix=self.prefix)
         errors = {}
         many = self.many if many is None else bool(many)
         if not many and utils.is_collection(obj) and not utils.is_keyed_tuple(obj):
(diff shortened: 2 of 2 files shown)


marshmallow/marshalling.py
```diff
@@ -113,9 +107,6 @@ def serialize(self, obj, fields_dict, many=False,
         .. versionchanged:: 1.0.0
             Renamed from ``marshal``.
         """
-        # Reset errors dict if not serializing a collection
-        if not self._pending:
-            self.reset_errors()
```
deckar01, line 118: > ```py
> class Marshaller(ErrorStore):
>     """Callable class responsible for serializing data and storing errors.
> ```

It seems odd to me that the `Marshaller` classes inherit from `ErrorStore`. Serializing data and storing errors seem like two separate responsibilities.

marshmallow/schema.py
```diff
@@ -513,7 +509,7 @@ def dump(self, obj, many=None, update_fields=True, **kwargs):
                     **kwargs
                 )
             except ValidationError as error:
-                errors = self._marshal.errors
+                errors = marshal.errors
```
deckar01, line 512: The marshaller passes its errors to the `messages` argument of `ValidationError()`, so we should be able to access it via `error.messages` instead of reaching back through the marshaller.

marshmallow/schema.py
```diff
@@ -533,10 +529,10 @@ def dump(self, obj, many=None, update_fields=True, **kwargs):
                 self.__error_handler__(errors, obj)
             exc = ValidationError(
                 errors,
-                field_names=self._marshal.error_field_names,
-                fields=self._marshal.error_fields,
+                field_names=marshal.error_field_names,
+                fields=marshal.error_fields,
```
deckar01, line 533: `error_field_names` and `error_fields` should also be available on the `ValidationError` instance.

marshmallow/schema.py
```diff
@@ -533,10 +529,10 @@ def dump(self, obj, many=None, update_fields=True, **kwargs):
...
                 data=obj,
-                **self._marshal.error_kwargs
+                **marshal.error_kwargs
```
deckar01, line 535: This is the only field that the marshaller isn't passing to the `ValidationError`. The `ValidationError` constructor will accept kwargs and expose them for convenient access through a `kwargs` attribute.

deckar01 (commented):
What do you think about having the `Marshaller` classes instantiate a new `ErrorStore` each time they are called, then just pull all of the data out of the `ValidationError`?

lafrech:
Thanks @deckar01 for the review. I tried to quickly apply suggested changes and had issues. It takes more work in schema.py than I expected. I don't have much time for this right now. You may send a PR based on mine, if you wish, and we'll close this one.

However, since 2.x is the stable branch, I think we'd better patch 2.x with minimal changes as a hotfix, merge into MA3, than rework this further in MA3.

deckar01:
I did not consider the `run_validator` method in `Unmarshaller` in my initial review. It needs to add to the error store after calling `deserialize`, but there doesn't seem to be a good way to manage the error store state outside of the unmarshaller instance.

This looks like this PR is the simplest change. Sorry for the confusion @lafrech.

lafrech:
@sloria, if you get the time, please consider this.

Current implementation has a thread-safety issue (#783) which results in clients receiving validation errors from a concurrent thread when using webargs, for instance.

This kind of things can be a pain to isolate and reproduce. We were lucky to find a case that would reproduce easily.

It is also a security issue. The fact that marshmallow fields don't write the value in the error message mitigates the issue, but we could imagine a schema with a custom validator returning `'"apssword" is not the correct password for account "user"'`, for instance.

This change is quite minimal. It shouldn't bring any issue.

sloria:
Will do. Apologies for the delay in getting to this. I'll hopefully get a chance to take a closer look at this after work.

taion:
It might be better to switch to composition over inheritance here. If you take a look at e.g. https://github.com/lyft/toasted-marshmallow/pull/16, the "no code in Marshmallow" approach to JITing the schemas uses the Marshaller as a natural point to inject the JIT logic.

However, this approach really wants something shared across sessions. It might be sufficient to indirect the construction of `Marshaller` and `Unmarshaller` to e.g. `create_marshaller` and `create_unmarshaller` methods, though.

taion:
That said, thanks for this great fix, @lafrech – we are doing the same concurrent request handling with the same schema, and I was scratching my head a bit wondering how what we were doing could be safe given how the marshaller worked, but I didn't see any actual issues and didn't follow up.

This will probably save us some trouble, I'm sure.

taion:
ref https://github.com/marshmallow-code/marshmallow/pull/629 as the issue that more directly integrates the JIT hooks, which would be a larger change

PR #789, opened 2018-05-02
Remove Number.serialize and check as_string in Number._serialize

### LINKED ISSUE #788: fields.List of fields.Integer doesn't work as expected with as_string
Currently this does not seem correct:

```
>>> o = type('Object', (), {'ids': [1,2,3]})()
>>> field = fields.List(fields.Integer(as_string=True))
>>> field.serialize('ids', o)
[1, 2, 3]
```

I was expecting `['1', '2', '3']`

It does work as expected for single values:

```
>>> o = type('Object', (), {'id': 1})()
>>> single_field = fields.Integer(as_string=True)
>>> single_field.serialize('id', o)
'1'
```

lafrech:
Fixes #788.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -629,20 +629,13 @@ class Number(Field):
         except (TypeError, ValueError):
             self.fail('invalid')
 
-    def serialize(self, attr, obj, accessor=None):
-        """Pulls the value for the given key from the object and returns the
-        serialized number representation. Return a string if `self.as_string=True`,
-        othewise return this field's `num_type`. Receives the same `args` and `kwargs`
-        as `Field`.
-        """
-        ret = Field.serialize(self, attr, obj, accessor=accessor)
-        return self._to_string(ret) if (self.as_string and ret not in (None, missing_)) else ret
-
     def _to_string(self, value):
         return str(value)
 
     def _serialize(self, value, attr, obj):
-        return self._validated(value)
+        """Return a string if `self.as_string=True`, otherwise return this field's `num_type`."""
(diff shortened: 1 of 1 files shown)


lafrech:
There is a design issue behind this. The fact that `List` calls `container._serialize` means that for each `field`, everything should happen in there and overriding `serialize` in a `field` is smelly.

Should we state this explicitly somewhere?

I don't think it makes sense to add a test for this specific use case, or we might as well add equivalent tests for all features of all fields (test that each feature works when used inside a `List` field).

Should we test that each field does not override `serialize`? This would be easy to do, but it sounds bad.

deckar01:
I think the underlying issue here is that `Field` is being given the responsibility of accessing schema level objects and plucking its own data out. That seems like it should be the responsibility of the schema or the marshaller. The only place that `serialize` is used internally is in `Marshaller.serialize`.

Maybe we should move the plucking+serializing logic to `marshaller.serialize_field`?

lafrech:
This makes sense. Same for `deserialize`.

This goes much further than this PR, though.

I'm open to rework suggestions on this PR but I think this is the right thing to do and it should be done on 2.x as bugifx. Hence https://github.com/marshmallow-code/marshmallow/pull/790.

Then in `dev` branch, we can try to implement the change you suggest for MA3. Would you like to open an issue about it?

deckar01:
I opened #799 to discuss refactoring serialize.

lafrech:
Applied to `2.x` (https://github.com/marshmallow-code/marshmallow/pull/789) then merged into `dev`.

PR #806, opened 2018-05-15
Clean up code for processors

taion:
I'm doing this as a preliminary change for https://github.com/marshmallow-code/marshmallow/issues/367. There was a bit of cruft and I picked some fairly unfortunate names for those functions when I first wrote this.


taion:
Scratch that, I'm going to drop this PR for now and do everything against `dev` first, per the contributing guidelines.

PR #807, opened 2018-05-15
Support @validates(field_obj)

### LINKED ISSUE #367: RFC: Improved API for registering field validator methods

taion:
Closes https://github.com/marshmallow-code/marshmallow/issues/367

### DIFF
--- a/tests/test_decorators.py
+++ b/tests/test_decorators.py
@@ -229,14 +229,26 @@ def test_pre_dump_is_invoked_before_implicit_field_generation():
 class ValidatesSchema(Schema):
     foo = fields.Int()
 
-    @validates('foo')
+    @validates(foo)
     def validate_foo(self, value):
         if value != 42:
             raise ValidationError('The answer to life the universe and everything.')
 
 class TestValidatesDecorator:
+    def test_validates_field_obj(self):
+        class VSchema(Schema):
+            s = fields.String()
 
-    def test_validates(self):
+            @validates(s)
+            def validate_string(self, data):
+                raise ValidationError('nope')
+
+        with pytest.raises(ValidationError) as excinfo:
+            VSchema().load({'s': 'bar'})
+
+        assert excinfo.value.messages == {'s': ['nope']}
+
+    def test_validates_field_name(self):
         class VSchema(Schema):
             s = fields.String()
 
@@ -254,7 +266,7 @@ class TestValidatesDecorator:
         class S1(Schema):
             s = fields.String(attribute='string_name')
 
-            @validates('s')
+            @validates(s)
             def validate_string(self, data):
                 raise ValidationError('nope')
         with pytest.raises(ValidationError) as excinfo:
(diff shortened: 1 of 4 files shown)


tests/test_decorators.py
```diff
@@ -316,11 +340,21 @@ def validate_bar(self, value):
         assert '"bar" field does not exist.' in str(excinfo)
 
     def test_precedence(self):
-        class Schema2(ValidatesSchema):
+        class Schema2Base(Schema):
+            foo = fields.Int()
+
+            @validates('foo')
```
taion, line 346: This is an interesting/tricky edge case.

I can't actually use the original `ValidatesSchema` here, because `@validates(foo)` tries to validate the `foo` field declared on `ValidatesSchema`.

However, that field gets overridden in `Schema2`, and thus can't be found when trying to validate it.

sloria: So if I'm understanding this correctly, `@validates('foo')` and `@validates(foo)` behave differently:

- `@validates(foo)` will always validate the `foo` field instance passed in
- `@validates('foo')` searches for `foo` according to the MRO.

So users would have to be careful when doing `@validates(foo)` on base schema classes because subclasses won't be able to override `foo`.

sloria (commented):
I like the name changes and refactoring done in cc69995...b822cb2. Perhaps we should get those merged into `dev` in separate PR so we can review the `@validates(field_obj)` code in isolation.

taion:
I've moved the internal refactoring code to https://github.com/marshmallow-code/marshmallow/pull/814.

tests/test_decorators.py
taion: yes. thinking about it some more, this is pretty weird DX-wise.

a different way to do this might be to bind the fields to their names in `SchemaMeta`.

that way, assuming the field has been bound to its name, then we can look up the name of the field from the field object, and not have to care about the MRO. it would also let us get rid of much of the extra code here around building the extra mapping.

the downside, then, is that this would limit the re-use of field objects. so for example, right now i believe this is okay:

```python
field = fields.Integer()

class FooSchema(Schema):
    foo = field

class BarSchema(Schema):
    bar = field
```

the straightforward way to do this (binding the field object to the assigned name in `SchemaMeta`) would prohibit the above, though it would still allow

```python
field = fields.Integer()

class FooSchema(Schema):
    foo = field

class BarSchema(Schema):
    foo = field
```

taion:
Rebased.

tests/test_decorators.py
taion: @sloria Any thoughts on the above?

sloria:
https://github.com/marshmallow-code/marshmallow/issues/367#issuecomment-530502790

Issue #809, opened 2018-05-15
Revamp handling of fields without declared field


taion:
This is a follow-up from https://github.com/marshmallow-code/marshmallow/issues/783#issuecomment-389027969.

Currently, there's a fair amount of complexity in updating `self.fields` on `Schema`.

The current logic is quite odd and has a number of edge cases. It looks a little bit like:

1. When dumping
    1. Check if the object type is in `_types_seen`
    2. If it's not in `_types_seen`, infer the field types for anything in e.g. `Schema.Meta.fields` but without a declared field from the types on the object to be dumped
    3. If the object is not of a mapping type, add its type to `_types_seen`
        1. Note that this includes `list`, which is what we get whenever dumping with `many=True`
2. When loading
    1. Just use whatever is in `self.fields` from the last run
3. For `fields.Nested`
    1. Only calculate `fields` once for every instance of `fields.Nested`

Assuming my understanding is correct, this leads to a number of weird things happening.

For example, if I try to dump an object of type `Foo`, then try to dump a `dict`, then try to dump another object of type `Foo`, the `fields` used will be the ones corresponding to the `dict`.

Likewise, if I keep trying to dump additional `dict`s, the fields well get recalculated every time. However, this won't happen if I try to repeatedly dump `list`s with `many=True`.

But when using a `fields.Nested`, the field inference will only happen once.

Additionally, this is all potentially non-thread-safe.

-----

This is all quite complicated. My proposal for a simpler option is to add a new `fields.Default` that has the relevant logic, and instead only infer the fields once at schema instantiation time.

Issue #778, opened 2018-04-19
RFC: Replace missing/default with default_load/default_dump/default


lafrech:
Quoting @deckar01 in https://github.com/marshmallow-code/marshmallow/issues/775#issuecomment-382130144:

> It does seem arbitrary for `default` to only be used for serialization. It isn't immediately obvious that `missing` is the deserialization complement. It would be nice if there was a little more symmetry here. Maybe something like `default_load`, `default_dump`, and `default` to set both at once.

taion:
I've definitely gotten tripped up by this when I started using Marshmallow.

However, I'm not sure making `default` do both is less surprising. I don't think I've ever _wanted_ to set both, and it seems like it'd be error-prone for the "obvious" flag to set a dump-time default, which is generally a fairly rare occurrence when dumping e.g. objects.

lafrech:
So should it be `default_load` and `default_dump` only?

If we do that we can keep and deprecate the old names:

- `missing` (deprecated) -> `load_default`
- `default`  (deprecated) -> `dump_default`

(I prefer `load_default` to `default_load`.)

sloria:
I'm lukewarm about this, only because I'm hesitant to add any more drastic backwards-incompatible changes to 3.0. Sure, the difference between `default` and `missing` isn't immediately obvious, but it's one of those things you learn just once.

tl;dr I'm open to something like `load_default` and `dump_default`, but I don't think it's worth delaying the 3.0 final release.

lafrech:
I agree. I'm eager to see marshmallow 3 out.

I just sent a PR with the change. If this makes it to marshmallow 3, fine. Otherwise, no problem.

I just realized I used `default_dump` while I preferred `dump_default`. Well done...

I can fix that with a find&replace. @sloria, @taion, @deckar01 feel free to suggest better names before I do so.

taion:
I think `dump_default` and `load_default` are good names. This default/missing distinction is something that every developer here, myself included, has gotten tripped up over. That said, emitting a `FutureWarning` and making this a deprecation is not much worse than what's here.

lafrech:
IMHO, there are so may changes every app will be broken when going from MA2 to MA3, so I don't mind hard changes (as opposed to deprecation warnings) if the fix is obvious (more or less find&replace). I don't know the good practices.

sloria:
[...]

PR #810, opened 2018-05-15
Revamp field type inference

taion:
Fixes https://github.com/marshmallow-code/marshmallow/issues/809


tests/test_schema.py
```diff
@@ -654,8 +654,8 @@ def test_only_in_init(SchemaClass, user):
     assert 'age' in s
 
 def test_invalid_only_param(user):
```
taion: The below demonstrates an actual change in behavior.

However, I think this is valid. If we do `only=('foo',)` and `foo` is actually a declared field, then we do not get any sort of exception. For example, doing:

```python
UserSchema(only=('name',)).dump({})
```

would not cause an exception.

Thus, it seems odd that exceptions only get raised when the field is not declared on the schema.

marshmallow/fields.py
```diff
@@ -1344,6 +1339,45 @@ def _deserialize(self, value, *args, **kwargs):
         return self.constant
 
 
+class Default(Field):
```
taion: `fields.Inferred` might be a better name

we could also define this in `schema.py`; this is really more of a private utility and should not be directly used

sloria: +1 for renaming to `fields.Inferred`. I don't feel strongly about where this goes, but it should be documented that this should be considered private. Perhaps add this to the docstring:

```
.. warning::

    This class is treated as private API.
    Users should not need to use this class directly.
```

marshmallow/fields.py
```diff
@@ -515,8 +510,8 @@ def get_value(self, obj, attr, accessor=None):
             return self.container.get_value(value, self.container.attribute)
         return value
 
-    def _add_to_schema(self, field_name, schema):
-        super(List, self)._add_to_schema(field_name, schema)
+    def _bind_to_schema(self, field_name, schema):
```
sloria, line 563: This is a better name. +1.

tests/test_schema.py
sloria: I'm not sure about this. `only` is typically used for filtering use cases. If an invalid field name is passed, an error is returned to the client. Wouldn't this break that?

sloria (commented):
I like the approach being taken here.

I'm slightly concerned about the [changed behavior with only](https://github.com/marshmallow-code/marshmallow/pull/810/files#r188339298), but I could be convinced it's the right thing to do.

This is a large enough change that I'd like at least one more set of eyes on this, so requesting a review from @lafrech or @deckar01 .

tests/test_schema.py
deckar01: I think this error provides important information to the developer. Having a typo silently produce truncated output could be hard to debug and easy to miss.

taion:
@sloria @deckar01 @lafrech 

The question of missing fields here is interesting and ties closely into https://github.com/marshmallow-code/marshmallow/issues/778.

The current behavior is sort of weird and inconsistent, though. Against `dev`, given:

```python
class Schema1(Schema):
    foo = fields.Integer()

class Schema2(Schema):
    class Meta:
        fields = ('foo',)
```

We get the following behavior:

```python
>>> Schema1().dump({})
{}
>>> Schema2().dump({})
{}
>>> Schema1().dump({'bar': 9})
{}
>>> Schema2().dump({'bar': 9})
KeyError: '"foo" is not a valid field for {\'bar\': 9}.'
>>> Schema1(only=('bar',)).dump({})
{}
>>> Schema1(only=('bar',)).dump({'foo': 9})
KeyError: '"bar" is not a valid field for {\'foo\': 9}.'
>>> Schema1(only=('bar',)).dump({'bar': 9})
{'bar': 9}
```

So... what should happen? I personally only use Marshmallow to dump SQLAlchemy instances, and I never actually run into cases where fields are missing on the objects I'm trying to dump (and I actually declare all my fields one way or another anyway), but...

The most straightforward thing to do is just to throw at schema instantiation time if there are any fields in `only` or `exclude` that aren't on the schema class.

I don't understand the breadth of use cases, but in my case, I would also be in favor of raising an exception by default on any missing fields at serialization time, rather than silently dropping them.

taion:
As a demonstration, I've implemented:
> The most straightforward thing to do is just to throw at schema instantiation time if there are any fields in `only` or `exclude` that aren't on the schema class.

in https://github.com/marshmallow-code/marshmallow/pull/810/commits/443aa4ea00ad3183c71011f87ef5497e91df4736 and the subsequent commits.

tests/test_schema.py
taion: With https://github.com/marshmallow-code/marshmallow/pull/810/commits/443aa4ea00ad3183c71011f87ef5497e91df4736, we now throw at schema instantiation time if a field in `only` or `exclude` is not present on the schema class.

taion:
Any thoughts on the new `only` behavior here?

lafrech:
I like this.

Do you mind rebasing? It would make things clearer about the change discussed in https://github.com/marshmallow-code/marshmallow/pull/810#discussion_r188159998. IIUC, the behaviour introduced by this PR about `only` is not a change anymore as it was changed in `dev` already.

marshmallow/fields.py
```diff
@@ -1355,6 +1350,45 @@ def _deserialize(self, value, *args, **kwargs):
...
+        if field_type is Field:
+            return super(Default, self)._serialize(value, attr, obj)
+
+        field = self._field
+        if type(field) is not field_type:
+            self._field = field = field_type()
+            field._bind_to_schema(self.name, self.parent)
+
+        # Use the local field rather than self._field for thread safety.
+        return field._serialize(value, attr, obj)
```
lafrech, line 1499: I don't understand the need for `self._field`. Is it just some sort of optimization: avoid binding to schema at every call to deserialize, caching the instantiated and bound field instead? If so, is it worth doing? Would it be worth going further and rather than caching only the field for the last type, use a `dict` to cache a field instance for each new type that is encountered?

Please forgive me if I am totally mistaken. I may have missed the obvious.

taion: It's to avoid instantiating a new field instance every time we try to use this field.

A dict would be more general, but the idea is that in practice the field type is unlikely to change, so it's a nice pragmatic trade-off between complexity and performance to just cache the last field used.

taion:
Okay, I see the changes here with `only` and `exclude` have indeed already been merged earlier.

This is now merged against `dev` again.

Please take another look.

marshmallow/fields.py
lafrech: OK, what I thought but I wasn't sure.

You've been investigating performance issues much more than I did and I have no idea of the penalty induced by instantiation.

The `dict` could also be filled at import time to avoid any penalty when serializing.

taion: What do you mean? The fields here are bound to the schema, so they can't be shared across schemas.

taion: Oh, I don't have performance numbers or anything, it just seemed like having a single-field memo would give us what we wanted in most cases, while not adding more than a few lines of code.

lafrech: > What do you mean? The fields here are bound to the schema, so they can't be shared across schemas.

Yeah, scratch that. Sorry.

> Oh, I don't have performance numbers or anything, it just seemed like having a single-field memo would give us what we wanted in most cases, while not adding more than a few lines of code.

Here's a mapping implementation (untested). I don't think it is worse than the implementation you propose, both in performance and readability:

```py
    def __init__(self):
        super(Inferred, self).__init__()
        self._field_cache = {}

        val_type = type(value)
        field_type = self.TYPE_MAPPING.get(val_type, None)
        if field_type is None:
            return super(Inferred, self)._serialize(value, attr, obj)

        field = self._field_cache.get(val_type)
        if field is None:
            self._field_cache[val_type] = field = field_type()
            field._bind_to_schema(self.name, self.parent)

        # Use the local field rather than self._field for thread safety.
        return field._serialize(value, attr, obj)
```

I also replaced `Field` with `None`, as `Field` is not used anyway.

Further rework (still untested):

```py
  def __init__(self):
        super(Inferred, self).__init__()
        self._field_cache = {}

        val_type = type(value)

        # Search field in cache
        field = self._field_cache.get(val_type)
        if field is None:
            # Cache miss, instantiate appropriate field
            field_type = self.TYPE_MAPPING.get(val_type, None)
            if field_type is None:
                field = super(Inferred, self)
            else:
                field = field_type()
                field._bind_to_schema(self.name, self.parent)
            # Cache field instance
            self._field_cache[val_type] = field

        # Use the local field rather than self._field for thread safety.
        return field._serialize(value, attr, obj)
```

taion: Sure, that works – the reason I didn't do it that way is because, in practice, you'd expect the type of the field to almost never change, so in general there wouldn't be much of a point to keeping around a map.

I can add a comment documenting that assumption if you'd like.

lafrech: I don't really mind either way.

When reviewing your code, I tried to understand (see my first question) and found it surprising. I think the implementation I propose is marginally superior from a functional point of view as it covers that corner case when the type would change, with no extra performance cost (or am I wrong?) and I find it easier to grasp at first glance, but that's just my POV.

Anyway, I don't want to nitpick. This is just about optimizing or not a corner case. What matters the most to me is to keep the code simple and easy to understand.

I'm fine with whatever you and the other reviewer(s) find best.

taion: Your example is actually not entirely safe against certain edge cases, incidentally.

In the event where the user dynamically generates new classes on-the-fly, the cache could potentially grow without bound and lead to excess memory usage. The correct caching key needs to be either the _field_ class, or else the cache needs to be a weakmap.

Granted, this is extremely unlikely to cause issues in practice, but I do think it's a point in favor of going for a simpler cache here.

taion:
Updated to add that implementation note.

marshmallow/fields.py
lafrech: Good catch.

Here's an alternative implementation that should address this (using the field class as key).

```py
    def __init__(self):
        super(Inferred, self).__init__()
        self._field_cache = {}

        field_cls = self.TYPE_MAPPING.get(type(value), None)
        if field_cls is None:
            field = super(Inferred, self)
        else:
            field = self._field_cache.get(field_cls, None)
            if field is None:
                field = field_cls()
                field._bind_to_schema(self.name, self.parent)
                self._field_cache[field_cls] = field

        # Use the local field rather than self._field for thread safety.
        return field._serialize(value, attr, obj)
```

Other minor differences with your implementation:

- use `None` instead of unused `Field`
- `_serialize(value, attr, obj)` appears only once at the end of the method

marshmallow/schema.py
```diff
@@ -726,120 +685,76 @@ def __apply_nested_option(self, option_name, field_names, set_operation):
...
             raise ValueError(message)
 
-        ret = self.__filter_fields(field_names, obj, many=many)
-        # Set parents
-        self.__set_field_attrs(ret)
-        self.fields = ret
-        return self.fields
+        fields_dict = self.dict_class()
+        for field_name in field_names:
+            field_obj = self.declared_fields.get(field_name, fields.Inferred())
```
deckar01, line 713: Defaulting to `Inferred` here should fix #921. :+1:

marshmallow/fields.py
deckar01: Benchmark: https://gist.github.com/deckar01/7643bb78a2dd04375fd2c16f90b9f052

10k iterations with python 3.7.0 dumping 3 inferred fields:

| (s) | dev | revamp-infer-fields | lafrech-cache |
|---|---|---|---|
| **static** | 1.01s | 0.316s | 0.331s |
| **changing** | 1.03s | 0.908s | 0.329s |

| (%) | dev | revamp-infer-fields | lafrech-cache |
|---|---|---|---|
| **static** | 100% | 31.3% | 32.8% |
| **changing** | 100% | 88.2% | 31.9% |

The speed up for dumping inferred fields is significant, so the optimization seems to justify the added complexity. Since the type cache @lafrech proposed optimizes both use cases without increasing the code complexity or penalizing the primary use case, it would advocate using it.

taion: I'm okay with either if you all feel strongly about this. At the margins I look at the non-dict based solution as being slightly simpler, but I'd be okay with using something dict-based as well.

taion:
Is there anything else here other than switching to a different caching scheme for the real fields?

deckar01:
I suspect there is some dead marshalling code now that `declared_fields` will always have field objects for all the fields. Since this change fixes a bug and any dead code would just be tech debt, I would recommend landing this and creating an issue to investigate the dead code in a separate pass.  See https://github.com/marshmallow-code/marshmallow/issues/920#issuecomment-414683394.

lafrech:
@taion Sorry for the delay. I just reviewed this and once the caching scheme switch is done, and after a rebase, it is good to merge.

Thank you for this work.

lafrech:
I just made the modif + merged dev branch.

One last check + update changelog and it should be good to merge.

taion:
Sorry, I've been swamped. I'll try to take a look in the next day or so.

lafrech:
No pb @taion.

I just added the changelog update.

This looks good to me. Thanks for taking another look if you get the time.

taion:
Okay, I can't hit "approve" since this is my PR, but this look good to me. Thanks for picking up the caching bit.

lafrech:
Merged, thanks!

cancan101:
I think this PR is breaking `marshmallow-sqlalchemy`. I am seeing this on marshmallow 3.0.0b18 (it worked on b12):
```
  File "/Users/alex/.pyenv/versions/my-app/lib/python3.7/site-packages/marshmallow/schema.py", line 110, in __new__
    dict_cls=dict_cls,
  File "/Users/alex/.pyenv/versions/my-app/lib/python3.7/site-packages/marshmallow_sqlalchemy/schema.py", line 60, in get_declared_fields
    declared_fields = mcs.get_fields(converter, opts, base_fields, dict_cls)
  File "/Users/alex/.pyenv/versions/my-app/lib/python3.7/site-packages/marshmallow_sqlalchemy/schema.py", line 94, in get_fields
    dict_cls=dict_cls,
  File "/Users/alex/.pyenv/versions/my-app/lib/python3.7/site-packages/marshmallow_sqlalchemy/convert.py", line 107, in fields_for_model
    field = base_fields.get(prop.key) or self.property2field(prop)
  File "/Users/alex/.pyenv/versions/my-app/lib/python3.7/site-packages/marshmallow_sqlalchemy/convert.py", line 127, in property2field
    field_class = field_class or self._get_field_class_for_property(prop)
  File "/Users/alex/.pyenv/versions/my-app/lib/python3.7/site-packages/marshmallow_sqlalchemy/convert.py", line 188, in _get_field_class_for_property
    field_cls = self._get_field_class_for_column(column)
  File "/Users/alex/.pyenv/versions/my-app/lib/python3.7/site-packages/marshmallow_sqlalchemy/convert.py", line 155, in _get_field_class_for_column
    return self._get_field_class_for_data_type(column.type)
  File "/Users/alex/.pyenv/versions/my-app/lib/python3.7/site-packages/marshmallow_sqlalchemy/convert.py", line 174, in _get_field_class_for_data_type
    if python_type in self.type_mapping:
  File "/Users/alex/.pyenv/versions/my-app/lib/python3.7/site-packages/marshmallow_sqlalchemy/convert.py", line 85, in type_mapping
    return self.schema_cls.TYPE_MAPPING
AttributeError: type object 'WidgetSchema' has no attribute 'TYPE_MAPPING'
```

lafrech:
Should we keep `TYPE_MAPPING` in the Schema and access it from `Inferred` or consider this is private and let marshmallow-sqlalchemy define its own?

IIUC, putting it in `Schema` allows users to extend it with their own types / custom fields associations. I don't use the inference feature but it sounds like a legit case.

taion:
This seems generic enough that it would make sense on `Schema`. I didn't realize this was part of the public API.

PR #811, opened 2018-05-15
Simplify utils.get_value

taion:
This is good for like a 10% speedup in the benchmark. It's not going to make a big difference, but it's a free win.

### DIFF
--- a/marshmallow/utils.py
+++ b/marshmallow/utils.py
@@ -16,7 +16,7 @@ from decimal import Decimal, ROUND_HALF_EVEN, Context, Inexact
 from email.utils import formatdate, parsedate
 from pprint import pprint as py_pprint
 
-from marshmallow.compat import basestring, binary_type, text_type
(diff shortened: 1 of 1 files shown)


marshmallow/utils.py
```diff
@@ -338,7 +338,7 @@ def _get_value_for_keys(obj, keys, default):
 def _get_value_for_key(obj, key, default):
     try:
         return obj[key]
-    except (TypeError, AttributeError):
```
sloria: Based on the failing Travis build, it looks like this might be necessary to support custom iterables that don't explicitly implement `__getitem__` in Python 2.

sloria (changes requested):
Once the failing Travis build is addressed, this should be good to merge.

taion:
updated and fixed – i didn't realize that python2 throws `AttributeError` when subscripting into un-subscriptable new-style class instances rather than `TypeError`

sloria (approved):

PR #814, opened 2018-05-15
Clean up code for schema hooks

taion:


taion:
These are just the refactorings from from https://github.com/marshmallow-code/marshmallow/pull/807.

sloria (approved):
Thanks for separating these into another PR. Looks good to me.

PR #816, opened 2018-05-16
Make UUIDs valid if input is string representation or 16 byte rep (continued)

sloria:
Same as #625 , with added tests.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -623,7 +623,10 @@ class UUID(String):
         if isinstance(value, uuid.UUID):
             return value
         try:
-            return uuid.UUID(value)
+            if isinstance(value, bytes) or len(value) == 16:
+                return uuid.UUID(bytes=value)
+            else:
+                return uuid.UUID(value)
         except (ValueError, AttributeError):
             self.fail('invalid_uuid')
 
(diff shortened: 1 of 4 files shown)


marshmallow/fields.py
```diff
@@ -623,7 +623,10 @@ def _validated(self, value):
         if isinstance(value, uuid.UUID):
             return value
         try:
-            return uuid.UUID(value)
+            if isinstance(value, bytes):
```
sloria: @JeffBerger I removed the length check here so that a `ValidationError` would be raised for input with improper length instead of an uncaught TypeError.

marshmallow/fields.py
```diff
@@ -623,8 +623,11 @@ def _validated(self, value):
         if isinstance(value, uuid.UUID):
             return value
         try:
-            return uuid.UUID(value)
-        except (ValueError, AttributeError):
+            if isinstance(value, bytes) and len(value) == 16:
```
sloria, line 592: Scratch my previous comment: I see now why you added the explicit length check--Python 2 `bytes` and `str` are the same type 😞 . So I just added exception handling for `TypeError` when an invalid input is passed to `uuid.UUID(value)` below.

PR #819, opened 2018-05-18
Consider None as return value from pre-processor rather than assume mutation

### LINKED ISSUE #347: Unenveloping None values doesn't work as expected

lafrech:
Fixes #347

TODO:

- [x] Update doc (incl. upgrading.rst)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,19 +6,25 @@ Changelog
 
 Features:
 
-* Clean up code for schema hooks (:issue:`814`). Thanks :user:`taion`.
-* Minor performance improvement from simplifying ``utils.get_value`` (:issue:`811`). Thanks again :user:`taion`.
-* Add ``require_tld`` argument to ``fields.URL`` (:issue:`749`). Thanks
+- Clean up code for schema hooks (:issue:`814`). Thanks :user:`taion`.
+- Minor performance improvement from simplifying ``utils.get_value`` (:issue:`811`). Thanks again :user:`taion`.
+- Add ``require_tld`` argument to ``fields.URL`` (:issue:`749`). Thanks
   :user:`DenerKup` for reporting and thanks :user:`surik00` for the PR.
-* ``fields.UUID`` deserializes ``bytes`` strings using ``UUID(bytes=b'...')`` (:issue:`625`).
+- ``fields.UUID`` deserializes ``bytes`` strings using ``UUID(bytes=b'...')`` (:issue:`625`).
   Thanks :user:`JeffBerger` for the suggestion and the PR.
 
 Bug fixes:
 
-* Fields nested within ``Dict`` correctly inherit context from their
+- Fields nested within ``Dict`` correctly inherit context from their
   parent schema (:issue:`820`). Thanks :user:`RosanneZe` for reporting
   and :user:`deckar01` for the PR.
-* Includes bug fix from 2.15.3.
+- Includes bug fix from 2.15.3.
+
+Other changes:
+
+- *Backwards-incompatible*: Pre/Post-processors must return modified data.
+  ``None`` does not mean data was mutated (:issue:`347`). Thanks
+  :user:`tdevelioglu` for reporting.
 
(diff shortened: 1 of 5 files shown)


tests/test_decorators.py
```diff
@@ -75,9 +78,44 @@ def decrement_value(self, item):
...
+        @pre_dump
+        def dump_none(self, item):
+            return None
+
+    schema = PostSchema()
+    assert schema.dump({'value': 3}) is None
+    assert schema.load({'value': 3}) is None
+    schema = PreSchema()
+    assert schema.dump({'value': 3}) == {}
+    with pytest.raises(ValidationError) as excinfo:
```
lafrech, line 111: I feel weird about this test.

Returning `None` from ~`pre_dump`~`pre_load` does not make much sense.

sloria: Do you mean `pre_load`? 

Agreed that it doesn't make sense for users to do that, but I think this is the expected behavior.

lafrech: Yes, `pre_load`, sorry.

The test could be catching a `ValidationError` for some other reason, but there is no way to narrow it without testing private interface, so we can keep it this way.

sloria (approved):
Looks good to me. Once docs are updated, this is good to merge.

lafrech:
Doc updated and PR rebased.

CHANGELOG.rst
```diff
@@ -6,19 +6,25 @@ Changelog
 
 Features:
 
-* Clean up code for schema hooks (:issue:`814`). Thanks :user:`taion`.
-* Minor performance improvement from simplifying ``utils.get_value`` (:issue:`811`). Thanks again :user:`taion`.
-* Add ``require_tld`` argument to ``fields.URL`` (:issue:`749`). Thanks
+- Clean up code for schema hooks (:issue:`814`). Thanks :user:`taion`.
```
sloria, line 9: Thanks for making these consistent.

docs/upgrading.rst
```diff
@@ -456,6 +456,14 @@ It is not possible to specify a different key for serialization and deserializat
 Also, when ``data_key`` is specified on a field, only ``data_key`` is checked in the input data. In marshmallow 2.x the field name is checked if ``load_from`` is missing from the input data.
 
 
+Pre/Post-processors must return modified data
+*********************************************
+
+In marshmallow 2.x, ``None`` returned by a pre or post-processor is interpreted as "the data was mutated". In marshmallow 3.x, the return value is considered as processed data even if it is ``None``.
+
+Processors that mutate the data should be updated to also return it.
+
```
sloria, line 465: Would be nice to have a code example of how to update existing code, but that doesn't have to block merging of this PR.

sloria (approved):
Looks good. Thanks a lot for updating the changelog and upgrading guide.

CHANGELOG.rst
```diff
@@ -6,19 +6,25 @@ Changelog
...
-* Fields nested within ``Dict`` correctly inherit context from their
+- Fields nested within ``Dict`` correctly inherit context from their
   parent schema (:issue:`820`). Thanks :user:`RosanneZe` for reporting
   and :user:`deckar01` for the PR.
-* Includes bug fix from 2.15.3.
+- Includes bug fix from 2.15.3.
+
+Other changes:
+
+- *Backwards-incompatible*: Pre/Post-processors must return modified data.
```
sloria, line 25: This should be in 3.0.0b12. I'll fix this.

lafrech: Right. I think it got screwed up during the rebase and I didn't notice.

PR #823, opened 2018-05-20
Pluck Field

### LINKED ISSUE #800: Nested single string 'only' behavior edge case
When using Nested fields, passing a single string field to `only` does not work as expected when applying a transform with `on_bind_field`

```python


def test_nested_only_on_bind_field():

    def to_camel_case(snake_str):
        components = snake_str.split('_')
        return components[0] + ''.join(x.title() for x in components[1:])

    class CamelCaseSchema(Schema):
            field_obj.data_key = to_camel_case(field_name)

    class UserRoleSchema(CamelCaseSchema):
        user_id = fields.Number()
        role_name = fields.String()

    class UserSchema(CamelCaseSchema):
        id = fields.Number()
        roles = fields.Nested(UserRoleSchema, many=True, only='role_name')

    user = {
        'id': 1,
        'roles': [
            {'user_id': 1, 'role_name': 'admin'},
            {'user_id': 1, 'role_name': 'test'},
        ]
    }

    serialized_user = UserSchema().dumps(user)

    assert serialized_user == {"id": 1, "roles": ["admin", "test"]}
```

This results in the following error:

```
/env/lib/python3.6/site-packages/marshmallow/schema.py:476: in dumps
    serialized = self.dump(obj, many=many, update_fields=update_fields)
/env/lib/python3.6/site-packages/marshmallow/schema.py:428: in dump
    **kwargs
/env/lib/python3.6/site-packages/marshmallow/marshalling.py:147: in serialize
    index=(index if index_errors else None)
/env/lib/python3.6/site-packages/marshmallow/marshalling.py:68: in call_and_store
    value = getter_func(data)
/env/lib/python3.6/site-packages/marshmallow/marshalling.py:141: in <lambda>
    getter = lambda d: field_obj.serialize(attr_name, d, accessor=accessor)
/env/lib/python3.6/site-packages/marshmallow/fields.py:250: in serialize
    return self._serialize(value, attr, obj)
/env/lib/python3.6/site-packages/marshmallow/fields.py:455: in _serialize
    return utils.pluck(ret, key=self.only)
/env/lib/python3.6/site-packages/marshmallow/utils.py:318: in pluck
    return [d[key] for d in dictlist]
KeyError: 'role_name'
```

This seems to be only affecting the single string case, as passing `only=['role_name']` works as expected.

deckar01:
Cherry pick #822 to `dev` and encapsulate `Nested`'s plucking logic in a subclass named `Pluck` with a more explicit interface.

Fixes #800

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -349,7 +349,7 @@ class Nested(Field):
 
         user = fields.Nested(UserSchema)
         user2 = fields.Nested('UserSchema')  # Equivalent to above
-        collaborators = fields.Nested(UserSchema, many=True, only='id')
+        collaborators = fields.Nested(UserSchema, many=True, only=('id',))
         parent = fields.Nested('self')
 
     When passing a `Schema <marshmallow.Schema>` instance as the first argument,
@@ -401,12 +401,6 @@ class Nested(Field):
             Renamed from `serializer` to `schema`
         """
         if not self.__schema:
-            # Ensure that only parameter is a tuple
-            if isinstance(self.only, basestring):
-                only = (self.only,)
-            else:
-                only = self.only
-
             # Inherit context from parent.
             context = getattr(self.parent, 'context', {})
             if isinstance(self.nested, SchemaABC):
@@ -416,7 +410,7 @@ class Nested(Field):
                     issubclass(self.nested, SchemaABC):
                 self.__schema = self.nested(
                     many=self.many,
-                    only=only, exclude=self.exclude, context=context,
+                    only=self.only, exclude=self.exclude, context=context,
                     load_only=self._nested_normalized_option('load_only'),
                     dump_only=self._nested_normalized_option('dump_only'),
                 )
@@ -424,7 +418,7 @@ class Nested(Field):
                 if self.nested == 'self':
                     parent_class = self.parent.__class__
                     self.__schema = parent_class(
-                        many=self.many, only=only,
+                        many=self.many, only=self.only,
                         exclude=self.exclude, context=context,
                         load_only=self._nested_normalized_option('load_only'),
                         dump_only=self._nested_normalized_option('dump_only'),
@@ -433,7 +427,7 @@ class Nested(Field):
                     schema_class = class_registry.get_class(self.nested)
                     self.__schema = schema_class(
                         many=self.many,
-                        only=only, exclude=self.exclude, context=context,
+                        only=self.only, exclude=self.exclude, context=context,
                         load_only=self._nested_normalized_option('load_only'),
                         dump_only=self._nested_normalized_option('dump_only'),
                     )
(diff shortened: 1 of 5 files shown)


marshmallow/fields.py
```diff
@@ -405,22 +399,22 @@ def schema(self):
...
                     parent_class = self.parent.__class__
                     self.__schema = parent_class(
-                        many=self.many, only=only,
+                        many=self.many, only=self.only,
                         exclude=self.exclude, context=context,
                         load_only=self._nested_normalized_option('load_only'),
                         dump_only=self._nested_normalized_option('dump_only'))
                 else:
                     schema_class = class_registry.get_class(self.nested)
                     self.__schema = schema_class(
```
lafrech: Do you think the schema instantiation could be factorized to avoid repeating the schema kwargs? Or maybe just declare the kwargs dict only once?

marshmallow/fields.py
```diff
@@ -445,35 +439,50 @@ def _serialize(self, nested_obj, attr, obj):
...
-                only_field = self.schema.fields[self.only]
-                key = ''.join([self.schema.prefix or '', only_field.data_key or self.only])
-                if self.many:
-                    return utils.pluck(ret, key=key)
-                else:
-                    return ret[key]
-        return ret
 
-    def _deserialize(self, value, attr, data):
+    def test_collection(self, value):
```
lafrech: Why not make this method private `_test_collection`?
Same with `load` -> `_load`.

lafrech: OK, I just realized you're using them in `Pluck`. Maybe that's the reason you're exposing them as public API. I don't think using them in a child class here means they need to be exposed publicly (or does it?), but maybe your point is to expose them for other potential child classes to use?

tests/test_schema.py
```diff
@@ -1913,6 +1887,48 @@ class ParentSchema(Schema):
...
+
+    @pytest.fixture
+    def user(self):
+        return User(name="Monty", age=81)
+
+    @pytest.fixture
+    def blog(self, user):
+        col1 = User(name="Mick", age=123)
+        col2 = User(name="Keith", age=456)
+        blog = Blog("Monty's blog", user=user, categories=["humor", "violence"],
```
lafrech: Why not write `user=User(name="Monty", age=81)` here rather than create a dedicated single-use `user` fixture?

marshmallow/fields.py
deckar01: Ya, they are exactly the same.

tests/test_schema.py
```diff
@@ -1913,6 +1887,48 @@ class ParentSchema(Schema):
...
+                         collaborators=[col1, col2])
+        return blog
+
+    def test_pluck(self, blog):
+        class FlatBlogSchema(Schema):
+            name = fields.String()
```
lafrech: This `name` field is never used. (Same in `FlatBlogSchema` below.)

tests/test_schema.py
deckar01: It was copy pasta from copying the setup fixtures from `TestNestedSchema`. Inlining it makes sense to me.

tests/test_schema.py
```diff
@@ -1971,8 +1987,8 @@ class Meta:
 
     def test_multiple_nested_self_fields(self, user):
```
lafrech: Rename as `test_multiple_pluck_self_field` and move away from `Nested` tests?

(Nitpicking: I think `rels = fields.Pluck('self', 'name', many=True, attribute='relatives')` fits in one line.)

lafrech:
Nothing really mandatory in my comments. This looks pretty good to me.

Thanks @deckar01 for tackling this.

lafrech:
When this is merged, we can add a test to raise `StringNotCollectionError` if `only` or `exclude` is passed to `Nested` as string, not list of strings.

See https://github.com/marshmallow-code/marshmallow/pull/896.

sloria:
Looks good to me as well. @deckar01 Can you please get this up to date? Once that's done, feel free to merge.

marshmallow/fields.py
deckar01: In C++ it would be considered a "protected" member, but it was really just needed to decompose `_deserialize` into reusable pieces. Prefixing it with `_` is probably fine since it doesn't need to be used outside subclasses.

marshmallow/fields.py
```diff
@@ -459,37 +443,52 @@ def _serialize(self, nested_obj, attr, obj):
...
             raise ValidationError(exc.messages, data=data, valid_data=exc.valid_data)
         return valid_data
 
+    def _deserialize(self, value, attr, data):
+        self._test_collection(value)
+        return self.load(value, data)
+
+
+class Pluck(Nested):
+    def __init__(self, nested, field_name, **kwargs):
```
deckar01: This needs docs.

deckar01:
I updated the existing docs, added docs for pluck, and added an entry to the upgrading docs. This could probably use a quick review.

The code changes were all small refactors with good test coverage, so another full code review probably isn't necessary. A quick sanity check should suffice.

lafrech:
Great!

LGTM.

Maybe rename `Nested.load` -> `Nested._load`.

sloria (approved):

Issue #636, opened 2017-05-30
only is not bound by declared and additional fields on serializaton


jan-23:
The only parameter is correctly bound to the fields that are defined in Meta.fields (#183):
```python
class MySchema(Schema):
    class Meta:
        fields = ('a', )

MySchema(only=('b', )).dump({'a': 1, 'b': 2}).data == {}
```

Strangely it is not bound to the declared and additional fields:
```python
class MySchema(Schema):
    a = fields.Str()
    class Meta:
        additional = ('b', )

MySchema(only=('c', )).dump({'a': 1, 'b': 2, 'c': 3}).data == {'c': 3}
```

It would be more consistent if the second example also returns an empty dict.

lafrech:
I can reproduce with latest 3.x revision.

sloria:
I was also able to reproduce this on the `2.x-line` branch.

sloria:
That said, it looks like the fix for this in #824 is backwards-incompatible, so this might need to go in 3.x without a backport. See the removed tests in #824. Instead of erroring upon serialization when an invalid field is passed to only, it passes silently.

deckar01:
I'm not sure passing silently is a desirable behavior. If a developer has a typo in one of their `only` parameters, it is useful to communicate this to them with an error instead of hoping they notice it and requiring them debug the source of the data loss. That is how it currently works as long as the "invalid" key doesn't happen to match some data that is being dumped.

I think the bug here is that the second example isn't errorring. There is nothing in the documentation to indicate that `only` parameters that are not in the schema should act like `additional` fields and pass through without validation. I think we should proactively validate the `only` values against the fields instead of relying on missing data to indirectly cause an error.

jeetjitsu:
I agree throwing an error is definitely better. But another test `test_only_bounded_by_fields` will start failing in #824.

```
def test_only_bounded_by_fields():
    class MySchema(Schema):

        class Meta:
            fields = ('foo', )

    sch = MySchema(only=('baz', ))
    assert sch.dump({'foo': 42}) == {}
```
I tried to minimize the number of tests I had to remove/change.

deckar01:
`fields` is part of the declared fields, so it or `additional` would need to be considered when validating `only`.

jeetjitsu:
@deckar01 sorry I couldn't quite get the point you're making.

[...]

PR #824, opened 2018-05-20
Fix GH#636:  only not bound by declared additional

jeetjitsu:
This is a fix to the issue raised [here](https://github.com/marshmallow-code/marshmallow/issues/636)

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -653,9 +653,6 @@ def test_only_in_init(SchemaClass, user):
     assert 'name' in s
     assert 'age' in s
 
-def test_invalid_only_param(user):
-    with pytest.raises(AttributeError):
-        UserSchema(only=("_invalid", "name")).dump(user)
 
 def test_can_serialize_uuid(serialized_user, user):
     assert serialized_user['uid'] == str(user.uid)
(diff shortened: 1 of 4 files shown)


tests/test_schema.py
```diff
@@ -653,9 +653,6 @@ def test_only_in_init(SchemaClass, user):
     assert 'name' in s
     assert 'age' in s
 
-def test_invalid_only_param(user):
```
sloria, line 656: Why was this test removed? Does this change introduce backwards-incompatible behavior?

jeetjitsu: Yes there is a minor change. The current behaviour of only is not consistent. To make it consistent I ignored all fields passed in with only which are not present on the schema. This affected two tests.

1. test_only_with_invalid_attribute
2. test_invalid_only_param

Both pass in a value for `only` which is not present in the schema and then expect AttributeError to be raised because those fieds are not present in the model being dumped. With these changes, those fields which are not present on the schema will be ignored, therefore no erros will be raise and the tests will fail.

For fields present on the Schema but not in the model(being dumped) the behavior remains the same.

deckar01:
I checked this branch out, added the changes discussed in #636, and opened #826. Hopefully this saves you from my scope creep @ikilledthecat. 😺

sloria:
This is continued in #826. Let's proceed with that PR.

PR #826, opened 2018-05-22
Validate Only and Exclude

### LINKED ISSUE #636: only is not bound by declared and additional fields on serializaton

deckar01:
Validate `only` and `exclude` against the declared fields to give developers feedback when they use invalid field names in the schema configuration.

Fixes #636


marshmallow/schema.py
```diff
@@ -706,10 +701,32 @@ def _update_fields(self, obj=None, many=False):
...
+
+        if invalid_fields:
+            invalid_names = ['"{}"'.format(name) for name in invalid_fields]
+            invalid_names = utils.iterable_to_string(invalid_names)
+            if len(invalid_fields) > 1:
+                message = '{0} are not valid fields for {1}.'
+            else:
+                message = '{0} is not a valid field for {1}.'
+            message = message.format(invalid_names, self.__class__.__name__)
+            raise AttributeError(message)
```
sloria: `AttributeError` doesn't feel like the correct error. From the [docs](https://docs.python.org/3/library/exceptions.html#AttributeError).

> Raised when an attribute reference (see Attribute references) or assignment fails.

I think `ValueError` is more appropriate.

marshmallow/utils.py
```diff
@@ -402,3 +402,16 @@ def get_func_args(func):
...
+    """Format an iterable as a human readable string suitable for a sentence.
+    """
+    values = list(values)
+    if len(values) > 2:
+        return '{}, and {}'.format(', '.join(values[:-1]), values[-1])
+    if len(values) == 2:
+        return ' and '.join(values)
+    if len(values) == 1:
+        return values[0]
+    return 'None'
```
sloria: Would it make more sense to raise an error instead of return `'None'` here? We *should*n't ever hit this, but it would be incorrect to get `'None is not a valid field for MySchema'`.

sloria (dismissed):
Behavior looks correct. Only suggestion is to change the error type.

marshmallow/schema.py
deckar01: That makes sense. I was thinking of fields as schema class attributes, but that isn't really how they are used.

marshmallow/utils.py
deckar01: As a generic utility it seemed like a logical representation of an empty list. It is currently impossible to hit this line from `Schema._update_fields()`.

Maybe we don't need to format the message as a single string though. Developers seem to be just fine with lists and dictionaries as the "message" value for validation exceptions. Any preference for the message structure?

marshmallow/schema.py
```diff
@@ -244,12 +244,13 @@ class Meta:
...
-    :param tuple|list only: Whitelist of fields to select when instantiating the Schema.
-        If None, all fields are used.
-        Nested fields can be represented with dot delimiters.
-    :param tuple|list exclude: Blacklist of fields to exclude when instantiating the Schema.
-        If a field appears in both `only` and `exclude`, it is not used.
-        Nested fields can be represented with dot delimiters.
+    :param tuple|list only: Whitelist of the declared fields to select when
+        instantiating the Schema. If None, all fields are used. Nested fields
+        can be represented with dot delimiters.
+    :param tuple|list exclude: Blacklist of of the declared fields to exclude
```
lafrech: > Blacklist of of the declared fields

"of" is repeated

marshmallow/schema.py
```diff
@@ -706,10 +701,32 @@ def _update_fields(self, obj=None, many=False):
...
         # If "exclude" option or param is specified, remove those fields
         excludes = set(self.opts.exclude) | set(self.exclude)
         if excludes:
             field_names = field_names - excludes
+            invalid_excludes = excludes - declared_fields
+            invalid_fields |= invalid_excludes
+
+        if invalid_fields:
+            invalid_names = ['"{}"'.format(name) for name in invalid_fields]
+            invalid_names = utils.iterable_to_string(invalid_names)
```
lafrech: (Pylint nazi inside)

Reusing variable name is considered bad practice (R0204 (redefined-variable-type)).

You could append _list or _str to one of them, or just go for the one-liner equivalent.

marshmallow/utils.py
lafrech: > Maybe we don't need to format the message as a single string though. Developers seem to be just fine with lists and dictionaries as the "message" value for validation exceptions. Any preference for the message structure?

Yep. I wouldn't bother creating a string.

```python
    if invalid_fields:
        raise AttributeError("Invalid fields for {0}: {1}".format(self.__class__.__name__, invalid_fields))
```

Validation errors may be exposed to end users (like when using webargs) or even parsed by API clients, so they should be consistent, but this is just meant to be seen by the developer after a typo, so don't overthink it.

sloria:
@deckar01 Would you mind updating the changelog and possibly the upgrading docs as well?

LMK if you don't have time right now; I can do it when I have some time this weekend or next week.

Issue #524, opened 2016-09-09
Add option to throw validation error on extra keys


tuukkamustonen:
I'm not sure if I'm missing something, but I would like to throw error if extra keys are provided, as in:

``` python
>>> class AlbumSchema(Schema):
...     title = fields.Str()

>>> AlbumSchema(strict=True).load({'extra': 2}, allow_extra=False)
Traceback (most recent call last):
...
marshmallow.exceptions.ValidationError: {'_schema': ["Extra arguments passed: ['extra']"]}
```

I've been using the following implementation (taken from https://github.com/sloria/webargs/issues/87#issuecomment-183949558):

``` python
class BaseSchema(Schema):

    @pre_load
    def validate_extra(self, in_data):
        if not isinstance(in_data, dict):
            return

        extra_args = [key for key in in_data.keys() if key not in self.fields]
        if extra_args:
            raise ValidationError('Extra arguments passed: {}'.format(extra_args))
```

I would expect this to be a common need, however, so could it be supported by the library out-of-the-box?

lafrech:
FYI, there's another example of how to do this in the docs:

http://marshmallow.readthedocs.io/en/latest/extending.html#validating-original-input-data

tuukkamustonen:
@lafrech Oh, right. That's practically the same, just using `@validates_schema` instead. I suppose that would indeed be a better decorator to use as the check throws `ValidationError`.

As it's in the docs, I assume maintainers have a negative stand on supporting this through a parameter?

lafrech:
Duplicate of https://github.com/marshmallow-code/marshmallow/issues/515.

sloria:
We do not currently have plans to add this feature. I recommend using the recipe in the docs for now. 

I am not completely opposed to adding this in the future, given a strong case and enough user support.

Closing this for now, but will reopen if there is need for further discussion.

barakalon:
+1 for adding built in support for this.

tuukkamustonen:
I now need to support this together with `many=True`.

With something like:

```python
class Sch(Schema):
    foo = fields.Int()

    @validates_schema(pass_original=True)
    def v(self, data, orig_data):
        ...
        raise ValidationError('Extra arguments passed.', ...)
```

With `Sch(many=True).load([{'foo': 1}, {'bar': 1}])`, `@validates_schema(pass_original=True)` gets passed the array.

[...]

PR #838, opened 2018-06-11
Handle unknown fields with EXCLUDE, INCLUDE or RAISE

ramnes:
This is a rework of #595 so that it supports the API discussed in #524.

By default, only the fields described in the schema are returned when data is deserialized.

This commit implements an `unknown` option, so that `Schema(unknown=ALLOW)` or `Schema().load(data, unknown=ALLOW)` permit users to receive the unknown fields from the data.

`Schema(unknown=RAISE)` or `Schema().load(data, unknown=RAISE)` will raise a `ValidationError` for each unknown field.

`Schema(unknown=IGNORE)` or `Schema().load(data, unknown=IGNORE)` keep the original behavior.

Edit: `ALLOW` will finally be `INCLUDE`, and `IGNORE` will be `EXCLUDE`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,9 +4,17 @@ Changelog
 3.0.0b12 (unreleased)
 +++++++++++++++++++++
 
+Features:
+
+- The behavior to apply when encountering unknown fields while deserializing can be controlled with the ``unknown`` option (:issue:`524`).
+  It makes it possible to either "include", "exclude" or "raise".
+
+Other changes:
+
 - *Backwards-incompatible*: Pre/Post-processors MUST return modified data.
   Returning ``None`` does not imply data were mutated (:issue:`347`). Thanks
   :user:`tdevelioglu` for reporting.
+
 
 3.0.0b11 (2018-05-20)
 +++++++++++++++++++++
--- a/docs/extending.rst
+++ b/docs/extending.rst
@@ -37,6 +37,36 @@ By default, pre- and post-processing methods receive one object/datum at a time,
 In cases where your pre- and post-processing methods need to receive the input collection  when ``many=True``, add ``pass_many=True`` to the method decorators. The method will receive the input data (which may be a single datum or a collection) and the boolean value of ``many``.
 
 
+Using Original Input Data
++++++++++++++++++++++++++
+
+By default, unspecified field names are ignored by the validator. If you want to use the original, unprocessed input, you can add ``pass_original=True`` to your call to `post_load <marshmallow.decorators.post_load>`.
+
+.. code-block:: python
+    :emphasize-lines: 7
+
+    from marshmallow import Schema, fields, validates_schema, ValidationError
+
+    class MySchema(Schema):
+        foo = fields.Int()
+        bar = fields.Int()
+
+        @post_load(pass_original=True)
+        def add_baz_to_bar(self, data, original_data):
+            baz = original_data.get('baz')
+            if baz:
+                data['bar'] = data['bar'] + baz
+            return data
+
+    schema = MySchema()
+    schema.load({'foo': 1, 'bar': 2, 'baz': 3})
+    # {'foo': 1, 'bar': 5}
+
+.. seealso::
+
+   The default behavior for unspecified fields can be controlled with the ``unknown`` option, see :ref:`Handling Unknown Fields <unknown>` for more information.
+
+
 Example: Enveloping
 +++++++++++++++++++
 
@@ -263,34 +293,6 @@ You can register schema-level validation functions for a :class:`Schema` using t
     except ValidationError as err:
         err.messages['_schema']
     # => ["field_a must be greater than field_b"]
-
-
-Validating Original Input Data
-++++++++++++++++++++++++++++++
-
-Normally, unspecified field names are ignored by the validator. If you would like access to the original, raw input (e.g. to fail validation if an unknown field name is sent), add ``pass_original=True`` to your call to `validates_schema <marshmallow.decorators.validates_schema>`.
-
-.. code-block:: python
-    :emphasize-lines: 7
-
-    from marshmallow import Schema, fields, validates_schema, ValidationError
-
-    class MySchema(Schema):
-        foo = fields.Int()
-        bar = fields.Int()
-
-        @validates_schema(pass_original=True)
-        def check_unknown_fields(self, data, original_data):
-            unknown = set(original_data) - set(self.fields)
-            if unknown:
-                raise ValidationError('Unknown field', unknown)
-
-    schema = MySchema()
-    try:
-        schema.load({'foo': 1, 'bar': 2, 'baz': 3, 'bu': 4})
-    except ValidationError as err:
-        err.messages
-    # {'baz': 'Unknown field', 'bu': 'Unknown field'}
 
 
 Storing Errors on Specific Fields
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -338,6 +338,45 @@ Or you can ignore missing fields entirely by setting ``partial=True``.
     # OR UserSchema(partial=True).load({'age': 42})
     result  # => ({'age': 42}, {})
 
+.. _unknown:
+
+Handling Unknown Fields
++++++++++++++++++++++++
+
+By default, :meth:`load <Schema.load>` will exclude any field that has not been defined in the schema.
+
+This behavior can be modified with the ``unknown`` option, which accepts three values:
+
+- `EXCLUDE <marshmallow.utils.EXCLUDE>`, which is the default
+- `INCLUDE <marshmallow.utils.INCLUDE>`, to accept and include the unknown fields inside the returned data
+- `RAISE <marshmallow.utils.RAISE>`, to raise a :exc:`ValidationError <marshmallow.exceptions.ValidationError>` if there is any unknown field
+
+There are several places where you can apply that option. You can define it in the *class Meta* of your :class:`Schema`:
+
+.. code-block:: python
+    :emphasize-lines: 2,3
+
+    class UserSchema(Schema):
+        class Meta:
+            unknown = INCLUDE
+
+You can apply it at instantiation time:
+
+.. code-block:: python
+
+    schema = UserSchema(unknown=INCLUDE)
+
+Or you can force it directly when loading the data:
+
+.. code-block:: python
+
+    UserSchema().load(data, unknown=INCLUDE)
+
+The ``unknown`` option value set in :meth:`load <Schema.load>` will always override the value applied at instantiation time, which itself will override the value defined in the *class Meta*.
+
+This order of precedence allows you to change the behavior of a schema for different contexts.
+
+
 Schema.validate
 +++++++++++++++
 
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -12,6 +12,7 @@ import decimal
 
 from marshmallow import validate, utils, class_registry
 from marshmallow.base import FieldABC, SchemaABC
+from marshmallow.utils import EXCLUDE
 from marshmallow.utils import missing as missing_
 from marshmallow.compat import text_type, basestring
 from marshmallow.exceptions import ValidationError
@@ -370,6 +371,8 @@ class Nested(Field):
         value will be returned as output instead of a dictionary.
         This parameter takes precedence over ``exclude``.
     :param bool many: Whether the field is a collection of objects.
+    :param unknown: Whether to exclude, include, or raise an error for unknown
+        fields in the data. Use `EXCLUDE`, `INCLUDE` or `RAISE`.
     :param kwargs: The same keyword arguments that :class:`Field` receives.
     """
 
@@ -382,6 +385,7 @@ class Nested(Field):
         self.only = only
         self.exclude = exclude
         self.many = kwargs.get('many', False)
+        self.unknown = kwargs.get('unknown', EXCLUDE)
         self.__schema = None  # Cached Schema instance
         self.__updated_fields = False
         super(Nested, self).__init__(default=default, **kwargs)
@@ -473,7 +477,7 @@ class Nested(Field):
             else:
                 value = {self.only: value}
         try:
-            valid_data = self.schema.load(value)
+            valid_data = self.schema.load(value, unknown=self.unknown)
         except ValidationError as exc:
             raise ValidationError(exc.messages, data=data, valid_data=exc.valid_data)
         return valid_data
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -10,7 +10,9 @@ and from primitive types.
 
 from __future__ import unicode_literals
 
-from marshmallow.utils import is_collection, missing, set_value
+from marshmallow.utils import (
+    EXCLUDE, INCLUDE, RAISE, is_collection, missing, set_value,
+)
 from marshmallow.compat import text_type, iteritems
 from marshmallow.exceptions import (
     ValidationError,
@@ -81,6 +83,7 @@ class ErrorStore(object):
             return self.store_error(field_name, error, field_obj, index)
         return value
 
+
 class Marshaller(ErrorStore):
     """Callable class responsible for serializing data and storing errors.
 
@@ -92,7 +95,8 @@ class Marshaller(ErrorStore):
         ErrorStore.__init__(self)
 
     def serialize(self, obj, fields_dict, many=False,
-                  accessor=None, dict_class=dict, index_errors=True, index=None):
+                  accessor=None, dict_class=dict, index_errors=True,
+                  index=None):
         """Takes raw data (a dict, list, or other object) and a dict of
         fields to output and serializes the data based on those fields.
 
@@ -161,6 +165,7 @@ class Marshaller(ErrorStore):
 # Key used for schema-level validation errors
 SCHEMA = '_schema'
 
+
 class Unmarshaller(ErrorStore):
     """Callable class responsible for deserializing data and storing errors.
 
@@ -207,7 +212,7 @@ class Unmarshaller(ErrorStore):
                     errors.setdefault(field_name, []).append(text_type(err))
 
     def deserialize(self, data, fields_dict, many=False, partial=False,
-            dict_class=dict, index_errors=True, index=None):
+            unknown=EXCLUDE, dict_class=dict, index_errors=True, index=None):
         """Deserialize ``data`` based on the schema defined by ``fields_dict``.
 
         :param dict data: The data to deserialize.
@@ -217,6 +222,8 @@ class Unmarshaller(ErrorStore):
         :param bool|tuple partial: Whether to ignore missing fields. If its
             value is an iterable, only missing fields listed in that iterable
             will be ignored.
+        :param unknown: Whether to exclude, include, or raise an error for unknown
+            fields in the data. Use `EXCLUDE`, `INCLUDE` or `RAISE`.
         :param type dict_class: Dictionary class used to construct the output.
         :param bool index_errors: Whether to store the index of invalid items in
             ``self.errors`` when ``many=True``.
@@ -227,8 +234,9 @@ class Unmarshaller(ErrorStore):
         if many and data is not None:
             self._pending = True
             ret = [self.deserialize(d, fields_dict, many=False,
-                        partial=partial, dict_class=dict_class,
-                        index=idx, index_errors=index_errors)
+                        partial=partial, unknown=unknown,
+                        dict_class=dict_class, index=idx,
+                        index_errors=index_errors)
                     for idx, d in enumerate(data)]
 
             self._pending = False
@@ -279,6 +287,21 @@ class Unmarshaller(ErrorStore):
             if value is not missing:
                 key = fields_dict[attr_name].attribute or attr_name
                 set_value(ret, key, value)
+
+        if unknown != EXCLUDE:
+            fields = {field_obj.data_key or field_name
+                      for field_name, field_obj in fields_dict.items()}
+            for key in set(data) - fields:
+                value = data[key]
+                if unknown == INCLUDE:
+                    set_value(ret, key, value)
+                elif unknown == RAISE:
+                    self.store_error(
+                        field_name=key,
+                        error=ValidationError('Unknown field.'),
+                        index=index
+                    )
+
         if self.errors and not self._pending:
             raise ValidationError(
                 self.errors,
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -46,6 +46,23 @@ class ErrorStore(object):
             errors = self.errors
         return errors
 
+    def store_error(self, field_name, error, field_obj=None, index=None):
+        self.error_kwargs.update(error.kwargs)
+        if field_obj is not None:
+            self.error_fields.append(field_obj)
+        self.error_field_names.append(field_name)
+        errors = self.get_errors(index=index)
+        # Warning: Mutation!
+        if isinstance(error.messages, dict):
+            errors[field_name] = error.messages
+        elif isinstance(errors.get(field_name), dict):
+            errors[field_name].setdefault(FIELD, []).extend(error.messages)
+        else:
+            errors.setdefault(field_name, []).extend(error.messages)
+        # When a Nested field fails validation, the marshalled data is stored
+        # on the ValidationError's valid_data attribute
+        return error.valid_data or missing
+
     def call_and_store(self, getter_func, data, field_name, field_obj, index=None):
         """Call ``getter_func`` with ``data`` as its argument, and store any `ValidationErrors`.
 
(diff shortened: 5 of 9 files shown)


marshmallow/marshalling.py
```diff
@@ -275,6 +283,22 @@ def deserialize(self, data, fields_dict, many=False, partial=False,
...
+                    set_value(ret, key, value)
+                elif unknown == RAISE:
+                    def _raise(_):
+                        raise ValidationError('Unkown key.')
+                    self.call_and_store(
+                        getter_func=_raise,
+                        data=value,
+                        field_name=key,
+                        field_obj=type(value)
+                    )
```
ramnes: I'm not very happy that I used `call_and_store` in a hackish way, but I could not find a better solution to achieve the same thing.

marshmallow/fields.py
```diff
@@ -365,7 +366,8 @@ class Nested(Field):
         will be marshalled. If a field name (string) is given, only a single
         value will be returned as output instead of a dictionary.
         This parameter takes precedence over ``exclude``.
-    :param bool many: Whether the field is a collection of objects.
```
lafrech: Did you mean to remove this line?

marshmallow/utils.py
```diff
@@ -18,6 +18,9 @@
 
 from marshmallow.compat import binary_type, text_type
 
+ALLOW = "allow"
```
lafrech: It would be simpler to use plain strings, like `'allow'`, `'ignore'` and `'raise'`, rather than having to import those globals. I don't see the benefit of the globals.

marshmallow/marshalling.py
lafrech: Maybe the logic that stores the error can be extracted in a `store_error` method and `call_and_store` can be changed call `store_error` upon `ValidationError`.

```python
    def call_and_store(self, getter_func, data, field_name, field_obj, index=None):
        try:
            return getter_func(data)
            self.store_error(...)

    def store_error(...):
            self.error_kwargs.update(err.kwargs)
            self.error_fields.append(field_obj)
            self.error_field_names.append(field_name)
            # Warning: Mutation!
            if isinstance(err.messages, dict):
                errors[field_name] = err.messages
            elif isinstance(errors.get(field_name), dict):
                errors[field_name].setdefault(FIELD, []).extend(err.messages)
            else:
                errors.setdefault(field_name, []).extend(err.messages)
            # When a Nested field fails validation, the marshalled data is stored
            # on the ValidationError's valid_data attribute
            value = err.valid_data or missing
```

Then you can call `store_error` directly. And at this point, we could be tempted to remove `call_and_store` completely.

Also, I'm not sure about `field_obj=type(value)`. We might as well pass `None` or some arbitrary constant. And `store_error` could be modified to not append to `error_fields` if `None`, for instance.

Question is, what do we want to store in the `ValidationError` `fields` and `field_names`? Currently, we only have field errors or schema errors. Error for unknown keys are not really schema errors, but they don't correspond to a field either. Passing the data key as field name makes sense but I don't know what to pass as field object.

I don't think I ever used `field_names` and `fields` anyway. I've always been using `messages`. So I'm not sure what to expect from them.

If we decide not to append to `fields` and `field_names` for unknown key error, then user can't rely on their length being zero to assert there was no error. (But who would do that anyway?)

@sloria, @deckar01, thoughts?

tests/test_deserialization.py
```diff
@@ -1315,6 +1315,39 @@ class MySchema(Schema):
...
+
+        schema = MySchema(unknown=ALLOW, many=True)
+        data = schema.load([{'foo': 3, 'bar': 5}])
+        assert 'foo' in data[0]
+        assert 'bar' in data[0]
+
+        with pytest.raises(ValidationError) as excinfo:
+            MySchema(unknown=RAISE).load({'foo': 3, 'bar': 5})
+        assert 'bar' in str(excinfo)
+
```
sloria: The code isn't handling list deserialization. The following test fails:

```python
            MySchema(unknown=RAISE, many=True).load([
                {'foo': 'abc'},
                {'foo': 3, 'bar': 5},
            ])
        err = excinfo.value
        assert 0 in err.messages
        assert 'foo' in err.messages[0]
        assert err.messages[0]['foo'] == ['Not a valid integer.']
        assert 1 in err.messages
        assert 'bar' in err.messages[1]
        assert err.messages[1]['bar'] == ['Unknown key.']
```

```
|| ______________________________________________________________________ TestSchemaDeserialization.test_unknown_fields_deserialization _______________________________________________________________________
|| tests/test_deserialization.py:1360: in test_unknown_fields_deserialization
||     assert 1 in err.messages
|| E   AssertionError: assert 1 in {0: {'foo': ['Not a valid integer.']}, 'bar': ['Unkown key.']}
|| E    +  where {0: {'foo': ['Not a valid integer.']}, 'bar': ['Unkown key.']} = ValidationError({0: {'foo': ['Not a valid integer.']}, 'bar': ['Unkown key.']},).messages
|| ========================================================================================= 1 failed in 0.11 seconds =========================================================================================
```

tests/test_deserialization.py
```diff
@@ -1315,6 +1315,39 @@ class MySchema(Schema):
...
+        assert 'foo' in str(excinfo)
+
+        schema = MySchema(unknown=ALLOW, many=True)
+        data = schema.load([{'foo': 3, 'bar': 5}])
+        assert 'foo' in data[0]
+        assert 'bar' in data[0]
+
+        with pytest.raises(ValidationError) as excinfo:
+            MySchema(unknown=RAISE).load({'foo': 3, 'bar': 5})
+        assert 'bar' in str(excinfo)
```
sloria: This can be more specific:

```python
err = excinfo.value
assert 'bar' in err.messages
assert err.messages['bar'] == ['Unknown key.']
```

marshmallow/marshalling.py
```diff
@@ -275,6 +283,22 @@ def deserialize(self, data, fields_dict, many=False, partial=False,
...
                 set_value(ret, key, value)
+
+        if unknown != IGNORE:
+            for key in set(data) - set(fields_dict):
+                value = data[key]
+                if unknown == ALLOW:
+                    set_value(ret, key, value)
+                elif unknown == RAISE:
+                    def _raise(_):
+                        raise ValidationError('Unkown key.')
```
sloria: Typo here: "Unknown". Also, we might want to make the error message `'Unknown field.'` to make it more language-agnostic.

marshmallow/schema.py
```diff
@@ -517,6 +526,10 @@ def loads(self, json_data, many=None, partial=None, **kwargs):
             the value for `self.partial` is used. If its value is an iterable,
             only missing fields listed in that iterable will be ignored.
         :return: A dict of deserialized data
+        :param unknown: Whether to include, ignore or alert for unknown
```
sloria: Minor revision suggestion:

```
Whether to include, ignore, or raise an error for unknown...
```

marshmallow/schema.py
```diff
@@ -486,7 +490,7 @@ def dumps(self, obj, many=None, update_fields=True, *args, **kwargs):
         serialized = self.dump(obj, many=many, update_fields=update_fields)
         return self.opts.render_module.dumps(serialized, *args, **kwargs)
 
-    def load(self, data, many=None, partial=None):
+    def load(self, data, many=None, partial=None, unknown=None):
```
sloria, line 496: Is it necessary/useful to add `unknown` as a parameter to `load`, or will the constructor parameter suffice?

sloria (changes requested):
Initial pass finished.

Thanks for taking this on!

marshmallow/marshalling.py
sloria: Hm, maybe we should factor out the error storage code into its own method, so this could look something like:

```python
self.store_error(
    'Unknown field.',
    data=value,
    field_name=key,
)
```

marshmallow/marshalling.py
```diff
@@ -275,6 +283,22 @@ def deserialize(self, data, fields_dict, many=False, partial=False,
...
+                    self.call_and_store(
+                        getter_func=_raise,
+                        data=value,
+                        field_name=key,
+                        field_obj=type(value)
```
sloria: `field_obj` is supposed to be the `Field` instance associated with the invalid key. In this case, there isn't a `field_obj`, so `call_and_store` should probably be modified to allow `field_obj` to be omitted.

ramnes:
Thanks a lot for your fast review, it answered most of the questions I had. I'll iterate later today, and the result should be much better than this.

lafrech (commented):
I never used that review feature before. I entered a few review comments last night but did not submit correctly. Here they are. Some of them overlap with @sloria's.

sloria:
> Question is, what do we want to store in the ValidationError fields and field_names? Currently, we only have field errors or schema errors. Error for unknown keys are not really schema errors, but they don't correspond to a field either. Passing the data key as field name makes sense but I don't know what to pass as field object.

My initial thought was to append the field name but not append the field object, but that might not be correct. Since we're saying the fields are "unknown", they probably shouldn't be stored. Also, `field_names` and `fields` should probably be kept the same length to make indexing into them easier.

> If we decide not to append to fields and field_names for unknown key error, then user can't rely on their length being zero to assert there was no error. (But who would do that anyway?)

I don't think that's a case we should handle. The user gets an exception when there are errors, so there's no need to check the length of `fields` or `field_names`.

marshmallow/marshalling.py
ramnes: fixed

tests/test_deserialization.py
ramnes: fixed, thanks

marshmallow/fields.py
ramnes: Nop, thanks! I think I've deleted this during the rebase of my old PR on the current HEAD of dev, which wasn't that easy. It's back now.

marshmallow/schema.py
ramnes: done

marshmallow/utils.py
ramnes: @sloria what's your opinion on this? I was just following your proposition there, and I feel it's a great one, but honestly I'm good with both solutions.

ramnes: Also, note to myself: if we keep it that way, I should replace double quotes by single quotes there.

marshmallow/marshalling.py
ramnes: Thank you guys, I was a bit to shy to change this but that's indeed a better way of doing it.

ramnes:
Also, ignore or raise by default?

ramnes:
And in the end, I think that it would be great to have this option available in `Meta` too, but still overridable at instanciation and/or at load time. That would give a lot of flexibility, and thus things like this:

```python
from marshmallow import ALLOW, RAISE
from marshmallow import Schema as BaseSchema


# define a default class because in our application
# we want to raise an error by default
class Schema(BaseSchema):
    class Meta:
        unknown = RAISE


# define some schemas
class FooSchema(Schema):
    ...


def some_context(data, ...):
    schema = FooSchema()
    if ...:
        # oh wait, for this schema in that particular
        # context, we want to allow unknown keys
        schema.load(data, unknown=ALLOW)
```

marshmallow/schema.py
ramnes: I've done it just like `partial` (and `many`), which I think are doing it the right way. I'd say that the more flex you can get, the better it is.

marshmallow/marshalling.py
ramnes: Done. What do you both think of it that way?

tests/test_deserialization.py
ramnes: Fixed, thanks.

ramnes:
All requested changes are done.

ramnes:
> My initial thought was to append the field name but not append the field object, but that might not be correct. Since we're saying the fields are "unknown", they probably shouldn't be stored. Also, field_names and fields should probably be kept the same length to make indexing into them easier.

I'm just realizing that the naive method you're describing here at first is how I've done it. If the `field_names` aren't stored for unknown fields, how is the user supposed to know that there's a problem with the data?

lafrech:
In `'raise'` mode, an `Exception` is raised and the faulty keys appear in `'messages'`. In `'allow'` or `'ignore'` mode, there is no problem.

marshmallow/utils.py
sloria: @lafrech Using constants prevents typos from passing silently, e.g. `MySchema(unknown='rais')`.

marshmallow/schema.py
```diff
@@ -517,6 +526,10 @@ def loads(self, json_data, many=None, partial=None, **kwargs):
             the value for `self.partial` is used. If its value is an iterable,
             only missing fields listed in that iterable will be ignored.
         :return: A dict of deserialized data
+        :param unknown: Whether to include, ignore, or raise an error for unknown
+            fields in the data. Use `ALLOW`, `IGNORE` or `RAISE`.
+            If `None`, the value for `self.unknown` is used.
+        :return: A tuple of the form (``data``, ``errors``)
```
sloria: This line should no longer be here.

sloria:
Looking into the `fields` and `field_names` attributes further, I don't think they're actually necessary, so I've opened #840 to propose removing them. That said, that issue doesn't need to block merging this PR, and I think the approach you've taken (adding the unknown fields to `field_names`) is fine.

I'll look at the rest of the PR more closely when I get a chance.

marshmallow/utils.py
lafrech: The point was to avoid having to import the constant. But I don't mind. In most use cases, this will be setup once in a base `Schema` class for the whole application.

marshmallow/schema.py
ramnes: Woops, another rebase error, sorry about this. It's fixed, thanks.

ramnes:
I've just added a commit so that `unknown` is also available in `Schema.Meta`.

@sloria how quick do you see this in the tree? We really want to take advantage of this at @numberly, and we have a technological iteration sprint on our applications coming in a week or so. I'd love to see this feature to be merged meanwhile so that we could add it to our sprint checklist; do you think that's feasible?

sloria:
@ramnes Thanks for adding the `Meta` option. 👍 

I can't promise that we'll get this in by the time you need it (we--the maintainers--are volunteering free time, without a set schedule). That said, this is looking good so far; I could see it landing soon. I'll cut a release after this is merged so that you can start using it right away.

sloria:
There are a few housekeeping things that need to be done:

- [ ] Replace the existing example in the docs: https://marshmallow.readthedocs.io/en/latest/extending.html#validating-original-input-data
- [ ] Add a new blurb in the upgrading guide about this feature

@ramnes Feel free to do any of these if you have time.

tests/test_deserialization.py
```diff
@@ -1315,6 +1315,77 @@ class MySchema(Schema):
...
+                {'foo': 3, 'bar': 5},
+            ])
+        err = excinfo.value
+        assert 0 in err.messages
+        assert 'foo' in err.messages[0]
+        assert err.messages[0]['foo'] == ['Not a valid integer.']
+        assert 1 in err.messages
+        assert 'bar' in err.messages[1]
+        assert err.messages[1]['bar'] == ['Unknown field.']
+
```
sloria, line 1367: Can you also add a test that `unknown` behaves correctly when `data_key` is set for a field?

ramnes:
Excellent, I'll do those changes, thank you for writing them down.

And by the way, you really don't need to justify yourself, I know how open source works; huge props to you for that voluntary work and for the quick answers!

ramnes:
Reading `upgrading.rst`, I feel that it's mostly the place where breaking changes are identified, and I don't feel that we're adding one here; unless we choose to raise by default, obviously. Should I modify `CHANGELOG.rst` first and foremost to explain the feature as a whole?

sloria:
Yeah, go ahead and start with the changelog. We can update the upgrading as necessary later.

ramnes:
There you go. I've added some documentation in the quickstart, moved and modifed things a bit in the `extending.rst` so that it's more logical given this modification, and added an entry to the changelog.

sloria:
Thanks for updating the docs.

Looks like we have a [unanimous vote](https://github.com/marshmallow-code/marshmallow/issues/524#issuecomment-397165731) for `RAISE` as the default. Can you change that here? Also, can you rename `ALLOW` -> `PASS` and `IGNORE` -> `DROP`, per https://github.com/marshmallow-code/marshmallow/issues/524#issuecomment-397413038 ?

ramnes:
Rename done. :+1: 

As for making `RAISE` the default, it's a bit trickier: it puts 21 tests in error… I feel it would overload this PR if I make all the changes required for this API change. Do you think we could make it in a second time, in another PR?

ramnes:
Also, does it still make sens to have the `pass_original` option if we raise by default? :thinking:

lafrech:
> I feel it would overload this PR if I make all the changes required for this API change. Do you think we could make it in a second time, in another PR?

Yes, absolutely (IMHO).

sloria:
Agreed. We can defer changing the default to `RAISE`.

As for `pass_original`: I think it's still needed. Users may need to do processing based on the original data, regardless of whether extra keys were passed.

tests/test_deserialization.py
ramnes: Well seen, it indeed doesn't work properly. I'm digging on it.

ramnes: Fixed. :+1:

tests/test_deserialization.py
```diff
@@ -1387,6 +1387,23 @@ class Meta:
...
+        class MySchema(Schema):
+            foo = fields.Integer(data_key="Foo")
+
+        data = MySchema().load({"Foo": 1})
+        assert data["foo"] == 1
+        assert "Foo" not in data
+
+        data = MySchema(unknown=RAISE).load({"Foo": 1})
+        assert data["foo"] == 1
+        assert "Foo" not in data
```
sloria, line 1399: Minor: Can you also add an assertion that `foo` is considered an unknown field and `MySchema(unknown=RAISE).load({"foo": 1})` should therefore raise an error?

nicktimko:
How about `INCLUDE` instead of `PASS`?  I'll stop bikeshedding...

docs/quickstart.rst
```diff
@@ -338,6 +338,45 @@ Or you can ignore missing fields entirely by setting ``partial=True``.
...
     result  # => ({'age': 42}, {})
 
+.. _unknown:
+
+Handling Unknown Fields
++++++++++++++++++++++++
+
+By default, :meth:`load <Schema.load>` will drop any field that has not defined in the schema.
+
+This behavior can be modified with the ``unknown`` option, which accept three values:
```
trubesv: accept**s***

docs/quickstart.rst
```diff
@@ -338,6 +338,45 @@ Or you can ignore missing fields entirely by setting ``partial=True``.
     # OR UserSchema(partial=True).load({'age': 42})
     result  # => ({'age': 42}, {})
 
+.. _unknown:
+
+Handling Unknown Fields
++++++++++++++++++++++++
+
+By default, :meth:`load <Schema.load>` will drop any field that has not defined in the schema.
```
trubesv: has not been*

sloria:
> How about INCLUDE instead of PASS? I'll stop bikeshedding...

I was actually thinking the same thing: INCLUDE might be clearer than PASS. Not a big deal, though, and it's certainly not a blocker to getting this merged.

ramnes:
Want me to rename to `INCLUDE`, `EXCLUDE` and `RAISE`? I actually hesitated to propose that, because I didn't want to nitpick on the naming, but since we're here...

sloria:
Sure. @lafrech feel free to chime in, but I think INCLUDE, EXCLUDE and RAISE are the clearest names proposed thus far.

lafrech:
I like drop/pass but exclude/include makes total sense as well. Fine. let's do this.

Let's do this.

ramnes:
Rename done. I still have to fix @trubesv typos and to add the test you asked, @sloria. Do you see anything left after that?

CHANGELOG.rst
```diff
@@ -4,10 +4,18 @@ Changelog
 3.0.0b12 (unreleased)
 +++++++++++++++++++++
 
+Features:
+
+- The behavior to apply when encountering unknown fields while deserializing can be controlled with the ``unknown`` option (:issue:`524`).
+  It makes it possible to either "drop", "pass" or "raise".
```
sloria: This needs to be updated to "exclude", "include", or "raise".

docs/quickstart.rst
```diff
@@ -338,6 +338,45 @@ Or you can ignore missing fields entirely by setting ``partial=True``.
...
+
+This behavior can be modified with the ``unknown`` option, which accept three values:
+
+- `DROP <marshmallow.utils.DROP>`, which is the default
```
sloria: These need to be updated.

sloria: Also, this should explain what the option does, for consistency.

```
- `EXCLUDE <marshmallow.utils.EXCLUDE>`, to exclude unknown fields from the returned data. This is the default behavior.
```

docs/quickstart.rst
```diff
@@ -338,6 +338,45 @@ Or you can ignore missing fields entirely by setting ``partial=True``.
...
+
+Or you can force it directly when loading the data:
+
+.. code-block:: python
+
+    UserSchema().load(data, unknown=PASS)
+
+The ``unknown`` option value set in :meth:`load <Schema.load>` will always override the value applied at instantiation time, which itself will override the value defined in the *class Meta*.
+
+This precedence allows you to define your own favorite behavior and to override it in a different context that could require a different behavior.
```
sloria: Revision suggestion:

```
This order of precedence allows you to change the behavior of a schema for different contexts.
```

sloria (commented):
Just a few more revision suggestions to the docs. Once the remaining nits are addressed, this should be good to merge!

docs/quickstart.rst
```diff
@@ -338,6 +338,45 @@ Or you can ignore missing fields entirely by setting ``partial=True``.
...
+- `RAISE <marshmallow.utils.RAISE>`, to raise a :exc:`ValidationError <marshmallow.exceptions.ValidationError>` if there is any unknown field
+
+There are several places where you can apply that option. You can define it in the *class Meta* of your :class:`Schema`:
+
+.. code-block:: python
+    :emphasize-lines: 2,3
+
+    class UserSchema(Schema):
+        class Meta:
+            unknown = PASS
```
sloria: Update these to `INCLUDE`.

docs/extending.rst
```diff
@@ -37,6 +37,36 @@ By default, pre- and post-processing methods receive one object/datum at a time,
 In cases where your pre- and post-processing methods need to receive the input collection  when ``many=True``, add ``pass_many=True`` to the method decorators. The method will receive the input data (which may be a single datum or a collection) and the boolean value of ``many``.
 
 
+Using Original Input Data
++++++++++++++++++++++++++
+
+By default, unspecified field names are ignored by the validator. If for some reason you want to use the original, raw input, you can add ``pass_original=True`` to your call to `post_load <marshmallow.decorators.post_load>`.
```
sloria: Minor revison suggestion:

```
If you want to use the original, unprocessed input, 
you can add pass_original=True` to your call to post_load.
```

ramnes:
Woops, looks like I forgot to update the docs when renaming from `PASS` / `DROP` to `INCLUDE` / `EXCLUDE`. You could have made only one comment for the whole documentation. :wink: 

I've done everything here. I've also simplified the history back to its original state because rebases where becoming more and more painful with all the extra context.

tests/test_deserialization.py
ramnes: Done

docs/extending.rst
ramnes: Done

docs/quickstart.rst
ramnes: Done

docs/quickstart.rst
ramnes: Done

docs/quickstart.rst
ramnes: Done

CHANGELOG.rst
ramnes: Done

docs/quickstart.rst
ramnes: Done

docs/quickstart.rst
ramnes: Done

sloria:
Looks great! I think this is good to merge. It would be nice to have a way to customize the error message, but that can happen in a follow-up PR. Let's get this merged.

sloria (approved):

lafrech:
A bit late to the party but if we want to enforce the use of the `EXCLUDE`, `INCLUDE` and `RAISE` singletons defined in `utils.py`, we could use `is` and `is not` rather than `==` and `!=` when testing the value of `unknown`.

This implementation allows the user to pass strings directly, although it is not explicitly supported.

sloria:
@lafrech I don't see a benefit to enforcing usage of the constants. It should suffice to document their usage. Also, Sphinx will render the default as a string.

api reference marshmallow 3 0 0b12 dev0 documentation 2018-07-01 13-39-38

lafrech:
That could arguably be a Sphinx issue. Anyway, I don't mind if we leave it as it is.

ramnes:
@sloria may I ask you to tag the current HEAD so that we can profit from this merge?

deckar01:
There are some followup issues that have not been resolved. (#851, #852)

`pip` can install packages directly from github. https://pip.pypa.io/en/stable/reference/pip_install/#git

I believe the URL would be `git+https://github.com/marshmallow-code/marshmallow#egg=marshmallow`.

ramnes:
> pip can install packages directly from github. 

Yeah @deckar01, I know that. We just want to go back on the PyPI version and stop dealing with pip eggs, which we use for very precise things and imply more work for us.

> There are some followup issues that have not been resolved. (#851, #852)

@sloria seemed okay to have a version tagged as soon as this was merged (https://github.com/marshmallow-code/marshmallow/pull/838#issuecomment-396917731, https://github.com/marshmallow-code/marshmallow/pull/838#issuecomment-397842618), hence my ping.

sloria:
There were a few bugs that were surfaced with this feature, so I wanted give this a little more time to "bake". But those bugs have been addressed. The followup issues above can be deferred to a near-future release. So yeah, let's ship it!

ramnes:
Awesome, thank you! :tada:

PR #842, opened 2018-06-14
Small typo

ramnes:


sloria:
Huh, I always thought a 2-tuple could be called a duple, but [the internet says otherwise](http://mathworld.wolfram.com/n-Tuple.html). TIL

PR #847, opened 2018-06-19
Enforce consistent quoting

sloria:
Use single quotes unless it requires escaping. Use triple
double quotes for docstrings

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -13,3 +13,4 @@ simplejson==3.15.0
 
 # Syntax checking
 flake8==3.5.0
+flake8-quotes==1.0.0
--- a/marshmallow/exceptions.py
+++ b/marshmallow/exceptions.py
@@ -43,7 +43,7 @@ class ValidationError(MarshmallowError):
         self.kwargs = kwargs
         MarshmallowError.__init__(self, message)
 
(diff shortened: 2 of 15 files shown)


linbrandon4:
his is a “style-only” PR, but it’s bigger than it looks since it hits 15 files and changes ~400 lines total. The goal is clear: standardize quoting so the repo stops bouncing between styles. Using single quotes by default (and triple double quotes for docstrings) makes diffs cleaner and keeps the code consistent across contributors.

A couple things I’d double-check in a sweep like this:

No behavioral changes: these should be pure quote swaps. I’d scan quickly for cases where the change might accidentally alter escapes (e.g., strings that previously relied on \" vs \', raw strings, regex patterns, JSON snippets, or strings with backslashes).

Docstrings stay readable: switching docstrings to triple double quotes is fine, but I’d make sure no docstring content/indentation got messed up in a way that affects docs generation or formatting.

Minimize future diff noise: this kind of PR is great to do once, but it’s best paired with a formatter/linter rule so it doesn’t drift back. If the project uses something like black/flake8/isort/etc., it’d be good if the quote convention is enforced automatically (or at least documented in contributing).

sloria:
@linbrandon4 This is one of several apparently AI-generated comments you've left on **very old PRs**. This is not helpful. Please stop.

Issue #438, opened 2016-04-22
Partial loading on nested fields


pablomolnar:
Given an object with a nested field, is it possible to partially load the object and all nested fields?
e.g.

```
#!/usr/bin/env python
# -*- coding: utf-8 -*-
from marshmallow import Schema, fields


class Address(Schema):
    street = fields.Str(required=True)
    city = fields.Str(required=True)
    state = fields.Str(required=True)
    zipcode = fields.Str(required=True)
    country = fields.Str(required=True)


class User(Schema):
    first_name = fields.Str(required=True)
    last_name = fields.Str(required=True)
    address = fields.Nested(Address)


payload = {
    'first_name': 'John',
    'address': {
        'zipcode': '90210'
    }
}

user_schema = User()
data, errors = user_schema.load(payload, partial=True)
if errors:
    raise Exception(errors)
```

Output:

```
Traceback (most recent call last):
  File "test_partial_nested.py", line 30, in <module>
    raise Exception(errors)
Exception: {'address': {'city': ['Missing data for required field.'], 'state': ['Missing data for required field.'], 'street': ['Missing data for required field.'], 'country': ['Missing data for required field.']}}
```

Input is coming from a REST API and ideally I would like to bind all incoming payload at once.

Thanks!

sloria:
There isn't built-in support for nested `partial` fields, but you could achieve this by using a schema factory. Something like

``` python
def PartialSchemaFactory(schema_cls):
    schema = schema_cls(partial=True)
    for field_name, field in schema.fields.items():
        if isinstance(field, fields.Nested):
            new_field = deepcopy(field)
            new_field.schema.partial = True
            schema.fields[field_name] = new_field
    return schema
```

A full gist is here: https://gist.github.com/sloria/9f6214b37aab3a6aa37e9a361a50cfd8

xingweicmu:
👍

lafrech:
The schema factory workaround does not provide a `Schema` with the ability to do both `partial` and `not partial` depending on what parameters are provided to `load`.

In other words, when the factory creates a `PartialSchema`, calling `schema.load(partial=False)` will only affect first level fields.

See this example. I added a `partial` parameter to the factory.

``` py
from copy import deepcopy

from marshmallow import Schema, fields

def PartialSchemaFactory(schema_cls, partial=True):
    schema = schema_cls(partial=partial)
    for field_name, field in schema.fields.items():
        if isinstance(field, fields.Nested):
[...]

PR #849, opened 2018-06-21
Allow partial loading of nested fields

arbor-dwatson:
This change passes the `partial` keyword argument from a `schema.load`
call down into nested fields (see #438).  In the context of a REST
api, this allows a POST to require a subset of the fields to be
specified, while a PATCH can use `partial=True` to suppress any
required fields.

This change also supports using a tuple as the `partial` argument
using dot notation to configure partial loading of nested schemas.
For example: `partial=('x', 'y.z')` would allow `x` in the top level
schema to be missing or for `z` in the sub-schema for field `y` to be
missing.

I believe this is a breaking change since subclasses of `fields.Nested`
need to handle the `partial` parameter to `_deserialize()`.  This seems
to follow the existing pattern for using dot delimiters to specify nested
fields (`exclude` and `only` in the Schema init).  I'm very open to
suggestions on a better way to approach this.


marshmallow/marshalling.py
```diff
@@ -264,7 +265,18 @@ def deserialize(self, data, fields_dict, many=False, partial=False,
                     (partial_is_collection and attr_name in partial)
                 ):
                     continue
-            getter = lambda val: field_obj.deserialize(val, field_name, data)
+            d_kwargs = {}
+            if isinstance(field_obj, fields.Nested):
+                # Allow partial loading of nested schemas.
+                prefix = field_name + "."
```
lafrech: Minor optimization.

```py
                if partial_is_collection:
                    prefix = field_name + "."
                    len_prefix = len(prefix)
                    sub_partial = [f[len_prefix:]
                                   for f in partial if f.startswith(prefix)]
```

tests/test_deserialization.py
```diff
@@ -1431,6 +1431,77 @@ class Sch(Schema):
...
+
+        b_dict = {
+            'z': {
+                'y': 42,
+            }
+        }
+        b_partial = ('z.x',)
+        result = SchemaB().load(b_dict, partial=b_partial)
+        assert result['z']['y'] == 42
+
```
lafrech, line 1714: Maybe add a failing test here to show the tuple is actually parsed, not just interpreted as `True`.

```py
        result = SchemaB().load(b_dict, partial=('z.x',))
        assert result['z']['y'] == 42

            SchemaB().load(b_dict, partial=('z.y',))
```

marshmallow/fields.py
```diff
@@ -306,6 +306,8 @@ def _deserialize(self, value, attr, data):
         :param value: The value to be deserialized.
         :param str attr: The attribute/key in `data` to be deserialized.
         :param dict data: The raw input data passed to the `Schema.load`.
+        :param bool|tuple partial: For nested schemas, the ``partial``
```
lafrech, line 319: So, should ~either `partial` or~ `**kwargs` appear in the signature?

lafrech:
Sorry @arbor-dwatson for the late reply. This looks good.

I guess we'd need to modify all the `_deserialize` signatures to add `**kwargs`.

Also, this implementation only passes `partial` (or subpartial tuple) to `Nested`, which makes sense and is more efficient. I'm wondering if it would be cleaner to pass it to any field, even though all the other fields will just ignore it. Singling out `Nested` prevents someone to implement his own nested field not deriving from `Nested`. But since `Nested` is already a special case everywhere in the code, it probably doesn't matter. I guess I'm overthinking it.

It just seems weird to modify all `_deserialize` signatures just for this special case, but it's not like it is breaking anything.

Also, adding `**kwargs` to all fields allows to add new parameters later on without it being a breaking change...

lafrech:
@sloria, @deckar01 feedback welcome. This is close to ready IMHO. I'm just not sure about the `_deserialize` signatures.

arbor-dwatson:
@lafrech Thanks for the feedback!  I believe I've addressed your more technical concerns.  Looks like there are some conflicts with master that I'm going to ignore for now but I'm happy to resolve when the design decisions are finished.

marshmallow/fields.py
arbor-dwatson: This feels like a project style decision so I didn't update it on my own.

My two cents: I think I can personally argue it both ways.  Adding it feels unnecessary from a client perspective.  It doesn't tell me anything about how I can directly use the function.  On the other hand, if I'm "using" the library by subclassing elements (which honestly feels like a typical use case), then I kind of want to know that the function will accept arbitrary keyword arguments.  The reason I didn't add it in the first place is that by the time I'm subclassing the function I'm already looking at the code so the documentation didn't seem as relevant.

sloria (approved):
@arbor-dwatson Sorry again for the late response on this. This looks great!

I don't feel strongly about adding `**kwargs` to all the field signatures, but I suppose it would be more future-proof. I don't think necessarily it needs to happen before this is merged, though.

For now, can you get this updated with `dev`, @arbor-dwatson ?

arbor-dwatson:
@sloria No problem with the delay!  Thank you for all your hard work!

Updated with dev.  The only real issue was the new Pluck class, which I think I updated correctly.

Let me know if there's anything else I can do.

sloria:
Thanks for updating it. Looks good to merge.

PR #855, opened 2018-06-26
Add pre-commit

sloria:
Adds [pre-commit](https://pre-commit.com/) for installing pre-commit hooks for formatting and linting staged files.

* Adds [trailing commas](https://github.com/asottile/add-trailing-comma)
* Ensures consistent quotes
* Trims trailing whitespace
* Ensures end of file has a single new line
* Runs flake8

The relevant configuration lives in `.pre-commit-config.yaml`.

If you already have a dev environment set up, you can set up pre-commit with:

```
pip install pre-commit
pre-commit install --allow-missing-config
```


sloria:
To whoever decides to review this: there's a lot of changes here that were automatically done by running `pre-commit run --all-files`. No need to review those. The relevant config was added in https://github.com/marshmallow-code/marshmallow/pull/855/commits/d6d40f832ab88de54d9d9d0e4c71cffebe4663b8 .

.pre-commit-config.yaml
```diff
@@ -0,0 +1,26 @@
...
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v1.3.0
+  hooks:
+  - id: trailing-whitespace
+    args: ['--markdown-linebreak-ext=md,rst']
+  - id: double-quote-string-fixer
+  - id: end-of-file-fixer
+  - id: flake8
+    # Need duplicated config due to known issues with flake8
+    # https://github.com/pre-commit/pre-commit-hooks/issues/112
```
asottile: note that that issue was (as far as I can tell) just about _exclude_

.pre-commit-config.yaml
```diff
@@ -0,0 +1,26 @@
...
+    exclude: >
+        (?x)^(
+          .git|
+          .ropeproject|
+          .tox|
+          marshmallow/compat.py|
+          build|
+          env|
+          venv
+        )$
```
asottile: the only thing you should need to exclude here is `mashmallow/compat.py` -- `pre-commit` will only ever pass filenames that are checked in (and at that point, you might as well just put `# flake8: noqa` at the top of the file and dump the exclude entirely)

.pre-commit-config.yaml
```diff
@@ -0,0 +1,26 @@
+repos:
+- repo: https://github.com/asottile/add-trailing-comma
+  rev: v0.6.4
+  hooks:
+  - id: add-trailing-comma
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v1.3.0
+  hooks:
+  - id: trailing-whitespace
+    args: ['--markdown-linebreak-ext=md,rst']
```
asottile: interesting! you may be the first I've seen using this setting 😆

asottile (commented):
👋 hi hi, sorry about the drive by -- noticed the star and got curious

.pre-commit-config.yaml
asottile: meaning: you shouldn't actually need to duplicate anything and can probably get away with the default here

.pre-commit-config.yaml
sloria: TBH, I wasn't sure I needed this, but figured I'd add it since we're using rST on this project. Might be worth supporting `rst` by default, since many Python projects use it.

.pre-commit-config.yaml
sloria: Oh, nice! That seems way better.

sloria:
@asottile I appreciate the drive-by. And big thanks for pre-commit. It's a very useful tool.

.pre-commit-config.yaml
asottile: the commandline option is for an _obscure_ (imo anti-feature) of markdown in that you can force a linebreak by *intentionally* adding trailing whitespace

I've yet to find actual code in the wild which uses it (I checked out your branch and removed this and it doesn't change anything for `marshmallow` either)

sloria: Ah ok, I'll just remove it then. Thanks.

asottile (approved):

lafrech:
With such a blessing, I guess this is good to go.

Issue #852, opened 2018-06-23
Make the error messages for "unknown fields" and "invalid data type" configurable


sloria:
As a follow-up to https://github.com/marshmallow-code/marshmallow/pull/838#issuecomment-399677937 : Make the error messages for passing unknown fields (currently `'Unknown field.'`) and invalid input type (currently `'Invalid input type.'`) configurable on a `Schema`.

lafrech:
Same with `"Invalid input type."` (https://github.com/marshmallow-code/marshmallow/pull/857#issuecomment-401613877).

sloria:
Good call. Renamed the issue.

Dunstrom:
Is this issue still relevant? If so I might be able to fix it.

sloria:
@Dunstrom Yes, this is still up for grabs!

Dunstrom:
Thanks for the info @sloria, I will start working on this.

Dunstrom:
@sloria, I would love it if you had a look at my PR when you get the time!

Issue #851, opened 2018-06-23
Change default for `unknown` to `RAISE`


sloria:
As a follow up to https://github.com/marshmallow-code/marshmallow/issues/524#issuecomment-397165731 and #838 : the default for `unknown` should be changed to `RAISE`.

sloria:
@lafrech Were you still planning to work on this?

lafrech:
Yep. I needed https://github.com/marshmallow-code/marshmallow/pull/857 merged first.

Here's the PR for this issue: https://github.com/marshmallow-code/marshmallow/pull/872.

PR #857, opened 2018-06-27
Manage wrong data type with unknown!=EXCLUDE

lafrech:


lafrech:
Not really a corner case since `RAISE` shall become default.

lafrech:
Hold this. It is a bit more complicated than I first thought.

It is not only the about the `None` case but about all wrong data types.

Current approach is to iterate on fields and get keys from input data. If data has no get method, then `AttributeError` is caught and an error is stored.

There are a few issues with this:

- When `unknown != EXCLUDE`, the code that handles unknown keys fails as `data` is not iterable and `set(data)` won't work. This is what brought me here in the first place.
- The error message for this is the "type" error message from the first field that triggers this, while this is a schema error. It does not really make sense IMHO to pick a random `Field` error message for a schema error.
- If the schema has no field, no error is triggered. A schema without field might not be an obvious use case, but why not? It could make sense when used with `unknown=INCLUDE`, or it could be a corner case of dynamically creating `Schema`s.

I added two commits showing two different approaches to this.

The first catches the `TypeError` when dealing with unknown fields and skips the treatment. It looks awkward and redundant. And it only addresses the first issue in the list above.

The second checks the type of `data` before doing anything. It adds a check to the normal execution path rather than relying solely on try/catch, which is not ideal in terms of performance, but not _that_ bad. Also, it works with `AttributeError`. If `data` has a `get` method but is not an iterable, it is bound to fail. This might not be an issue in practice. I mean it will only break if you try hard. And the `'Invalid input type'` error message is hardcoded.

This reveals underlying design issues. It is not the simple change I thought. (And I came there while trying to tackle https://github.com/marshmallow-code/marshmallow/issues/851...)

Thoughts welcome.

lafrech:
In fact, the whole wrong data type case management is smelly from the start:

```python
                msg = field_obj.error_messages['type'].format(
                    input=data, input_type=data.__class__.__name__,
                )
                self.error_field_names = [SCHEMA]
                errors = self.get_errors()
                errors.setdefault(SCHEMA, []).append(msg)
```

We're trying to fit a schema error in a mechanism that was meant for field errors. And the logic here is a bit twisted: the first call to `get_errors` mutates `self.errors` so it can't be removed, which is not obvious when reading this part of the code.

Perhaps `store_error` / `get_errors` need refactoring and should provide a clean interface to store schema errors.

This looks quite like a [WWdD](https://github.com/marshmallow-code/apispec/issues/207#issuecomment-388217912) issue...

sloria:
> We're trying to fit a schema error in a mechanism that was meant for field errors. 

I agree that it's awkward right now. We should come up with a solution to configure both "invalid type" and "unknown field" (#852) error messages on the Schema class.

> the first call to get_errors mutates self.errors so it can't be removed, which is not obvious when reading this part of the code.

Yeah, it's not ideal. Either `get_errors` should be renamed or we should separate the mutation of `self.errors` from it.

marshmallow/marshalling.py
```diff
@@ -250,57 +250,55 @@ def deserialize(
...
-        for attr_name, field_obj in iteritems(fields_dict):
-            if field_obj.dump_only:
-                continue
-            field_name = attr_name
-            if field_obj.data_key:
-                field_name = field_obj.data_key
-            try:
+        # Check data is a dict
+        try:
+            data.get('dummy', missing)
```
sloria: I think doing `if isinstance(data, collections.Mapping)` is more appropriate in this case.

marshmallow/marshalling.py
```diff
@@ -250,57 +250,55 @@ def deserialize(
...
+        except AttributeError:
+            errors = self.get_errors(index=index)
+            msg = 'Invalid input type.'
```
sloria: This is fine for now until #852 is addressed. Might as well remove `'type'` in `Field.default_error_messages` while you're at it.

sloria (changes requested):
Just a couple suggestions: using `isinstance` insead of EAFP for checking the data type, and removing `type` from default error messages.

lafrech:
Suggestions applied. Rebased.

sloria (approved):

PR #858, opened 2018-06-28
Officially support Python 3.7

sloria:
Tests against Python 3.7.

This can't be merged because Python 3.7 isn't supported by Travis yet: https://github.com/travis-ci/travis-ci/issues/9069

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,9 @@ install:
 script: invoke test
 jobs:
   include:
+  - python: 3.7
+    dist: xenial    # required for Python 3.7 (travis-ci/travis-ci#9069)
+    sudo: required  # required for Python 3.7 (travis-ci/travis-ci#9069)
   - stage: PyPI Release
     if: tag IS present
     python: "3.6"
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -5,6 +5,7 @@ language: python
 sudo: false
 cache: pip
 python:
+  - "3.7"
```
michael-k: It works this way:
```yml
matrix:
  include:
    - python: 3.7
      dist: xenial
      sudo: required
```

michael-k: @sloria The `matrix` block is correct, but you have to remove line 8 (the one I've commented on).

sloria:
Ah, right. Thanks for your help, @michael-k .

It still doesn't appear to be working (the 3.7 job is absent). I think `matrix.include` (or at least the way I'm using it) is incompatible with `jobs.include`. I'm unsure how to reconcile the two.

michael-k:
Oh, I've missed the `jobs.include`.  You can put the stuff from `matrix.include` in `jobs.include`.  Here's an example: https://github.com/django/channels/blob/master/.travis.yml

sloria:
That worked. Thanks a lot, @michael-k !

PR #862, opened 2018-07-01
Rework schema errors

lafrech:
I tried to address issues raised in https://github.com/marshmallow-code/marshmallow/pull/857#issuecomment-400737110 by refactoring the code.

The first commit (https://github.com/marshmallow-code/marshmallow/pull/862/commits/9f84247bad8693ec8e357252aa35d9b4e07093a8) changes the error structure in a corner case, but in a good way. It could be considered a bugfix. I can send a separate PR for this one if needed.

The other commits are about factorizing the code that stores schema errors by adapting `store_error` (that normally stores field errors) while it is currently duplicated in `Unmarshaller`. I'm happy with this rework (not "this is so perfect" happy, but rather "this is much better" happy), except one test is broken.

When `store_error` receives a `dict` as message, it does not merge it, nor does it append it to the current error messages list. It just uses it as error message.

```python
        if isinstance(messages, dict):
            errors[field_name] = messages
```

Hence the test I had to modify:

```python
-        assert errors['_schema'][0] == {'code': 'invalid_field'}
+        assert errors['_schema'] == {'code': 'invalid_field'}
```

This differs from the behaviour of the code duplicated in `Unmarshaller`. And changing this in `store_error` would break a lot of field errors tests.

While working on this, I ended up facing the issues described in https://github.com/marshmallow-code/marshmallow/issues/441. I think I'll either follow-up there or create a separate issue.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,11 +4,19 @@ Changelog
 3.0.0b13 (unreleased)
 +++++++++++++++++++++
 
+Bug fixes:
+
+- Errors reported by a schema-level validator for a field in a ``Nested`` field
+are stored under corresponding field name, not ``_schema `` key (:issue: `862`).
+
 Other changes:
 
 - *Backwards-incompatible*: The ``unknown`` option now defaults to ``RAISE``
   (`#524 (comment) <https://github.com/marshmallow-code/marshmallow/issues/524#issuecomment-397165731>`_,
   :issue:`851`).
+- *Backwards-incompatible*: When a schema error is raised with a ``dict`` as
+  payload, the ``dict`` overwrites any existing error list. Before this change,
+  it would be appended to the list.
 
 3.0.0b12 (2018-07-04)
 +++++++++++++++++++++
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -42,17 +42,20 @@ class ErrorStore(object):
     def get_errors(self, index=None):
         return self.errors if index is None else self.errors.setdefault(index, {})
 
-    def store_error(self, field_name, error, index=None):
-        self.error_kwargs.update(error.kwargs)
+    def store_error(self, field_name, messages, index=None):
         self.error_field_names.append(field_name)
         errors = self.get_errors(index=index)
         # Warning: Mutation!
-        if isinstance(error.messages, dict):
-            errors[field_name] = error.messages
+        if isinstance(messages, dict):
+            errors[field_name] = messages
         elif isinstance(errors.get(field_name), dict):
-            errors[field_name].setdefault(FIELD, []).extend(error.messages)
+            errors[field_name].setdefault(FIELD, []).extend(messages)
         else:
-            errors.setdefault(field_name, []).extend(error.messages)
+            errors.setdefault(field_name, []).extend(messages)
+
+    def store_validation_error(self, field_name, error, index=None):
+        self.error_kwargs.update(error.kwargs)
+        self.store_error(field_name, error.messages, index=index)
         # When a Nested field fails validation, the marshalled data is stored
         # on the ValidationError's valid_data attribute
         return error.valid_data or missing
@@ -70,7 +73,7 @@ class ErrorStore(object):
         try:
             value = getter_func(data)
         except ValidationError as error:
-            return self.store_error(field_name, error, index)
+            return self.store_validation_error(field_name, error, index)
         return value
 
 
@@ -248,10 +251,7 @@ class Unmarshaller(ErrorStore):
         ret = dict_class()
         # Check data is a dict
         if not isinstance(data, collections.Mapping):
-            errors = self.get_errors(index=index)
-            msg = 'Invalid input type.'
-            self.error_field_names = [SCHEMA]
-            errors.setdefault(SCHEMA, []).append(msg)
+            self.store_error(SCHEMA, ('Invalid input type.', ), index=index)
         else:
             for attr_name, field_obj in iteritems(fields_dict):
                 if field_obj.dump_only:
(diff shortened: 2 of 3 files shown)


lafrech:
Rebased.

sloria:
I like the refactoring done here; the addition of `store_validation_error` makes sense.

Re: the changed test: I *think* it's an OK change. There is a valid use case for using dictionaries (rather than strings) as individual error messages. marshmallow-jsonapi, for example, [does this](https://github.com/marshmallow-code/marshmallow-jsonapi/blob/289b4ab750e2d5781e50b4f8cc0c82e5ebfefe9b/marshmallow_jsonapi/schema.py#L147-L154). But this use case is still supported by this PR, since the user can pass a list of dictionaries to `ValidationError`. Also, I think you're right to point out that this change could set up #441 nicely.

Because this is a breaking change, it'd be good to have one more set of eyes on this.

lafrech:
Thanks for the feedback.

This PR does not really address #441. It is just a small step towards refactoring the error structure. Maybe it will help by centralizing the problem in one single piece of code. Regarding #441, I'd like to list all cases (different types like string, list, dict,... and which fields returns what) to get a clearer view and decide what should be valid or not. But that can be in a further step.

marshmallow/marshalling.py
```diff
@@ -42,17 +44,23 @@ def __init__(self):
...
         return self.errors if index is None else self.errors.setdefault(index, {})
 
-    def store_error(self, field_name, error, index=None):
-        self.error_kwargs.update(error.kwargs)
+    def store_error(self, field_name, messages, index=None):
         self.error_field_names.append(field_name)
         errors = self.get_errors(index=index)
         # Warning: Mutation!
-        if isinstance(error.messages, dict):
-            errors[field_name] = error.messages
```
deckar01, line 51: > When store_error receives a dict as message, it does not merge it ... This differs from the behaviour of the code duplicated in Unmarshaller. And changing this in store_error would break a lot of field errors tests.

It looks like fields were originally ([8340fca5][0]) allowed to append dicts to the error list, but in a seemingly unrelated bug fix ([4698fe51][1]) it was regressed.

It's probably fine to overwrite some schema errors since field errors have been overwritten for so long, but it might be worth investigating what errors this is hiding. When the field error tests fail, is it because there are extra errors that the tests don't expect, or does it violate core expectations about the structure of the error? Are the hidden errors important?

[0]: https://github.com/marshmallow-code/marshmallow/commit/8340fca5868346e2e20d721d87bbc892354e6d19#diff-c4881f63f8b05b8d3eadf046c3dc7912R110
[1]: https://github.com/marshmallow-code/marshmallow/commit/4698fe51e6d756dca92912a9097b11a7df8a57f2#diff-c4881f63f8b05b8d3eadf046c3dc7912R111

lafrech:
For the record, while working on the error structure (https://github.com/marshmallow-code/marshmallow/issues/441), I noticed that this PR fixes (what I think is) a corner case issue.

I modified a test in `test_deserialization` as a quick way of testing the output when errors are raised from different validators:

```python
    def test_nested_single_deserialization_to_dict(self):
        class SimpleBlogSerializer(Schema):
            title = fields.String()
            author = fields.Nested(SimpleUserSchema, required=True)
            @validates('author')
            def validate_author(self, value):
               if value != 42:
                   raise ValidationError('The answer to life the universe and everything.')
            @validates_schema(skip_on_field_errors=False)
            def validate_schema(self, data):
                raise ValidationError('This will never work', 'author')

        blog_dict = {
            'title': 'Gimme Shelter',
            'author': {'name': 'Mick', 'age': 'lol', 'email': 'mick@stones.com'},
        }
        result = SimpleBlogSerializer().load(blog_dict)
```

Without this PR, the error returned is 

`ValidationError: {'author': {'_field': ['The answer to life the universe and everything.'], 'age': ['Not a valid number.'], '_schema': ['This will never work']}}`

which is wrong: there shouldn't be `_schema`, here.

With this change, I get

`ValidationError: {'author': {'age': ['Not a valid number.'], '_field': ['The answer to life the universe and everything.', 'This will never work']}}`

which is the expected output.

sloria:
@deckar01 Do you see any blockers to merging this?

deckar01:
Feel free to ignore https://github.com/marshmallow-code/marshmallow/pull/862#discussion_r199907288. It is similar to existing behavior and it doesn't seem to be bothering anyone. Looks good to me otherwise.

deckar01 (approved):

sloria:
Thanks @deckar01. @lafrech, care to update the changelog and upgrading guide?

lafrech:
I rebased and updated the changelog. @sloria, do you see anything to add to the upgrading guide?

I still don't like the way `store_error` works, and the breaking change underlined by @deckar01 makes it slightly worse, but it is not that bad as long as we get this sorted for 3.0.

I think this should be dealt with in a wider reflection and I hope https://github.com/marshmallow-code/marshmallow/issues/879 will help clarifying things.

CHANGELOG.rst
```diff
@@ -4,11 +4,19 @@ Changelog
 3.0.0b13 (unreleased)
 +++++++++++++++++++++
 
+Bug fixes:
+
+- Errors reported by a schema leval validator for a field in a ``Nested`` field
```
sloria: Typos: "schema-level"; "is" -> "are"

sloria:
It would be good to document the changed behavior with `raise ValidationError({...})` in the upgrading guide. Certainly not a blocker to merging this if you don't have time right now, though.

CHANGELOG.rst
lafrech: Oh, sorry. Fixed.

lafrech:
Yeah, time and not sure how to formulate this.

I'd better leave this to you when you get the time.

PR #863, opened 2018-07-01
WIP - Fix 'error on Nested field is stored under "_field" key'

lafrech:
This is just a failing test case for the issue I described in https://github.com/marshmallow-code/marshmallow/pull/299#issuecomment-401607834.

> The required constraint triggers an error inside the Nested field, and `errors` is equal to:
> 
> ```python
> {'inner': {0: {'req': ['Missing data for required field.']}, '_field': ['not a chance']}}
> ```
> 
> This is the behaviour that is tested.
> 
> However, if the `required` constraint is removed from `Inner.req`, no error is raised from inside the Nested field, then we don't pass in 
> 
> ```python
>             elif isinstance(errors.get(field_name), dict):
>                 errors[field_name].setdefault(FIELD, []).extend(err.messages)
> ```
> 
> and error is :
> 
> ```python
> {'inner': ['not a chance']}
> ```
> 
> This is inconsistent.

The usage of `_field` seems like a good idea, but we need a more reliable way to determine that the field is a Nested field, or that `_field` should be used.

I didn't see any obvious solution. This could be part of a larger reflection about the error structure (including https://github.com/marshmallow-code/marshmallow/issues/441).

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -1941,9 +1941,10 @@ class TestNestedSchema:
         assert errors['inner']['_schema'] == ['Invalid input type.']
 
     # regression test for https://github.com/marshmallow-code/marshmallow/issues/298
-    def test_all_errors_on_many_nested_field_with_validates_decorator(self):
+    @pytest.mark.parametrize('required', (True, False))
+    def test_all_errors_on_many_nested_field_with_validates_decorator(self, required):
         class Inner(Schema):
-            req = fields.Field(required=True)
+            req = fields.Field(required=required)
 
         class Outer(Schema):
             inner = fields.Nested(Inner, many=True)


lafrech:
Milestoning this as 3.0 for now.

I'd really like to solve https://github.com/marshmallow-code/marshmallow/issues/879 but if we can't come up with something totally satisfying, at least we should make this consistent.

lafrech:
Unfortunately, I didn't find a solution to this.

I explored the issue in https://github.com/marshmallow-code/marshmallow/issues/879 and proposed a rework in https://github.com/marshmallow-code/marshmallow/pull/1026 but it does not address this specific issue.

AFAIU, to achieve consistency, we'd need to change the error structure (for instance using `_errors` in all fields). It might not be worth the trouble.

Let's close this for now and continue discussion in #879. We can always dig up that test if we ever manage to fix this behaviour.

lafrech:
Here's the test, before I delete the branch.

```py
    # regression test for https://github.com/marshmallow-code/marshmallow/issues/298
    @pytest.mark.parametrize('required', (True, False))
    def test_all_errors_on_many_nested_field_with_validates_decorator(self, required):
        class Inner(Schema):
            req = fields.Field(required=required)

        class Outer(Schema):
            inner = fields.Nested(Inner, many=True)

            @validates('inner')
            def validates_inner(self, data):
                raise ValidationError('not a chance')

        outer = Outer()
            outer.load({'inner': [{}]})
        errors = excinfo.value.messages
        assert 'inner' in errors
        assert '_field' in errors['inner']
```

PR #865, opened 2018-07-03
dump-only fields should be considered unknown

sloria:
See https://github.com/marshmallow-code/marshmallow/issues/127#issuecomment-71489839


lafrech:
Good catch. LGTM.

PR #868, opened 2018-07-03
field_b must be greater or equal field_a

jopasdev:
On "Storing Errors on Specific Fields": to raise an error field_b must be greater or equal field_a, not the inverse.


sloria:
Good catch!

Issue #758, opened 2018-03-26
DateTime deserializer cannot use dateutil


mchalek:
The DateTime deserializer cannot use the dateutil library because it contains an inaccessible code path.

For example, parsing the string `2018/03/26` fails:
```python
>>> import marshmallow as ma
>>> f = ma.fields.DateTime()
>>> f.deserialize('2018/03/26')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "marshmallow/fields.py", line 261, in deserialize
    output = self._deserialize(value, attr, data)
  File "marshmallow/fields.py", line 929, in _deserialize
    raise self.fail('invalid')
  File "marshmallow/fields.py", line 216, in fail
    raise ValidationError(msg)
marshmallow.exceptions.ValidationError: Not a valid datetime.
```
Even though the `dateutil`-based parser knows how to deserialize it:
```python
>>> ma.utils.dateutil_available
True
>>> ma.utils.from_datestring('2018/03/26')
datetime.datetime(2018, 3, 26, 0, 0)
```

The cause of this is located in [this code block](https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/fields.py#L923-L935).  Specifically: `dateutil` is only used when the field does not have a pre-defined date format, but the `deserialize` method itself sets the field's date format to a default value before that condition is evaluated.

I submitted #619 to address this last year, but I admit that the description was poorly worded and I never followed-up with it until recently, when user @znatty22 commented that she has run into the same issue.

Could someone please advise on steps to take in order to get #619 merged?

Thanks,

Kevin

deckar01:
What if we registered `dateutil` in `DATEFORMAT_DESERIALIZATION_FUNCS` instead?

lafrech:
What would it do when serializing? What would be the equivalent `to_datestring` function?

We could remove the dead code (including `from_datestring` function) for 3.0 and postpone the "how to use dateutil" issue.

lafrech:
`dateutil` is not really a format. It is more more like an auto/catch-all/non-strict option to let dateutil do whatever it thinks appropriate to provide a datetime from a string, which is in line with the `auto_deserialize_format` approach in #619. Although it looks weird, I don't object to adding it as an option.

[...]

PR #869, opened 2018-07-05
Differentiate between datetimeformat and dateformat

knagra:
Marshmallow currently uses `dateformat` in the `DateTime` field and does not support a custom format for `Date`. This PR changes `DateTime.dateformat` to `DateTime.DateTimeFormat` and adds `Date.dateformat`.

This is my first PR for this repo. Please let me know if I'm breaking any protocols or if other changes are necessary for this to be merged.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -955,42 +955,46 @@ class DateTime(Field):
             self.DEFAULT_FORMAT
 
     def _serialize(self, value, attr, obj):
-        self.format = self.format or self.DEFAULT_FORMAT
+        data_format = self.format or self.DEFAULT_FORMAT
         if value is None:
             return None
-        format_func = self.SERIALIZATION_FUNCS.get(self.format, None)
+        format_func = self.SERIALIZATION_FUNCS.get(data_format, None)
         if format_func:
             try:
                 return format_func(value, localtime=self.localtime)
             except (TypeError, AttributeError, ValueError):
                 self.fail('format', input=value, obj_type=self.OBJ_TYPE)
         else:
-            return value.strftime(self.format)
+            return value.strftime(data_format)
+
+    def _create_data_object_from_parsed_value(self, parsed):
+        """
+        Return a datetime from a parsed object that contains the needed info.
+        """
+        return dt.datetime(
+            parsed.year, parsed.month, parsed.day, parsed.hour, parsed.minute,
+            parsed.second,
+        )
 
     def _deserialize(self, value, attr, data):
-        self.format = self.format or self.DEFAULT_FORMAT
+        data_format = self.format or self.DEFAULT_FORMAT
         if not value:  # Falsy values, e.g. '', None, [] are not valid
             raise self.fail('invalid', obj_type=self.OBJ_TYPE)
-        func = self.DESERIALIZATION_FUNCS.get(
-            self.format,
-        )
+        func = self.DESERIALIZATION_FUNCS.get(data_format)
         if func:
             try:
                 return func(value)
             except (TypeError, AttributeError, ValueError):
                 raise self.fail('invalid', obj_type=self.OBJ_TYPE)
-        elif self.format:
+        elif data_format:
             try:
-                return dt.datetime.strptime(value, self.format)
+                return dt.datetime.strptime(value, data_format)
             except (TypeError, AttributeError, ValueError):
                 raise self.fail('invalid', obj_type=self.OBJ_TYPE)
         elif utils.dateutil_available:
             try:
                 parsed = utils.from_datestring(value)
-                return dt.datetime(
-                    parsed.year, parsed.month, parsed.day, parsed.hour,
-                    parsed.minute, parsed.second,
-                )
+                return self._create_data_object_from_parsed_value(parsed)
             except (TypeError, ValueError):
                 raise self.fail('invalid', obj_type=self.OBJ_TYPE)
         else:
@@ -1069,6 +1073,12 @@ class Date(DateTime):
     OBJ_TYPE = 'date'
 
     SCHEMA_OPTS_VAR_NAME = 'dateformat'
+
+    def _create_data_object_from_parsed_value(self, parsed):
+        """
+        Return a date from a parsed object that contains the need information.
+        """
+        return dt.date(parsed.year, parsed.month, parsed.day)
 
 
 class TimeDelta(Field):
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -944,21 +944,23 @@ class DateTime(Field):
     def __init__(self, format=None, **kwargs):
         super(DateTime, self).__init__(**kwargs)
         # Allow this to be None. It may be set later in the ``_serialize``
-        # or ``_desrialize`` methods This allows a Schema to dynamically set the
+        # or ``_deserialize`` methods This allows a Schema to dynamically set the
         # format, e.g. from a Meta option
         self.format = format
 
     def _add_to_schema(self, field_name, schema):
         super(DateTime, self)._add_to_schema(field_name, schema)
-        self.format = self.format or \
-            getattr(schema.opts, self.SCHEMA_OPTS_VAR_NAME) or \
+        self.format = (
+            self.format or
+            getattr(schema.opts, self.SCHEMA_OPTS_VAR_NAME) or
             self.DEFAULT_FORMAT
+        )
 
     def _serialize(self, value, attr, obj):
-        data_format = self.format or self.DEFAULT_FORMAT
         if value is None:
             return None
-        format_func = self.SERIALIZATION_FUNCS.get(data_format, None)
+        data_format = self.format or self.DEFAULT_FORMAT
+        format_func = self.SERIALIZATION_FUNCS.get(data_format)
         if format_func:
             try:
                 return format_func(value, localtime=self.localtime)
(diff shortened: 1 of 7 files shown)


marshmallow/fields.py
```diff
@@ -1016,32 +1027,86 @@ def _deserialize(self, value, attr, data):
 class Date(Field):
     """ISO8601-formatted date string.
 
+    :param strict: If True, only accept Date objets. If False, will accept any
```
sloria: Drive-by: There is some discussion about removing `strict` flags on fields here: https://github.com/marshmallow-code/marshmallow/pull/755#issuecomment-377662237 , so I have reservations about adding it to `Date`.

sloria:
Thanks @knagra for the PR. I think you're onto something here. I don't have time to review this deeply right now, but that's not to say it won't get reviewed/merged. I'm currently focused on moving marshmallow 3 forward, which means reviewing backwards-incompatible API changes.

If anyone else would like to review this, all feedback is welcome.

knagra:
I can remove the `strict` flag. Also, this is a backwards-incompatible change. Currently, `DateTimeField` uses `dateformat` in `meta`. Under these changes, `DateTimeField` will use `datetimeformat` in `meta` and `DateField` will use `dateformat`. Thanks for the quick initial look, @sloria.

sloria:
Yeah, go ahead and remove the `strict` option for now. That can be discussed separate from this PR. I'm good with the overall approach here. Since this is a breaking change, I'd like at least one more review of this.  cc @deckar01 @lafrech

marshmallow/fields.py
```diff
@@ -1016,32 +1027,86 @@ def _deserialize(self, value, attr, data):
...
+            self.fail('format', input=value)
         try:
-            return value.isoformat()
-        except AttributeError:
+            return dt.date(value.year, value.month, value.day)
+        except (TypeError, AttributeError, ValueError):
             self.fail('format', input=value)
-        return value
+
+    def _serialize(self, value, attr, obj):
```
deckar01: A lot of this new code seems to be duplicated from `DateTime`. Can we use inheritance to avoid duplicating this code?

marshmallow/fields.py
```diff
@@ -1016,32 +1027,86 @@ def _deserialize(self, value, attr, data):
...
+        'iso': utils.from_iso_date,
+        'iso8601': utils.from_iso_date,
+    }
+
+    DEFAULT_FORMAT = 'iso'
+
+    def __init__(self, strict=False, format=None, **kwargs):
+        super(Date, self).__init__(**kwargs)
+        self.strict = strict
+        self.dateformat = format
```
deckar01: It might be better to avoid field name specific prefixes to make these properties nicer to override with inheritance. They are already scoped under a class, so there is no benefit to further namespacing them.

- `self.dateformat` -> `self.format`
- `DATEFORMAT_SERIALIZATION_FUNCS` -> `SERIALIZATION_FUNCS`
- `DATEFORMAT_DESERIALIZATION_FUNCS ` -> `DESERIALIZATION_FUNCS `

marshmallow/utils.py
```diff
@@ -309,6 +309,11 @@ def from_iso_date(datestring, use_dateutil=True):
     else:
         return datetime.datetime.strptime(datestring[:10], '%Y-%m-%d').date()
 
+
+def to_iso_date(date):
+    return date.strftime('%Y-%m-%d')
```
fuhrysteve: This functionality already exists in the stdlib as `datetime.date.isoformat` - so we should be able to use that instead of adding another utils function.

```python
In [1]: import datetime

In [2]: datetime.date.today().isoformat()
Out[2]: '2018-07-12'

In [3]: datetime.date.isoformat(datetime.date.today())
Out[3]: '2018-07-12'
```

knagra: I'm fine with that. I'll make the change.

marshmallow/fields.py
knagra: That makes sense to me. I'll make the changes.

marshmallow/fields.py
knagra: The only major impediment is the distinction between `dateformat` and `datetimeformat` in Meta, but I think I can write a function to change that between the two classes.

How are you envisioning the class inheritance graph for this? `class Date(DateTime)` or `class Date(FormattedField)` and `class DateTime(FormattedField)` or something similar?

deckar01: I don't think python has a defined hierarchy between the two modules. I think it would be valid to treat `Date` as a specialized form of `DateTime`. `class DateTime(Field)` and `class Date(DateTime)` would probably generate the smallest diff also.

> I can write a function to change [the meta format] between the two classes

:+1: Good catch.

lafrech:
@knagra thanks for this work, already. Are you still working on this?

knagra:
Yeah, I'm about to work on this today and tomorrow. Just been busy with work.

lafrech:
No pb, no pressure. Thanks.

knagra:
@deckar01 @fuhrysteve I've incorporated your suggested changes. Implementing `Date(DateTime)` inheritance required adding a couple of extra class variables to those fields to maintain cleanliness of error messages and to allow schemas to have both `dateformat` and `datetimeformat` in the `Meta` class.

I converted the `to_iso_format` function to use `datetime.date.isoformat` as suggested but decided to leave to function to absorb the extra `self.localtime` keyword argument passed to it by `DateTime`.

Let me know if you have other suggestions. I should be able to address them more quickly now.

marshmallow/fields.py
```diff
@@ -931,62 +931,74 @@ class DateTime(Field):
...
 
     def _add_to_schema(self, field_name, schema):
         super(DateTime, self)._add_to_schema(field_name, schema)
-        self.dateformat = self.dateformat or schema.opts.dateformat
+        self.format = self.format or \
+            getattr(schema.opts, self.SCHEMA_OPTS_VAR_NAME) or \
+            self.DEFAULT_FORMAT
 
     def _serialize(self, value, attr, obj):
+        self.format = self.format or self.DEFAULT_FORMAT
```
lafrech: This will conflict with https://github.com/marshmallow-code/marshmallow/pull/934.

Please don't mutate `self.format` and use a local `format` variable.

knagra: Okay. It seems that change was merged just a few days ago. I'll go ahead and change `self.format` to just `format` in this case. Do you want me to change the name of the `self.format` variable back to `self.dateformat` as well? I changed it since `dateformat` in the `DateTime` field doesn't make sense.

tests/test_schema.py
```diff
@@ -1468,19 +1468,32 @@ class BadExclude(Schema):
...
             fields = ('created', 'updated')
-            dateformat = fmt
-    serialized = DateFormatSchema().dump(user)
+            datetimeformat = fmt
+    serialized = DateTimeFormatSchema().dump(user)
     assert serialized['created'] == user.created.strftime(fmt)
     assert serialized['updated'] == user.updated.strftime('%m-%d')
 
+def test_dateformat_option(user):
+    fmt = '%Y-%m'
```
lafrech: Very minor: it is strange to define this format here and no the other one.

marshmallow/fields.py
lafrech: > It seems that change was merged just a few days ago. I'll go ahead and change self.format to just format in this case.

Yes, exactly. I though I'd mention that to help you with the rebase.

> Do you want me to change the name of the self.format variable back to self.dateformat as well?

No, not at all. `format` is better now.

tests/test_schema.py
knagra: It was the pattern saw I above at line 1471. I'll go ahead name them in both tests.

marshmallow/fields.py
```diff
@@ -931,62 +931,74 @@ class DateTime(Field):
...
             except (TypeError, AttributeError, ValueError):
-                raise self.fail('invalid')
+                raise self.fail('invalid', obj_type=self.OBJ_TYPE)
         elif utils.dateutil_available:
             try:
-                return utils.from_datestring(value)
-            except TypeError:
-                raise self.fail('invalid')
+                parsed = utils.from_datestring(value)
+                return dt.datetime(
```
lafrech: I don't understand this change. I don't think this would return a `dt.date` when called from `Date`.

Isn't this code unreachable anyway ? (https://github.com/marshmallow-code/marshmallow/issues/758)

(I don't mean to throw #758 in your way, this is another story. If the code is indeed unreachable, which I think it is, don't worry too much about it. I just don't understand the change.)

tests/test_schema.py
lafrech: Oh, right, of course. Thanks. No pb otherwise.

marshmallow/fields.py
lafrech: Sorry, I just realized this PR is rebased already. Then please change `self.format` to `format` as in #934.

marshmallow/fields.py
knagra: The object creation is an oversight. I'll fix that. I think #758 is beyond the scope of this PR. I can look at it after this change, though, if you need someone to work on it.

knagra:
@lafrech Thanks for the review. I'll implement changes tonight and have it ready for another take.

knagra:
@lafrech I implemented your suggestions. Yes, the code block you mentioned is inaccessible. That's why we weren't getting any errors from the tests. I don't think there's a clear quick fix for #758. There are some behavior questions that need to be answered.

lafrech:
Thanks @knagra for your work on this.

I think this is ready to merge.

lafrech:
@knagra, if you have time, would you mind adding a paragraph to docs/upgrading.rst?

lafrech:
I merged and updated the docs.

Thanks again @knagra. Please double-check the doc update if you get the time.

PR #872, opened 2018-07-05
Change default for `unknown` to `RAISE`

### LINKED ISSUE #851: Change default for `unknown` to `RAISE`

lafrech:
Closes #851 

The change to the code is minimal but a lot of tests needed to be fixed.

Most tests are fixed by just passing `unknown=EXCLUDE` not out of slopiness but because they check that good keys are passed and bad keys are not, so raising wouldn't be practical there.

I removed the unknown keys where they were not relevant to the test.

I also modified `test_load_resets_error_fields` as I think it was broken and it is better this way.

### DIFF
--- a/tests/test_decorators.py
+++ b/tests/test_decorators.py
@@ -541,20 +541,13 @@ class TestValidatesSchemaDecorator:
     def test_allow_arbitrary_field_names_in_error(self):
 
         class MySchema(Schema):
-            foo = fields.Int()
-            bar = fields.Int()
 
-            @validates_schema(pass_original=True)
-            def strict_fields(self, data, original_data):
-                for key in original_data:
-                    if key not in self.fields:
-                        raise ValidationError('Unknown field name', key)
+            @validates_schema
+            def validator(self, data):
+                raise ValidationError('Error message', 'arbitrary_key')
 
-        schema = MySchema(unknown=EXCLUDE)
-        errors = schema.validate({'foo': 2, 'baz': 42})
-        assert 'baz' in errors
-        assert len(errors['baz']) == 1
-        assert errors['baz'][0] == 'Unknown field name'
+        errors = MySchema().validate({})
+        assert errors['arbitrary_key'] == ['Error message']
 
     def test_skip_on_field_errors(self):
 
--- a/tests/test_decorators.py
+++ b/tests/test_decorators.py
@@ -135,7 +135,7 @@ class TestPassOriginal:
                 ret['_post_dump'] = obj['sentinel']
                 return ret
 
-        schema = MySchema()
+        schema = MySchema(unknown=EXCLUDE)
         datum = {'foo': 42, 'sentinel': 24}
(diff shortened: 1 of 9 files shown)


tests/test_deserialization.py
```diff
@@ -1354,15 +1354,15 @@ def test_unknown_fields_deserialization(self):
         class MySchema(Schema):
             foo = fields.Integer()
 
-        data = MySchema().load({'foo': 3, 'bar': 5})
+        data = MySchema(unknown=EXCLUDE).load({'foo': 3, 'bar': 5})
```
sloria, line 1357: I think this test was meant to test the default behavior when `unknown` isn't passed. `unknown=EXCLUDE` is tested a few lines down. So I think this should test that an error is raised.

tests/test_decorators.py
```diff
@@ -550,7 +550,7 @@ def strict_fields(self, data, original_data):
                     if key not in self.fields:
                         raise ValidationError('Unknown field name', key)
 
-        schema = MySchema()
+        schema = MySchema(unknown=EXCLUDE)
```
sloria: This test is fine for now, but it'd be good to change it soon, now that there's a built-in way to handle unknown fields.

sloria (commented):
Just a minor change suggested. Otherwise, this looks good to go.

tests/test_deserialization.py
lafrech: Default is tested a few lines later.

Maybe this test could be split in several tests and/or parametrized, but I think it does the right thing. At least I didn't change it, I just switched the `EXCLUDE` and `RAISE` cases.

tests/test_decorators.py
lafrech: I'm not sure I understand.

Do you mean that it is good as a test but bad as an example, so we should change it to a more useful or at least more neutral example?

tests/test_deserialization.py
sloria: Oh, I missed that. OK, I think this if fine as is.

tests/test_decorators.py
sloria: Exactly: the `strict_fields` implementation is a bad example that should be replaced by a more neutral one.

sloria (approved):

tests/test_decorators.py
lafrech: Done.

sloria:
Thanks @lafrech . Would you like to update the changelog and the upgrading guide?

If you don't have time right now, I'll try to find some time during the weekend.

lafrech:
Changelog + upgrading guide updated in dev branch. Feel free to reword.

sloria:
Thanks!

Issue #907, opened 2018-08-08
None with many=True sets valid_data to {} instead of []


tuukkamustonen:
With `3.0.0.b13` this now passes:

```python
class Sch(Schema):
    foo = fields.Str()

with pytest.raises(ValidationError) as e:
    Sch(many=True).load(None)
assert e.value.messages == {'_schema': ['Invalid input type.']}
assert e.value.valid_data == {}
```

This sets `valid_data` to `{}` rather than `[]` as usual with `many=True`. I understand that we're not really interested about `valid_data` here, but isn't this kinda confusing because we're expecting a list (for `many=True`) and valid data is suddenly dictionary instead (invalid)?

sloria:
Good catch. I agree that `valid_data` should be a list in that case. PRs welcome.

Issue #906, opened 2018-08-08
TypeError thrown with many=True on non-iterable types


tuukkamustonen:
With `3.0.0b13` this passes:

```python
class Sch(Schema):
    foo = fields.Str()

for p in [False, 1, 1.2]:
    with pytest.raises(TypeError):
        Sch(many=True).load(p)
```

I think `Schema.load()` should rather catch invalid types and raise `ValidationError`?

sloria:
I think your suggestion makes sense, given that we raise a `ValidationError` for `None`

```python
from marshmallow import Schema, fields

class Sch(Schema):
    foo = fields.Str()


Sch(many=True).load(None)
# marshmallow.exceptions.ValidationError: {'_schema': ['Invalid input type.']}
```

We should probably do the same for other non-iterable types. PRs welcome.

PR #910, opened 2018-08-09
Fixes for #906 (TypeError with many=True) and #907 (return [] not {} on many=True error) and #930 (disallows strings and dicts on many=True)

tuukkamustonen:
See #906 and #907. In same PR, as latter one re-uses fix and test of the former.

Do you want me to move the test from `test_schema.py` to `test_marshalling.py` (as the fixes are in `marshalling` module?

Also, I wonder if these need to be reflected in `Marshaller.serialize()`, too? What is the intended behavior if you try to marshall a primitive, for example?

I took liberty to thank myself in changelog :)

### DIFF
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -12,7 +12,7 @@ from __future__ import unicode_literals
 import collections
 
 from marshmallow.utils import (
-    EXCLUDE, INCLUDE, RAISE, is_collection, missing, set_value,
+    EXCLUDE, INCLUDE, RAISE, is_collection, missing, set_value, is_iterable_but_not_string
 )
 from marshmallow.compat import iteritems, basestring
 from marshmallow.exceptions import (
@@ -210,7 +210,7 @@ class Unmarshaller(ErrorStore):
             serializing a collection, otherwise `None`.
         :return: A dictionary of the deserialized data.
         """
-        if many and data is not None:
+        if many and data is not None and is_iterable_but_not_string(data):
             self._pending = True
             ret = [
                 self.deserialize(
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -12,7 +12,7 @@ from __future__ import unicode_literals
 import collections
 
 from marshmallow.utils import (
-    EXCLUDE, INCLUDE, RAISE, is_collection, missing, set_value, is_iterable_but_not_string
+    EXCLUDE, INCLUDE, RAISE, is_collection, missing, set_value
 )
 from marshmallow.compat import iteritems, basestring
 from marshmallow.exceptions import (
@@ -211,7 +211,7 @@ class Unmarshaller(ErrorStore):
         :return: A dictionary of the deserialized data.
         """
         ret = dict_class() if not many else []
-        if many and data is not None and is_iterable_but_not_string(data):
+        if many and data is not None and is_collection(data):
             self._pending = True
             ret.extend([
                 self.deserialize(
@@ -231,7 +231,7 @@ class Unmarshaller(ErrorStore):
                     data=ret,
                 )
             return ret
-        if not isinstance(data, collections.Mapping):
+        if not isinstance(data, collections.Mapping) or many:
             self.store_error(SCHEMA, ('Invalid input type.', ), index=index)
         else:
             partial_is_collection = is_collection(partial)
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -203,16 +203,17 @@ class Unmarshaller(ErrorStore):
             will be ignored.
         :param unknown: Whether to exclude, include, or raise an error for unknown
             fields in the data. Use `EXCLUDE`, `INCLUDE` or `RAISE`.
-        :param type dict_class: Dictionary class used to construct the output.
+        :param type dict_class: Dictionary class used to construct the output when many=False.
         :param bool index_errors: Whether to store the index of invalid items in
             ``self.errors`` when ``many=True``.
         :param int index: Index of the item being serialized (for storing errors) if
             serializing a collection, otherwise `None`.
         :return: A dictionary of the deserialized data.
         """
+        ret = dict_class() if not many else []
         if many and data is not None and is_iterable_but_not_string(data):
             self._pending = True
-            ret = [
+            ret.extend([
                 self.deserialize(
                     d, fields_dict, many=False,
                     partial=partial, unknown=unknown,
@@ -220,7 +221,7 @@ class Unmarshaller(ErrorStore):
                     index_errors=index_errors,
                 )
                 for idx, d in enumerate(data)
-            ]
+            ])
 
             self._pending = False
             if self.errors:
@@ -230,12 +231,10 @@ class Unmarshaller(ErrorStore):
                     data=ret,
                 )
             return ret
-        partial_is_collection = is_collection(partial)
-        ret = dict_class()
-        # Check data is a dict
         if not isinstance(data, collections.Mapping):
             self.store_error(SCHEMA, ('Invalid input type.', ), index=index)
         else:
+            partial_is_collection = is_collection(partial)
             for attr_name, field_obj in iteritems(fields_dict):
                 if field_obj.dump_only:
                     continue
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,8 @@ Bug fixes:
 
 - Raise ``ValidationError`` instead of ``TypeError`` when non-iterable types are
   validated with ``many=True`` (:issue:`851`). Thanks :user:`tuukkamustonen`.
+- Return ``[]`` as ``ValidationError.valid_data`` instead of ``{}`` with
+  ``many=True`` (:issue:`907`). Thanks :user:`tuukkamustonen`.
 
 3.0.0b13 (2018-08-04)
 +++++++++++++++++++++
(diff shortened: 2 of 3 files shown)


marshmallow/marshalling.py
```diff
@@ -190,7 +190,7 @@ def run_validator(
 
     def deserialize(
         self, data, fields_dict, many=False, partial=False,
-        unknown=RAISE, dict_class=dict, index_errors=True, index=None,
+        unknown=RAISE, dict_class=dict, list_class=list, index_errors=True, index=None,
```
tuukkamustonen: Not sure about the use case, but `dict_class` is configurable, so...

marshmallow/marshalling.py
```diff
@@ -230,12 +232,10 @@ def deserialize(
                     data=ret,
                 )
             return ret
-        partial_is_collection = is_collection(partial)
-        ret = dict_class()
-        # Check data is a dict
         if not isinstance(data, collections.Mapping):
             self.store_error(SCHEMA, ('Invalid input type.', ), index=index)
```
tuukkamustonen: The error for non-iterables is logged here. I don't like the flow of code now - it's somewhat confusing - but I didn't want to nest the code to leave minimal diff.

tuukkamustonen: Oh, and I moved `partial_is_collection` slightly, because it's not needed on the main level. I can restore original code if you want.

sloria:
Thanks for the PR. I'll try to review this sometime over the weekend.

marshmallow/marshalling.py
sloria: I don't think `list_class` is necessary. The only reason we have `dict_class` is to support the `ordered = True` option (in which case `OrderedDict` is used instead of `dict`).

marshmallow/marshalling.py
sloria: > The error for non-iterables is logged here. I don't like the flow of code now - it's somewhat confusing - but I didn't want to nest the code to leave minimal diff.

Fair enough. Feel free to make further improvements to the flow here in a future PR.

marshmallow/marshalling.py
```diff
@@ -203,24 +203,26 @@ def deserialize(
...
+        :param type list_class: List class used to construct the output when many=True.
         :param bool index_errors: Whether to store the index of invalid items in
             ``self.errors`` when ``many=True``.
         :param int index: Index of the item being serialized (for storing errors) if
             serializing a collection, otherwise `None`.
         :return: A dictionary of the deserialized data.
         """
-        if many and data is not None:
+        ret = dict_class() if not many else list_class()
+        if many and data is not None and is_iterable(data):
```
sloria: Why not use `is_iterable_but_not_string` here?

marshmallow/utils.py
```diff
@@ -52,6 +52,11 @@ def is_generator(obj):
     return inspect.isgeneratorfunction(obj) or inspect.isgenerator(obj)
 
 
+def is_iterable(obj):
```
sloria: See previous comment: not sure this is necessary--I think you can just use `is_iterable_but_not_string`.

sloria (changes requested):
Thanks for the PR! I appreciate that you updated the changelog.

Just a few nits to address; when those are done, this is good to merge.

marshmallow/marshalling.py
tuukkamustonen: Alright, I'll remove the option.

marshmallow/marshalling.py
tuukkamustonen: In my understanding, the library regards strings as iterable (which they are) so I didn't want to make an exception here. If you pass string with many=True, marshmallow will iterate over string and pick individual characters as elements, so I think we should act the same way here.

(However, I think it's confusing and weird, because you cannot validate "primitives" with marshmallow anyway (afaik), and accidentally passing in string rather than "actual" iterable such as list should rather raise than pass, but that's just my perspective. I actually have a PR on this in https://github.com/marshmallow-code/marshmallow/pull/622 but there was no clear outcome.)

marshmallow/utils.py
tuukkamustonen: Will remove if it's not needed.

marshmallow/marshalling.py
sloria: > However, I think it's confusing and weird, because you cannot validate "primitives" with marshmallow anyway (afaik), and accidentally passing in string rather than "actual" iterable such as list should rather raise than pass, but that's just my perspective

Exactly. Passing in a string is invalid usage, so I think raising an error is the correct behavior.

tuukkamustonen: > Exactly. Passing in a string is invalid usage, so I think raising an error is the correct behavior.

But marshmallow doesn't currently give error when you pass string as data for `many=True`? Or 2.x at least didn't, was it changed for 3.x?

sloria: I don't believe 2.x raises an error, but I don't think that should be dispositive. 3.x became more strict in its handling of `None`, so I think it should do the same for strings (and other invalid types).

tuukkamustonen: Ok, I'll check that too tomorrow. In a way regarding strings like any iterable makes sense, but it's normally not what the user intended so raising error makes more sense. So we agree on that, nice.

tuukkamustonen: @sloria Checked this, marshmallow 3.x still considers a `str` valid for `many=True`:

```python
In [1]: from marshmallow import Schema

In [3]: from marshmallow import fields

In [4]: class Sch(Schema):
   ...:     foo = fields.Int()

In [7]: Sch(many=True).validate([{'foo': 1}, {'foo': 2}])
Out[7]: {}

In [8]: Sch(many=True).validate('asf')
Out[8]: 
{0: {u'_schema': [u'Invalid input type.']},
 1: {u'_schema': [u'Invalid input type.']},
 2: {u'_schema': [u'Invalid input type.']}}
```

Rather than giving out error `{'_schema': ['Invalid input type.']}`.

Could you check #622 (the discussion there, basically). What is your take on it?

tuukkamustonen:
@sloria Made the adjustments:

- Removed `list_class` option
- Changed to `is_iterable_but_not_string()`

The latter adjustment CHANGES behavior in that passing `str` to `many=True` will now raise `{'_schema': ['Invalid input type.']}`. I think we should write a changelog entry also on this...

tuukkamustonen:
@sloria I added a commit that tests `many=True` and a CHANGELOG entry (new commit). Also opened #930 for it. If we now would block strings, should we also reject dicts?

sloria:
@tuukkamustonen Hm, yes, I suppose `dicts` should be dis-allowed, too.

tuukkamustonen:
Ok, added `collections.Mapping` exception (instead of raw `dict`). Squashed into previous commit.

For consistence, should we tackle strings and mappings also on `dump()`? Trying to dump string with `many=True` now gives `AttributeError` and dumping dict just works. I guess `dump()` is so different beast we don't need to touch it?

tuukkamustonen:
Resolved conflicts.

tuukkamustonen:
@sloria @lafrech Should we try to push it in? Are there any adjustments still that you want me to apply?

sloria:
This is looking good. Since these are bug fixes, though, I wonder if they should be sent to `2.x-line`. @lafrech  Thoughts?

marshmallow/marshalling.py
```diff
@@ -203,24 +203,25 @@ def deserialize(
             will be ignored.
         :param unknown: Whether to exclude, include, or raise an error for unknown
             fields in the data. Use `EXCLUDE`, `INCLUDE` or `RAISE`.
-        :param type dict_class: Dictionary class used to construct the output.
+        :param type dict_class: Dictionary class used to construct the output when many=False.
```
lafrech, line 206: The class is used also when many=True, to construct the output of each element. I'd remove this change.

marshmallow/marshalling.py
```diff
@@ -203,24 +203,25 @@ def deserialize(
...
         :param bool index_errors: Whether to store the index of invalid items in
             ``self.errors`` when ``many=True``.
         :param int index: Index of the item being serialized (for storing errors) if
             serializing a collection, otherwise `None`.
         :return: A dictionary of the deserialized data.
         """
-        if many and data is not None:
+        ret = dict_class() if not many else []
+        if many and data is not None and is_collection(data):
```
lafrech, line 214: `data is not None` is useless, here, as `is_collection(None)` is `False`.

lafrech:
Although I had worked on this code already, I found the logic hard to grab at first sight.

```py
if many and data is not None and is_collection(data):
    [...]  # Normal many=True case
if not isinstance(data, collections.Mapping) or many:
    [...]  # (many=False and not dict) or (many=True and not collection) case
else:
    [...]  # Normal many=False case
```
I reworked it in https://github.com/marshmallow-code/marshmallow/pull/954 while also addressing my remarks above.

My version duplicates `self.store_error(SCHEMA, ('Invalid input type.', ), index=index)` but I don't think it is that band and overall I find it easier to understand.

No strong opinion, @sloria, pick the one you prefer.

> Since these are bug fixes, though, I wonder if they should be sent to 2.x-line. 

I don't think anyone would [reasonably](https://xkcd.com/1172/) rely on current behaviour, so fixing this in 2.x makes sense. How would you proceed? Would you apply the same patches manually? Cherry-pick the commits? Not sure which is easier.

sloria:
Thanks @lafrech. I agree with your changes on that PR. Let's go with that.

> How would you proceed? Would you apply the same patches manually? Cherry-pick the commits? Not sure which is easier.

I can take care of cherry-picking the commits into a branch off of `2.x-line`.

Closing. Let's continue in #954.

Issue #875, opened 2018-07-06
dump-only fields are passed when unknown = INCLUDE


lafrech:
When unknown=INCLUDE, dump-only fields are passed as unknown (and not validated).

This is due to https://github.com/marshmallow-code/marshmallow/pull/865 and is a side-effect of being considered unknown, but is this really what we want?

Modified test:

```python
    def test_dump_only_fields_considered_unknown(self):
        class MySchema(Schema):
            foo = fields.Field(dump_only=True)

        with pytest.raises(ValidationError) as excinfo:
            MySchema().load({'foo': 42})
        err = excinfo.value
        assert 'foo' in err.messages
        assert err.messages['foo'] == ['Unknown field.']

        # Test with unknown=INCLUDE
        data = MySchema(unknown=INCLUDE).load({'foo': 42})
        assert 'foo' not in data  # fail
```

Excluding them seems inconsistent (why exclude a dump-only field and not a real unknown field if dump-only are treated as unknown?) but less risky.

I'm not saying this is a bug, but I just got surprised by it and it is not explicitly tested, so I figured I'd ask.

sloria:
Good catch. I agree that excluding dump-only fields seems inconsistent. So I think the current behavior is correct. It'd be a good idea to explicitly test this behavior. Go ahead and add that assertion if you have time; otherwise, I can do it after work today.

rileycitizennet:
I'm going to put a vote in favor of excluding dump-only fields when they're passed in. The dump-only field is "known," it's just a one-way sort of deal.

Of course, it would be _really_ nice to have a way to configure this, so we can each have our own opinions, but maybe that's ugly.

rileyjohngibbs:
I'm going to put a vote in favor of excluding dump-only fields when they're passed in. The dump-only field is "known," it's just a one-way sort of deal.

Also, if dump-only fields aren't excluded, it makes `unknown=INCLUDE` essentially override (without any validation) `dump_only=True`. This is particularly troublesome since `unknown` is declared on the Schema, and `dump_only` is declared on a field; specific (a field parameter) should take precedence over general (a schema parameter).

Of course, it would be _really_ nice to have a way to configure this, so we can each have our own opinions, but maybe that's ugly.

tuukkamustonen:
> I'm going to put a vote in favor of excluding dump-only fields when they're passed in. The dump-only field is "known," it's just a one-way sort of deal.

I think this describes it well.

[...]

PR #911, opened 2018-08-09
Fixes #908: Nested field default unknown to RAISE

### LINKED ISSUE #908: RAISE should be default for Nested()
Isn't it a bit weird that `Schema` defaults to `RAISE` but `Nested` defaults to `EXCLUDE`?

It's probably not what one expects (at least I didn't). How about changing the default? Or what might be the reasoning?

tuukkamustonen:
Ref #908.

Do we want more tests? I now just updated tests to stick with `EXCLUDE`, where necessary.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,12 @@ Features:
 - Raise a `StringNotCollectionError` if ``only`` or ``exclude`` is
   passed as a string to ``fields.Nested`` (:pr:`931`).
 
+Other changes:
+
+- *Backwards-incompatible*: ``Nested`` field now defaults to ``unknown=RAISE``
+  instead of ``EXCLUDE``. This harmonizes behavior with ``Schema`` that
+  already defaults to ``RAISE`` (:issue:`908`). Thanks :user:`tuukkamustonen`.
+
 3.0.0b13 (2018-08-04)
 +++++++++++++++++++++
 
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -12,7 +12,7 @@ import decimal
 
 from marshmallow import validate, utils, class_registry
 from marshmallow.base import FieldABC, SchemaABC
-from marshmallow.utils import EXCLUDE, is_collection
+from marshmallow.utils import RAISE, is_collection
 from marshmallow.utils import missing as missing_
 from marshmallow.compat import text_type, basestring
 from marshmallow.exceptions import ValidationError, StringNotCollectionError
@@ -393,7 +393,7 @@ class Nested(Field):
         self.only = only
         self.exclude = exclude
         self.many = kwargs.get('many', False)
-        self.unknown = kwargs.get('unknown', EXCLUDE)
+        self.unknown = kwargs.get('unknown', RAISE)
         self.__schema = None  # Cached Schema instance
         self.__updated_fields = False
         super(Nested, self).__init__(default=default, **kwargs)
--- a/tests/test_decorators.py
+++ b/tests/test_decorators.py
@@ -471,7 +471,7 @@ class TestValidatesSchemaDecorator:
                 raise ValidationError('Method called')
 
         class MySchema(Schema):
-            nested = fields.Nested(NestedSchema, required=True, many=True)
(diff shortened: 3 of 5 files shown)


sloria:
It might be good to have a test for the default behavior. Perhaps the existing tests should be updated to pass with `RAISE` instead of changing `unknown` to `EXCLUDE`?

tuukkamustonen:
Sorry for late response. All the tests that I modified are actually testing something else than the unknown keyword validation (`@schema_validates` decorator, deserialization, ``dump_only``/``load_only``) so that's why I left their behavior as was.

Of course, it would be best to test them with combination of all ``unknown`` options (where it makes sense). I'll try to find where the generic ``Nested``/``unknown`` tests are and update/add there.

While updating the tests I bumped into https://github.com/marshmallow-code/marshmallow/issues/875#issuecomment-412347945, you may want to check it.

lafrech:
+1 @tuukkamustonen

In https://github.com/marshmallow-code/marshmallow/pull/872, I also passed `unknown=EXCLUDE` in most tests for the same reason as they are meant to test something else.

Indeed, one test with all three values would be nice.

sloria:
@tuukkamustonen Do you have time to update the tests?

tuukkamustonen:
Yeah, sorry. I got a bit stuck elsewhere :)

Anyway, will do and ping then.

tuukkamustonen:
@sloria Added `test_unknown_validation` in `test_schema.py`, to test the combinations.

It ruthlessly exceeds 80 char limit (because it's test data, not code). Is that a problem?

I didn't touch other tests (due to reasoning I wrote earlier).

tuukkamustonen:
> It ruthlessly exceeds 80 char limit (because it's test data, not code). Is that a problem?

Flake8 test failed so I think that's a yes :D I'll fix.

(Does running flake8 on test code make sense? I always include tests, because you want to sometimes bend the rules there for flexibility.)

tuukkamustonen:
Ok, should be good now.

lafrech:
We do intend to flake8 the tests. flake8 enforce a set of rules that are not too restrictive.

I agree that some pylint default rules are difficult to respect in the tests and irrelevant: function names, number of variables,... But flake8 is easy to please.

Note that in marshmallow, the max length is set to 100 in [setup.cfg](https://github.com/marshmallow-code/marshmallow/blob/dev/setup.cfg).

lafrech:
Thanks for that new test.

Was the grandchild necessary to test `Nested`? AFAIU, the test on the first level should be enough.

tuukkamustonen:
> Was the grandchild necessary to test Nested? AFAIU, the test on the first level should be enough.

Not really, but I assumed you'd rather want it than not. Anyway, I can simplify it, if you want?

What I mean is that I don't know the library that well, so I thought it might reveal something and ended up adding grand child to the test. Well, it didn't reveal anything but I left it there anyhow.

lafrech:
Yeah, I understand. It tend to do that as well.

As the tests are piling up, it is nice to keep them as focused as possible, to make them clear about their intent, and to avoid complexifying refactors unnecessarily.

Thanks for your work already. If you have time, you may simplify it, it'll be appreciated. Otherwise, I can try to find some time to do it later, even after this is merged.

Also, please rename the test as `test_nested_unknown_validation`.

Thanks again.

tuukkamustonen:
@lafrech Sorry again, I've been afk. I'll simplify the test.

About naming though: it's already member of `TestNestedSchema` class, that's why I left "nested" away from method name. But yeah, I see all the other methods have "nested" in their name so I'll add it.

lafrech:
No pb.

I think the rationale about the naming is that you may want to run 

    py.test -k my_test_function

I can't tell for marshmallow, but in my application tests, I write long function names including all the hierarchy (like `test_module_class_function_behaviour`) to avoid duplicate function names and let `-k` be more selective.

tuukkamustonen:
Good to know about `-k` conventions - I just normally run tests within IDE and rarely need selective runs on command-line.

Anyway, pushed the changes.

Edit: Resolved conflicts.

lafrech:
Looks good to me. Thanks!

PR #913, opened 2018-08-14
Update docs to reflect change in default behaviour for unknown fields

declantraynor:
#872 introduced a change in the default behaviour of the `unknown` keyword argument passed to `Schema.load`. Looks like the docs update was missed.

Thanks for all the great work so far!


sloria:
Good catch. Thanks @declantraynor !

PR #917, opened 2018-08-17
Closes #916: Excluding fields during deserialization of object

### LINKED ISSUE #916: Excluding fields during deserialization of object
The tuple notation seems to cause problems excluding fields during the serialization process. That documentation states that the following should work:

``` python
summary_schema = UserSchema(exclude=('name'))
summary_schema.dump(user)
# {"email": "monty@python.org", ...}
```

Unfortunately, specifying a tuple does not work if it only contains one element (in this case `name`). What works is to simply pass a list instead of a tuple, like this:

``` python
summary_schema = UserSchema(exclude=['name'])
summary_schema.dump(user)
# {"email": "monty@python.org", ...}
```
I think the documentation should be adapted to avoid confusion.

milost:
Adaptation of the documentation so that the exclude case is described more clearly.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -72,11 +72,17 @@ You may not need to output all declared fields every time you use a schema. You 
 
 .. code-block:: python
 
-    summary_schema = UserSchema(only=('name', 'email'))
+    summary_schema = UserSchema(only=['name', 'email'])
     summary_schema.dump(user)
     # {"name": "Monty Python", "email": "monty@python.org"}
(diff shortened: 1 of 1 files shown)


deckar01:
I don't think this is necessary. There are a lot of other places tuples are used in the docs. It might be a pitfall for those unfamiliar with the single item tuple syntax, but it is the correct language feature to use.

sloria:
I agree with @deckar01 ; this is

PR #926, opened 2018-08-28
Fix Nested deserialization for only=string and data_key

timakro:


lafrech:
`only=string` is being removed in https://github.com/marshmallow-code/marshmallow/pull/823 to be replace by a dedicated `Pluck` field, so the problem this PR addresses should disappear as soon as #823 is finalized and merged.

timakro:
Cool, didn't know that.

Issue #520, opened 2016-08-31
Add load_as_tz_aware (and dump_as_tz_aware ?) parameter to DateTime


lafrech:
As said [in the docs](http://marshmallow.readthedocs.io/en/latest/api_reference.html#marshmallow.fields.DateTime),

> Schema.load returns datetime objects that are timezone-aware.

I'm using webargs to parse input to an API and the code expects naive `datetime`s. I would like to get naive `datetime`s, rather than having to remove the TZ from every date attribute in each resource method.

Maybe I'm wrong working with naive `datetime`s in the first place. I have no opinion about it. But sometimes you're working with a lib that expects naive `datetime`s, and you don't want to patch it. 

I guess the most flexible approach would be to provide a field parameter allowing to choose which format a `DateTime` field should output.

Could it be possible to have this configured with a flag/metadata in the `DateTime` field?

Or is there a design reason not to allow this?

(Apparently, I'm not the first one falling into this: https://github.com/marshmallow-code/marshmallow/issues/309.)

maximkulkin:
I wonder if the problem is in Marshmallow DateTime field implementation or in third-party software that in 2016 still uses naive `datetime`s and can't handle time zone aware ones. Maybe you need to be knocking their door to finally support that.

Anyways, it is easy to implement:

``` python
import marshmallow.fields as mf

class MyNaiveDateTime(mf.DateTime):
    def _deserialize(self, value, attr, data):
        result = super(MyNaiveDateTime, self)\
            ._deserialize(value, attr, data)
        return result.replace(tzinfo=None)
```

lafrech:
I was wrong, as explained [here](https://github.com/marshmallow-code/marshmallow/issues/309#issuecomment-244076069): `DateTime` deserializes a naive datetime string as a naive `datetime`. The doc is a bit misleading.

Now, assuming
- my API exchanges datetimes as UTC according to ISO_8601 (e.g. `1966-05-24T00:00:00+00:00`)
- my business code assumes everything is UTC and works with naive `datetime`s

I still think it would be nice to be able to enforce a format for deserialization.

And in fact, I could extend this request to serialization as well.

Maybe my business code should be all TZ aware datetimes (Is this what you suggest?). 

Thanks for the subclass hint.

> Maybe you need to be knocking their door to finally support that.

Done: https://github.com/lepture/flask-oauthlib/issues/294

lafrech:
[...]

PR #933, opened 2018-09-02
WIP - Add AwareDateTime and NaiveDateTime

lafrech:
Tackling https://github.com/marshmallow-code/marshmallow/issues/520.

This provides `AwareDateTime` and `NaiveDateTime` fields.

Currently, only deserialization is supported. Those fields enforce awareness/naiveness when deserializing (which IMHO is critical if application code only deals with one type of datetime, see #520).

They take a timezone as parameter to use as default, but `AwareDateTime` keeps input timezone if there is one and `NaiveDateTime` uses it only to convert a datetime expressed with a non-UTC timezone to UTC before removing the timezone.

Deserialization is only tested with iso8601, not rfc822.

Docstrings are drafts and should be reworded.

I've been using `AwareDateTime` for a while in my applications.

Serialization is a bit harder because it conflicts with utils functions `isoformat` and `rfcformat`. I didn't look into rfc but `isoformat`, for instance, forces serialization as aware. A greater refactor would be needed.

The whole datetime management is pretty complicated due to several formats, plus the fact that dateutil may be available or not.

Adding those fields with only deserialization would be quite useful already (I never needed serialization), but it looks like job half done.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -999,6 +999,56 @@ class LocalDateTime(DateTime):
     localtime = True
 
 
+class AwareDateTime(DateTime):
+    """A formatted timezone aware datetime string.
+
+        Example: ``'2014-12-22T03:12:58.019077+00:00'``
+
+    Takes the same arguments as :class:`DateTime <marshmallow.fields.DateTime>`,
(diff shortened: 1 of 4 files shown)


lafrech:
Obsolete.

Addressed in #1234.

PR #936, opened 2018-09-03
Fixes #899: Make from_iso_date method use isoparse from dateutils

### LINKED ISSUE #899: fields.Date accepts input in non ISO8601 format. This leads to unaccepted behaviour.
The doc of `fields.Date` suggests that it accepts only ISO8601 date format.

```python
class Date(Field):
    """
```

It works with ISO8601 format fine:
```python
In [2]: fields.Date().deserialize('2018-10-08')
Out[2]: datetime.date(2018, 10, 8)
```

However unfortunately it also works with non ISO8601 formats:
```python
In [3]: fields.Date().deserialize('08-10-2018')
Out[3]: datetime.date(2018, 8, 10)
```
And in this case it seems like it parses the format as `%m-%d-%Y`
I feel, it should raise a ValidationError with `invalid_date` format incase the input is not in the iso8601 format.

Digging further into the issue I found that DateField uses `utils.from_iso_date`; which inturn uses dateutil (if available) parser to parse the date.
I think that util should be modified to use [dateutil.parser.isoparse](https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.isoparse)

dushr:
Ref #899 

I have made `utils.from_iso_date` util method use `dateutil.parser.isoparse`

However I also noticed that `utils.from_iso_datetime` and `utils.from_isotime` also use the incorrect parser. Should I make the changes to them in this PR itself or should I make separate PRs.

### DIFF
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -349,6 +349,8 @@ class TestFieldDeserialization:
             42,
             '',
             [],
+            dt.datetime.now().strftime('%H:%M:%S %Y-%m-%d'),
+            dt.datetime.now().strftime('%m-%d-%Y %H:%M:%S'),
         ],
     )
     def test_invalid_datetime_deserialization(self, in_value):
@@ -541,6 +543,7 @@ class TestFieldDeserialization:
             '',
             123,
             [],
+            dt.date(2014, 8, 21).strftime('%d-%m-%Y'),
         ],
     )
     def test_invalid_date_field_deserialization(self, in_value):
--- a/marshmallow/utils.py
+++ b/marshmallow/utils.py
@@ -284,7 +284,7 @@ def from_iso_datetime(datestring, use_dateutil=True):
         raise ValueError('Not a valid ISO8601-formatted datetime string')
     # Use dateutil's parser if possible
     if dateutil_available and use_dateutil:
-        return parser.parse(datestring)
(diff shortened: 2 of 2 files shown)


lafrech:
Great.

> However I also noticed that utils.from_iso_datetime and utils.from_isotime also use the incorrect parser. Should I make the changes to them in this PR itself or should I make separate PRs.

Same PR is fine.

Thanks.

dushr:
Cool, done. 
I have changed `from_iso_date` and `from_iso_datetime`

Unfortunately `dateutil.parser.isoparse` only works for dates and not time, so I haven't changed `from_iso_time`; and I believe it has the correct behaviour.

tests/test_deserialization.py
```diff
@@ -349,6 +349,8 @@ def test_boolean_field_deserialization_with_custom_falsy_values(self):
             42,
             '',
             [],
+            dt.datetime.now().strftime('%H:%M:%S %Y-%m-%d'),
```
lafrech, line 352: I think this datetime string was rejected before your change as well. It is not the use of `isoparse` that makes it rejected, but the regex. See [my comment](https://github.com/marshmallow-code/marshmallow/issues/899#issuecomment-418042719) and my proposal to also uses Django regexes for date and time.

sloria:
This was integrated in #968

PR #941, opened 2018-09-09
Documents attribute retrieval

### LINKED ISSUE #776: Documentation error:  utils.get_value attempts to __getitem__ before getattr
I had a problem with a very basic use case (see https://stackoverflow.com/questions/49889233) and it turns out the explanation was that `Field.get_value()` -> `utils.get_value()` tries `__getitem__` and only falls back to `getattr` in event of an error. 

The first section of the Quickstart reads: 

> Create a schema by defining a class with variables mapping attribute names to Field objects.

Strictly speaking, this is not true, since `__getitem__` keys are not attribute names. This should be clarified somewhat exhaustively in the docs, as any object that does not raise an error on an invalid `__getitem__` key will not be marshallable.

remeika:
- Documents how attributes are retrieved from objects by default during serialization.
- Warn users that objects must raise an exception on missing key (`x[‘I’]` before the default accessor will try to retrieve an object attribute (`x.i`).

Fixes #776

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -105,3 +105,4 @@ Contributors (chronological)
 - Prasanjit Prakash `@ikilledthecat <https://github.com/ikilledthecat>`_
 - Guillaume Gelin `@ramnes <https://github.com/ramnes>`_
 - Maxim Novikov `@m-novikov <https://github.com/m-novikov>`_
+- James Remeika `@remeika <https://github.com/remeika>`_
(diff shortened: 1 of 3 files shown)


sloria:
Thanks!

cancan101:
Following up on this, is there any easy way to override the `accessor` on a field much like you can set the `attribute`?

sloria:
@cancan101 You can override the `get_value` value on a `Field` class.

PR #945, opened 2018-09-10
Add allow_nan to Float field

### LINKED ISSUE #944: RFC: extend allow_nan parameter to all number fields
`Decimal` field has an `allow_nan` parameter (`False` by default):

> If `True`, `NaN`, `Infinity` and `-Infinity` are allowed, even though they are illegal according to the JSON specification.

Any objection to extend it to all numbers?

Currently, `Int` will reject those as they are not integers, but they pass `Float` which just pawned me.

Should we add the same validation in `Float`? Move it up to `Number`?

Even if it is useless for `Int`, I'd be tempted to extend it to `Number`.

- Those are special values with a special meaning so a different error message makes sense.
- Any `Number` subclass would inherit it -> less duplication.

This is a breaking change as the default would be `False` so `Float` would need to be explicitly passed `allow_none=True`.

lafrech:
Closes #944.

TODO:

- [x] Test
- [x] Doc

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,12 +6,17 @@ Changelog
 
 Features:
 
-- Add ``fields.Pluck`` for serializing a single field from a nested object (:issue:`800`). Thanks :user:`timc13` for the
-  feedback and :user:`deckar01` for the implementation.
-- *Backwards-incompatible*: Passing a string argument as ``only`` to ``fields.Nested`` is no longer supported.
-  Use ``fields.Pluck`` instead (:issue:`800`).
+- Add ``fields.Pluck`` for serializing a single field from a nested object
+  (:issue:`800`). Thanks :user:`timc13` for the feedback and :user:`deckar01`
+  for the implementation.
+- *Backwards-incompatible*: Passing a string argument as ``only`` to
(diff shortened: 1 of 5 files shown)


lafrech:
The error message is not that explicit:

    'Special numeric values are not permitted.'

Perhaps change it to 

    'Special numeric values (nan or infinity) are not permitted.'

sloria (approved):
Change looks good to me. Good to merge when tests and docs are added.

lafrech:
Done.

I'm not absolutely found of the `allow_nan` naming (could be `allow_specials`,...) and the error message (see above). In fact I'm tied between too generic (special values), too restrictive (nan) and too comprehensive (nan, infinity,...).

I'm fine if we keep them as is.

sloria:
I don't feel strongly about the `allow_nan` name. `allow_specials` *could* work, but it's I think it's less self-descriptive.

Changing the error message to your suggestion ('Special numeric values (nan or infinity) are not permitted.') seems like a good change. Let's go for it.

sloria:
Thanks for merging, @lafrech . I'll plan to release b14 this weekend, once I add the docs stuff for Tidelift.

PR #952, opened 2018-09-14
Pass render kwargs

Tecktron:
This adds 2 new params to the Meta class, `render_dumps_kwargs` and `render_loads_kwargs`. These are then passed to the `render_module` when `.dumps()` or `.loads()` is invoked . 
Test and documentation included.

### DIFF
--- a/tests/test_options.py
+++ b/tests/test_options.py
@@ -6,7 +6,7 @@ import pytest
 
 from marshmallow import fields, Schema, EXCLUDE
 
-from tests.base import User
+from tests.base import User, Item, DecimalAsFloatEncoder
 
 
 class TestUnordered:
@@ -209,3 +209,22 @@ class TestIncludeOption:
         assert 'email' in s._declared_fields.keys()
         assert 'from' in s._declared_fields.keys()
         assert isinstance(s._declared_fields['from'], fields.Str)
+
+
+class TestJsonKwargs:
+
+    class SimpleSchema(Schema):
+        name = fields.Str()
+        cost = fields.Decimal(places=2)
+
+        class Meta:
+            render_dumps_kwargs = {
+                'cls': DecimalAsFloatEncoder,
+                'sort_keys': True,
+            }
+
+    def test_dump_returns_cost_as_float(self):
+        schema = self.SimpleSchema()
+        i = Item('Marshmallow', cost=3.50)
+        result = schema.dumps(i)
+        assert result == '{"cost": 3.5, "name": "Marshmallow"}'
--- a/tests/test_options.py
+++ b/tests/test_options.py
@@ -211,7 +211,7 @@ class TestIncludeOption:
         assert isinstance(s._declared_fields['from'], fields.Str)
 
 
-class TestJsonKwargs:
+class TestRenderKwargs:
 
     class SimpleSchema(Schema):
         name = fields.Str()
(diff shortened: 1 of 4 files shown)


lafrech:
Looks good to me.

Please add a test for `render_loads_kwargs`. And maybe name the test with the feature, like `test_render_dumps_kwargs` rather than `test_dump_returns_cost_as_float`.

deckar01:
Can render modules be configured before being passed into the schema to avoid the need for arg passing?

deckar01:
Is there a reason the existing `*args` and `**kwargs` arguments on `Schema.loads` and `Schema.dumps` can't be used?

Tecktron:
Test added and renamed.

To answer your first question, yes you could do that, but what if you want different params for each schema, now you'll have to create multiple version of a render_module.

As for the second, again you could just always pass these in to very single `.dumps/loads()` call that you make, but am trying to make it easy to avoid having to always do that. Also the way this works, those will still allow you to override these, should you need to do so.

tests/test_options.py
```diff
@@ -222,9 +223,19 @@ class Meta:
...
+            }
 
-    def test_dump_returns_cost_as_float(self):
+    def test_render_dumps_kwargs(self):
         schema = self.SimpleSchema()
         i = Item('Marshmallow', cost=3.50)
         result = schema.dumps(i)
         assert result == '{"cost": 3.5, "name": "Marshmallow"}'
+
+    def test_render_load_kwargs(self):
```
lafrech: s/render_load/render_loads

Tecktron: nice catch. fixed

deckar01:
This still feels unnecessary to me. How many places are you calling `dump`/`load` per schema? How many schemas are in your app? In my experience most apps call `dump`/`load` once per schema and have many schemas, which makes using the render args on the constructor unnecessary. If you have several schemas that need the same custom renderer, then you would be better of configuring the renderer once before passing it to all the schemas. If you have one schema or several unique renderers, I think that is not a common enough use case to justify the added complexity to the API. Your use case is supported by marshmallow, it just doesn't have a dedicated API.

I think a custom renderer class should satisfy this use case.

```py
class CustomRenderer(object):
    renderer = json
    dump_kwargs = None
    load_kwargs = None

    @classmethod
    def dumps(self, *args, **kwargs):
        kwargs.update(self.dump_kwargs or {})
        self.renderer.loads(*args, **kwargs)

    @classmethod
    def loads(self, *args, **kwargs):
        kwargs.update(self.load_kwargs or {})
        self.renderer.loads(*args, **kwargs)
```

```py
class MyRenderer(CustomRenderer):
    dumps_kwargs = {
        'cls': DecimalAsFloatEncoder,
        'sort_keys': True,
    }
    loads_kwargs = {
        'cls': NaNAsZeroDecoder,
    }
```

Tecktron:
Well it's here, take it or leave it. 

There are many other ways of solving this as well, for example overriding the entire class and implementing what I did that way. That's what I'm currently doing and just thought I'd share it. 

Overall my thought was more about if you accept an object and call it's functions, why not have a way to accept its function's parameters as well ("_Explicit is better than implicit._" type of thing) 

I use several of these things, mostly because I have to deal with multiple apis, many of which are weird and legacy and have issues I'm not even going to get into (from date formats to key ordering), so to be able to base a schema off of another schema and update the Meta class so that it can pass lets say `sort_keys`, to me, meant less and more consolidated code.

I realize this may not be the common case, but I not gonna fuss over 12 lines of code (outside of the tests) if you feel that strongly about not giving users this option.

sloria:
I agree with @deckar01 ; there are sufficient ways to achieve the OP without adding API surface.

Thanks anyway for taking the time to work on this @Tecktron and informing us of your use case.

Issue #930, opened 2018-08-30
Don't accept any iterable with many=True


tuukkamustonen:
With `many=True`, it probably doesn't make sense to accept any iterable (such as string or dict), simply because the user probably didn't intend to pass it.

Accepting any iterable (such as some "string") produces weird outcome with no good. It's better to fail fast and record error on invalid type.

So, maybe reject some specific iterables:

- `str`
- `dict` (is iterable, but would be weird and  you cannot have dict as key anyway)

(Originating this from #622 PR, which suggested even more restrictive behavior.)

Issue #779, opened 2018-04-19
RFC: Remove Nested(many=True) to keep only the List(Nested()) form


lafrech:
See discussion in https://github.com/marshmallow-code/marshmallow/issues/493#issuecomment-380468027.

We have two ways to do the same thing. They sometimes behave differently, which generally means one of them is buggy.

IMHO, and from what I have seen from users around me, it is more natural for a new user to write `List(Nested)` than `Nested(many)`. Besides, it is consistent with `Dict(values=Nested)`.

I realize this is quite a breaking change, both from user perspective, because `Nested(many=True)` is probably the most commonly used form, and from developer perspective, because it involves a lot of code changes (but also a lot of code removal, which is nice).

There could be things that can be done with `Nested(many=True)` but not with `List(Nested)`, because in the former case, the nested Schema has access to the whole data, while in the latter, it is called on each element. Perhaps some `pre_load` edge cases might suffer from the change. Any real-life example, anyone?

Anyway, since the question was raised in a comment in another issue, I figured I'd give it more visibility.

seansfkelley:
Is there any documentation outlining the differences between these two? I find the distinction confusing as well.

deckar01:
I don't think there is supposed to be a difference.

sloria:
I think deprecating `Nested(many=True)` is a good idea. Would someone be up for doing an analysis of the current differences between the two APIs?

lafrech:
Do you mean deprecate in 2.X and remove in 3 or deprecate in 3?

Here are a few differences:


- Validation error issue with `Nested(many=True)` (https://github.com/marshmallow-code/marshmallow/issues/493#issuecomment-380468027) (`List(Nested)` does the right thing.)
- Different error message when deserializing an invalid input (not a list) (https://github.com/marshmallow-code/marshmallow/pull/954#issuecomment-421916563) (`List(Nested)` does the right thing.)
- `List(Nested)` prevents pre/post processors in nested schema to work on the whole data collection] (https://github.com/marshmallow-code/marshmallow/issues/1801)

------------------------------

- [Partial data if an item fails validation](https://github.com/marshmallow-code/marshmallow/issues/766) [#766] (Fixed in #1052)
- [`List` is not propagating the context between a nested field and the parent schema](https://github.com/marshmallow-code/marshmallow/issues/922) [#922] (Fixed)
[...]

PR #954, opened 2018-09-16
Fixes for #906 (TypeError with many=True) and #907 (return [] not {} on many=True error) and #930 (disallows strings and dicts on many=True)

lafrech:
#910 + minor rework

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -336,7 +336,6 @@ def test_dumps_many():
     assert len(data) == 2
     assert data[0] == s.dump(u1)
 
-
 def test_load_returns_an_object():
     s = UserSchema()
     result = s.load({'name': 'Monty'})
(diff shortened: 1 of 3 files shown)


tests/test_schema.py
```diff
@@ -1633,6 +1677,22 @@ def test_validate_with_custom_error_handler(self):
 
 class TestFieldValidation:
 
+    @pytest.mark.parametrize('val', (None, False, 1, 1.2, object()))
+    def test_non_iterables_with_many(self, val):
```
sloria: This test appears misplaced: it's not really about field validation, as the class name suggests. Perhaps this can just be a top-level test function.

tests/test_schema.py
```diff
@@ -350,6 +350,50 @@ def test_load_many():
     assert type(result[0]) == User
     assert result[0].name == 'Mick'
 
+def test_load_many_input_types():
```
sloria: This is largely redundant with `test_non_iterables_with_many`. Maybe remove this in favor of that one, and separate out the empty collection cases (`s.load([])` and `s.load(set())`) into a different test function.

sloria (changes requested):
Looks good. Just a few suggestions re: tests. Once those are addressed, this can be merged.

lafrech:
See updated tests.

I noticed another discrepancy between List(Nested) and Nested(many=True). See the TODO in the tests.

```py
@pytest.mark.parametrize('val', (False, 1, 1.2, object(), {}, {'1': 2}, 'lol'))
def test_load_many_in_nested_invalid_input_type(val):
    class Inner(Schema):
        name = fields.String()
    class Outer(Schema):
        list1 = fields.List(fields.Nested(Inner))
        list2 = fields.Nested(Inner, many=True)
    with pytest.raises(ValidationError) as e:
        Outer().load({'list1': val, 'list2': val})
    # TODO: Error messages should be identical (#779)
    assert e.value.messages == {'list1': ['Not a valid list.'], 'list2': ['Invalid type.']}
```

Unless you think of a quick fix, we can add it to #779 and merge this.

sloria (approved):

sloria:
Let's defer the error message discrepancy for now and get this merged.

sloria:
Hm, on further review, I don't think this should be backported to 2.x.

We decided that raising an error when deserializing None (#511) was a backwards-incompatible change, and this is a similar change.

As for `valid_data`--that doesn't exist in 2.x.

lafrech:
I think #906 is a bug and could be fixed in 2.x.

```py
        if many and data is not None:
            self._pending = True
            ret = [self.deserialize(d, fields_dict, many=False,
                        partial=partial, dict_class=dict_class,
                        index=idx, index_errors=index_errors)
                        for idx, d in enumerate(data)]
```

If `data` is not "a sequence, an iterator, or some other object which supports iteration", enumerate will raise a `TypeError`. Fixing this wouldn't be a breaking change (unless someone relies on a `TypeError` being raised...).

This only affects `Schema(many=True)`, not `Nested(many=True)`. The test in `Nested(many=True)` is

```py
        if self.many and not utils.is_collection(value):
            self.fail('type', input=value, type=value.__class__.__name__)
```

AFAIU, this is not a security issue. I probably won't affect most APIs as on POST or PUT, you don't expect a plurality of items to be sent. Worst case scenario, you get a 500 error, not some data being leaked.

sloria:
@lafrech Good point. Backport for the #906 fix is in #959 .

PR #959, opened 2018-09-20
[backport] Prevent TypeError for non-collection with many=True

### LINKED ISSUE #906: TypeError thrown with many=True on non-iterable types

sloria:
fixes #906
this is manual backport from #954

### DIFF
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -220,20 +220,26 @@ class Unmarshaller(ErrorStore):
         :return: A dictionary of the deserialized data.
         """
         if many and data is not None:
-            self._pending = True
-            ret = [self.deserialize(d, fields_dict, many=False,
-                        partial=partial, dict_class=dict_class,
-                        index=idx, index_errors=index_errors)
-                    for idx, d in enumerate(data)]
+            if not is_collection(data):
+                errors = self.get_errors(index=index)
+                self.error_field_names.append(SCHEMA)
+                errors[SCHEMA] = ['Invalid input type.']
+                ret = []
+            else:
+                self._pending = True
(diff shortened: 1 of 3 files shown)


marshmallow/marshalling.py
```diff
@@ -220,20 +220,26 @@ def deserialize(self, data, fields_dict, many=False, partial=False,
...
-            self._pending = False
-            if self.errors:
-                raise ValidationError(
-                    self.errors,
-                    field_names=self.error_field_names,
-                    fields=self.error_fields,
-                    data=ret,
-                )
+                self._pending = False
+                if self.errors:
```
lafrech, line 236: I would have left this indented one level less, as it was before, so as to raise the `ValidationError`. That would be consistent with the rest of the method.

I was surprised the test passed anyway, so I investigated and it works because in `Schema._do_load` there is [a test at the end](https://github.com/marshmallow-code/marshmallow/blob/2.x-line/marshmallow/schema.py#L686) and a `ValidationError` is raised if there are errors.

It's a bit blurry to me but I find it smelly that we can do things two different ways. Looks like we don't need to raise `ValidationError` at all in this method, and just setting errors should be enough.

Maybe something to investigate on MA3. For now, we just need to get the bug fixed.

lafrech:
I wrote that review but didn't notice I had to _submit_ it...

Never mind. Let's follow this in #961.

PR #960, opened 2018-09-20
Fix Collection Root Inheritance

### LINKED ISSUE #956: Root reference is broken for nested/container fields when using schema inheritance
I use schema inheritance a lot to reuse shared parts of schemas. Additionally I have some custom fields with validation methods which use references to the field's root schema. When these custom fields are used in nested schemas or inside lists, then the same field instance is used in different schema classes and thus the root reference is "broken":

```python
>>> import marshmallow
>>> marshmallow.__version__
'3.0.0b13'
>>> from marshmallow.schema import Schema
>>> class A(Schema):
...     f1 = fields.List(fields.Integer())
... 
>>> class B(A):
...     pass
... 
>>> a = A()
>>> b = B()
>>> a_f1 = a.declared_fields['f1']
>>> b_f1 = b.declared_fields['f1']
>>> a_f1.root == a
True
>>> b_f1.root == b
True
>>> a_f1.container.root == a
False
>>> a_f1.container.root == b
True
```
So the problem is that the container root of field `f1` of schema instance `a` is schema instance `b`. In fact `a_f1.container` and `b_f1.container` point to the same object.

Is this a problem of how I use marshmallow or is this a bug?

I find the possibility to reference a field's root schema very valuable and the root property looks like it can be used like that. However, I think the problem here is that nested (or container) fields are not unique instances when used in multiple classes.

deckar01:
Fixes #956 

The schema uses `copy.deepcopy()` to fork field instances before mutating them, but that operation does not fork a collection field's container attribute. This causes the attribute to be overwritten when the forked fields mutate the shared container instance.

`Field._add_to_schema()` is called after copying the parent schema's fields and is where the containers are mutated, so the containers must be forked there.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -4,6 +4,7 @@
 from __future__ import absolute_import, unicode_literals
 
 import collections
+import copy
 import datetime as dt
 import numbers
 import uuid
@@ -568,6 +569,7 @@ class List(Field):
 
     def _add_to_schema(self, field_name, schema):
         super(List, self)._add_to_schema(field_name, schema)
+        self.container = copy.deepcopy(self.container)
         self.container.parent = self
         self.container.name = field_name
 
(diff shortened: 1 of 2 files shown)


lafrech:
LGTM.

The `List` fix could be backported to 2.x. (I just checked it is affected too.)

chrisbune:
Thanks for the quick fix. Looks good.

PR #961, opened 2018-09-21
Refactor unmarshall: don't raise ValidationError in deserialize

lafrech:
As I said in #959 review comment, since `Schema._do_load` will raise on errors, we don't need to raise `ValidationError` in `Unmarshaller.deserialize`.

I removed what looks redundant and the only test that is broken tests `Unmarshaller` directly. Otherwise, the `Schema` interface is unchanged.

This is just a quick test I did out of curiosity. This looks like a design issue and maybe it calls for deeper reflection.

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -620,18 +620,15 @@ class BaseSchema(base.SchemaABC):
         else:
             processed_data = data
         if not errors:
-            try:
-                result = unmarshal(
-                    processed_data,
-                    self.fields,
-                    many=many,
-                    partial=partial,
-                    unknown=unknown,
-                    dict_class=self.dict_class,
-                    index_errors=self.opts.index_errors,
-                )
-            except ValidationError as error:
-                result = error.data
+            result = unmarshal(
+                processed_data,
+                self.fields,
+                many=many,
+                partial=partial,
+                unknown=unknown,
+                dict_class=self.dict_class,
+                index_errors=self.opts.index_errors,
+            )
             self._invoke_field_validators(unmarshal, data=result, many=many)
             errors = unmarshal.errors
             # Run schema-level validation.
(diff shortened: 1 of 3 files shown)


sloria (approved):
Huh, good catch.

I missed your review on #959. Did you submit it?

Anyway, LGTM.

marshmallow/schema.py
```diff
@@ -620,18 +620,15 @@ def _do_load(
...
+                self.fields,
+                many=many,
+                partial=partial,
+                unknown=unknown,
+                dict_class=self.dict_class,
+                index_errors=self.opts.index_errors,
+            )
             self._invoke_field_validators(unmarshal, data=result, many=many)
             errors = unmarshal.errors
             # Run schema-level validation.
```
deckar01, line 634: `_invoke_schema_validators()` is calling `unmarshal.run_validator()` and expecting `ValidationError` even though `unmarshal.run_validator()` catches `ValidationError` and does not raise it. It looks like all the error collecting/raising in `_invoke_schema_validators()` is dead code. All those errors build up in the error store and get raised at the end of `_do_load()`.

deckar01 (approved):
I think these changes are good as is. If you think the code I pointed out can be removed in this MR also, even better.

lafrech:
I removed some more useless try/catch.

I still think this part is a bit convoluted. It works because the unmarshaller mutates the errors dictionary, but I don't find it really clear. I'm not sure it was really intended to work like this.

This deserves a refactor, deciding what is the role of unmarshaller and processors, and which is responsible for raising errors and holding the error structure.

Meanwhile, we can merge this as a first step.

sloria:
I agree that the unmarshaller code could use further refactoring.

This is already better than the status quo, though, so let's get it merged.

PR #963, opened 2018-09-22
[BUG] If fields.Nested.unknown undefined it should fallback to schema.unkown instead RAISE

vgavro:
If fields.Nested.unknown is undefined, it should fallback to schema defaults instead of RAISE.

consider following example:
```
class Nested(Schema):
    class Meta:
        unknown = EXCLUDE

class MySchema(Schema):
    class Meta:
        unknown = EXCLUDE
    nested = fields.Nested(Nested)
```
This will raise ValidationError on unknown field if fields.Nested(unknown=EXCLUDE) is not set, and it's not obvious. With this fix fields.Nested can override unknown, but will fallback to NestedSchema default if argument not provided.

### DIFF
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 import pytest
 
-from marshmallow import fields, Schema, ValidationError, EXCLUDE
+from marshmallow import fields, Schema, ValidationError, EXCLUDE, INCLUDE, RAISE
 from marshmallow.marshalling import missing
 from marshmallow.exceptions import StringNotCollectionError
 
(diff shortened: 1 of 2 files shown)


lafrech:
I agree with this fix. Nested should not set its default value (`RAISE`) to the nested schema. The value of the nested schema should be used. This patch fixes that the right way.

Do you think you could add a test case?

vgavro:
@lafrech yes, of course, i'll commit tests asap. thanks.

tests/test_fields.py
```diff
@@ -222,3 +222,20 @@ class TestNestedField:
     def test_nested_only_and_exclude_as_string(self, param):
         with pytest.raises(StringNotCollectionError):
             fields.Nested(Schema, **{param: 'foo'})
+
+    @pytest.mark.parametrize('unknown', (None, 'exclude', 'include', 'raise'))
+    def test_nested_unknown_override(self, unknown):
+        class NestedSchema(Schema):
+            class Meta:
+                unknown = 'exclude'
```
lafrech: Would it make sense to also parametrize the Meta attribute, so that all combinations are covered?

Current test would also pass if unknown defaulted to `EXCLUDE` instead of `RAISE`.

Also, in the rest of the tests, we use the capitalized global (`INCLUDE`) rather than its string value (`'include'`). Do you think you could change that?

Thanks.

lafrech:
Good. Once my two nitpicks are addressed, I think this is good to go.

lafrech:
Perfect. Thanks @vgavro.

PR #968, opened 2018-09-26
Enforce ISO8601 when deserializing date and time

### LINKED ISSUE #899: fields.Date accepts input in non ISO8601 format. This leads to unaccepted behaviour.

lafrech:
Closes #899, #936.

Built upon @dushr's #936. I picked the regexes from Django for time and date like it has been done for datetime.

It makes `Date` and `Time` fields more picky about the input format, so it could be considered a breaking change. Or a fix, since the docs say "ISO8601-formatted date/time string".

Anyway, it makes things consistent between `DateTime`, `Date` and `Time`.

TODO:

- [x] Update CHANGELOG
- [x] ~Update docs (upgrading.py) ? Not sure, if it is a bugfix.~

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -107,3 +107,4 @@ Contributors (chronological)
 - Maxim Novikov `@m-novikov <https://github.com/m-novikov>`_
 - James Remeika `@remeika <https://github.com/remeika>`_
 - Karandeep Singh Nagra `@knagra <https://github.com/knagra>`_
+- Dushyant Rijhwani `@dushr<https://github.com/dushr>`_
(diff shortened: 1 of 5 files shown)


dushr:
👍

lafrech:
@sloria I think this is ready to merge.

It makes things more consistent and reliable: input should match ISO 8601 rather than comply with undefined external lib behaviour.

sloria (approved):
Thanks @dushr and @lafrech !

PR #970, opened 2018-09-30
Update docstring to `post_load` from `make_object`

eihli:
`make_object` was deprecated

https://marshmallow.readthedocs.io/en/3.0/upgrading.html?highlight=make_object#use-post-load-instead-of-make-object


sloria:
Good catch. For consistency with the docstring of `Schema.dump`, I just removed the `make_object` bit altogether.

PR #982, opened 2018-10-08
Allow username without password in basic auth part of the url.

alefnula:
Some URLs, for example gitlab ssh URLs (ssh://git@gitlab.domain.com:2222/user/project.git) don't have a password in itself. This pull request allows validating URLs with only user in basic auth part of the url.


lafrech:
Thanks.

lafrech:
I documented this as a feature in the CHANGELOG. Could be a bugfix. @sloria, feel free to edit.

sloria:
Yeah, this is a bugfix. I've backported in #983

PR #984, opened 2018-10-10
Fix Python 3.3 Travis builds

sloria:
Use system-installed pip rather than upgrading.
Newer versions of pip don't support Python 3.3


sloria:
Didn't work...I'll try to look into this later today.

sloria:
Bah, we should probably just drop Python 3.3 altogether. It's EOL. See #987

PR #987, opened 2018-10-10
Drop Python 3.3 support

sloria:
We typically don't drop support for Python versions in minor updates, but I think we should make an exception for Python 3.3. 3.3 went EOL in 2017-09-29, and is no longer receiving bug fixes nor security updates. Furthermore, 3.3 is no longer supported by pip, which is causing the build failures against the 2.x-line branch.

close #984


lafrech (approved):
Totally agree.

Here's a link to an EOL announcement: https://www.python.org/downloads/release/python-337/.

PR #990, opened 2018-10-11
Fix Pluck load with data_key

lafrech:
The `data_key` must be managed also when deserializing.

@deckar01 do you think you could check this is correct?


sloria (approved):
Behavior looks correct.

Without this change, the added assertion fails with an uncaught error:

```python
marshmallow.exceptions.ValidationError: {'collaborators': {0: {'name': ['Unknown field.']}, 1: {'name': ['Unknown field.']}}, 'user': {'name': ['Unknown field.']}}
```

This restores the dump/load symmetry. 👌

PR #1000, opened 2018-10-14
Rename default/missing Field params into default_dump/default_load

### LINKED ISSUE #778: RFC: Replace missing/default with default_load/default_dump/default

lafrech:
Closes #778.

TODO:

- [ ] Changelog
- [ ] Docs

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -60,9 +60,6 @@ class Field(FieldABC):
     data does not need to be formatted before being serialized or deserialized.
     On error, the name of the field will be returned.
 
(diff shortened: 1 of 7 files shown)


sloria:
Closing this as it's gone stale. We can revisit/remake this for 4.0

Issue #612, opened 2017-04-16
fields.DateTime - load/dump to timestamp


vgavro:
I'm currently using this code to parse/dump timestamp, but I believe this should be in standard library. Should I create pull request, or have I missed another way to work with datetime/timestamp format?
```
class DateTime(ma.fields.DateTime):
    """
    Class extends marshmallow standart DateTime with "timestamp" format.
    """

    DATEFORMAT_SERIALIZATION_FUNCS = \
        ma.fields.DateTime.DATEFORMAT_SERIALIZATION_FUNCS.copy()
    DATEFORMAT_DESERIALIZATION_FUNCS = \
        ma.fields.DateTime.DATEFORMAT_DESERIALIZATION_FUNCS.copy()

    DATEFORMAT_SERIALIZATION_FUNCS['timestamp'] = lambda x: x.timestamp()
    DATEFORMAT_DESERIALIZATION_FUNCS['timestamp'] = datetime.fromtimestamp
```

sloria:
@vgavro I think this is a fair use case. I anticipate that users will typically want to deserialize to timezone-aware datetimes, so the API should make this simple. I haven't given to much thought to what it would look like though. Maybe it would make sense as a separate field that receives a `tzinfo` param,

```python
my_timestamp = fields.Timestamp(tzinfo=pytz.utc)
```

disconnect3d:
Bump, this would be nice to have :+1:

stephan-rayner:
Is there a plan to do this thing? I can't tell you how happy this would make me?

disconnect3d:
@stephan-rayner You can do it even better: work on that and send a PR? :)

Note that you can also attend in https://hacktoberfest.digitalocean.com and win a t-shirt ;D.

sloria:
I would gladly review and merge a PR adding a `Timestamp` field.

vgavro:
I will gladly work on this PR, but maybe it's better to implement this as custom format for AwareDateTime and NaiveDateTime instead of separate field https://github.com/marshmallow-code/marshmallow/pull/933 ? Like I see it's good decision to force conversion to aware datetime, but still explicitly allow naive datetimes.

disconnect3d:
> but still explicitly allow naive datetimes

*explicitly* is the key here. If the user wants it, let him do that.

I am not a maintainer of the project but I am up for this as we wouldn't make the bad thing (naive datetimes) to be the default ones.

sloria:
[...]

PR #1003, opened 2018-10-15
[WIP] fields.DateTime timestamp, timestamp_ms, timezone, timezone_naive added

vgavro:
UPDATE: [ THIS MERGE REQUEST SHOULD BE CANCELED WHEN WE HAVE SOME PROGRESS WITH NEW #1009  ] 

----
Please see related comment https://github.com/marshmallow-code/marshmallow/issues/612#issuecomment-429947895

TODO: (if this concept will be approved)
* tests
* timezone conversion from string needed
* timestamp overflow raises ValueError/OverflowError/OSError
* fields.DateTime subclasses doesn't need `timezone` and `timezone_naive`, but may want timestamp format (some api are storing dates in timestamps for some reason, maybe timezone replace for proper date from timestamp may be needed there?)

### DIFF
--- a/marshmallow/utils.py
+++ b/marshmallow/utils.py
@@ -317,10 +317,37 @@ def to_iso_date(date, *args, **kwargs):
     return datetime.date.isoformat(date)
 
 
+def from_timestamp(timestamp, tzinfo=UTC):
+    return datetime.fromutctimestamp(float(timestamp)).replace(tzinfo=tzinfo)
+
+
+def to_timestamp(dt, localtime=False, utc_offset=True):
+    """Converts to timestamp, preserves offset if utc_offset=False"""
+    # TODO: do we need localtime?
+    if not dt.tzinfo or not utc_offset:
+        # We must ensure datetime is timezone-aware,
+        # because it will be converted with local time offset otherwise.
+        # If utc_offset is False - datetime will be converted "as is",
+        # offset related to timezone will be added otherwise
+        return dt.replace(tzinfo=UTC).timestamp()
+    # If we're not replacing tzinfo to UTC, timestamp will be converted
+    # with utc offset by default
+    return dt.timestamp()
+
+
+def from_timestamp_ms(value, tzinfo=UTC):
+    return from_timestamp(float(value) / 1000, tzinfo)
+
+
+def to_timestamp_ms(value, localtime=False, utc_offset=True):
+    return to_timestamp(value, localtime, utc_offset) * 1000
+
+
 def ensure_text_type(val):
     if isinstance(val, binary_type):
         val = val.decode('utf-8')
     return text_type(val)
+
 
 def pluck(dictlist, key):
     """Extracts a list of dictionary values from a list of dictionaries.
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -932,6 +932,8 @@ class DateTime(Field):
         'iso8601': utils.isoformat,
         'rfc': utils.rfcformat,
         'rfc822': utils.rfcformat,
+        'timestamp': utils.to_timestamp,
+        'timestamp_ms': utils.to_timestamp_ms,
     }
 
     DESERIALIZATION_FUNCS = {
@@ -939,6 +941,8 @@ class DateTime(Field):
         'iso8601': utils.from_iso_datetime,
         'rfc': utils.from_rfc,
         'rfc822': utils.from_rfc,
+        'timestamp': utils.from_timestamp,
+        'timestamp_ms': utils.from_timestamp_ms,
     }
 
     DEFAULT_FORMAT = 'iso'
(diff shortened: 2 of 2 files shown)


lafrech:
Using a format is the right approach for this. Typically in marshmallow, the name of the field class refers to the object type, not the serialized type. We're dealing here with datetimes serialized as timestamps so this PR is the right way and I'm closing https://github.com/marshmallow-code/marshmallow/pull/1009.

From a quick glance, I think what you're trying to achieve here could be done relatively easily and in a non-breaking change now that https://github.com/marshmallow-code/marshmallow/pull/1278 is merged.

sloria:
Are you still interested in this? If so, could you please get this up to date with `dev`?

vgavro:
@sloria This merge request (and idea behind it) is with conflict with current AwareDateTime/NaiveDateTime implementation, so decisions should be made before I can work further on timestamp functionality.
The problem with current decision with Naive/Aware DateTime fields
that we're forcing naivness or awareness only for deserialization,
not expecting that serialization target may *implicitly* assume timezone.
For `timestamp` format cases this is crucial, because some API (serialization target) timestamps are in their local timezones.
```py
# Consider problems with current iso format
AwareDateTime("iso", default_tz=UTC)
deserialize("2017-12-12 17:00:00")  # dt(2017,12,12,17,0,0,tzinfo=UTC)
serialize(...)  # "2017-12-12 17:00:00+Z"
# problem here - what if serialization target (api) is not expecting +X part?

NaiveDateTime("iso", tz=UTC)
deserialize("2017-12-12 17:00:00+01")  # dt(2017,12,12,16,0,0)
serialize(...)  # "2017-12-12 16:00:00"
# problem here - what if serialization target (api) has "default_tz" set to UTC+01,
# so next time we can get 2017-12-12 16:00:00+01 ?

# Let's assume we want to add timestamp format with other than UTC timezone..
AwareDateTime("timestamp", default_tz=UTC+01)
deserialize(0)  # dt(1970,1,1,0,0,0, tzinfo=UTC+01)
serialize(...)  # -3600  # because timestamps expected to be in UTC...
```

SO - there are 4 ways to go:

1. Not to break current Naive/Aware/Datetime logic:
to rely that timestamps are always UTC, but in such cases we must not allow
NaiveDateTime(timezone) and AwareDateTime(timezone) be other than UTC.
ValueError on `__init__` and None defaults to UTC on timestamp formats? 
2. Make "timestamp" formats SPECIAL usecases (AwareDateTime converts to default_timezone on timestamp serialization, NaiveDateTime converts to "timezone" on timestamp serialization). [bad idea, as we don't solve problems for other formats + breaking consistency and adding not obvious behaviour]
3. Make TimestampField.
4. *[MY VOTE GOES HERE]* (and this was the idea behind this PR before there was other decision in related ticket, because of that I started to work on separate field) Consider DateTimeField refactoring, which will be working AS CLOSE AS EXPECTED with all formats, but timestamp will be "special case" for serialization anyway.


```py
class SmartDateTime:
# timezone=None - default timezone for naive datetimes on serialization/deserialization;
# naive=False - should datetimes be serialized/deserialized without timezone,
# if true - aware-datetimes will be converted to "timezone" before serialization/deserialization.

def deserialize(value):
    if is_aware(value):
        if self.naive:
            assert self.timezone
            return value.astimezone(self.timezone).replace(tzinfo=None)
        else:
            return value
    else:
        if self.naive:
            return value
        else:
            assert self.timezone
            return value.replace(self.timezone)

def serialize(value):
    if is_aware(value):
        if self.naive or self.format in timestamp_formats:
            assert self.timezone
            return value.astimezone(self.timezone).replace(tzinfo=None)
        return value
    else:
        if self.naive or self.format in timestamp_formats:
            return value
        else:
            assert self.timezone
            return value.replace(self.timezone)
```

sloria:
Sorry I haven't had time to review this. @deckar01 or @lafrech can you take a look?

lafrech:
Thans @vgavro for investigating this.

Here's my understanding for now.

You're right about the API being incomplete. One could want to enforce awareness in the app and naiveness in the API and this is not possible currently, as you shown.

A timestamp is TZ aware (https://stackoverflow.com/questions/23062515/do-unix-timestamps-change-across-timezones) so it would be rather like

```py
AwareDateTime("timestamp", default_tz=UTC+01)
deserialize(0)  # dt(1970,1,1,0,0,0, tzinfo=UTC)  # default_tz is not used since timestamps is aware
serialize(...)  # 0
```

However it fails in the `NaiveDateTime` case

```py
NaiveDateTime("timestamp", tz=UTC-01)
deserialize(3600)  # dt(1970,1,1,0,0,0)
serialize(...)  # 0
```

We could add specific timestamp logic in the field or in timestamp dedicated util functions, but it would be better to address the root cause, which is the API not allowing to specify awareness on serialization.

marshmallow/utils.py
```diff
@@ -317,11 +317,38 @@ def to_iso_date(date, *args, **kwargs):
     return datetime.date.isoformat(date)
 
 
+def from_timestamp(timestamp, tzinfo=UTC):
+    return datetime.fromutctimestamp(float(timestamp)).replace(tzinfo=tzinfo)
```
shaggyrogers, line 321: Per the [documentation](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp), `datetime.fromtimestamp(value, tz=datetime.timezone.utc)` is preferred over `datetime.fromutctimestamp(value)`.

lafrech:
Achieved in #2022.

PR #1007, opened 2018-10-17
Let all fields _deserialize method accept **kwargs

lafrech:
Following up on https://github.com/marshmallow-code/marshmallow/pull/849. I think this makes things more consistent.

Adding a note to the docs saying `_deserialize` may receive arbitrary `kwargs` should allow us to add other field-specific `kwargs` later on without it being a breaking change.

I'm tempted to do the same to `_serialize`, preventively...

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -310,16 +310,18 @@ class Field(FieldABC):
         """
         return value
 
-    def _deserialize(self, value, attr, data):
+    def _deserialize(self, value, attr, data, **kwargs):
         """Deserialize value. Concrete :class:`Field` classes should implement this method.
 
         :param value: The value to be deserialized.
         :param str attr: The attribute/key in `data` to be deserialized.
         :param dict data: The raw input data passed to the `Schema.load`.
-        :param bool|tuple partial: For nested schemas, the ``partial``
-            parameter passed to `Schema.load`.
+        :param dict kwargs': Field-specific keyword arguments.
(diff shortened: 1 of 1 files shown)


sloria:
Seems fine. I think doing the same to `_serialize` makes sense for consistency's sake. I don't feel strongly about it, though. Omitting `**kwargs` makes for a stricter interface, which could be good.

deckar01:
I thought new kwargs were a minor version change, not a major version change. Preventing old minor versions from erroring when a new kwarg features is used does not seem like a desirable behavior to me.

lafrech:
So you'd rather have a strict interface where `Nested` and `Pluck` fields receive `partial` and no other field accepts any kwargs?

pylint complains when changing a method signature when subclassing as it is considered bad practice, I suppose.

> I thought new kwargs were a minor version change, not a major version change. 

It will break custom fields not accepting kwargs. Doesn't this qualify as a breaking change?

deckar01:
I see what you mean. I was stuck thinking about field inheritance and subclasses calling our base class methods. The issue is the marshaller calling custom fields' `_deserialize` method with kwargs that they don't support. That is a major change unless they are required to ignore extra kwargs. Thanks for bearing with me.

Since these methods are only expected to be called by us internally, there is no need to protect ourselves from passing a kwargs to a custom field that doesn't do anything with it. :+1:

deckar01 (approved):

PR #1008, opened 2018-10-17
Improve error message for data_key and attribute collision

### LINKED ISSUE #1005: Attribute Kwarg Error
```
class TestSchema(Schema):
    title = fields.Str()
    title_sub = fields.Str(attribute='title')
```
If I try to instantiate this class in 3.0.0b17 and above, I get `ValueError: Duplicate attributes: set(['title'])`.  I've inherited some Marshmallow test code using similar attributes.  Seems awfully like a bug to me given the documentation for the attribute kwarg for Field, but I'm not sure.

sloria:
close #1005

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -721,14 +721,24 @@ class BaseSchema(base.SchemaABC):
         ]
         if len(dump_data_keys) != len(set(dump_data_keys)):
             data_keys_duplicates = {x for x in dump_data_keys if dump_data_keys.count(x) > 1}
-            raise ValueError('Duplicate data_keys: {}'.format(data_keys_duplicates))
+            raise ValueError(
+                'The data_key argument for one or more fields collides '
+                "with another field's name or data_key argument. "
+                'Check the following field names and '
+                'data_key arguments: {}'.format(list(data_keys_duplicates)),
+            )
 
         load_attributes = [
             obj.attribute or name for name, obj in iteritems(fields_dict) if not obj.dump_only
         ]
         if len(load_attributes) != len(set(load_attributes)):
             attributes_duplicates = {x for x in load_attributes if load_attributes.count(x) > 1}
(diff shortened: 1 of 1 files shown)


sloria:
Example error message:

```
ValueError: Multiple fields are using the same data_key. Change the following data_key arguments: ['bar', 'foo']
```

lafrech:
Fine with me. At least more explicit than current phrasing.

Just one information missing: one of the fields may not have a `data_key` attribute. It can also be the field name if no `data_key` is specified. But i don't see a way to express this with enough concision for an error message.

And to nitpick, the user doesn't have to change all those attributes, rather all but one. But that's fine with me if we keep this as is. It's enough for the reader to understand and fix his code.

sloria:
How about:

> The data_key argument for one or more fields collides with another field. Check the following data_key arguments: ['foo', 'bar']

lafrech:
Yes.

Technically, the data_key attributes can collide together even if no field with the same name is involved, so it could be

> collides together (maybe "with one another" is more correct grammatically) or with another field. Check following data_key arguments and fields (or "field names")

It sounds complicated but that's the way it is.

I don't mind a simpler and less 100% correct form. The developer should understand.

sloria:
I meant "field" to mean "a field's data_key or field name". I could revise to be more precise:

> The data_key argument for one or more fields collides with another field's name or data_key. Check the following data_key arguments and field names: ['foo', 'bar']

But I agree--not worth splitting hairs. If you think of something better than the above, let me know. I'll plan to merge this later today.

lafrech:
Great.

The only thing that might be missing is the fact that it can be solved by making some of them dump_only. In the docs, this is only stated in upgrading.py.

But if this must be explained, it would rather be in the docs than in the error message so let's merge this and we can always complete the docs if needed. The error message is not meant to paraphrase the docs.

PR #1009, opened 2018-10-17
Timestamp field added

### LINKED ISSUE #612: fields.DateTime - load/dump to timestamp

vgavro:
closes #612 
Actually in case of naive=True timezone is used only for timezone-aware datetimes conversion on serialization, maybe we should drop option naive=True and allow timezone=None instead? In this case on datetime-aware serialization we should throw ValueError, what do you think?

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -1149,6 +1149,41 @@ class TimeDelta(Field):
             self.fail('invalid')
 
 
+class Timestamp(Field):
+    """Timestamp field, converts to datetime.
+
+    :param timezone: Timezone of timestamp (defaults to UTC), should be tzinfo object.
+    :param bool ms: Milliseconds instead of seconds, defaults to `False`. For javascript
+        compatibility.
+    :param bool naive: Should deserialize to timezone-naive or timezone-aware datetime.
+        Defaults to `False`, so all datetimes will be timezone-aware with `timezone`.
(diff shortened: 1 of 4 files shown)


sloria:
Thanks for your work! Sorry for the delay on reviewing this; we're currently focused on getting a stable marshmallow 3 release out the door, so we're prioritizing getting the last few backwards-incompatible changes released.

We'll return to new features like this one after marshmallow 3 is released.

lafrech:
Closing. See https://github.com/marshmallow-code/marshmallow/pull/1003#issuecomment-512152610.

PR #1011, opened 2018-10-18
Let all fields _serialize method accept **kwargs

lafrech:
Same as https://github.com/marshmallow-code/marshmallow/pull/1007 but for `_serialize`.


sloria (approved):
Looks good. Just need to change the "Creating A Field Class" docs to be consistent with this change. I can do this after I merge.

PR #1024, opened 2018-10-26
Minor Schema._do_load rework

lafrech:
- Add comments
- Indent the postprocessors part correctly to save a test
- Move the `errors = unmarshal.errors` statement to make things clearer (before the change it is unclear as it is unmarshal.errors is mutated during schema-level validation
- Simplify logic and escape sooner when skipping schema validation due to field errors


lafrech:
(As soon as the push CI tests passed, I merged and removed the branch, hence the failure in PR CI tests...)

PR #1025, opened 2018-10-26
Use tox for running tests in CI and running dev commands

### LINKED ISSUE #1020: RFC: Use tox for testing in travis and local dev?
Should we use tox to run tests?

Example: https://github.com/python-attrs/attrs/blob/master/.travis.yml

Pros:

* Better parity between local dev and travis
* Can run tests without having to set up an entire environment

Cons:

* Slightly more config in .travis.yml
* Some learning curve for those of us used to using pytest directly

Strong opinions welcome, esp. those of frequent contributors. @lafrech @deckar01 @taion

sloria:
...for better parity between local dev and Travis. This also
simplifies setting up a local dev environment.

Implements the proposal in https://github.com/marshmallow-code/marshmallow/issues/1020#issuecomment-433544037

Close #1020


lafrech:
I'm fine with this change.

- Maybe add a pre-commit action at the end of tox.ini to run `pre-commit run --all-files` locally. Better consistency, everything available from tox.

- I generally prefer dev-requirements.txt to `EXTRAS_REQUIRE` setup.py, but the difference is cosmetic. When using a new lib, before even reading the docs, I visually identify the requirements.txt file. It is more straightforward than peeking into setup.py. Maybe I should read the docs... Also, you could add all the dependencies under a `dev` entry. I mean I wouldn't bother separating reco, lint, and tests. The difference can be conveyed with a simple comment. In practice, someone willing to develop will install all the dev deps.

BTW, what is the benefit of hard-pinning versions in dev-requirements? It generates a lot of dependabot noise. Is it about being able to reproduce an older passing setup in case tests begin to fail, therefore identify quickly which dependency update broke the tests?

sloria:
> Maybe add a pre-commit action at the end of tox.ini to run pre-commit run --all-files locally. Better consistency, everything available from tox.

Yes, I'll do that on `dev`. We're not using pre-commit on `2.x-line`. On `dev`, `tox -e lint` will run `pre-commit run ...`.

> I generally prefer dev-requirements.txt to EXTRAS_REQUIRE setup.py,

The main advantage of using `extras_require` is that the tox environments only need to install the extras they need,  e.g. test environments install `[test]` and the lint environment installs `[lint]`, which should speeds things up a bit.

> BTW, what is the benefit of hard-pinning versions in dev-requirements?

For build reproducibility. It prevents a breaking change in a dependency from suddenly producing failing builds under our feet, like what happened with flake8 this past week.

lafrech (approved):

Issue #879, opened 2018-07-11
How to report errors for container fields?


lafrech:
Generally, a field name is associated a list of errors as strings. The exception to this is container fields. Those may return errors for the container and the contained field(s). In this case, rather than a list, a `dict` is used  where the keys are the field names or the list index ([as string or as integer](https://github.com/marshmallow-code/marshmallow/issues/493)):

```python
errors = {
    '_schema': [...],
    'field': [...],
    'nested': {
        # What about errors in the Nested field itself?
        '_schema': [...],
        'field': [...],
    },
    'nested_many': {
        # What about errors in the Nested field itself?
        '0': {
            '_schema': [...],
            'field': [...],
        },
    },
    'list': {
        # What about errors in the List field itself?
        '0': [...],
    }
    'list_nested': {
        # What about errors in the List field itself?
        '0': {
            # What about errors in the Nested field itself?
            '_schema': [...],
            'field': [...],
        },
    },
    'dict': {
        # What about errors in the Dict field itself?
        '0': {
            'key': [...],
            'value': [...],
        }
    }
    'dict_nested': {
        # What about errors in the Dict field itself?
        '0': {
            'key': [...],
            'value': {
                # What about errors in the Nested field itself?
                '_schema': [...],
                'field': [...],
            },
        }
    }
}
```

This has been working relatively fine for a while. It leaves a hole as you can't return errors relative to the `Nested` field itself, for instance, while returning errors for the fields in the nested schema.

It is _generally_ not blocking, as the most frequent `Nested` field level error is the `missing` error, and when the field is missing, there is _generally_ no nested error, so we return

```python
    nested: ['Missing data for required field.']
```

It is a shame that the structure (`dict` or `list`) depends on the error, but we could live with this.

Except we sometimes need to return errors for both the field and the nested schema, so we came up with this (https://github.com/marshmallow-code/marshmallow/pull/299):

```python
    'nested': {
        '_field': [...],
        '_schema': [...],
        'field': [...],
    },
```

where `_field` stores errors about the field itself.

This is inconsistent, as I said in https://github.com/marshmallow-code/marshmallow/pull/299#issuecomment-401607834 and showed in https://github.com/marshmallow-code/marshmallow/pull/863 because the errors on the nested field itself may be reported in different places depending on whether there are errors in the nested schema. In other words, a `dict` with a `_field` attribute is used only if there are also errors in the nested schema, otherwise, the errors are returned as a strings, like for any field.

[...]

PR #1026, opened 2018-10-26
Rework validation error structure

lafrech:
This PR implements #441 using the merge function implemented in #442, and it addresses a few of the concerns expressed in #441 and #879.

---------------------------------------------------------------------------------------------------------------------------------

New validation error specification:

- A `ValidationError` has a `message` and a `field_name`. If `field_name` is not specified, the error applies to the schema and is stored under the `_schema` key.

- `message` can be a `dict`, a `string` or a `list`. Passing a `string` is a shortcut for passing a `list` with that `string` as unique element. When `message` is a `dict`, its keys are supposed to be field names, or an equivalent hierarchy (like in the `dict` messages generated by `List` or `Field`), so that several `dict` messages can be deep-merged together.

- It is not possible anymore to raise a `ValidationError` with a message applying to several fields by passing a message as `string` and a list of filed names. This is a special case of a schema-level error with the messages expressed as a `dict`.

- When a `Nested` field has both errors in its fields and errors in itself, its own errors are not stored under `_field` but under `_schema`. From client perspective, schema or field is an implementation detail.

---------------------------------------------------------------------------------------------------------------------------------

Fixes:

- Fixes the case where a pre/post_dump/load validator returns an error as `dict`.
- Allows two schema-level validators to return errors as `dict` on the same field. The errors messages are merged.
- Rewrite `test_passing_original_data` to test only what it is meant to and not use a wrong error structure. There was a long argument  in #441/#442 about whether the `{'code': 'invalid_field'}` message was wrong or not, and whether forbidding such a message would be a breaking change. It is wrong according to the specifications described above, which may be a breaking change (we're changing the error structure, the values expected in a `ValidationError`) but that's fine, as it is "breaking changes" season.
- `index = index if index_errors else None` is factorized which should provide a ridiculously marginal performance gain in most cases but also ensure `index_errors` is respected all along, even in case of invalid input type error.

### DIFF
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -33,21 +33,15 @@ class ErrorStore(object):
         #: Dictionary of extra kwargs from user raised exception
         self.error_kwargs = {}
 
-    def get_errors(self, index=None):
-        return self.errors if index is None else self.errors.setdefault(index, {})
-
-    def set_errors(self, errors, index=None):
-        if index is None:
-            self.errors = errors
-        else:
-            self.errors[index] = errors
-
     def store_error(self, messages, field_name=SCHEMA, index=None):
-        errors = self.get_errors(index=index)
-        if field_name != SCHEMA:
-            errors[field_name] = merge_errors(errors.get(field_name), messages)
-        else:
-            self.set_errors(merge_errors(errors, messages), index=index)
+        # field error  -> store/merge error messages under field name key
+        # schema error -> if string or list, store/merge under _schema key
+        #              -> if dict, store/merge with other top-level keys
+        if field_name != SCHEMA or not isinstance(messages, dict):
+            messages = {field_name: messages}
+        if index is not None:
+            messages = {index: messages}
+        self.errors = merge_errors(self.errors, messages)
 
     def call_and_store(self, getter_func, data, field_name, index=None):
         """Call ``getter_func`` with ``data`` as its argument, and store any `ValidationErrors`.
--- a/marshmallow/marshalling.py
+++ b/marshmallow/marshalling.py
@@ -14,7 +14,7 @@ from marshmallow.utils import (
     EXCLUDE, INCLUDE, RAISE, is_collection, missing, set_value,
 )
 from marshmallow.compat import iteritems, Mapping
-from marshmallow.exceptions import ValidationError
+from marshmallow.exceptions import ValidationError, SCHEMA
 from marshmallow.fields import Nested
 
 __all__ = [
@@ -22,10 +22,6 @@ __all__ = [
     'Unmarshaller',
 ]
 
-# Key used for schema-level validation errors
-SCHEMA = '_schema'
-# Key used for field-level validation errors on nested fields
-FIELD = '_field'
 
 class ErrorStore(object):
 
@@ -46,7 +42,7 @@ class ErrorStore(object):
         if isinstance(messages, dict):
             errors[field_name] = messages
         elif isinstance(errors.get(field_name), dict):
-            errors[field_name].setdefault(FIELD, []).extend(messages)
+            errors[field_name].setdefault(SCHEMA, []).extend(messages)
         else:
             errors.setdefault(field_name, []).extend(messages)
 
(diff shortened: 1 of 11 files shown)


sloria:
On first glance, the implemented behavior looks desirable. I think this is the right direction to take.

> Fixes the case where a pre/post_dump/load validator returns an error as dict.

Can you clarify this?

Should we prevent keys for unknown fields to passed to `ValidationError`?

This PR allows arbitrary keys to be passed:

```python
from marshmallow import Schema, fields, pre_load, ValidationError

class MySchema(Schema):
    foo = fields.Field()

    @pre_load
    def always_error(self, data):
        raise ValidationError({
            'foo': 'foo error',
            'unknown': '<-- unknown key',
        })


MySchema().load({'foo': 42})
# marshmallow.exceptions.ValidationError: {'foo': 'foo error', 'unknown': '<-- unknown key'}
```

lafrech:
> >  Fixes the case where a pre/post_dump/load validator returns an error as dict.
>
> Can you clarify this?

This is the third commit of the PR ("Fix handling dict error on field in pre/post_dump/load", https://github.com/marshmallow-code/marshmallow/pull/1026/commits/0aeee447ad172ebb8a3cb6548302858d5b53fe4b).

I figured a decorator (pre/post_dump/load) could raise errors similar to schema errors, by passing a message and an optional field name, and the message could be a `dict`. Without the change I made in `normalized_message`, the field name is lost if the message is a `dict`.

I realized afterwards that the logic is the same in `normalized_message` and `store_error`, so the comment in `store_error` applies.

```py
        # field error  -> store/merge error messages under field name key
        # schema error -> if string or list, store/merge under _schema key
        #              -> if dict, store/merge with other top-level keys
```



> Should we prevent keys for unknown fields to passed to ValidationError?

I don't see any problem with this. I think this is something you purposely allowed (https://github.com/marshmallow-code/marshmallow/issues/273).

sloria:
> Without the change I made in normalized_message, the field name is lost if the message is a dict.

Ah, I see. Thanks for the clarification.

> I don't see any problem with this. I think this is something you purposely allowed (#273).

Ah, good catch. Forgot about that issue. OK, I think the behavior is fine as is.

lafrech:
Note that as opposed to the proposal in #442, I wouldn't consider `merge_errors` public API as I don't see the benefit for this. I didn't underscore the name, but it is not documented and not advertised as public.

Also I didn't copy the `ValidationErrorBuilder`. It's never too late to add something like that if we feel the need, but I wanted to keep this minimal, it is already an important change, and in most cases, creating the `dict` should be easy enough not to need this.

sloria:
> I wouldn't consider merge_errors public API as I don't see the benefit for this. 

Agreed. All of `marshmallow.marshalling` is considered private API.

> It's never too late to add something like that if we feel the need, but I wanted to keep this minimal,

👍

sloria:
@lafrech Is there anything left to do with this PR except for update the upgrading docs and changelog?

lafrech:
Nothing, I believe. I wasn't totally satisfied until the last rework. Now, I think it is complete.

It would be nice to have @deckar01's eye on this since he commented on #441/#442.

lafrech:
Rebased.

Shall we merge this or shall we wait for another reviewer?

lafrech:
Rebased.

I'd be glad if we could merge this so as to limit merge conflicts.

The sooner it is merged and beta released, the sooner we get feedback...

marshmallow/marshalling.py
```diff
@@ -254,11 +241,61 @@ def deserialize(
...
+
+
+def merge_errors(errors1, errors2):
+    """Deeply merge two error messages
+
+    Error messages can be string, list of strings or dict of error messages
+    (recursively). Format is the same as accepted by :exc:`ValidationError`.
+    Returns new error messages.
+    """
+    if errors1 is None:
```
sloria, line 261: I think this can be simplified to

```suggestion
    if not errors1:
```

Same with L263.

marshmallow/marshalling.py
```diff
@@ -254,11 +241,61 @@ def deserialize(
...
+        return errors2
+    if errors2 is None:
+        return errors1
+    if isinstance(errors1, list):
+        if not errors1:
```
sloria, line 266: This conditional shouldn't be necessary.

sloria (approved):
Took a second look and the behavior looks correct.

Had a couple nits but those definitely aren't blockers.

Let's get this merged!

lafrech:
Thanks @sloria for reviewing and merging!

PR #1029, opened 2018-10-29
Clean up Travis and tox config a bit

taion:


.travis.yml
```diff
@@ -12,13 +12,13 @@ jobs:
   - python: '2.7'
     env: TOXENV=py27
   - python: '3.5'
-    env: TOXENV=py35
+    env: TOXENV=py
```
taion, line 15: No need to explicitly specify `TOXENV` unless there are specific overrides relating to the env.

sloria: I prefer passing the full TOXENV name for better visual consistency.

Also, this change makes the Travis display a bit awkward:

image

sloria:
Thanks for the PR. I think I'm going to stick with explicitly passing the envs.

I'm going clean things up a bit like https://github.com/encode/django-rest-framework/pull/5519, which looks nice.

Issue #1027, opened 2018-10-29
Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working


nkonin:
Affected version: marshmallow-2.16.1

Warning examples:
```python
======================================= warnings summary ========================================
.../python3.7/site-packages/marshmallow/schema.py:5: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import defaultdict, Mapping, namedtuple

.../python3.7/site-packages/marshmallow/orderedset.py:26: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  class OrderedSet(collections.MutableSet):

.../python3.7/site-packages/marshmallow/utils.py:54: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  (isinstance(obj, collections.Iterable) and not hasattr(obj, "strip")) or is_generator(obj)

.../venv/lib/python3.7/site-packages/tornado/httputil.py:107: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  class HTTPHeaders(collections.MutableMapping):
```

Maybe it worth to use something like:
```python
try:
    from collections.abc import defaultdict, Mapping, namedtuple
except ImportError:
    from collections import defaultdict, Mapping, namedtuple
```

jmargeta:
@nkonin I came across the same deprecation warnings.
Does the proposed PR solve it for you?

sloria:
This is released in 2.16.2

PR #1032, opened 2018-10-30
Fix deprecated imports of collections.abc classes 

jmargeta:
This PR proposes to move all collections.abc imports into the compat module.

In Python 3.3, Collections Abstract Base Classes such as Mapping, Iterable, or MutableSet were moved to the collections.abc module.
For backwards compatibility, are still visible in the collections module through Python 3.7. In the following versions they will be entirely removed.

This should fix Issue https://github.com/marshmallow-code/marshmallow/issues/1027
See also: https://docs.python.org/3.7/library/collections.html

### DIFF
--- a/marshmallow/compat.py
+++ b/marshmallow/compat.py
@@ -23,7 +23,11 @@ if PY2:
         from .ordereddict import OrderedDict
     else:
         from collections import OrderedDict
+    from collections import Mapping, Iterable, MutableSet
     OrderedDict = OrderedDict
+    Mapping = Mapping
+    Iterable = Iterable
+    MutableSet = MutableSet
     def get_func_args(func):
         if isinstance(func, functools.partial):
             return list(inspect.getargspec(func.func).args)
@@ -45,6 +49,13 @@ else:
     zip_longest = itertools.zip_longest
     from collections import OrderedDict
     OrderedDict = OrderedDict
+    try:
+        from collections import Mapping, Iterable, MutableSet
+    except ImportError:
+        from collections.abc import Mapping, Iterable, MutableSet
+    Mapping = Mapping
+    Iterable = Iterable
+    MutableSet = MutableSet
     def get_func_args(func):
         if isinstance(func, functools.partial):
             return list(inspect.signature(func.func).parameters)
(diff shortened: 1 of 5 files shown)


lafrech:
I suppose the import in schema.py should be changed as well: https://github.com/marshmallow-code/marshmallow/blob/2.x-line/marshmallow/schema.py#L5.

------------------------------------------------------------

I don't see the need for those statements:

```py
    OrderedDict = OrderedDict
    Mapping = Mapping
    Iterable = Iterable
    MutableSet = MutableSet
```

Readability, I guess.

Anyway, this is not specific to this PR. You conformed to the existing convention.

jmargeta:
Thanks for looking into the PR @lafrech . I forgot to stage that file in the first commit. Should be fixed now.

I also find that convention unnecessary but thought that there might have been a reason for that.
Will open another PR with that change in a moment (when/if this one gets merged) and we can discuss it there.

marshmallow/compat.py
```diff
@@ -45,6 +49,13 @@ def get_func_args(func):
     zip_longest = itertools.zip_longest
     from collections import OrderedDict
     OrderedDict = OrderedDict
+    try:
```
sloria: This try-except should not be necessary, since we don't support Python 3.3.

marshmallow/compat.py
```diff
@@ -45,6 +49,13 @@ def get_func_args(func):
...
+        from collections import Mapping, Iterable, MutableSet
+    except ImportError:
+        from collections.abc import Mapping, Iterable, MutableSet
+    Mapping = Mapping
```
sloria: I'm fine getting rid of this convention. I think it was done in order to avoid flake8 errors, but we are now ignoring this whole file in the flake8 config, so no use in keeping it.

sloria (changes requested):
Thanks @jmargeta for the PR!

Just had a few nits. I can take care of those myself.

sloria (approved):

marshmallow/compat.py
jmargeta: Ah, ok, that's why! Thank you @sloria for the explanation, for doing the remaining changes, and for merging them in :)

PR #1034, opened 2018-10-31
pass partial to validate_schema; kill pass_original

### LINKED ISSUE #994: Accessing partial within validates_schema

cancan101:
Closes https://github.com/marshmallow-code/marshmallow/issues/994

Still needs docs mentioning breaking api change.

/CC @lafrech


tests/test_decorators.py
```diff
@@ -556,13 +558,14 @@ class MySchema(Schema):
             bar = fields.Int()
 
             @validates_schema(pass_original=True)
-            def validate_original(self, data, original_data):
+            def validate_original(self, data, original_data, partial):
                 if isinstance(original_data, dict) and isinstance(original_data['foo'], str):
                     raise ValidationError('foo cannot be a string')
 
             # See https://github.com/marshmallow-code/marshmallow/issues/127
             @validates_schema(pass_many=True, pass_original=True)
```
lafrech: Remove `pass_original`

lafrech:
What do you guys think about deciding that 
- `many`, `partial` and `original` (or `original_data`) are always passed as kwargs
- validators should accept unknown kwargs (even kwargs we may add in the future)

?

A validator would only have in its signature the kwargs it needs.

```py

        @validates_schema
        def validate_email(self, data, **kwargs):
            if len(data['email']) < 3:
                raise ValidationError('Email must be more than 3 characters', 'email')

        @validates_schema(pass_many=pass_many)
        def check_unknown_fields(self, data, original_data, many=True, **kwargs):
            def check(datum):
                for key, val in datum.items():
                    if key not in self.fields:
                         raise ValidationError({'code': 'invalid_field'})
             if many:
                 for each in original_data:
                     check(each)
             else:
                 check(original_data)
```

vgavro:
I really agree with @lafrech and think that `**kwargs` is more pythonic and intuitive way to do this, django.signals use this approach, design of requesting arguments that should be passed in decorator makes implementation more complex and (imho) counterintuitive, and as it breaks complatibility this really should be in initial 3.0 release.

sloria:
I also agree with @lafrech 's suggestion of requiring `**kwargs` in method signatures and always passing `partial`, etc.

Though if we go that route, the same treatment should be applied to all the decorators.

cancan101:
@sloria / @lafrech / @vgavro so what would you like to see changed in this PR ?

sloria:
- [ ] Implement the proposal in the [comment above](https://github.com/marshmallow-code/marshmallow/pull/1034#issuecomment-436556833): Always pass `partial`, `original_data`, etc. to schema validators, as well as post/pre_load/dump methods. Note: `pass_many` behavior should stay the same.
- [ ] Update tests
- [ ] Optional: Clean up commit history (rebase and squash).

sloria:
Let us know if you don't have time to do any of the above, and someone else may be able to pick up where you leave off.

sloria:
@cancan101 Will you have time to update this, or would you rather someone else work on it?

cancan101:
@sloria I have some time to wrap this up but am still a bit confused as to what is missing from the PR that is up and what is asked for in the proposal, at least as far as the changes to schema validators (I hadn't made any changes to  post/pre_load/dump methods).

lafrech:
@cancan101 see https://github.com/marshmallow-code/marshmallow/pull/1034#issuecomment-454254580.

AFAIR, the change involved is not related to this PR but could be a prerequisite. It can be done in another PR for clarity. If you don't feel like implementing this, someone else could do it, and then you rebase this PR.

antgel:
@cancan101 Thanks for this interesting PR, looking forward to seeing it tidied up and merged. Echoing comments about squash / rebase, I found it a bit difficult to follow the commit history.

sloria:
I'm carrying on this work in https://github.com/marshmallow-code/marshmallow/pull/1220 . I've decided to keep `pass_original` for the reasons stated in https://github.com/marshmallow-code/marshmallow/pull/1219#discussion_r285353391.

Let's continue discussion on https://github.com/marshmallow-code/marshmallow/pull/1220 .

Issue #1048, opened 2018-11-14
TypeError when serializing a Pluck field valuated to None


toffan:
Using `fields.Pluck` on an attribute valuated to `None` raises a `TypeError`, cf. the example below.
```python
from marshmallow import Schema, fields
from dataclasses import dataclass

@dataclass
class Nested:
    key: str

@dataclass
class Main:
    nested: Nested

class NestedSchema(Schema):
    key = fields.String()

class MainSchema(Schema):
    nested = fields.Pluck(NestedSchema, 'key')

m = Main(None)
dumped = MainSchema().dump(m)  # ERROR
# File ".../marshmallow/fields.py", line 521, in _serialize
#   return ret[self._field_data_key]
# TypeError: 'NoneType' object is not subscriptable
```

PR #1049, opened 2018-11-14
Fix TypeError when serializing None with Pluck

toffan:
Serialize None with fields.Pluck returns None. Add a unit test for it.


toffan:
Fix #1048

sloria:
Good catch. Thanks for the PR!

PR #1050, opened 2018-11-14
Don't raise ValidationError in Unmarshaller.serialize

lafrech:
We did the same for deserialization in #961.

No use raising/catching.

Besides, I think there was an error as we used `data` instead of `valid_data`. Since the error was made consistently in both `Marshaller` and `Schema`, there was no impact.


sloria (approved):
Yep, this is correct.

PR #1052, opened 2018-11-15
Fix List and Dict valid_data

### LINKED ISSUE #766: List field does does not put anything in valid_data if an item fails validation
```python
class TestSchema(Schema):
    ints = fields.List(fields.Int())

try:
    u = TestSchema().load({'ints': [1, 2, 3, 4, 'a']})
except ValidationError as exc:
    print(exc)
    print(exc.valid_data)

# {'ints': {4: ['Not a valid integer.']}}
# {}
```

I naively expected to find 

    {'ints': [1, 2, 3, 4]}

in `exc.valid_data`.

Is this intended?

I'm not saying it is a bug. Removing elements breaks the indexing, so it may be safer to trash everything.

Note, however, that `List(Nested)` differs from `Nested(many=True)` on this. With `Nested(many=True)`, you get `valid_data` filled with partially deserialized data.

lafrech:
I've been trying to fix a few issues about `valid_data` in `List` and `Dict`.

Fixes https://github.com/marshmallow-code/marshmallow/issues/766.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -14,7 +14,7 @@ import math
 from marshmallow import validate, utils, class_registry
 from marshmallow.base import FieldABC, SchemaABC
 from marshmallow.utils import is_collection, missing as missing_
-from marshmallow.compat import basestring, text_type, Mapping
+from marshmallow.compat import basestring, text_type, Mapping, iteritems
 from marshmallow.exceptions import ValidationError, StringNotCollectionError
 from marshmallow.validate import Validator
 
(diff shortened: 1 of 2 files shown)


lafrech:
A few words of explanation.

In `List` field, I applied the changes described in #766.

Since it is used in a `Dict` test, I ended up modfying both fields in the same PR.

In `Dict` field, I passed the valid data as `valid_data` kwarg rather than `data` (bugfix), and I modified the logic to ensure valid data contains items for which the key deserializes correctly (even if value deserializes correctly, it is worthless if the key is incorrect) and the value deserializes at least partially (`valid_data` is not `None`). I think it makes for a more useful valid data structure.

In the code before this change, the `keys` array is initialized with serialized keys and then each key is replaced with its deserialized version, but in case of `ValidationError`, the serialized key remains in the result `dict`.

One thing worth mentioning, and I noted it in a comment: [|de]serializetion outputs a simple `dict` whatever the input `dict` type. We could allow specifying the `dict` type, or even use a trick to return the same type (I'm thinking of using `copy` and `clear`, but that sounds terribly hackish).

lafrech:
Just reviewed and removed the WIP in title as I think this is good to go.

sloria (approved):

PR #1058, opened 2018-11-29
is_collection should test for collections.abc.Collection, as documented, not Mapping

twosigmajab:
I noticed the docstring of `is_collection` says "Return True if `obj` is a collection type, e.g list, tuple, queryset.", which is actually a lie since the implementation is actually testing `isinstance(obj, Mapping)`. This PR updates the implementation to match the docstring, assuming that is what is desired.

I think this might also improve interop with data structures such as [werkzeug's Headers](https://github.com/pallets/werkzeug/blob/master/werkzeug/datastructures.py#L909) type. Apparently due to this code using `Mapping`, @barakalon had to resort to wrapping the werkzeug Headers object in a shim here before it could be used with marshmallow:
https://github.com/plangrid/flask-rebar/pull/43/files#r237132730

### DIFF
--- a/marshmallow/utils.py
+++ b/marshmallow/utils.py
@@ -17,7 +17,7 @@ from email.utils import formatdate, parsedate
 from pprint import pprint as py_pprint
 
 from marshmallow.compat import binary_type, text_type
-from marshmallow.compat import Mapping, Iterable
+from marshmallow.compat import Collection, Iterable
 
 EXCLUDE = 'exclude'
 INCLUDE = 'include'
@@ -67,7 +67,7 @@ def is_indexable_but_not_string(obj):
 
 def is_collection(obj):
     """Return True if ``obj`` is a collection type, e.g list, tuple, queryset."""
-    return is_iterable_but_not_string(obj) and not isinstance(obj, Mapping)
+    return is_iterable_but_not_string(obj) and not isinstance(obj, Collection)
 
 
 def is_instance_or_subclass(val, class_):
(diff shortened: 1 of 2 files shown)


marshmallow/utils.py
```diff
@@ -67,7 +67,7 @@ def is_indexable_but_not_string(obj):
 
 def is_collection(obj):
     """Return True if ``obj`` is a collection type, e.g list, tuple, queryset."""
-    return is_iterable_but_not_string(obj) and not isinstance(obj, Mapping)
+    return is_iterable_but_not_string(obj) and not isinstance(obj, Collection)
```
deckar01, line 70: This logic was saying "is collection if iterable but not a string and not a mapping". Now it says "is collection if iterable but not a string and not a collection", which is not right. I think you might was something more like "is collection if a collection but not a string".

deckar01: The reason it is done this way is probably, because there is no `Collection` abstract class in python 2. Starting with a more generic base class and excluding the bits that aren't collections should be a pretty inclusive strategy.

twosigmajab: Thanks @deckar01, I somehow completely misread this. In that case, I'll cancel this PR in favor of whatever changes you think should go in to improve interop (e.g. with werkzeug Headers) if any. At the very least maybe add a comment in the code that Python 2's lack of collections.Collection is a reason this exists?

barakalon:
Its worth noting that werkzeug's headers would no longer work, because it's `__iter__` method yields tuples of (key, value). The preemptive `isinstance` check caught this early.

marshmallow/utils.py
deckar01: I'm not familiar with the werkzeug headers problem. Can you create an issue and provide an example that reproduces the bug?

PR #1062, opened 2018-12-01
WIP - Add __eq__ method to Schema

### LINKED ISSUE #1061: RFC - Add __eq__ method to Schema
I'm thinking of adding an `__eq__` method to `Schema`, to compare schema instances.

~The only use case I have in mind is apispec (https://github.com/marshmallow-code/apispec/pull/333#issuecomment-443206311), where we need to compare `Schema` instances and modifiers matter. If this doesn't make it to marshmallow, it can be done in an util function in apispec.~

I figured it would be nice to offer this, and here is the best place to keep the list of modifiers up to date. But maybe I overlooked some difficulties and corner cases.

Edit: On second thought, this is not a good idea to use that for apispec. I'm keeping this open for now in case anyone wants to comment, but I'm starting to think this feature is not such a good idea.

lafrech:
Closes #1061.

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -213,6 +213,9 @@ class SchemaOpts(object):
         self.unknown = getattr(meta, 'unknown', RAISE)
 
 
+MODIFIERS = {'only', 'exclude', 'many', 'context', 'load_only', 'dump_only', 'partial', 'unknown'}
+
+
 class BaseSchema(base.SchemaABC):
     """Base schema class with which to define custom schemas.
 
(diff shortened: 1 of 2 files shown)


lafrech:
Let's forget this for now.

lafrech:
Copying the code here for the record before deleting the branch.

```py
MODIFIERS = {'only', 'exclude', 'many', 'context', 'load_only', 'dump_only', 'partial', 'unknown'}

    def __eq__(self, other):
        if not isinstance(other, base.SchemaABC):
            return NotImplemented
        if self.__class__ != other.__class__:
            return False
        return all(getattr(self, m) == getattr(other, m) for m in MODIFIERS)

    def __ne__(self, other):
        return not self.__eq__(other)
```

PR #1063, opened 2018-12-01
Fix getting values from dicts with properties

### LINKED ISSUE #1060: Marshmallow 3 can't dump attributes from dict-likes
I'm trying to port [umongo](https://github.com/Scille/umongo) to marshmallow 3 and I stumbled upon this:

```py
import marshmallow as ma


class MyDict(dict):
    @property
    def prop(self):
        return 12


class MySchema(ma.Schema):
    prop = ma.fields.Int()


print(MySchema().dump(MyDict()))

# MA 2
MarshalResult(data={'prop': 12}, errors={})

# MA 3
{}
```

It is due to @taion's https://github.com/marshmallow-code/marshmallow/pull/811.

In `get_value`, we try to get the value as key and in case of `KeyError`, `default` is returned before even trying to get as attribute.

This is wrong for structures that can have keys and attributes.

I think this is an issue in marshmallow.

taion:
Fixes #1060

The `hasattr` check shortcut also speeds up the benchmark by another 10% or so. Python exception handling is sort of slow, given how much this sort of pattern is recommended.

### DIFF
--- a/marshmallow/utils.py
+++ b/marshmallow/utils.py
@@ -362,12 +362,13 @@ def _get_value_for_keys(obj, keys, default):
 
 
 def _get_value_for_key(obj, key, default):
+    if not hasattr(obj, '__getitem__'):
+        return getattr(obj, key, default)
+
     try:
         return obj[key]
-    except (TypeError, AttributeError):
+    except (KeyError, IndexError, TypeError, AttributeError):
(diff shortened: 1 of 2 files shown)


sloria (approved):
Looks good. Verified that this issue is only on 3.0.

PR #1067, opened 2018-12-04
#852 configurable error messages

### LINKED ISSUE #852: Make the error messages for "unknown fields" and "invalid data type" configurable

Dunstrom:
Here is a suggested way of implementing custom error messages for unknown field and invalid type ValidationErrors in Schema. Please tell me if there is anything I could improve!  Resolves #852


sloria:
Thank for sending the PR @Dunstrom . I'm busy with personal things these days and haven't been able to put as much time into OSS as I'd like, so I probably won't get to it this week.

@lafrech or @deckar01 may be able to look at this, but they're working on marshmallow in their free time, too, and I don't know their availability.

deckar01:
Having to pass custom error messages through the unmarshaller constructor seems odd to me. This makes me think that the unmarshaller is doing something that should be the responsibility of the schema.

I don't fully understand why these error messages need to be configured per schema (or globally tbh). There are a lot of error messages in the schema class and none of them are configurable.

Dunstrom:
@deckar01 as specified in the issue #852 by @sloria it was on the `Schema` you should be able to configure the error message but the error message is thrown in the `Unmarshaller`. So I needed to pass the customized error messages somehow. If you have a better idea on how to do that I'd love to hear it.

To why this is needed, you have to ask @sloria I was not involved in creating the issue unfortunately.

sloria:
> I don't fully understand why these error messages need to be configured per schema (or globally tbh). There are a lot of error messages in the schema class and none of them are configurable.

@deckar01 All field-level error messages are configurable. "Invalid fields" and and "invalid type" errors are the only end user-facing error messages that aren't configurable.

> Having to pass custom error messages through the unmarshaller constructor seems odd to me...

I agree with this. I don't have time to code dive right now, but it'd be worth exploring alternative implementation(s).

lafrech:
I don't see how we could achieve this any better as long as the separation between `Schema` and `Marshaller`/`Unmarshaller` is what it is.

My intuition is that we should keep the `ErrorStore` class separated, as @deckar01 already suggested in the past, and integrate methods from `Marshaller`/`Unmarshaller`  in `Schema`. The validation/serialization/deserialization methods belong to the `Schema`, not to the error store.

(Be sure to instantiate `ErrorStore` each time we load/dump to avoid thread-safety issues.)

I think we need this fix (configurable messages) for 3.0 as failing to do that would be a regression for libraries providing i18n, for instance (e.g. [umongo](https://github.com/Scille/umongo)). If we can't refactor, maybe this PR in its current state is acceptable.

lafrech:
I just sent a PR with the refactor described in my comment above: https://github.com/marshmallow-code/marshmallow/pull/1070.

sloria:
> I think we need this fix (configurable messages) for 3.0 as failing to do that would be a regression for libraries providing i18n, for instance (e.g. umongo)

@lafrech  Just for my understanding: why would this be a regression for umongo (or any other library using the gettext approach)? Wouldn't you just include the hard-coded error messages in your translation file?

lafrech:
@sloria:

> why would this be a regression for umongo (or any other library using the gettext approach)? Wouldn't you just include the hard-coded error messages in your translation file?

I'm not familiar with the feature so I might be mistaken (I didn't code it and never used it). You can put the translation in the translation file, but you need to replace the string with the same string wrapped in a call to the translation function. I think it happens [here](https://github.com/Scille/umongo/blob/master/umongo/abstract.py#L103) in umongo's base Field. When porting to marshmallow 3, one would need to do the same with the base Schema, overriding

```py
    _default_error_messages = {
        'type': 'Invalid input type.',
        'unknown': 'Unknown field.',
    }
```

with

```py
    _default_error_messages = {
        'type': _('Invalid input type.'),
        'unknown': _('Unknown field.'),
    }
```

(Or using [`I18nErrorDict`](https://github.com/Scille/umongo/blob/master/umongo/abstract.py#L68) like what is done in `BaseField`.)

@Dunstrom Could you please rebase? We refactored marshalling so that you don't need to pass error messages to `Marshaller`: it doesn't exist anymore.

Dunstrom:
Sorry for late answer, holidays.. Right now I am studying for exams, I can have a look after my exams but that will not be until 15 jan.

deckar01:
I'm not sure what the etiquette is for updating PRs. I merged `dev` into this MR to preserve the commit history. No material changes were made in the merge commit, because I either kept both or only kept the changes from `dev`. I added commits to account for the marshalling refactor and made a few personal refactors.

lafrech:
May I suggest we merge the custom error messages with the default ones into `self.error_messages` at `__init__`, like [what is done in `Field`](https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/fields.py#L169)?

This would avoid doing the "get custom message / fallback on default" logic everytime and duplicating it in the code.

deckar01:
Ya, there is some logic for making this play nicely with inheritance we need also.

sloria (approved):

sloria:
Thanks @Dunstrom for the PR and thanks @deckar01 for bringin it home 👌 ✊ 👍

PR #1070, opened 2018-12-17
Refactor marshmalling.py

lafrech:
Quick implementation of my suggestion in https://github.com/marshmallow-code/marshmallow/pull/1067#issuecomment-447763942: removing `Marshaller`/`Unmarshaller` and adding their methods to `Schema`, keeping `ErrorStore` as a separate class.

I had to remove all the test on `Marshaller` and `Unmarshaller`. I didn't check if they were redundant or if we needed to reproduce the same use cases from `Schema`. I left them commented in he first commit (https://github.com/marshmallow-code/marshmallow/commit/094e7ff5c64620fe5b54a22050d61416ec894614) in case you want to check.


deckar01:
The code changes look good. I combed through the tests that didn't get copied and they all seemed to have corresponding tests in schema, serialization, or deserialization that touch the same features. Could we keep the tests with minimal refactoring if we stubbed a schema instance and called the new methods through it?

lafrech:
> The code changes look good. I combed through the tests that didn't get copied and they all seemed to have corresponding tests in schema, serialization, or deserialization that touch the same features. 

Wow, thank you so much for reviewing this.

> Could we keep the tests with minimal refactoring if we stubbed a schema instance and called the new methods through it?

Since you checked the features are tested at `Schema` level, I'd say it's fine to drop those tests. Better than keeping duplicate tests.

marshmallow/error_store.py
```diff
@@ -0,0 +1,83 @@
+# -*- coding: utf-8 -*-
+"""Utility classes and values used for marshalling and unmarshalling objects to
```
sloria: This needs to be changed. How about 

```
Utilities for storing collections of error messages.
```

sloria (commented):
Nice. Yeah, I think this is the way to go.

Just a small nit about the docstring. Otherwise, LGTM.

marshmallow/error_store.py
lafrech: Good catch. Fixed.

sloria (approved):

Issue #580, opened 2017-02-02
Add "on" and "off" values to Boolean truthy/falsy sets


marcellarius:
Rationale: Checkboxes in HTML forms use the value "on" when checked. This would be convenient as it would allow checkbox fields to be parsed as booleans when using Webargs with HTML forms.

podhmo:
this is work-around solution.

```python
from marshmallow import Schema, fields


class MyBoolean(fields.Boolean):
    truthy = fields.Boolean.truthy.union({"on"})
    falsy = fields.Boolean.falsy.union({"off"})


class S(Schema):
    v = MyBoolean()


print(S().load({"v": True}))
print(S().load({"v": False}))
print(S().load({"v": "on"}))
print(S().load({"v": "off"}))

# UnmarshalResult(data={'v': True}, errors={})
# UnmarshalResult(data={'v': False}, errors={})
# UnmarshalResult(data={'v': True}, errors={})
# UnmarshalResult(data={'v': False}, errors={})
```

justanr:
A better solution might to allow iterables of truthy and falsey values in the constructor. Or using the function/method field to pipe the value through a custom truth check

sloria:
I am open to this. Looks like this was implemented in Django-REST-framework's BooleanField (https://github.com/tomchristie/django-rest-framework/commit/276ed80fd302184dddb3af01e53d43be4aef15e4).

PRs welcome.

PR #1081, opened 2019-01-03
feat: fields.Boolean parses yes, no

r1b:
DRF does this [1]. Postgres accepts yes / no literals [2].

My actual use case is that checkboxes in Oracle Netsuite can be rendered as Yes
/ No when exported to CSV [3].

[1] https://github.com/encode/django-rest-framework/blob/0eb2dc1137189027cc8d638630fb1754b02d6cfa/rest_framework/fields.py#L662-L677
[2] https://www.postgresql.org/docs/current/datatype-boolean.html
[3] https://docs.oracle.com/cloud/latest/netsuitecs_gs/NSCVI/NSCVI.pdf (p. 285)

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -856,6 +856,8 @@ class Boolean(Field):
         't', 'T',
         'true', 'True', 'TRUE',
         'on', 'On', 'ON',
+        'y', 'Y',
+        'yes', 'Yes', 'YES',
         '1', 1,
         True,
     }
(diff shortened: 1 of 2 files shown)


r1b:
See https://github.com/marshmallow-code/marshmallow/issues/580

sloria (approved):
Makes sense. `yes` and `no` are also used by YAML and INI files.

r1b:
Thanks!

PR #1083, opened 2019-01-03
write a more optimized get_values function

gtmanfred:
Instead of checking the type of value, just catch exceptions if they are
not working.

This performs significantly better with large datasets being dumped from
schemas.

### DIFF
--- a/marshmallow/utils.py
+++ b/marshmallow/utils.py
@@ -335,7 +335,7 @@ def pluck(dictlist, key):
 
 # Various utilities for pulling keyed values from objects
 
-def get_value(obj, key, default=missing):
+def get_value(obj, key, default=missing, delimiter='.'):
     """Helper for pulling a keyed value off various types of objects. Fields use
     this method by default to access attributes of the source object. For object `x`
     and attribute `i`, this method first tries to access `x[i]`, and then falls back to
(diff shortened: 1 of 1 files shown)


gtmanfred:
This might make sense to put in the 2.x branch if there are still going to be release for that.

Thanks,
Daniel

lafrech:
Copying @taion as he's been working on this part of the code.

taion:
This isn't faster in the general case. First, the hasattr-before-try/except pattern is actually faster when working with objects in Marshmallow, which is likely the more common case (e.g. dumping ORM objects), as exception handling in Python is slow:

image

Second, this is also slower per the bundled benchmark:

```
(marshmallow):~/vendor/marshmallow (gtmanfred-dev)$ git co dev
Switched to branch 'dev'
Your branch is up to date with 'upstream/dev'.
(marshmallow):~/vendor/marshmallow (dev=)$ python performance/benchmark.py
Benchmark Result: 1143.14 usec/dump
(marshmallow):~/vendor/marshmallow (dev=)$ python performance/benchmark.py
Benchmark Result: 1090.66 usec/dump
(marshmallow):~/vendor/marshmallow (dev=)$ python performance/benchmark.py
Benchmark Result: 1159.73 usec/dump
(marshmallow):~/vendor/marshmallow (dev=)$ git co gtmanfred-dev
Switched to branch 'gtmanfred-dev'
(marshmallow):~/vendor/marshmallow (gtmanfred-dev)$ python performance/benchmark.py
Benchmark Result: 1249.06 usec/dump
(marshmallow):~/vendor/marshmallow (gtmanfred-dev)$ python performance/benchmark.py
Benchmark Result: 1312.89 usec/dump
(marshmallow):~/vendor/marshmallow (gtmanfred-dev)$ python performance/benchmark.py
Benchmark Result: 1293.49 usec/dump
```

taion:
So broadly what's happening here is that while the `hasattr` check slows down operations on dictionaries a bit, the exception handling slows down operations on non-dictionary objects by quite a lot.

The more optimal thing to do would be to figure out which value getter to use for any given object (since the result of that `hasattr` call isn't going to change), but that would require a bit of a refactor.

The even more optimal thing to do would be to just use something like toastedmarshmallow plus https://github.com/lyft/toasted-marshmallow/pull/16, though, which speeds things up dramatically more, especially on serialization.

deckar01:
Running a `for` loop on a `split` that results in a single item is ~10% slower than using the `in` operator to avoid the `for` loop.

Inlining `_get_value_for_keys` with a `for` loop to avoid the overhead of method invocation of recursion is less than 1% faster.

gtmanfred:
ahh, yeah when i originally wrote this for testing, I was only doing dictionary stuff, which is also what is being done in `salt` where the idea for this was lifted and where only dictionaries are used.

https://github.com/saltstack/salt/blob/2018.3/salt/utils/data.py#L451-L465

I have looked at toastedmarshmallow, but we have a bunch of custom validation 🙄 which is causing a huge slow down as well.

gtmanfred:
I did some testing with the benchmark script and with the python timeit module, and am opening a completely different pr that makes dictionaries faster.  #1084

Thanks!

PR #1087, opened 2019-01-07
Change order of keys and values arguments to fields.Dict()

sloria:
"keys" before "values" is a more natural ordering and
is more consistent with type annotations, e.g.
`typing.Dict[str float]`.

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -1178,15 +1178,15 @@ class Dict(Field):
 
     Example: ::
 
-        numbers = fields.Dict(values=fields.Float(), keys=fields.Str())
+        numbers = fields.Dict(keys=fields.Str(), values=fields.Float())
 
-    :param Field values: A field class or instance for dict values.
     :param Field keys: A field class or instance for dict keys.
+    :param Field values: A field class or instance for dict values.
     :param kwargs: The same keyword arguments that :class:`Field` receives.
 
     .. note::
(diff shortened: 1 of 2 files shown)


marshmallow/fields.py
```diff
@@ -1178,7 +1178,7 @@ class Dict(Field):
 
     Example: ::
 
-        numbers = fields.Dict(values=fields.Float(), keys=fields.Str())
+        numbers = fields.Dict(keys=fields.Str(), values=fields.Float())
 
     :param Field values: A field class or instance for dict values.
     :param Field keys: A field class or instance for dict keys.
```
deckar01, line 1183: The order of the doc params should probably be switched as well.

deckar01 (changes requested):
I assumed the value type would need to be constrained much more often than the key type. The order was chosen to allow the value type to be specified as the first positional argument. I don't feel strongly about keeping this convenience. Borrowing from the typing module makes sense to me.

marshmallow/fields.py
sloria: Good catch. Fixed.

deckar01 (approved):

sloria:
> I assumed the value type would need to be constrained much more often than the key type. The order was chosen to allow the value type to be specified as the first positional argument.

I see what you mean, but key-first makes more sense when passing two positional arguments.

Issue #1092, opened 2019-01-10
Adding a custom loader `field.Dict(loader=OrderedDict, dumper=OrderedDict)`


sayanarijit:
Hi
As mentioned in #1085 marshmallow will not support OrderedDict in near future. As an alternative can we have a parameter `loader` to pass a custom loader?
We can pass loader and dumper like `field.Dict(loader=MyCustomOrderedDict, dumper=OrderedDict)`.

PR #1093, opened 2019-01-10
Custom loader and dumper for field.Dict

sayanarijit:
Issue: #1092 

Below params added to fields.Dict

    :param type loader: The loader to use for loading objects. Defaults to dict
    :param type dumper: The dumper to use for dumping objects. Defaults to dict

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -1182,6 +1182,8 @@ class Dict(Field):
 
     :param Field keys: A field class or instance for dict keys.
     :param Field values: A field class or instance for dict values.
+    :param type loader: The loader to use for loading objects. Defaults to dict
+    :param type dumper: The dumper to use for dumping objects. Defaults to dict
     :param kwargs: The same keyword arguments that :class:`Field` receives.
 
     .. note::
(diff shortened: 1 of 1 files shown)


sloria:
Thanks for the PR.

I'm lukewarm on this because of the added API surface as well as the possible performance regression (due to an intermediate list being created). I wonder if we could expose dict creation via a method hook 🤔 .

Not going to merge this immediately, while we think of alternative approaches..

sayanarijit:
Hi @sloria 
Thanks for considering this. I think performance concerns should be solved by 2nd commit.
If it is added, it will help with use cases [like this](https://github.com/rapidstack/RESTEasyCLI/blob/master/resteasycli/schema/common.py#L7) where ordering is critical to maintain.

sloria:
We generally prefer separate field classes over parameters for (de)serializing to different types, i.e. `OrderedDict` > `Dict(loader=OrderedDict)`. Also, is there a need for the "load" type to be different from the "dump" type? I'm wondering if we should make the API similar to `Number`, with a `dict_type` class variable.

sayanarijit:
This is also a good idea. Right now I cant's think of a reason of separating loader and dumper. However, as it will be parent field for all dict types, and serve as a generic field for customized needs, maybe we can let it be a little more flexible. I don't know, maybe we can have a single dict_type field also to go with the convention as no other field is using separate loader and dumper.

sayanarijit:
Have a look at #1095

Issue #1090, opened 2019-01-09
Custom error message for unknown when dump only


choco:
I'm just getting started with marshmallow so if there's a more elegant way to solve the issue please let me know.
I have the following simple model and schema.

```python
class Instance(db.Model):
    __tablename__ = 'instances'

    id = db.Column(db.Integer, primary_key=True)
    hostname = db.Column(db.Unicode, index=True, unique=True, nullable=False)
    port = db.Column(db.Integer, nullable=False)

    def __repr__(self):
        return '<Instance: {}>'.format(self.hostname)

class InstanceSchema(Schema):
    id = fields.Integer(dump_only=True)
    hostname = fields.String(
        required=True,
        error_messages={'required': {'message': 'Hostname required', 'code': 400}}
    )
    port = fields.Integer(
        required=True,
        error_messages={'required': {'message': 'Port required', 'code': 400}}
    )
    @validates_schema
    def at_least_one(self, data):
        if self.partial and not len(data.keys()):
            raise ValidationError('Provide at least one valid key')
```

As you can see `id` is `dump_only`. If I try to load a simple dict (in the real application this would be data of a PATCH request)
```python
try:
    result = InstanceSchema(partial=True).load({'id': 4})
    print(result)
except ValidationError as err:
    print(err.messages)
```
I'll get that id is an unknown field. So I have to manually check for it to provide a better error message like
```python
    if "id" in err.messages:
        err.messages["id"] = "id is readonly"
```

It would be cool if it was possible to have a custom error message like hostname or port like
```python
id = fields.Integer(dump_only=True,
    error_messages={'dump_only': {'message': 'id is readonly', 'code': 400}}
)

deckar01:
This seems like it would be consistent with the other field validation error messages since the `dump_only` option can be declared when creating a field instance.

We may also want to consider using a more specific default error message for dump only errors since unknown isn't exactly true if the consumer is aware of the schema output.

This should be a lot easier to implement now that the marshalling logic is in the schema. Previously the marshaller only had access to a filtered set of fields.

choco:
I'm not familiar with your codebase, but something like this seems to work
```diff
diff --git a/marshmallow/fields.py b/marshmallow/fields.py
index 01cff48..29f7831 100644
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -122,6 +122,7 @@ class Field(FieldABC):
         'required': 'Missing data for required field.',
[...]

PR #1094, opened 2019-01-10
Don't threat dump_ony fields as unknowns

choco:
...instead give them their own logic to provide more useful error messages

The current implementation pitfalls described in #1090 still stand, i.e. just try-catching a known failing function doesn't seem clean.

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -623,8 +623,6 @@ class BaseSchema(base.SchemaABC):
         else:
             partial_is_collection = is_collection(partial)
             for attr_name, field_obj in iteritems(fields_dict):
-                if field_obj.dump_only:
-                    continue
                 field_name = attr_name
                 if field_obj.data_key:
                     field_name = field_obj.data_key
@@ -636,6 +634,13 @@ class BaseSchema(base.SchemaABC):
                         (partial_is_collection and attr_name in partial)
                     ):
                         continue
+                elif field_obj.dump_only:
+                    try:
+                        field_obj.fail('dump_only')
+                    except ValidationError as err:
+                        error_store.store_error(err.messages, attr_name, index=index)
+                    continue
+
                 d_kwargs = {}
                 if isinstance(field_obj, Nested):
                     # Allow partial loading of nested schemas.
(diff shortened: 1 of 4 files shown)


marshmallow/schema.py
```diff
@@ -665,7 +670,6 @@ def _deserialize(
                 fields = {
                     field_obj.data_key or field_name
                     for field_name, field_obj in fields_dict.items()
-                    if not field_obj.dump_only
```
deckar01, line 668: This would change the behavior of the `if unknown == INCLUDE:` branch which is supposed to keep all values.

deckar01: See #875.

sloria:
We decided in https://github.com/marshmallow-code/marshmallow/issues/875 that dump-only fields should be treated as unknown.

As for https://github.com/marshmallow-code/marshmallow/issues/1090, 3.0.0rc3 will allow you to customize the "unknown field" error message, albeit on a per-Schema (rather than per-Field) basis.

choco:
After reading #875 I still don't understand why `unknown` and `dump-only` would  be treated the same. It doesn't seem intuitive at all in my opinion.
You've got this strong per-field property, `dump-only`, which explicitly prohibits loading but then workaround that with an option targeted to `unknown` values, doesn't ring correct to me.
If I understand correctly there's no validation performed either when using `INCLUDE` so why wouldn't someone just declare a schema without that field altogether?
And if someone still wants such a wonky field you could just add similar logic for `unknown` with `INCLUDE`, `EXCLUDE`, `RAISE` in the `_serialize` method mirroring `_deserialize`

I'm sorry if I reopened that debate with this pull request, you already discussed it a lot and I only started looking at marshmallow yesterday, so my understanding is probably limited.

marshmallow/schema.py
choco: Even if I kept that line, the error is already stored above; I'd have to check above for `INCLUDE` before storing the error. Am I wrong? Also see my comment below with my opinion on `INCLUDE` for `dump_only`

deckar01:
It's not really a field property, it is a schema property with a field interface for convenience. If the field validated `dump_only` itself, there wouldn't be any ambiguity.

PR #1095, opened 2019-01-10
Map field added

sayanarijit:
Issue: #1092 

Added a Map field to be used as a parent by Dict field. Other dict-like fields can also be defined by passing `map_type`. e.g: `map_type=OrderedDict`.

Dict now extends Map class with map_type=dict

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -1182,6 +1182,8 @@ class Dict(Field):
 
     :param Field keys: A field class or instance for dict keys.
     :param Field values: A field class or instance for dict values.
+    :param type loader: The loader to use for loading objects. Defaults to dict
+    :param type dumper: The dumper to use for dumping objects. Defaults to dict
     :param kwargs: The same keyword arguments that :class:`Field` receives.
 
     .. note::
@@ -1195,8 +1197,10 @@ class Dict(Field):
         'invalid': 'Not a valid mapping type.',
     }
 
-    def __init__(self, keys=None, values=None, **kwargs):
+    def __init__(self, keys=None, values=None, loader=dict, dumper=dict, **kwargs):
         super(Dict, self).__init__(**kwargs)
+        self.loader = loader
+        self.dumper = dumper
         if keys is None:
             self.key_container = None
         elif isinstance(keys, type):
@@ -1259,14 +1263,13 @@ class Dict(Field):
             }
 
         # Serialize values
-        # Note: the dict type (dict, OrderedDict,...) of the value is lost
         if self.value_container is None:
-            result = {keys[k]: v for k, v in iteritems(value) if k in keys}
+            result = self.loader([(keys[k], v) for k, v in iteritems(value) if k in keys])
         else:
-            result = {
-                keys[k]: self.value_container._serialize(v, None, None, **kwargs)
+            result = self.loader([
+                (keys[k], self.value_container._serialize(v, None, None, **kwargs))
                 for k, v in iteritems(value)
-            }
+            ])
 
         return result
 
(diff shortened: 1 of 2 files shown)


marshmallow/fields.py
```diff
@@ -1259,14 +1261,14 @@ def _serialize(self, value, attr, obj, **kwargs):
             }
 
         # Serialize values
-        # Note: the dict type (dict, OrderedDict,...) of the value is lost
+        result = self.map_type()
         if self.value_container is None:
-            result = {keys[k]: v for k, v in iteritems(value) if k in keys}
```
deckar01, line 1264: You could keep these comprehensions by passing item pairs to the constructor.

```py
result = self.mapping_type((keys[k], v) for k, v in iteritems(value) if k in keys)

File: marshmallow/fields.py [hunk 2]
```diff
@@ -1172,14 +1172,15 @@ def _deserialize(self, value, attr, data, **kwargs):
             self.fail('invalid')
 
 
-class Dict(Field):
-    """A dict field. Supports dicts and dict-like objects. Optionally composed
+class Map(Field):
```
deckar01: The typing and collections modules call this abstract class `Mapping`.

https://docs.python.org/3/library/typing.html#typing.Mapping

https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping

marshmallow/fields.py
```diff
@@ -1311,6 +1314,20 @@ def _deserialize(self, value, attr, data, **kwargs):
...
+    Example: ::
+
+        numbers = fields.Map(keys=fields.Str(), values=fields.Float())
+
+    :param kwargs: The same keyword arguments that :class:`Map` receives.
+
+    .. versionadded:: 2.1.0
+    """
+    def __init__(self, **kwargs):
+        super(Dict, self).__init__(map_type=dict, **kwargs)
```
deckar01, line 1330: You could reduce the boilerplate by making this declarative and allowing the `Mapping` field to default to the `Mapping` collection type.

```py
class Dict(Mapping):
    mapping_type=collections.Mapping
    ...

class Dict(Mapping):
    mapping_type=dict
```

marshmallow/fields.py
sayanarijit: Tried similar approach in #1093 but as @sloria pointed out, i think there will be additional overhead of creating tuple for each pair.

marshmallow/fields.py
sayanarijit: Let's call this field Mapping and make alias Map = Mapping?

marshmallow/fields.py
sayanarijit: Issue with this is if I want to create my own custom dict field, I need to create class and change field by inheriting it. If we pass the mapping type as argument, it will be as easy as `myfield = fields.Mapping(map_type=MyCustomDict)`

sloria: I agree with @deckar01 's suggestion. As pointed out in the previous PR, we generally prefer new field types to parameters because it simplifies introspection by libraries like apispec. Creating a new class is not much more boilerplate than a constructor parameter, especially if you will reuse the field

```python
class OrderedDictField(Mapping):
    mapping_type = OrderedDict

my_field = OrderedDictField()
```

marshmallow/fields.py
sloria: I don't think the `Map` alias is necessary since it's there's no corresponding "Map" type built in to Python.

marshmallow/fields.py
sloria: Yeah, creating tuples incurs a ~2x performance penalty.

image

sloria (changes requested):

marshmallow/fields.py
sayanarijit: Actually I was thinking from the user's perspective i.e. how easily an user can define a field in their schema.
In case of Dict, user doesn't need to pass any extra argument... only keys, values and kwargs. User need to pass map_type only while initializing Mapping field.

If we take the alternate approach, we need to define Mapping as abstract class. Check #1098

sloria: Implementing a custom class in userland is the way to go, and consistent with how `Number` is implemented. OK, let's continue on #1098.

sloria:
Closing in favor of #1098

PR #1098, opened 2019-01-11
Mapping field

sayanarijit:
Issue: #1092

Added a Mapping field as an abstract class with mapping_type as an abstract property as suggested in #1095  

Dict now extends Mapping class with mapping_type=dict

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -117,3 +117,4 @@ Contributors (chronological)
 - `@toffan <https://github.com/toffan>`_
 - Hampus Dunström `@Dunstrom <https://github.com/Dunstrom>`_
 - Robert Jensen `@r1b <https://github.com/r1b>`_
+- Arijit Basu `@sayanarijit <https://github.com/sayanarijit>`_
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -10,7 +10,6 @@ import numbers
 import uuid
 import decimal
 import math
-from abc import ABCMeta, abstractproperty
 
 from marshmallow import validate, utils, class_registry
 from marshmallow.base import FieldABC, SchemaABC
@@ -23,6 +22,7 @@ __all__ = [
     'Field',
     'Raw',
     'Nested',
+    'Mapping',
     'Dict',
     'List',
     'String',
@@ -1191,11 +1191,8 @@ class Mapping(Field):
 
     .. versionadded:: TODO: specify version
     """
-    __metaclass__ = ABCMeta
 
-    @abstractproperty
-    def mapping_type(self):
-        pass
+    mapping_type = _Mapping
 
     default_error_messages = {
         'invalid': 'Not a valid mapping type.',
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -1174,12 +1174,7 @@ class TimeDelta(Field):
 
 
 class Mapping(Field):
-    """An abstract class for objects with key-value pairs. Optionally composed
-    with another `Field` class or instance.
-
-    Example: ::
-
-        numbers = fields.Mapping(keys=fields.Str(), values=fields.Float())
+    """An abstract class for objects with key-value pairs.
 
     :param Field keys: A field class or instance for dict keys.
     :param Field values: A field class or instance for dict values.
@@ -1315,6 +1310,7 @@ class Mapping(Field):
 
         return result
 
+
 class Dict(Mapping):
     """A dict field. Supports dicts and dict-like objects. Extends
     Mapping with dict as the mapping_type.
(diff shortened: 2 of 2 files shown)


marshmallow/fields.py
```diff
@@ -1188,15 +1189,20 @@ class Dict(Field):
         When the structure of nested data is not known, you may omit the
         `keys` and `values` arguments to prevent content validation.
 
-    .. versionadded:: 2.1.0
+    .. versionadded:: TODO: specify version
     """
+    __metaclass__ = ABCMeta
```
sloria: This won't work on Python 3. You could use `with_metaclass` from `marshmallow.compat`, but it's not really necessary. We've eschewed the `abc` module for the sake of simplicity. `Number` doesn't use it, so we should keep `Mapping` consistent.

AUTHORS.rst
```diff
@@ -117,3 +117,4 @@ Contributors (chronological)
 - `@toffan <https://github.com/toffan>`_
 - Hampus Dunström `@Dunstrom <https://github.com/Dunstrom>`_
 - Robert Jensen `@r1b <https://github.com/r1b>`_
+- Arijit Basu `@sayanarijit <https://github.com/sayanarijit>`_
```
sloria, line 120: Thanks for doing this 👍

sloria (changes requested):
This is getting closer! Just one comment to address.

marshmallow/fields.py
sayanarijit: collection.Mapping does't support initialization.
still keep `mapping_type = collection.Mapping?` in Mapping field? or keep dict as default?

sayanarijit: Alright... I understood the concept here... Mapping will provide the need for abstraction itself.

sayanarijit: Done in last commit

marshmallow/fields.py
```diff
@@ -1172,13 +1173,13 @@ def _deserialize(self, value, attr, data, **kwargs):
...
-class Dict(Field):
-    """A dict field. Supports dicts and dict-like objects. Optionally composed
+class Mapping(Field):
+    """An abstract class for objects with key-value pairs. Optionally composed
     with another `Field` class or instance.
 
     Example: ::
 
-        numbers = fields.Dict(keys=fields.Str(), values=fields.Float())
+        numbers = fields.Mapping(keys=fields.Str(), values=fields.Float())
```
sloria: We should probably drop this example, since `Mapping` isn't meant to be instantiated.

sloria (changes requested):

marshmallow/fields.py
```diff
@@ -1172,13 +1173,13 @@ def _deserialize(self, value, attr, data, **kwargs):
             self.fail('invalid')
 
 
-class Dict(Field):
-    """A dict field. Supports dicts and dict-like objects. Optionally composed
+class Mapping(Field):
+    """An abstract class for objects with key-value pairs. Optionally composed
     with another `Field` class or instance.
```
sloria: > Optionally composed with another `Field` class or instance.

I realize you didn't add this, but it should be corrected, since `Mapping` classes can be composed with both a "key" and a "value" container. How about we just drop that sentence?

marshmallow/fields.py
sayanarijit: Right. I missed to correct this.

marshmallow/fields.py
sayanarijit: Done

marshmallow/fields.py
sayanarijit: One doubt. If we compare it with the Number field, Number defined float as the default type. Not numbers.Number. Following this pattern we should have used dict as the default mapping type for Mapping.

sayanarijit:
I think we should have used dict as the default type as Number uses float.

marshmallow/fields.py
sloria: Yeah, `dict` as the default mapping_type makes sense.

sayanarijit: Done and pushed.

sloria (approved):

lafrech:
LGTM.

Note there's a TODO left in the code:

    versionadded:: TODO: specify version

Is it worth adding a test demonstrating how it works with e.g. an `OrderedDict`?

Would `mapping_cls` or `mapping_class` be more explicit than `mapping_type`?

sloria:
> Note there's a TODO left in the code:

Yep, noted. Was going to fill that in after merging.

> Is it worth adding a test demonstrating how it works with e.g. an OrderedDict?

Not strictly necessary. Since `Dict` is tested, one could argue that testing `Mapping` is testing an implementation detail. ¯\\_(ツ)_/¯

> Would mapping_cls or mapping_class be more explicit than mapping_type?

Not necessarily. A type (as in, a `type` object) of `Mapping` is passed. Also, `Number` uses `num_type`.

sloria:
This is released in 3.0.0rc4.

PR #1099, opened 2019-01-11
Fix "missing" singleton copy/deepcopy

lafrech:
I had troubles while deep-copying objects and came to realize that `missing` was duplicated.

It was not obvious from the `print`s because the `repr` is the same but when printing the variable's address (`hex(id(variable))`) I could see that it was duplicated.

This PR implements [this answer from SO](https://stackoverflow.com/a/9887720/4653485).

We want a singleton because there are tests in the code like

```py
    if value is missing:
```

### DIFF
--- a/marshmallow/utils.py
+++ b/marshmallow/utils.py
@@ -35,6 +35,12 @@ class _Missing(object):
 
     __nonzero__ = __bool__  # PY2 compat
 
+    def __copy__(self):
+        return self
+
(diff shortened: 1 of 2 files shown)


sloria:
This is a bugfix. Can you resend to `2.x-line`?

lafrech:
Closing in favor of 2.x version: https://github.com/marshmallow-code/marshmallow/pull/1100.

Issue #644, opened 2017-06-09
tuple field


rizplate:
is there a tuple filed?

douglas-treadwell:
There doesn't seem to be one, but there is fields.List.  What's your use case @rizplate?

stefanoborini:
Some applications expect the tuple data type (e.g. for array shapes), so if you can't specify that a field is a tuple, you will lose the information and end up with a list. json has only lists, but marshmallow could know the specific requirement for type and do the conversion itself.

sloria:
There is no tuple field out of the box, but you could easily add it as a custom field in your application:

```python
from marshmallow import fields


class TupleField(fields.List):
    def _serialize(self, value, attr, obj):
        ret = super()._serialize(value, attr, obj)
        if ret is None:
            return None
        return tuple(ret)


field = TupleField(fields.Int())
assert field.serialize('data', {'data': [1, 2]}) == (1, 2)
```

taion:
@sloria This isn't quite the same. A tuple is a fixed-length collection of potentially heterogenous fields. You'd use it for something like `fields.Tuple((fields.String, fields.Int))`.

jo-soft:
```python
class TupleField(fields.Field):

    def __init__(self, tuple_entries, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.tuple_entries = tuple_entries

    def _serialize(self, attr, obj, accessor=None):
        return tuple(
            (field._serialize(val, obj, accessor) for field, val in zip(self.tuple_entries, attr))
        )

    def _deserialize(self, value, attr=None, data=None):
        return (
            (field._deserialize(val, attr, data) for field, val in zip(self.tuple_entries, value))
        )
```

usage:

```python

class SchemaWithTupleOfNumbers(Schema):
    tuples  = TupleField(
        (fields.Number(required=True), fields.Number(required=True))
    )
```

lafrech:
@jo-soft you may want to check the length of `value` matches the length of `tuple_entries`.

taion:
Right – this is definitely doable, but it's nontrivial yet still common enough that it seems to make sense to add here.

jo-soft:
@lafrech true. it's only a POC thats why I did not add any error handling.

Maybe it also should not relay on length because the above implementation accepts any kind of iterable which may not have a length defined.
It may be better to throw an error if there is no field for a given value.

one more point to be discussed:
the value `attr`, `accessor` and `data` make less sense like it's used above.

PR #1103, opened 2019-01-14
Add tuple field

zblz:
This is a proposal to add a tuple field, defined as a fixed-length, heterogeneous list. This `Tuple` field allows to define a list that, e.g., is of length 3 and always has a string and two floats.

AFAIK, It is not possible to build such a field or schema out of existing marshmallow fields. Some of the need for this type of field has been discussed in #644, and my particular need for it stems from trying to build a schema for a [Pandas Dataframe in the `orient=split` format](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.to_json.html), which has a `data` field with a list of such tuples:

```json
{
  "columns": ["col 1", "col 2"], 
  "index": ["row 1", "row 2"], 
  "data": [["a", 10], ["b", 20]]
}
```

In the example above column 1 if of type `string` and column 2 of type `int`, so the data field could be written in a mm schema as:

```
data = fields.Tuple([fields.String(), fields.Integer()], many=True)
```

The implementation below tries to mimic most behaviour of the `List` field to avoid doing things in different was for similar container fields.

There are still a couple of things to be done, but I thought it would be good to have a discussion on whether such a field is seen as a necessary addition and the implementation.

TODO:

- [x] Deserialization tests
- [x] Mention in documentation

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -686,13 +686,9 @@ class Tuple(Field):
         if not utils.is_collection(value):
             self.fail('invalid')
 
-        self.validate_length(value)
-
         return tuple(
-            (
-                container._serialize(each, attr, obj, **kwargs)
-                for container, each in zip(self.tuple_fields, value)
-            )
+            container._serialize(each, attr, obj, **kwargs)
+            for container, each in zip(self.tuple_fields, value)
         )
 
     def _deserialize(self, value, attr, data, **kwargs):
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -622,8 +622,8 @@ class Tuple(Field):
 
     .. note::
         Because of the structured nature of `collections.namedtuple` and
-        `typing.NamedTuple`, using a nested schema for them is more appropriate
-        than using a `Tuple` field.
+        `typing.NamedTuple`, using a Schema within a Nested field for them is
+        more appropriate than using a `Tuple` field.
 
     :param Iterable[Field] tuple_fields: An iterable of field classes or
         instances.
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -699,8 +699,6 @@ class Tuple(Field):
     def _serialize(self, value, attr, obj, **kwargs):
         if value is None:
             return None
-        if not utils.is_collection(value):
-            self.fail('invalid')
 
         return tuple(
             container._serialize(each, attr, obj, **kwargs)
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -13,7 +13,11 @@ import math
 
 from marshmallow import validate, utils, class_registry
 from marshmallow.base import FieldABC, SchemaABC
-from marshmallow.utils import is_collection, missing as missing_
+from marshmallow.utils import (
+    is_collection, missing as missing_,
+    FieldInstanceResolutionError,
+    resolve_field_instance,
+)
 from marshmallow.compat import basestring, text_type, Mapping as _Mapping, iteritems
 from marshmallow.exceptions import ValidationError, StringNotCollectionError
 from marshmallow.validate import Validator, Length
@@ -646,10 +650,18 @@ class Tuple(Field):
                 'tuple_fields must be an iterable of Field classes or '
                 'instances.',
             )
-        self.tuple_fields = [
-            self._get_instance_from_field_class_or_instance(cls_or_instance)
-            for cls_or_instance in tuple_fields
-        ]
+
+        try:
+            self.tuple_fields = [
+                resolve_field_instance(cls_or_instance)
+                for cls_or_instance in tuple_fields
+            ]
+        except FieldInstanceResolutionError:
+            raise ValueError(
+                'Elements of "tuple_fields" must be subclasses or '
+                'instances of marshmallow.base.FieldABC.',
+            )
+
         self.validate_length = Length(equal=len(self.tuple_fields))
 
     def _bind_to_schema(self, field_name, schema):
@@ -661,23 +673,6 @@ class Tuple(Field):
             new_container.name = field_name
             new_tuple_fields.append(new_container)
         self.tuple_fields = new_tuple_fields
-
-    @staticmethod
-    def _get_instance_from_field_class_or_instance(cls_or_instance):
-        if isinstance(cls_or_instance, type):
-            if not issubclass(cls_or_instance, FieldABC):
-                raise ValueError(
-                    'Elements of "tuple_fields" must be subclasses or '
-                    'instances of marshmallow.base.FieldABC.',
-                )
-            return cls_or_instance()
-        else:
-            if not isinstance(cls_or_instance, FieldABC):
-                raise ValueError(
-                    'Elements of "tuple_fields" must be subclasses or '
-                    'instances of marshmallow.base.FieldABC.',
-                )
-            return cls_or_instance
 
     def get_value(self, obj, attr, accessor=None):
         """Return the value for a given key from an object."""
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -703,12 +703,12 @@ class Tuple(Field):
         for idx, (container, each) in enumerate(zip(self.tuple_fields, value)):
             try:
                 result.append(container.deserialize(each))
-            except ValidationError as e:
-                result.append(e.data)
-                errors.update({idx: e.messages})
-
+            except ValidationError as error:
+                if error.valid_data is not None:
+                    result.append(error.valid_data)
+                errors.update({idx: error.messages})
         if errors:
-            raise ValidationError(errors, data=result)
+            raise ValidationError(errors, valid_data=result)
 
         return tuple(result)
 
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -637,9 +637,8 @@ class Tuple(Field):
         super(Tuple, self).__init__(*args, **kwargs)
         if not utils.is_collection(tuple_fields):
             raise ValueError(
-                'The tuple elements argument must be an iterable of '
-                'subclasses or instances of subclasses of '
-                'marshmallow.base.FieldABC',
+                'tuple_fields must be an iterable of Field classes or '
+                'instances.',
             )
         self.tuple_fields = [
             self._get_instance_from_field_class_or_instance(cls_or_instance)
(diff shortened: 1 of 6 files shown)


zblz:
Pinging @acroz as I had some of the initial discussions on the need for this field and might want to comment on implementation.

marshmallow/fields.py
```diff
@@ -615,6 +616,110 @@ def _deserialize(self, value, attr, data, **kwargs):
...
+    default_error_messages = {
+        'invalid': 'Not a valid tuple.',
+    }
+
+    def __init__(self, tuple_fields, *args, **kwargs):
+        super(Tuple, self).__init__(*args, **kwargs)
+        if not utils.is_collection(tuple_fields):
+            raise ValueError(
+                'The tuple elements argument must be an iterable of '
+                'subclasses or instances of subclasses of '
```
sloria: How about:

```
tuple_fields must be an iterable of Field classes or instances.
```

marshmallow/fields.py
```diff
@@ -615,6 +616,110 @@ def _deserialize(self, value, attr, data, **kwargs):
...
+
+        for idx, (container, each) in enumerate(zip(self.tuple_fields, value)):
+            try:
+                result.append(container.deserialize(each))
+            except ValidationError as e:
+                result.append(e.data)
+                errors.update({idx: e.messages})
+
+        if errors:
+            raise ValidationError(errors, data=result)
```
sloria: I believe `data` should be `valid_data`. See https://github.com/marshmallow-code/marshmallow/pull/1052

marshmallow/fields.py
```diff
@@ -615,6 +616,110 @@ def _deserialize(self, value, attr, data, **kwargs):
...
+            return tuple(result)
+        return value
+
+    def _serialize(self, value, attr, obj, **kwargs):
+        if value is None:
+            return None
+        if not utils.is_collection(value):
+            self.fail('invalid')
+
+        self.validate_length(value)
```
sloria: Validation should not happen on serialization, since `obj` is assumed to be valid prior to serialization.

marshmallow/fields.py
```diff
@@ -615,6 +616,110 @@ def _deserialize(self, value, attr, data, **kwargs):
...
+            for cls_or_instance in tuple_fields
+        ]
+        self.validate_length = Length(equal=len(self.tuple_fields))
+
+    @staticmethod
+    def _get_instance_from_field_class_or_instance(cls_or_instance):
+        if isinstance(cls_or_instance, type):
+            if not issubclass(cls_or_instance, FieldABC):
+                raise ValueError(
+                    'The type of the tuple elements '
```
sloria: How about:

```
Elements of "tuple_fields" must be subclasses or instances of marshmallow.base.FieldABC.
```

marshmallow/fields.py
```diff
@@ -615,6 +616,110 @@ def _deserialize(self, value, attr, data, **kwargs):
...
+    def _serialize(self, value, attr, obj, **kwargs):
+        if value is None:
+            return None
+        if not utils.is_collection(value):
+            self.fail('invalid')
+
+        self.validate_length(value)
+
+        return tuple(
+            (
```
sloria: Is this extra level of nesting necessary?

sloria (changes requested):
Thanks for the PR!

Just a few drive-by comments for now.

Will look more closely at this when I have time.

zblz:
@sloria -- Thanks for the drive-by comments. I've fixed those so when you have more time you can look at the rest of the implementation.

justanr:
One thing that could be considered here is supporting namedtuples (both collections and typing) by accepting a "factory" argument that defaults to regular tuple.

This could also be handled by a post function but it'd be nice from an ease of use perspective.

Also :+1: this is a good addition. Seems intuitive to use from both end user and code generation perspectives.

sloria:
> One thing that could be considered here is supporting namedtuples (both collections and typing) by accepting a "factory" argument that defaults to regular tuple.

`Number` and `Mapping` have the `num_type` and `mapping_type` class variables. `tuple_type` would make sense here.

zblz:
Casting a `namedtuple` to a JSON list on serialization means dropping the information about the field names, so isn't it more naturally serialized as a JSON object rather than a fixed length list? In that case a `NamedTuple` field could be a subclass of `Mapping` rather that related to this `Tuple` field. I do think that a `NamedTuple` field would be another worthwhile addition to marshmallow. 

I agree though that using `tuple` as an object factory is not obvious, but the only alternative I see with the current serialization format of a JSON list would be a python list.

justanr:
> Casting a namedtuple to a JSON list on serialization means dropping the information about the field names, so isn't it more naturally serialized as a JSON object rather than a fixed length list? In that case a NamedTuple field could be a subclass of Mapping rather that related to this Tuple field.

In marshmallow-annotations we treat `typing.NamedTuple` as a nested schema. 

That might be the actual suggested advice here as well. Similar to the documentation about structured dictionary data which suggests using a nested scheme instead of `fields.Dict` in which case exposing the tuple factory might not make sense.

zblz:
> In marshmallow-annotations we treat `typing.NamedTuple` as a nested schema.
> That might be the actual suggested advice here as well. 

Do you mean that this `Tuple` field should return a nested schema when instantiated via a `NamedTuple` or that a future `NamedTuple` field might be implemented that way? I feel the former to be quite confusing, and the latter is a different pattern from any other fields in `marshmallow.fields`, but would work well with typed named tuples.

justanr:
@zblz what I mean is that when you create a marshmallow-annotations schema setup for typing.NamedTuple it generates a fields.Nested. 

And since named tuples are structured data you're right that a json list of the wrong thing to serialize to.

My sentence was a little confusing. What I actually mean is leave this field as is and create docs for it that suggest using fields.Nested for named tuples (either collections or typing)

zblz:
@justanr, @sloria -- I've added deserialization tests and schema binding functionality and tests. I believe this is now ready for a full review.

I've not found a place in the documentation where this field might be explained or mentioned beyond the API reference generated from the docstring. Is that correct or am I overlooking something?

marshmallow/fields.py
```diff
@@ -616,6 +617,118 @@ def _deserialize(self, value, attr, data, **kwargs):
...
+                    result.append(
+                        container.get_value(each, container.attribute),
+                    )
+            return tuple(result)
+        return value
+
+    def _serialize(self, value, attr, obj, **kwargs):
+        if value is None:
+            return None
+        if not utils.is_collection(value):
```
sloria: Validation should not happen on serialization, for performance reasons. `value` is already assumed to be valid.

sloria (commented):
Just one comment; otherwise looks good.

@justanr @lafrech @deckar01 Would be great to have one more pass on this before merging. I'll plan on merging this later this week.

marshmallow/fields.py
zblz: Done.

tests/test_deserialization.py
```diff
@@ -51,6 +51,10 @@ def test_list_field_deserialize_none_to_empty_list(self):
         field = fields.List(fields.String(allow_none=True), allow_none=True)
         assert field.deserialize(None) is None
 
+    def test_tuple_field_deserialize_none_to_nont(self):
```
lafrech: Typo

s/nont/none

marshmallow/fields.py
```diff
@@ -616,6 +617,116 @@ def _deserialize(self, value, attr, data, **kwargs):
...
+        new_tuple_fields = []
+        for container in self.tuple_fields:
+            new_container = copy.deepcopy(container)
+            new_container.parent = self
+            new_container.name = field_name
+            new_tuple_fields.append(new_container)
+        self.tuple_fields = new_tuple_fields
+
+    @staticmethod
+    def _get_instance_from_field_class_or_instance(cls_or_instance):
```
lafrech: This logic is copied from `List` and `Dict` fields. It could be factorized in utils.py.

utils.py

```py
def resolve_schema_instance():
        if isinstance(cls_or_instance, type):
            if not issubclass(cls_or_instance, FieldABC):
                raise SchemaTypeError()
            return cls_or_instance()
        else:
            if not isinstance(cls_or_instance, FieldABC):
                raise SchemaTypeError()
            return cls_or_instance
```

fields.py

```py
    try:
        self.tuple_fields = [
            resolve_schema_instance(cls_or_instance)
            for cls_or_instance in tuple_fields
        ]
    except SchemaTypeError:
        raise ValueError(
            'Elements of "tuple_fields" must be subclasses or '
            'instances of marshmallow.base.FieldABC.',
        )
```

Maybe this can be done later in another PR. The duplication issue is not new: the code was already duplicated in `List` and `Dict`.

The exception needs a better name.

marshmallow/fields.py
```diff
@@ -616,6 +617,116 @@ def _deserialize(self, value, attr, data, **kwargs):
...
 
+class Tuple(Field):
+    """A tuple field, composed of a fixed number of other `Field` classes or
+    instances
+
+    Example: ::
+
+        row = Tuple((fields.String(), fields.Integer(), fields.Float()))
+
+    .. note::
```
lafrech, line 623: I didn't understand that note until I read the conversation in the PR.

I don't understand if using a Nested field would work out of the box for this use case. If so, then we could at least rephrase as "using a Nested field". Otherwise, if a dedicated field must be written, then I'm not sure what to suggest.

marshmallow/fields.py
zblz: There is indeed quite a bit of duplication. I've factored that code out into `utils.resolve_field_instance` and modified List and Dict to use it as well.

marshmallow/fields.py
zblz: I've reworded it to "using a Schema within a Nested field", which reflects better my initial intention of indicating that the user needed to build a schema for the specific named tuple, which is not as obvious if we only mention Nested fields.

sloria:
@lafrech Are you good with the latest changes?

zblz:
@sloria, @lafrech -- I've added myself to the `AUTHORS.rst`. Anything else you need before merging this?

lafrech:
Looks good to me.

The exception should be declared in exceptions.py. I'll do that after merging.

sloria:
This is released in 3.0.0rc4.

mbrucher:
Is there a way of having a non-required tuple? I tried `allow_none`, `required=False`, but it seems that not having the parameter makes the parsing fail?

PR #1116, opened 2019-01-31
Fix doc block for schema.dump

snjypl:
schema.dump now returns only the data. not a tuple. fixed doc block


lafrech:
Good catch. Thanks.

Do you think you could add a commit to add yourself to AUTHORS.rst?

snjypl:
> 
> 
> Good catch. Thanks.
> 
> Do you think you could add a commit to add yourself to AUTHORS.rst?

Done !

Issue #885, opened 2018-07-22
Better default error messages for validators


mcgfeller:
The default error message for validators ('Not a valid choice.') is not really helpful. 

I suggest better default messages, for example:
- OneOf / ContainsOnly: 'invalid value: {input}; use one of {choices}.'
- Range: 'invalid value: {input}; must be between {min} and {max}.'

The error argument could be a callable as well, for example to restrict the enumeration of valid choices to a certain length. 

Thanks for Marshmallow and best regards,
Martin

sloria:
Thanks @mcgfeller for the feedback.

We removed input values from the default error messages in https://github.com/marshmallow-code/marshmallow/commit/27992463466a7d0936f4cef000402f853ed6f944#diff-6b084e078a8ab4b51fd1616021309a42 to address #255 and to prevent cases where clients render unsafe user input.

However, I don't see why we couldn't show static values such as `{choices}`, `{min}`, and `{max}`. 

* OneOf: `Not one of {choices}.`
* ContainsOnly: `One or more of the choices you made was not in {choices}.`

sloria:
To add to my previous point: DRF recently removed input values from error messages: https://github.com/encode/django-rest-framework/pull/5881 .

I still think it should be possible to include input values--it just shouldn't be the default.

hdoupe:
@sloria said:
> I still think it should be possible to include input values--it just shouldn't be the default.

This makes sense to me. I'm working on a [project](https://github.com/hdoupe/ParamTools) where the error messages really need to include the input in order to be helpful. I'm willing to take a crack at implementing a non-default option to include the input values, if that is still desired behavior. If not, I can find a work around to satisfy my own needs.

P.S. Marshmallow has been a joy to work with. Thanks for this awesome project.

sloria:
@hdoupe Feel free to send a PR to make it possible to include input values.

P.S. Thanks for the kind words!

sloria:
This was done in #1155.

PR #1129, opened 2019-02-10
Show invalid value if '{input}' included in 'invalid' error message

hdoupe:
This PR gives the user the option to show the invalid value for a subset of the fields: Boolean, Number (and its descendants), and DateTime (and its descendants). 

My motivation for displaying this information is that I want to show very helpful error messages to the users of [ParamTools][1], a parameter processing and validation library for computational modeling projects. Right now, I match the error message given by Marshmallow with the invalid value and tack the invalid value onto the end of the error message. I can continue to do this if Marshmallow isn't interested in having this functionality. 

This is related to #885.

[1]: https://github.com/PSLmodels/ParamTools

### DIFF
--- a/marshmallow/fields.py
+++ b/marshmallow/fields.py
@@ -796,7 +796,7 @@ class Number(Field):
         try:
             return self._format_num(value)
         except (TypeError, ValueError):
-            self.fail('invalid')
+            self.fail('invalid', input=value)
 
     def _to_string(self, value):
         return str(value)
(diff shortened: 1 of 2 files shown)


sloria (approved):
Looks good. Thanks for the PR!

We'll probably want to do this for all fields eventually, but no reason to block merging this PR.

hdoupe:
Sure, thanks for merging it!

PR #1131, opened 2019-02-12
Update documentation for partial schema parameter

rileyjohngibbs:
It seemed unclear at first glance whether the partial parameter would effectively set required to False for all fields on the schema. Updated the documentation to clarify this.


sloria:
Thanks!

Issue #1108, opened 2019-01-21
Nested field raises warning with strict=True


Aegdesil:
When using a Nested field in a `Schema` instantiated with `strict=True`, a deprecation warning is logged because the nested schema is instantiated without setting `strict`. I think the value of strict should be the same as the parent schema, so that the warning is only emitted when the parent schema is not strict.
 
```
from marshmallow import Schema, fields


class ChildSchema(Schema):
    pass

class ParentSchema(Schema):
    child = fields.Nested(ChildSchema)

parent_schema = ParentSchema(strict=True)
parent_schema.load({"child": {}})  # emits a ChangedInMarshmallow3Warning
```

Using marshmallow version `2.18.0`.

deckar01:
Changing `strict` to be inherited would be a breaking change that would go in the next major release, but v3 makes schemas always strict, so this is no longer an issue.

Aegdesil:
Well since it isn't documented (at least I didn't find anything about it), it depends if you consider it as a bug or a feature :)
V3 fixes it indeed, but it could still take some time before it is released.

deckar01:
I initially interpreted this as a proposal to change the way the strict property was inherited. After looking into it, strict only affects the behavior of the root schema, so it is already implicitly inherited.

```py
from marshmallow import Schema, fields


class ChildSchema(Schema):
    pass

class ParentSchema(Schema):
    child = fields.Nested(ChildSchema(strict=True))

parent_schema = ParentSchema()
parent_schema.load({"child": {}})  # Does not error
```

The warning does seem unnecessary since the nested schema is never loaded directly. Silencing the warning for this use case seems like it would be backwards compatible.

If you added a way to silence the warning when constructing a nested schema you would still get the error when Nested is passed a schema instance. If you moved the warning to the load method you would also need to add it to the validate method. The warning would occur every time one of these methods were called, so you would need to setup a flag and use it in both methods to determine if the warning has already been emitted.

I think the intended way to make that warning go away is to make all of your schemas strict, not just the ones that get loaded directly.

Aegdesil:
> After looking into it, strict only affects the behavior of the root schema, so it is already implicitly inherited.

So setting strict in the child schema instance from the parent instance would not be breaking right?

> If you added a way to silence the warning when constructing a nested schema you would still get the error when Nested is passed a schema instance.

[...]

PR #1135, opened 2019-02-13
Don't raise ChangedInMarshmallow3Warning for Nested schemas

### LINKED ISSUE #1134: ChangedInMarshmallow3Warning raised for nested schemas
From https://github.com/marshmallow-code/marshmallow/issues/1110#issuecomment-462901519

> This warning appears to be happening on fields.Nested, even if the parent is using strict=True.

sloria:
Fix #1134 #1108

This changes the default of the `strict` Meta option from `False` to `None` because there is no other way to tell if a Nested Schema's parent has `strict` set.

This did require a test change. Unsure if this classifies as a breaking change. Thoughts, @lafrech @deckar01 ?

### DIFF
--- a/marshmallow/schema.py
+++ b/marshmallow/schema.py
@@ -195,7 +195,7 @@ class SchemaOpts(object):
         self.exclude = getattr(meta, 'exclude', ())
         if not isinstance(self.exclude, (list, tuple)):
             raise ValueError("`exclude` must be a list or tuple.")
-        self.strict = getattr(meta, 'strict', False)
+        self.strict = getattr(meta, 'strict', None)
         if hasattr(meta, 'dateformat'):
             warnings.warn(
                 "The dateformat option is renamed to datetimeformat in marshmallow 3.",
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -2458,7 +2458,7 @@ class TestStrictDefault:
         pass
 
     def test_default(self):
-        assert self.SchemaWithoutMeta().strict is False
+        assert self.SchemaWithoutMeta().strict is None
 
     def test_meta_true(self):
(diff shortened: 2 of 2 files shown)


marshmallow/schema.py
```diff
@@ -195,7 +195,7 @@ def __init__(self, meta):
         self.exclude = getattr(meta, 'exclude', ())
         if not isinstance(self.exclude, (list, tuple)):
             raise ValueError("`exclude` must be a list or tuple.")
-        self.strict = getattr(meta, 'strict', False)
+        self.strict = getattr(meta, 'strict', None)
```
deckar01, line 198: This prevents warnings on top level schemas also. Unless `strict` is explicitly set to `False`. I suspect the majority of users are relying on the default value, so this would have the same effect as removing the warning for most users.

lafrech: As a side note, I'm not sure the warning is _that_ useful. There are tons of changes between 2 and 3 anyway, we can't warn for everything. If the last resort solution is to remove the warning, that's fine IMHO. If you find an elegant solution to sort this out and keep the warning, that's fine with me as well.

sloria: Yeah, I thought about just removing this particular warning, since it can add a fair amount of noise (I ended up [silencing it in flask-resty's tests](https://github.com/4Catalyzer/flask-resty/pull/232)). The change in `strict`  behavior is arguably the most significant change from 2 to 3 and is hard to miss when upgrading. Setting `strict=True` will not suffice, since you need to account for the change in return value type as well.

Anyway, I think this PR isn't the way to go for the reason @deckar01 pointed out, so closing.

So question is:  does this warning provide a significantly positive nudge to upgrade to marsmallow 3, or do the false positives produce too much noise to be useful? 🤔

PR #1138, opened 2019-02-15
Switch to src/marshmallow layout

### LINKED ISSUE #1017: RFC: Use src/ directory
Put the package in `src/marshmallow`, as recommended in these articles:

* https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure
* https://hynek.me/articles/testing-packaging/

The benefits makes sense, but I haven't yet run into the pitfalls talked about in those articles...

Strong opinions welcome!

miniscruff:
Remaking PR for 2.x-line, same changes as before.
There was a flake8 exclude file that needed its path changed as well but that is the only difference.

Closes #1017


sloria:
Thanks!

PR #1139, opened 2019-02-15
Dump Error Keys

### LINKED ISSUE #1132: Remove validation on serialization
When using `Dict(values=Nested())` I expect errors in dictionary values to be nested in such a way that I can locate exactly what element is causing the problem. This seems to work as expected for `load()`, but not for `dump()`.

Minimal example:

- Marshmallow version: 3.0.0rc4
- Python version: 3.7.2

```python
from marshmallow import Schema, ValidationError, fields


class InnerSchema(Schema):
    x = fields.Int()


class OuterSchema(Schema):
    inner = fields.Dict(keys=fields.Str(), values=fields.Nested(InnerSchema))


foo = {"inner": {"some-key": {"x": "abc"}}}

try:
    OuterSchema().load(foo)
except ValidationError as e:
    print(e)
    # >>> {'inner': defaultdict(<class 'dict'>, {'some-key': {'value': {'x': ['Not a valid integer.']}}})}
    # Very helpful - and what I expect

class Inner:
    def __init__(self, x):
        self.x = x


class Outer:
    def __init__(self, inner):
        self.inner = inner


bar = Outer(inner={"some-key": Inner(x="abc")})

try:
    OuterSchema().dump(bar)
except ValidationError as e:
    print(e)
    # >>> {'inner': {'x': ['Not a valid integer.']}}
    # Surprisingly - 'some-key' and 'value' are not present here,
    # which makes it harder to pinpoint the problem in large objects
```

deckar01:
Store errors by key when dumping collections to prevent item errors from being reported as their parent.

Fixes #1132

---

Example:

```py
from marshmallow import Schema, ValidationError, fields


class Test(Schema):
    foo = fields.List(fields.Int)

Test().dump({'foo': ['five']})
# Before: {'foo': ['Not a valid integer.']}
# After: {'foo': {0: ['Not a valid integer.']}}
```

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -600,9 +600,20 @@ class List(Field):
     def _serialize(self, value, attr, obj, **kwargs):
         if value is None:
             return None
-        if utils.is_collection(value):
-            return [self.inner._serialize(each, attr, obj, **kwargs) for each in value]
-        return [self.inner._serialize(value, attr, obj, **kwargs)]
+        if not utils.is_collection(value):
+            return [self.inner._serialize(value, attr, obj, **kwargs)]
+        result = []
+        errors = {}
+        for idx, each in enumerate(value):
+            try:
+                result.append(self.inner._serialize(each, attr, obj, **kwargs))
+            except ValidationError as error:
+                if error.valid_data is not None:
+                    result.append(error.valid_data)
+                errors.update({idx: error.messages})
+        if errors:
+            raise ValidationError(errors, valid_data=result)
+        return result
 
     def _deserialize(self, value, attr, data, **kwargs):
         if not utils.is_collection(value):
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -597,40 +597,32 @@ class List(Field):
             self.inner.only = self.only
             self.inner.exclude = self.exclude
 
+    @staticmethod
+    def _marshal(operation, value, attr, root, **kwargs):
+        result = []
+        errors = {}
+        for idx, each in enumerate(value):
+            try:
+                result.append(operation(each, attr, root, **kwargs))
+            except ValidationError as error:
+                if error.valid_data is not None:
+                    result.append(error.valid_data)
+                errors.update({idx: error.messages})
+        if errors:
+            raise ValidationError(errors, valid_data=result)
+        return result
+
     def _serialize(self, value, attr, obj, **kwargs):
         if value is None:
             return None
         if not utils.is_collection(value):
             return [self.inner._serialize(value, attr, obj, **kwargs)]
-        result = []
-        errors = {}
-        for idx, each in enumerate(value):
-            try:
-                result.append(self.inner._serialize(each, attr, obj, **kwargs))
-            except ValidationError as error:
-                if error.valid_data is not None:
-                    result.append(error.valid_data)
-                errors.update({idx: error.messages})
-        if errors:
-            raise ValidationError(errors, valid_data=result)
-        return result
+        return self._marshal(self.inner._serialize, value, attr, obj, **kwargs)
 
     def _deserialize(self, value, attr, data, **kwargs):
         if not utils.is_collection(value):
-            self.fail("invalid")
-
-        result = []
-        errors = {}
-        for idx, each in enumerate(value):
-            try:
-                result.append(self.inner.deserialize(each, **kwargs))
-            except ValidationError as error:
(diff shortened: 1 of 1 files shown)


marshmallow/fields.py
```diff
@@ -1348,8 +1363,19 @@ def _serialize(self, value, attr, obj, **kwargs):
...
+            for key, val in iteritems(value):
+                try:
+                    deser_val = self.value_container._serialize(val, None, None, **kwargs)
+                except ValidationError as error:
+                    errors[key]['value'] = error.messages
+                    if error.valid_data is not None and key in keys:
+                        result[keys[key]] = error.valid_data
+                else:
+                    if key in keys:
+                        result[keys[key]] = deser_val
```
deckar01: I'm not really happy with copy and pasting this much code. It is effectively ~~6~~ ~~8~~ 4 copies of the same thing, with slight variations. I suspect there is a utility interface that can abstract this functionality while also improving the code's readability.

deckar01: I took a first pass at deduplicating the logic. I didn't see a clean way to abstract the operations across field classes, but I avoided introducing any duplicate logic in this PR and deduplicated a few operations that were already present.

I need to test what the performance implications are for this abstraction. It should be minimal, because the primary difference is `getattr` being used in place of static attribute access once in each method. An alternative would be to use a lambda instead, but I don't have an intuition for which is preferable or faster.

A few methods were passing `None` into  the `attr` and `obj`/`data` arguments instead of passing them through like the rest. Normalizing this did not affect the tests, but I suspect the original behavior was an obscure bug. I need to look at the implications of this change a little closer.

marshmallow/fields.py
```diff
@@ -585,22 +584,13 @@ def _bind_to_schema(self, field_name, schema):
...
-        if utils.is_collection(value):
-            return [self.container._serialize(each, attr, obj, **kwargs) for each in value]
-        return [self.container._serialize(value, attr, obj, **kwargs)]
-
-    def _deserialize(self, value, attr, data, **kwargs):
-        if not utils.is_collection(value):
-            self.fail('invalid')
-
+    @staticmethod
+    def _marshall(operation, value, attr, root, **kwargs):
```
sloria: Spelling should be `_marshal`.

sloria (approved):
Implementation looks correct; just a minor spelling change suggested.

I'm also not sure about how to deduplicate the logic across field classes without introducing a confusing amount of indirection. This could happen after @lafrech's suggested [ContainerField](https://github.com/marshmallow-code/marshmallow/pull/1066#issuecomment-450500701) becomes a reality, but no need to block merging of this PR nor #1066 for it.

Granted there are no performance regressions with this, I think this is on the right track.

lafrech:
I pushed a `ContainerMixin` proposal in https://github.com/marshmallow-code/marshmallow/pull/1066.

marshmallow/fields.py
deckar01: > I suspect the original behavior was an obscure bug.

Confirmed. See #1176.

lafrech:
I finally dropped the `ContainerMixin` I proposed in #1066 and submitted a simpler PR (#1229) so don't count on it.

deckar01:
I will get this up to date now that #1229 has landed.

sloria:
@deckar01 Are you still working on this? Let us know if you need help with it.

deckar01:
**Benchmark**:

Setup: [benchmark.py](https://github.com/marshmallow-code/marshmallow/files/3393145/benchmark.py.txt)

```py
# dev
load 18.1108 µs ± 0.65%
dump 10.8205 µs ± 1.14%
load no err 18.0016 µs ± 0.69%
dump no err 10.7058 µs ± 2.38%
```

```py
# 1132-nested-dump-errors
load 18.7329 µs ± 0.85%
dump 10.8929 µs ± 0.42%
load no err 18.6097 µs ± 0.85%
dump no err 10.9445 µs ± 0.97%
```

```py
# change
load +3.43%
dump +0.66%
load +3.37%
dump +2.23%
```

sloria:
It's a minor performance regression, but I'm wondering if we should incur it at all...we've since decided that validation should only happen on deserialization and that the `dump` should be considered valid. Do we want to continue supporting validation-on-serialization?

deckar01:
The performance regression is a side effect of refactoring to avoid copy and pasting validation code. It could be omitted. I will continue the discussion about dump validation on the issue.

deckar01:
Closing in favor of #1304.

Issue #1146, opened 2019-02-21
How to pass parameters to a nested schema which is referred by name


Kamforka:
I would like to refer to a schema by its name when I use it as a nested one, but still I'd like to pass custom arguments to its constructor.

For example instead of doing this:
```python
class AuthorSchema(Schema):
    books = fields.Nested(BookSchema(enveloped=True, many=True))
    class Meta:
        fields = ('id', 'name', 'books')
```

I'd like to use something like this:

```python
class AuthorSchema(Schema):
    books = fields.Nested('BookSchema', enveloped=True, many=True)
    class Meta:
        fields = ('id', 'name', 'books')
```

Please note that in this case the `BookSchema`'s `__init__` method is looking for a custom parameter called `enveloped`.

Is there any official way that I can achieve this using only the native `Nested` field?

I'm using the following versions of marshmallow:
`marshmallow==2.16.3`
`flask-marshmallow==0.9.0`

deckar01:
There is currently not a way to pass custom arguments when the nested schema is referenced as a string. What you can do instead is subclass the schema with the custom `Meta` properties and use that schema name instead.

```py
class NestedBookSchema(BookSchema):
    class Meta:
        enveloped = True

...
    books = fields.Nested('NestedBookSchema', many=True)
```

I'm not exactly sure where `enveloped` comes from, so if this is not part of the `Meta` then you might have to get a little creative with passing the arg through the constructor.

It seems like this could be made a lot more convenient by adding an argument that passes schema constructor args through the nested field constructor.

Kamforka:
@deckar01 
Thanks for your reply!

Yeah I know I can do that, but I was looking for an inbuilt solution, as I thought I might overlook something. 

But still I feel like this feature would be a nice to have when you want to tweak nested schemas which are referenced by name.

lafrech:
> It seems like this could be made a lot more convenient by adding an argument that passes schema constructor args through the nested field constructor.

Absolutely.

Kamforka:
I've made some changes needed for this feature, however when I try to push it I receive back an unauthorized error message. Can you help me with that please?

lafrech:
I you want to propose a change to the code, you may fork the repository in Github, checkout a branch with your changes, push that branch in your fork, then create a pull-request to ask us to merge your changes in the dev branch here.

sloria:
[...]

PR #1151, opened 2019-02-23
Nested field to pass parameters for name referred schemas

Kamforka:
PR based on the enhancement identified in #1146

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -406,7 +406,10 @@ class Nested(Field):
         'type': 'Invalid type.',
     }
 
-    def __init__(self, nested, default=missing_, exclude=tuple(), only=None, **kwargs):
+    def __init__(
+        self, nested, default=missing_, exclude=tuple(),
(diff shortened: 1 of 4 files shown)


src/marshmallow/fields.py
```diff
@@ -416,6 +417,7 @@ def __init__(self, nested, default=missing_, exclude=tuple(), only=None, **kwarg
         self.exclude = exclude
         self.many = kwargs.get('many', False)
         self.unknown = kwargs.get('unknown')
+        self.schema_args = kwargs.get('schema_args', {})
```
lafrech: Don't keep `schema_args` in kwargs, they would be passed to `Field` init. Either pop from kwargs, or maybe better

```py
def __init__(self, nested, default=missing_, exclude=tuple(), only=None, schema_args=None, **kwargs):
    self.schema_args = schema_args or {}
```

Kamforka: Sure, fixed.

PR #1155, opened 2019-02-25
Better default error messages for OneOf and ContainsOnly

maxalbert:
This PR is related to #885. (Not sure if it fully closes it?).

It changes the default messages for `OneOf` and `ContainsOnly` to the ones suggested by @sloria:
- `OneOf`: `'Not one of {choices}.'`
- `ContainsOnly`: `'One or more of the choices you made was not in {choices}'`

Please let me know if I should add a CHANGELOG entry for this change.

P.S.: Many thanks for everyone's work on this project!

### DIFF
--- a/src/marshmallow/validate.py
+++ b/src/marshmallow/validate.py
@@ -410,16 +410,13 @@ class OneOf(Validator):
         interpolated with `{input}`, `{choices}` and `{labels}`.
     """
 
-    default_message_prefix = 'Not one of'
+    default_message = 'Must be one of: {choices}.'
 
     def __init__(self, choices, labels=None, error=None):
         self.choices = choices
         self.choices_text = ', '.join(text_type(choice) for choice in self.choices)
         self.labels = labels if labels is not None else []
         self.labels_text = ', '.join(text_type(label) for label in self.labels)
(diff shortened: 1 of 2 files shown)


src/marshmallow/validate.py
```diff
@@ -410,13 +410,16 @@ class OneOf(Validator):
         interpolated with `{input}`, `{choices}` and `{labels}`.
     """
 
-    default_message = 'Not a valid choice.'
+    default_message_prefix = 'Not one of'
```
sloria: To keep this consistent with other validators, better to make `default_message` a format template, like `Not one of [{choices}]`, then implement `_format_error`. See `Equal` for an example.

sloria: Also, for language consistency the other messages, how about

```
Must be one of: {choices}.
```

sloria (changes requested):

src/marshmallow/validate.py
maxalbert: Thanks, I have switched to your suggestion. Indeed, my approach felt very clumsy, so this works much better. I hadn't realised that `_format_error()` already provides all the means to correctly format this message.

sloria (approved):

PR #1156, opened 2019-02-25
Remove the deprecated fields param from ValidationError docs

rileyjohngibbs:
The `fields` keyword param is no longer used on the `ValidationError`. Remove that from the docstring for `ValidationError`.

### DIFF
--- a/src/marshmallow/exceptions.py
+++ b/src/marshmallow/exceptions.py
@@ -21,7 +21,6 @@ class ValidationError(MarshmallowError):
         error messages. If a dict, the keys are subitems and the values are error messages.
     :param str field_name: Field name to store the error on.
         If `None`, the error is stored as schema-level error.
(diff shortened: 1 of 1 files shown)


lafrech:
Good catch. This was forgotten in https://github.com/marshmallow-code/marshmallow/pull/845.

rileyjohngibbs:
Should the [comment](https://github.com/marshmallow-code/marshmallow/pull/845/files#diff-df7f7e6732f2439141eb0dae350a64eaR110) about accessing the errors from `fields.schema` also be removed?

lafrech:
Hmmm, I don't think so. Should it?

rileyjohngibbs:
Scratch my previous. I was misreading the intent of the comment. It's clear enough; I just didn't get enough sleep last night. 😛

PR #1168, opened 2019-03-08
Fix a warning under Python 3.7

felixonmars:
```
Warning: 'keywords' should be a list, got type 'tuple'
```
Related issue: https://bugs.python.org/issue19610


lafrech:
Thanks.

PR #1171, opened 2019-03-12
Use Xenial for all builds

kevin-brown:
This was special cased for Python 3.7, but it supports all of the versions of Python we test on so it makes sense to just use it for all builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+dist: xenial
 cache: pip
 install: travis_retry pip install -U tox
 script: tox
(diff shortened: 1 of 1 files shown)


sloria:
Thanks!

lafrech:
Great. I checked two days ago to do exactly the same thing but I was mislead by [the docs](https://docs.travis-ci.com/user/reference/xenial/#python-support): 

> Pre-installed Python versions: 2.7.15, 3.6.7, and 3.7.1.

I thought 3.5 wouldn't be covered.

It is not installed by default but it is downloaded when needed:

```
3.5 is not installed; attempting download
Downloading archive: https://travis-ci-python-archives.global.ssl.fastly.net/binaries/ubuntu/16.04/x86_64/python-3.5.tar.bz2
```

lafrech:
@sloria, I can duplicate this in webargs/apispec. Or do you plan to use all-repos to do that in all your repos?

sloria:
Yeah, was planning to use all-repos. Possibly this weekend.

lafrech:
Good. No rush.

Issue #980, opened 2018-10-05
RFC: Propagate `unknown` value through nested schemas? 


lafrech:
I'm wondering if the `unknown` value set on parent schema should propagate down through all nested schemas.

We raise by default, and to change that, one needs to let all his schemas specify another behaviour (~this is broken but will be fixed when #963 is merged~). This can be done with a custom schema to use as a base for all schemas, so it is feasible, and raise is a sensible default IMHO, but wouldn't it make sense to let the parent schema `unknown` setting propagate?

vgavro:
I think it's good idea to add fields.Nested.unknown="inherit", which can be set explicitly, but in most cases what I see you don't want to propagate by default, for example you have http json response, which can have various fields, but if field is nested it will be binded to specific entity, and you want to be sure that fields are strict there, so it's each entity (schema) responsibility to react to unknown fields, not related to parent.
Also (off-topic) I think it will be great to have unknown='warn' or unknown='debug' on schema, i will be glad to create pull request if you think this may be useful.

RosanneZe:
It would be nice to have the option to have all nested schemas inherit.

sloria:
For now, our stance is that class Meta options should not be automagically propagated to nested fields. Users should use an base schema class in order to share default options across classes.

zaraken:
> For now, our stance is that class Meta options should not be automatically propagated to nested fields. Users should use an base schema class in order to share default options across classes.

I do agree that class Meta options should not be automatically propagated.

However, what about when overriding `unknown` in the load method which should according to documentation override the value set at instantiation or in Meta. For this use case it would make sense to (conditionally) allow propagation to nested fields or ?

charlax:
At the very least, this behavior should be documented.

mirko:
I really agree with the previous pro-comments.
I'm using it on load() and it fails for Nested fields, although the doc say:

> The unknown option value set in load will override the value applied at instantiation time, which itself will override the value defined in the class Meta.

Which is very much what I'd have expected for nested fields, what the documentation confirms, which however doesn't happen.

Either way: How to solve that? I'm having List() and Item().
I'm sending an instance of a list having items to the server and the server loads and validates sent data via List_Schema().load(unknown=EXCLUDE).
Now an exception is thrown: `TypeError: 'X' is an invalid keyword argument for Item`

jemshit:
> I really agree with the previous pro-comments. I'm using it on load() and it fails for Nested fields, although the doc say:
[...]

PR #1172, opened 2019-03-15
Nested schemas should only inherit 'ordered' if available

li-wjohnson:
Currently, using `fields.Dict(values=Nested(Foo))`, Foo will never be ordered, even if it has `ordered = True` in its Meta. 
I'm not sure what the right precedence is here (could instead be `getattr(self.__schema, 'ordered', False) or getattr(self.parent, 'ordered', False)`, or `getattr(self.__schema, 'ordered', getattr(self.parent, 'ordered', False))`, or vice-versa?), but as is, it's impossible to order a Dict(Nested(Foo)).

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -447,7 +447,8 @@ class Nested(Field):
                     load_only=self._nested_normalized_option('load_only'),
                     dump_only=self._nested_normalized_option('dump_only'),
                 )
-            self.__schema.ordered = getattr(self.parent, 'ordered', False)
+            if hasattr(self.parent, 'ordered'):
+                self.__schema.ordered = self.parent.ordered
         return self.__schema
 
     def _nested_normalized_option(self, option_name):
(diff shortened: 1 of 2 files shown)


sloria:
Thanks! Can you please add a test for this?

li-wjohnson:
@sloria: Now that I look at it more, I think it makes more sense if nested schemas should always inherit order from the root schema.

As it is now:

```
class UnorderedSchema(Schema):
  a_user = fields.Nested(OrderedSchema)
  users = fields.Dict(keys=fields.Str(), values=fields.Nested(OrderedSchema))
```

`a_user` will be unordered, but `users`'s values is ordered.  Instead, if nested schemas always inherit order from the root schema, then both `a_user` and `users` will be unordered.

Somehow, this still doesn't seem ideal. Why shouldn't it be possible to have an ordered schema inside an unordered schema? IMO it makes sense if the nested schema becomes ordered if the parent schema is ordered, but *not* become unordered if the parent is unordered.

sloria:
Yeah, I think it's too restrictive to make all nested schemas inherit from the root without a way to override the behavior. Perhaps we should just remove the inheritance behavior. Users can use a base schema class if they want to inherit the `ordered` option.

There's a related discussion to propagate `unknown`: https://github.com/marshmallow-code/marshmallow/issues/980

sloria:
Discussed with @lafrech : our stance is that class Meta options should not be propagated to nested fields. Users should use a base schema class in order to share options across classes.

PR #1173, opened 2019-03-15
Update docs for Function and Method fields

gastlich:
Not sure, what is desired behaviour of this functionality, but updated docs to present at least working examples.


sloria:
The docs assume that you're serializing objects--not dictionaries--which is the common use case. Why change this to dictionary access?

gastlich:
Hi @sloria, thanks for your reply. I moved recently to marshmallow 3 and didn't know about difference between object and dict deserialisation. I assumed, that objects built are always dictionaries. Will have to catch up on this feature then, thanks. Closing it now.

PR #1177, opened 2019-03-26
bug: treat integers that overflow float as invalid

brycedrennan:
Found this case where Marshmallow throws an exception instead of validating the input.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -802,7 +802,7 @@ class Number(Field):
             return None
         try:
             return self._format_num(value)
-        except (TypeError, ValueError):
+        except (TypeError, ValueError, OverflowError):
             self.fail('invalid', input=value)
 
     def _to_string(self, value):
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -93,6 +93,13 @@ class TestField:
             strict_field.serialize('value', {'value': True})
         assert excinfo.value.args[0] == 'Not a valid number.'
 
+    def test_number_field_overflow(self):
+        strict_field = fields.Float()
+        with pytest.raises(ValidationError) as excinfo:
+            strict_field.serialize('value', {'value': 2**1024})
+        assert excinfo.value.args[0] == 'Not a valid number.'
+
+
 class TestParentAndName:
     class MySchema(Schema):
         foo = fields.Field()


tests/test_fields.py
```diff
@@ -93,6 +93,13 @@ def test_number_fields_prohbits_boolean(self):
             strict_field.serialize('value', {'value': True})
         assert excinfo.value.args[0] == 'Not a valid number.'
 
+    def test_number_field_overflow(self):
+        strict_field = fields.Float()
+        with pytest.raises(ValidationError) as excinfo:
+            strict_field.serialize('value', {'value': 999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999990})
```
deckar01: You could probably simplify this constant to `2**1024`. It would be nice if it was based on `sys.float_info.max`, but simply adding to that value doesn't seem to cause an overflow.

src/marshmallow/fields.py
```diff
@@ -794,7 +794,10 @@ def _format_num(self, value):
         # (value is True or value is False) is ~5x faster than isinstance(value, bool)
         if value is True or value is False:
             raise TypeError('value must be a Number, not a boolean.')
-        return self.num_type(value)
+        try:
+            return self.num_type(value)
+        except OverflowError:
+            raise TypeError('value too big for float value. Try Integer field.')
```
deckar01: I don't think this will actually pass the custom error out. `_validated` swallows the specific errors and just replaces them with "fail". I simpler fix would be to add `OverflowError` to the `except` tuple in `_validated`.

brycedrennan:
Thanks @deckar01. I've made the changes you suggested.

sloria:
Is this bug also on 2.x? If so, can you please resend this patch to the `2.x-line` branch?

Thanks for your work on this!

src/marshmallow/fields.py
```diff
@@ -802,7 +802,7 @@ def _validated(self, value):
             return None
         try:
             return self._format_num(value)
-        except (TypeError, ValueError):
+        except (TypeError, ValueError, OverflowError):
```
sloria, line 805: It might be good to handle this separately, with a new failure type, e.g. `'overflow'` or `'too_large'`.

sloria (changes requested):

sloria:
I've backported and carried this on in #1179.

Issue #1120, opened 2019-02-03
Drop Python 2 support


sloria:
~~marshmallow 3 will be the last major version to support Python 2.~~ Per discussion below, Python 2 support will be dropped in marshmallow 3.

rooterkyberian:
Why not drop Python 3 in marshmallow 3?

* Python 2 EoL is in 8 months, marshmallow 2 was being supported for 4 years and counting
* marshmallow 3 already contains multiple breaking changes
* supporting old Python version costs development time
  * duh
  * there are still open bugs for marshmallow 2 concerning Python 2-3 compatibility that there is low interest in fixing (#797, #493)
* ~python suffers from relatively [high function call overhead](https://gist.github.com/rooterkyberian/5fe6acc7c5126dc2befb81cf323bf6b1), and that is how [compat module works](https://github.com/marshmallow-code/marshmallow/blob/dc40e10419549d99ad0afdde51faa570b28c312c/src/marshmallow/compat.py) - not arguing it would be a game changer, but should provide measurable benefits~
  * ~performance is arguably one of biggest drawbacks of marshmallow and this seems like an easy win~
* switch to Python 3-only should be easy enough - AFAIK it just a matter of removal of `compat` module
* opens doors for more non-breaking improvements as part of 3.x line (which are NOT required to be a part of 3.0.0 release); for example - type annotations

_____________________
Initial tests after Python 2 compat removal #1204 show performance improvement that is well within margin of error using marshmallow benchmark. I.e. no actual improvement, so my statement about "easy win" has proven to be false.

deckar01:
Although the most recent python developer survey indicates that developers use python 3 **more than** python 2 80% of the time, python 2 is still used a lot. Assuming the developers on average use their less-used python version in 25% of their active projects, the stats would come out to python 3 65%, python 2 35%. I suspect the percentage of actual development activity would be even more skewed towards python 2 considering how many python 2 projects are still being actively maintained.

https://www.jetbrains.com/research/python-developers-survey-2018/#python-3-adoption

rooterkyberian:
If someone wants to use outdated software they can keep on doing that (Python interpreter, older marshmallow) - I'm just arguing that we shouldn't hold the project back for minority that is unlikely to use the newer version of a package that brings breaking changes on its own.

Which reminds me:
[...]

PR #1195, opened 2019-04-11
Upgrade Python syntax with pyupgrade

hugovk:
pyupgrade is a tool "to automatically upgrade syntax for newer versions of the language".

```
pyupgrade `find . -name "*.py" -type f`
```

https://github.com/asottile/pyupgrade

Once Python 2 is dropped (#1120), it can be run with the `--py3-plus` flag.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,3 +13,7 @@ repos:
   rev: 3.7.7
   hooks:
   - id: flake8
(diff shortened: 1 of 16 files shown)


sloria:
We are already using pre-commit, so we should run `pyupgrade` through that. Can you please update the config then run `tox -e lint` to run pre-commit?

hugovk:
Updated!

PR #1204, opened 2019-04-29
removed python 2 support

### LINKED ISSUE #1120: Drop Python 2 support

rooterkyberian:
resolves #1120


setup.py
```diff
@@ -65,13 +65,11 @@ def read(fname):
         'serialization', 'rest', 'json', 'api', 'marshal',
         'marshalling', 'deserialization', 'validation', 'schema',
     ],
-    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
+    python_requires='>=3, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
```
sloria: Can't this be simplified to `>=3.5`?

tests/test_validate.py
```diff
@@ -511,7 +510,7 @@ def test_predicate_repr():
     assert (
         repr(validate.Predicate(method='foo', error='bar', zoo=1)) ==
         '<Predicate(method={!r}, kwargs={!r}, error={!r})>'
-        .format('foo', {str('zoo') if PY2 else 'zoo': 1}, 'bar')
+        .format('foo', {'zoo': 1}, 'bar')
```
sloria, line 513: We'll probably want to switch to f-strings, but this is fine.

sloria: Oh scratch that. I forget that f-strings aren't in 3.5, which we're supporting.

sloria:
Can you add the `--py3-plus` flag to the `pyupgrade` hook in `.pre-commit-config.yaml` and run it?

tests/test_validate.py
rooterkyberian: 3.5 EoL in 2020-09-13 so it probably will be still a while

lafrech:
AUTHORS.rst needs to be updated.

sloria:
Thank you for the PR!

PR #1209, opened 2019-05-06
removed unused `error` argument

rooterkyberian:
seems like some kind of leftover


lafrech:
No idea what that is.

It is there in 2.x as well and it seems just as useless, so it must be a pretty old leftover, unrelated to the error structure rework we did in MA3. (I'm not saying we should change this in 2.x. I just looked there to try and understand.)

sloria:
Ah good catch! I think this is a holdover from marshmallow 1.

PR #1213, opened 2019-05-11
Pass 'value' instead of 'obj' to serialize_func 

nbreum15:
The parameter 'value' should be passed to the function `serialize_func` in `fields.Function` instead of 'obj'.
Otherwise, the value passed to the serialize function will just be `None`.


sloria:
The current behavior is correct. Both `Function` and and `Method` receive the object so that you can perform whatever property access you need to.

PR #1218, opened 2019-05-18
Use keyword-only arguments

### LINKED ISSUE #1216: Use keyword-only arguments
Use keyword-only arguments where it makes sense.

sloria:
Close #1216

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 Changelog
 ---------
 
+3.0.0 (unreleased)
+++++++++++++++++++
+
+- *Backwards-incompatible*: Use keyword-only arguments (:issue:`1216`).
+
 3.0.0rc6 (2019-05-05)
 +++++++++++++++++++++
 
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -129,6 +129,7 @@ class Field(FieldABC):
 
     def __init__(
         self,
+        *,
         default=missing_,
         attribute=None,
         data_key=None,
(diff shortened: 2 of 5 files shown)


lafrech:
Good. :+1:

I might have spotted a few omissions.

#### fields.py

`as_string`, `strict`, `allow_nan` in `Number`, `Integer`, `Float`.

And `format` in `DateTimeField`.

Perhaps `precision` in `TimeDelta`, `serialize`/`deserialize` in `Method`/`Function`. No strong opinion about them.

#### schema.py

Looks like you missed a few `_invoke_*`.

sloria:
Good call on those. Updated those and also functions in utils.py.

I deliberately omitted `format` in `DateTime`; I think it's fine to allow the format to be passed positionally, and some tests were doing that.

lafrech:
IIUC, this PR addresses https://github.com/marshmallow-code/marshmallow/pull/1034#issuecomment-436556833.

> - `many`, `partial` and `original` (or `original_data`) are always passed as kwargs

(Edit: at this point, only `many` and `original_data` exist. `partial` will be added in https://github.com/marshmallow-code/marshmallow/pull/1220.)

Should we document the fact that

> - validators should accept unknown kwargs (even kwargs we may add in the future)

?

sloria:
@lafrech Did you mean to make the above comment on https://github.com/marshmallow-code/marshmallow/pull/1220 ?

lafrech:
Both PRs are related. I meant to comment here, but I was about to edit and mention my comment was pointless because the change and documentation will be done in #1220 indeed. Sorry about the noise.

sloria:
Ah, I see, no problem. Indeed, this PR doesn't pass any additional arguments.

lafrech (approved):
LGTM.

The most important part IMHO is `Schema` `__init__` and `load[|s]` / `dump[|s]`.

Issue #1150, opened 2019-02-22
Decorators: wrong behaviour if pass_many=False, many=True, pass_original=True


lafrech:
Related to https://github.com/marshmallow-code/marshmallow/issues/315.

Looks like the fixes in #744 and #750 are somehow wrong.

It works in the general case where `original_data` is a collection. But it might not be the case if it is not. For instance, if it contains an envelope that is removed in a `pre_load` decorator.

For instance, run the `test_decorated_processors` test with a slight modification: add `pass_original=True` to the last `post_load`.

```py
        @post_load(pass_original=True)
        def decrement_value(self, item, **kwargs):
            item['value'] -= 1
            return item
```

It fails because this piece of code expects `original_data` to be a `list` (`[{'value': 4}, {'value': 4}]`), but it is a `dict` (`{'data': [{'value': '4'}, {'value': '6'}]`): 

```py
                if pass_original:
                    data = [
                        processor(item, original_data=original)
                        for item, original in zip(data, original_data)
                    ]
                else:
                    data = [processor(item) for item in data]
```

`zip` only grabs a single element (`'data'`), which is wrong.

I don't see any solution to this. I'm afraid the idea in #315 works in the typical case but can't work in all cases.

szeitlin:
I think I just ran into this on 3.0.0b8, though I'm not sure if I'm doing something else wrong. I have a `@validates_schema` that works as expected on a single object, but when I try to use it on a collection with `many=True`, I get an error about too many arguments.

phrfpeixoto:
I encountered the same issue o 3.0.0 stable, but I did not touch the `pass_original` attribute.
Here's a sample project showing my issue:
https://github.com/phrfpeixoto/marshmallow_bug/tree/master

It works ok o 2.x Branch. I'm holding to that version for now.

phrfpeixoto:
Any updates on this?

lafrech:
@phrfpeixoto your issue is unrelated to this one.

Looks like it is another bug and the fix could be the following:

```py
    def _invoke_processors(
        self, tag, *, pass_many, data, many, original_data=None, **kwargs
    ):
        key = (tag, pass_many)
        for attr_name in self._hooks[key]:
            # This will be a bound method.
            processor = getattr(self, attr_name)

            processor_kwargs = processor.__marshmallow_hook__[key]
            pass_original = processor_kwargs.get("pass_original", False)

            if pass_many:
                if pass_original:
[...]

PR #1219, opened 2019-05-18
[WIP] Pass partial, many, and original_data to all processor methods

### LINKED ISSUE #994: Accessing partial within validates_schema

sloria:
Carries on the work done in https://github.com/marshmallow-code/marshmallow/pull/1034 .

I rebased and squashed the commits in that PR and followed up on https://github.com/marshmallow-code/marshmallow/pull/1034#issuecomment-454254580.

Closes #994 

TODO:

- [ ] Decide whether we want to drop `pass_original`
- [ ] Update docs

### DIFF
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -1,5 +1,6 @@
 """The :class:`Schema` class, including its metaclass and options (class Meta)."""
 from collections import defaultdict, OrderedDict
+from itertools import zip_longest
 import datetime as dt
 import uuid
 import decimal
@@ -802,6 +803,7 @@ class BaseSchema(base.SchemaABC):
                     data,
                     many,
                     original_data=data,
+                    partial=partial,
                 )
             except ValidationError as err:
                 errors = err.normalized_messages()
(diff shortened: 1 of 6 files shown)


src/marshmallow/schema.py
```diff
@@ -1132,33 +1133,22 @@ def _invoke_processors(
...
-                if pass_original:
-                    data = [
-                        processor(item, original)
-                        for item, original in zip(data, original_data)
-                    ]
-                else:
-                    data = [processor(item) for item in data]
+                data = [
+                    processor(item, original_data=original, **kwargs)
+                    for item, original in zip_longest(data, original_data)
```
sloria, line 1148: I'm having second thoughts about dropping the `pass_original` parameter because it forces us to zip the processed data with `original_data` even if the user doesn't need the original data, which incurs a performance penalty for both dump-ing and load-ing.

sloria: Even worse, this behavior can be incorrect when `original_data` doesn't have the same shape as  `data` (e.g. when `original_data` is enveloped).

sloria:
Let's keep `pass_original` for now. Dropping this in favor of https://github.com/marshmallow-code/marshmallow/pull/1220

src/marshmallow/schema.py
lafrech: Reminds me of https://github.com/marshmallow-code/marshmallow/issues/1150.

The case where `original_data` is not iterable is not totally covered.

PR #1220, opened 2019-05-18
Pass many and partial to processor methods

### LINKED ISSUE #994: Accessing partial within validates_schema

sloria:
Replaces #1219 and #1034 . This PR is the same as #1219 except that it does not remove `pass_original` for the reasons in https://github.com/marshmallow-code/marshmallow/pull/1219#discussion_r285353391

I rebased and squashed the commits in #1034  and followed up on https://github.com/marshmallow-code/marshmallow/pull/1034#issuecomment-454254580.

Closes #994 .

TODO:

- [x] Add tests
- [x] Update docs

### DIFF
--- a/src/marshmallow/decorators.py
+++ b/src/marshmallow/decorators.py
@@ -106,6 +106,9 @@ def pre_dump(fn=None, pass_many=False):
     By default, receives a single object at a time, regardless of whether ``many=True``
     is passed to the `Schema`. If ``pass_many=True``, the raw data (which may be a collection)
     and the value for ``many`` is passed.
+
+    .. versionchanged:: 3.0.0
+        ``many`` is always passed as a keyword arguments to the decorated method.
     """
     return set_hook(fn, (PRE_DUMP, pass_many))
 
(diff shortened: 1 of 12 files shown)


src/marshmallow/decorators.py
```diff
@@ -116,6 +120,10 @@ def post_dump(fn=None, pass_many=False, pass_original=False):
 
     If ``pass_original=True``, the original data (before serializing) will be passed as
     an additional argument to the method.
+
+    .. versionchanged:: 3.0.0
+        ``partial`` and ``many`` are always passed as keyword arguments to
+        the decorated method.
```
lafrech, line 129: Shouldn't this be added to `pre_dump` and `pre_load` as well?

Apart from this, LGTM.

lafrech: Also, AFAIU, `partial` only applies to `pre_load` / `post_load`.

sloria: Fixed; good catch.

Bala2105 (commented):
Reviewed PR #1220 and it’s a major step for the Marshmallow 3.0 milestone. By ensuring that many and partial are consistently passed as keyword arguments to all decorated processor methods, the library finally gives developers full context during pre and postprocessing. While it requires adding kwargs to existing methods a significant breaking change the trade off is much more powerful and flexible schema logic.

Issue #1214, opened 2019-05-12
Add strict min/max Range validator


lafrech:
`Range` only offers "greater than or equal" (min) and "lower than or equal" (max).

We could add "greater than" and "lower than".

How should we call the new parameters?

- `min_strict` and `max_strict`?
- `strict_min` and `strict_max`?
- ...?

sloria:
Not sure this is necessary or desirable. We've had issues in the past with the "strict" verbiage, and I think we should avoid it when we can.

lafrech:
Example use case:

In my API, a resource allows to POST a building with an area. The area must be positive (a negative area doesn't make sense). And I want it to be not zero because it can't be zero either (and because I want to avoid zero division errors when computing ratios).

I can't express "> 0" with marshmallow so as a workaround I set ">= 0.01" which does the job in my use case but might not be that great if the dynamic was greater and the value could be arbitrary small.

I can live with the 0.01 trick but it looks like a sorry workaround for an issue that seems simple to fix.

sloria:
That makes sense. Perhaps we could use `greater_than`, `greater_than_eq`, `less_than`, `less_than_eq`. for parameter names.

kdop:
I have a similar problem. I think if we simply add `min_inclusive`, `max_inclusive` we can achieve all the scenarios described above.

So the signature will become:

`class marshmallow.validate.Range(min=None, max=None, error=None, min_inclusive=True, max_inclusive=True)`

so the default behavior will remain unchanged.

lafrech:
@kdop, would you like to propose a PR for this?

@sloria, any preference (`greater_than_eq` vs. `min_inclusive`)? We're allowed to introduce breaking changes and I find `min_inclusive` a slight bit more awkward.

~OTOH, `max`/`min` are more in line with `min`/`max_length`~ Scratch that, `Lengh` validator also uses `min`/`max` parameters. We should change them too.

kdop:
@lafrech sure, assign to me.

sloria:
Sure, `min_inclusive` and `max_exclusive` seem fine to me.

lafrech:
@kdop I didn't manage to assign you with the assign function, but please go for it!

Thanks!

kdop:
PR https://github.com/marshmallow-code/marshmallow/pull/1221

PR #1221, opened 2019-05-19
Add strict min/max Range validator

kdop:
Implements #1214

### DIFF
--- a/src/marshmallow/validate.py
+++ b/src/marshmallow/validate.py
@@ -184,8 +184,8 @@ class Range(Validator):
         value will not be checked.
     :param str error: Error message to raise in case of a validation error.
         Can be interpolated with `{input}`, `{min}` and `{max}`.
-    :param bool min_inclusive: Whether the `min` bound should be included in the range.
-    :param bool max_inclusive: Whether the `max` bound should be included in the range.
+    :param bool min_inclusive: Whether the `min` bound is included in the range.
+    :param bool max_inclusive: Whether the `max` bound is included in the range.
     """
 
     message_min = 'Must be at least {min}.'
--- a/src/marshmallow/validate.py
+++ b/src/marshmallow/validate.py
@@ -197,7 +197,7 @@ class Range(Validator):
     message_lte = 'less or equal to'
     message_lt = 'less than'
 
-    def __init__(self, min=None, max=None, min_inclusive=True, max_inclusive=True, *, error=None):
+    def __init__(self, min=None, max=None, *, min_inclusive=True, max_inclusive=True, error=None):
         self.min = min
         self.max = max
         self.error = error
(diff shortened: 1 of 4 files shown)


src/marshmallow/validate.py
```diff
@@ -171,41 +171,66 @@ def __call__(self, value):
...
 
-    message_min = 'Must be at least {min}.'
-    message_max = 'Must be at most {max}.'
-    message_all = 'Must be between {min} and {max}.'
+    message_min = 'Must be {min_op} {{min}}.'
+    message_max = 'Must be {max_op} {{max}}.'
+    message_all = 'Must be {min_op} {{min}} and {max_op} {{max}}.'
 
-    def __init__(self, min=None, max=None, error=None):
+    _msg_gte = 'greater or equal to'
```
sloria: No need to privatize these, and "message" shouldn't be abbreviated, i.e. `message_gte`.

src/marshmallow/validate.py
```diff
@@ -171,41 +171,66 @@ def __call__(self, value):
...
+    _msg_gt = 'greater than'
+    _msg_lte = 'less or equal to'
+    _msg_lt = 'less than'
+
+    def __init__(self, min=None, max=None, error=None, min_inclusive=True, max_inclusive=True):
```
sloria: `min_inclusive` and `max_inclusive` should be keyword-only arguments, along with `error` (which should be the last parameter, for consistency).

The latest `dev` branch uses keyword-only arguments, so you might want to get this branch up to date (rebase preferred, but merge is fine).

sloria (changes requested):
Thanks for the PR!

Looks good on a quick glance. Just a few drive-by comments for now.

lafrech:
This enhancement could be applied to `Length` as well for consistency, although it is not very useful since mix/max are integers in the length case.

src/marshmallow/validate.py
kdop: I also like rebase, but is --force pushing to the PR branch ok?

sloria: Yes, go ahead and force-push to your branch.

src/marshmallow/validate.py
```diff
@@ -170,41 +170,66 @@ def __call__(self, value):
...
+    message_max = 'Must be {max_op} {{max}}.'
+    message_all = 'Must be {min_op} {{min}} and {max_op} {{max}}.'
 
-    def __init__(self, min=None, max=None, *, error=None):
+    message_gte = 'greater or equal to'
+    message_gt = 'greater than'
+    message_lte = 'less or equal to'
+    message_lt = 'less than'
+
+    def __init__(self, min=None, max=None, min_inclusive=True, max_inclusive=True, *, error=None):
```
sloria: `min_inclusive` and `max_exclusive` should be keyword-only.

```suggestion
    def __init__(self, min=None, max=None, *, min_inclusive=True, max_inclusive=True, error=None):
```

sloria (approved):

PR #1222, opened 2019-05-21
Fix wrong get_declared_fields() param types

ccrisan:
It seems that `SchemaMeta.get_declared_fields()` takes arguments `cls_fields` and `inherited_fields` as `list` type. Yet the docstring params indicate their expected type as `dict`.

This PR adjusts the docstring declared types and corrects a couple of typos.


sloria:
Thanks!

PR #1227, opened 2019-06-03
Add partial parameter to Nested field

lafrech:
Investigating https://github.com/marshmallow-code/marshmallow/pull/1066#issuecomment-464207995.

I'm trying to unify the way modifiers are passed to nested schemas.

There may not be a real use case behind this PR, but the point is to make things more consistent.

Includes a nice multi-parameter test the way I like it!

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -415,7 +415,10 @@ class Nested(Field):
 
     default_error_messages = {'type': 'Invalid type.'}
 
-    def __init__(self, nested, *, default=missing_, exclude=tuple(), only=None, **kwargs):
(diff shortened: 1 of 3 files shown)


lafrech:
After further investigation, I don't think unification is really achievable (https://github.com/marshmallow-code/marshmallow/pull/1066#issuecomment-498262750). But I don't think this change hurts either.

Since `partial` is accepted as schema parameter, why not accept it in `Nested` as well?

Feel free to merge or reject. No big deal.

sloria:
Closing for now to avoid adding API surface. If this use case appears again for users, we can re-consider this.

PR #1230, opened 2019-06-04
Fix propagation of "partial" to Nested containers

lafrech:
Attempt to fix passing `partial` to `Nested` containers.

This passes the test case provided in https://github.com/marshmallow-code/marshmallow/issues/779#issuecomment-450499543 and does not break any other test.

I didn't think much about the implications.

TODO:

- [x] Check this is the right way (review welcome).
- [x] Add tests.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -600,7 +600,7 @@ class List(Field):
         errors = {}
         for idx, each in enumerate(value):
             try:
-                result.append(self.container.deserialize(each))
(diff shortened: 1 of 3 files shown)


lafrech:
I added tests and fixed the fix.

There might be a slight performance penalty when partial is a collection as the "remove prefix" logic is executed for all fields, not only `Nested` fields.

There will be a merge conflict with #1229 in test_fields.py. I can rebase when the first of the two is merged.

sloria (approved):

PR #1232, opened 2019-06-06
fields.Url: take into account the `schemes` parameter for validation

vincent-olivert-riera:
If we have this:

```
class Foo(Schema):
    url = fields.Url(schemes={"bar"})
```

The `schemes` parameter is not taken into account for doing the validation. Currently we need to use `fields.String` with `validate=validate.URL` as a workaround.

```
class Foo(Schema):
    url = fields.String(validate=validate.URL(schemes={"bar"}))
```

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1398,6 +1398,7 @@ class Url(String):
     :param str attribute: The name of the attribute to get the value from. If
         `None`, assumes the attribute has the same name as the field.
     :param bool relative: Allow relative URLs.
+    :param set schemes: Valid schemes. By default, http, https, ftp, and ftps.
     :param kwargs: The same keyword arguments that :class:`String` receives.
     """
 
@@ -1407,6 +1408,7 @@ class Url(String):
         String.__init__(self, **kwargs)
 
         self.relative = relative
+        self.schemes = schemes
         self.require_tld = require_tld
         # Insert validation into self.validators so that multiple errors can be
         # stored.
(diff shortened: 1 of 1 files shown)


sloria:
This actually should not make a difference because `_validated` no longer gets called anywhere. These should be removed, as I've done in https://github.com/marshmallow-code/marshmallow/pull/1236.

I've verified that the current behavior is working:

```python

class Foo(Schema):
    url = fields.Url(schemes={"bar"})


assert Foo().load({'url': 'bar://foo.bar'}) == {'url': 'bar://foo.bar'}
```

Note: `Url` field no longer validates on serialization (https://github.com/marshmallow-code/marshmallow/pull/712).

vincent-olivert-riera:
@sloria that example you put above fails for me:

```
$ ./python 
Python 2.7.16 (default, Apr 30 2019, 15:54:43) 
[GCC 9.0.1 20190312 (Red Hat 9.0.1-0.10)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from marshmallow import Schema, fields
>>> class Foo(Schema):
...     url = fields.Url(schemes={"bar"})
... 
>>> assert Foo().load({'url': 'bar://foo.bar'}) == {'url': 'bar://foo.bar'}
AssertionError
>>> 
```
I am using marshmallow 2.15.0

deckar01:
The result just needs to access `.data` in marshmallow 2.

```py
assert Foo().load({'url': 'bar://foo.bar'}).data == {'url': 'bar://foo.bar'}
```

PR #1239, opened 2019-06-12
Fixes #1223 dateformat is ignored for DateTime fields when not format…

### LINKED ISSUE #1223: dateformat is ignored for DateTime fields when not format parameter
As documentation explains about `dateformat`

> dateformat: Date format for all DateTime fields that do not have their
> date format explicitly specified.

I do not specify `format` for my datetime field:

```
class CartSchema(Schema):
    class Meta:
        # Tuple or list of fields to include in the serialized result
        fields = ("_id", "created_at")
        dateformat = '%Y-%m-%d %H:%M:%S'

    created_at = fields.DateTime(dump_only=True,
                                 default=lambda: datetime.datetime.utcnow(),
                                 allow_none=False
                                 )
```

My date format is ignored and it is saved like `2019-05-25T06:58:00.103904+00:00`, but when give the same format to `DateTime` field it formats date as expected. 

Is this a bug? What is this misbehavior?

joegasewicz:
… parameter. Adds to comments for dateformat & datetimeformat.

### DIFF
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -323,7 +323,9 @@ class BaseSchema(base.SchemaABC):
             `OrderedDict`.
         - ``exclude``: Tuple or list of fields to exclude in the serialized result.
             Nested fields can be represented with dot delimiters.
-        - ``dateformat``: Date format for all DateTime fields that do not have their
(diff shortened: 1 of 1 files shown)


sloria:
Thanks!

PR #1247, opened 2019-06-15
Support microseconds in from_iso_datetime when use_dateutil=False

### LINKED ISSUE #1147: DateTime loses microseconds on deserialization if dateutil is not installed
```py
fields.DateTime().deserialize('2019-02-22T17:56:43.820462')
datetime.datetime(2019, 2, 22, 17, 56, 43)
```

This is obvious in the last line of `from_iso_datetime`:

```py
def from_iso_datetime(datetimestring, use_dateutil=True):
    """Parse an ISO8601-formatted datetime string and return a datetime object.

    Use dateutil's parser if possible and return a timezone-aware datetime.
    """
    if not _iso8601_datetime_re.match(datetimestring):
        raise ValueError('Not a valid ISO8601-formatted datetime string')
    # Use dateutil's parser if possible
    if dateutil_available and use_dateutil:
        return parser.isoparse(datetimestring)
    else:
        # Strip off timezone info.
        return datetime.datetime.strptime(datetimestring[:19], '%Y-%m-%dT%H:%M:%S')
```

The case is addressed in `from_iso_time`:

```py
def from_iso_time(timestring, use_dateutil=True):
    """Parse an ISO8601-formatted datetime string and return a datetime.time
    object.
    """
    if not _iso8601_time_re.match(timestring):
        raise ValueError('Not a valid ISO8601-formatted time string')
    if dateutil_available and use_dateutil:
        return parser.parse(timestring).time()
    else:
        if len(timestring) > 8:  # has microseconds
            fmt = '%H:%M:%S.%f'
        else:
            fmt = '%H:%M:%S'
        return datetime.datetime.strptime(timestring, fmt).time()
```

It is easier because there is no timezone info.

`from_iso_datetime` should be able to discriminate TZ info from microseconds.

I noticed this because an unrelated test fails if python-dateutil is not installed.

lafrech:
Fixes #1147.

Also improves `tests.base.assert_datetime_equal` and `tests.base.assert_time_equal`.

### DIFF
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -1,7 +1,7 @@
 """Utility methods for marshmallow."""
 import collections
 import functools
-import datetime
+import datetime as dt
 import inspect
 import json
 import re
@@ -89,10 +89,10 @@ def pprint(obj, *args, **kwargs):
 
 
 # From pytz: http://pytz.sourceforge.net/
-ZERO = datetime.timedelta(0)
+ZERO = dt.timedelta(0)
 
 
-class UTC(datetime.tzinfo):
+class UTC(dt.tzinfo):
     """UTC
 
     Optimized UTC implementation. It unpickles using the single module global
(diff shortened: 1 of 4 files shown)


lafrech:
Probably affects 2.x too. I shall check that.

sloria:
#1249 is forward-ported.

PR #1254, opened 2019-06-17
python 3 lets you raise with extra context using 'from'

fuhrysteve:
Since we've dropped support for python 2 now, this should give folks some better stack traces


sloria:
Nice! TIL about PEP 3134.

PR #1258, opened 2019-06-18
add note regarding backwards compatibility change

fuhrysteve:
Not sure if I worded this right, but I ran into this when my build failed after attempting to modify a Schema's `exclude` attribute after `__init__` . I.e.:

```python
foo = MySchema()
foo.exclude = ['first_name']
foo._update_fields()
```

The above was failing on a subsequent call to (private) `_update_fields()`. Probably my own fault for making nasty use of private API in a rather... interesting implementation (basically I have some permissions overrides for different user types and rolled my own `exclude_only` pattern for different user types)

See stacktrace below:

```python
    def _update_fields(self, obj=None, many=False):
        """Update fields based on the passed in object."""
        if self.only is not None:
            # Return only fields specified in only option
            if self.opts.fields:
                field_names = self.set_class(self.opts.fields) & self.set_class(self.only)
            else:
                field_names = self.set_class(self.only)
        elif self.opts.fields:
            # Return fields specified in fields option
            field_names = self.set_class(self.opts.fields)
        elif self.opts.additional:
            # Return declared fields + additional fields
            field_names = (self.set_class(self.declared_fields.keys()) |
                            self.set_class(self.opts.additional))

       else:
            field_names = self.set_class(self.declared_fields.keys())
    
        # If "exclude" option or param is specified, remove those fields
>       field_names -= self.exclude
E       TypeError: unsupported operand type(s) for -=: 'set' and 'tuple'

/usr/local/lib/python3.6/site-packages/marshmallow/schema.py:770: TypeError
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -555,6 +555,9 @@ Bug fixes:
   multiple instantiations (:issue:`1212`). Thanks :user:`MHannila` for
   reporting.
 
+Other changes:
+- ``Schema.exclude`` is now a set (previously it was a list or tuple)
+
 2.19.2 (2019-03-30)
 +++++++++++++++++++
 


lafrech:
This is related to https://github.com/marshmallow-code/marshmallow/pull/1238.

The "modify `exclude` after `__init__`" case was not taken into account in this change.

That case could be the reason

```py
     excludes = set(self.opts.exclude) | set(self.exclude)
```

was done in `_update_fields` and `_normalize_nested_options` rather than in `__init__`.

We could change the line that errors in your stacktrace into

```py
field_names -= set(self.exclude)
```

to make your code pass, but there would still be another difference: `self.opts.exclude` would be used only in `__init__`, not when modifying `exclude` after `__init__`.

Another issue with modifying `exclude` after `__init__` is that `_normalize_nested_options` is not called so fields with dots are not passed to nested fields. This was the case before #1238 already.

I'm tempted to say `exclude` shouldn't be modified  after `__init__`. Should we underscore it?

fuhrysteve:
Hmm yeah we could do something like this:


```python

def __init__(self, ...):
    self._exclude = exclude

@property
def exclude(self):
    return self._exclude

@exclude.setter
def exclude(self, value):
    self._exclude = set(value)
```

however you'd lose the ability to do more complex set operations via public api, i.e. this would no longer work:

```python
# append password to exclude
foo.exclude |= {'password'}
```

lafrech:
I think it would work if 

```py
def __init__(self, ...):
    self._exclude = exclude
```

but it wouldn't address the other concerns about `self.opts.exclude` and `_normalize_nested_options`. 

The point of underscoring `exclude` in my comment was to explicitly prevent modifications after `__init__`, not to find a way to make them work.

I realize this would break your use case.

The issue in your case is modifying `exclude` after `__init__`, not calling `_update_fields`. `_update_fields` is private API but it can be triggered by setting `update_fields` to `True` in `dump`.

fuhrysteve:
For better or worse, I was able to refactor out my implementation of modifying `Schema.exclude` after `__ini__` (I'm still modifying `.only`, but that doesn't seem to be problematic yet).

Making `exclude` private API makes sense _especially_ if we are going to be making assumptions about its type (namely, in this case, `set`).

Anything that is public API should not be subject to undocumented type modification. `set` isn't documented as an acceptable input for `exclude`, so if we're starting to think about `exclude` as a `set` rather than a `list` or `tuple` (as per the docs) we should either (1) move `exclude` to private API, or (2) tell everyone that `exclude` is supposed to be a `set` and issue a `DeprecationWarning` or `TypeError` when it is treated as anything else

fuhrysteve:
@lafrech do you think we should make any decisions for the future about making `exclude` to private API at this time? Either way I'm inclined to just close this PR for now

lafrech:
I'd make `exclude` and `only` private API because they are used in `_normalize_nested_options` which is called only at `__init__`.

I'd also make `load_only` and `dump_only` private because setting them after `__init__` conflicts with

```py
        self.load_only = set(load_only) or set(self.opts.load_only)
        self.dump_only = set(dump_only) or set(self.opts.dump_only)
```

Likewise with `unknown` and `context` due to

```py
        self.unknown = unknown or self.opts.unknown
        self.context = context or {}
```

(setting them to `None` after `__init__` wouldn't be the same as in `__init__`)

At this point, I'm tempted to make everything private. I mean the attributes above plus `many`, `ordered` and `partial`.

Adding the underscore would break libs using marshmallow (like apispec) but that's life. Or is there another way to signify those are private? Like, just a comment in the docstring?

fuhrysteve:
I suggest we make an issue to make a decision about these fields for some future version. Maybe flagging it for 3.1 or 4.0 milestones would be appropriate?

Issue #663, opened 2017-07-28
mypy types annotations


JulienPalard:
I was trying mypy and on the same project I started using marshmallow. So I wrote an approximative and minimal (working for my usage) stub here: https://github.com/JulienPalard/typeshed/tree/marshmallow

I'm unsure if you prefer type annotations in your code or have a stub on python/typeshed.git or nothing so...

sloria:
Thanks for posting. We won't be adding type annotations in the near future. Maybe if/when we decide to support Python 3 only, but we don't have plans for that at the moment.

neilsh:
The marshmallow type annotations in typeshed would be useful for people using Python 3, but don't require the marshmallow project itself to be Py3-only.

sloria:
I would prefer to have type annotations co-located with the codebase rather than typeshed stubs, which run the risk of getting outdated. [1]

We can add annotations as part of https://github.com/marshmallow-code/marshmallow/issues/1120

[1] If someone is up for maintaining typeshed stubs, go for it! I just won't have the time to maintain them myself ATM.

lafrech:
Python 2 will be dropped in Marshmallow 3.

JulienPalard:
Since then, it's possible to package type annotations: https://www.python.org/dev/peps/pep-0561/ but I think the better way is to have them inline. In a nutshell there's three ways for distributing annotations:

1. Write type information inline.
2. Package stubs in the package.
3. Package stube in another package (like typeshed).

sloria:
Yes we will add these inline

lafrech:
Draft PR by @fuhrysteve: https://github.com/marshmallow-code/marshmallow/pull/1259.

sloria:
Types are now checked with mypy, and as of 3.1.0 type information is distributed with the package. Closing this for now; we'll continue to expand type coverage with proximate development, when it makes sense. PRs to expand type coverage are welcome.

PR #1259, opened 2019-06-18
Add type annotations and mypy checking

fuhrysteve:
How do you guys feel about [type annotations](https://docs.python.org/3/library/typing.html)? Maybe something like `mypy` could be part of the build in the future?

I don't know that I got all these types right, but it's a start. I'll put some more time into it if y'all are bullish on type annotations!


lafrech:
Thanks @fuhrysteve for tackling https://github.com/marshmallow-code/marshmallow/issues/663.

fuhrysteve:
@lafrech oh no sounds like you've volunteered me!

I actually think this could help prevent a lot of bugs, especially when considering our discussion on #1258

sloria:
Thanks @fuhrysteve ! Yes, we've been planning to add annotations. We'll eventually want to add `mypy` to our `pre-commit` setup and distribute the types in the package (https://mypy.readthedocs.io/en/stable/installed_packages.html?highlight=py.typed#making-pep-561-compatible-packages).

This looks like a good start. This can be an iterative effort--no need to get 100% type coverage before this is merged. I think we'll get 90% of the benefit by just typing `Schema` and `fields` anyway.

src/marshmallow/fields.py
```diff
@@ -416,7 +417,13 @@ class Nested(Field):
     default_error_messages = {"type": "Invalid type."}
 
     def __init__(
-        self, nested, *, default=missing_, exclude=tuple(), only=None, **kwargs
+        self,
+        nested: SchemaABC,
```
sloria: Can probably just be `Schema`. Users shouldn't really deal with `SchemaABC` in practice.

src/marshmallow/fields.py
```diff
@@ -416,7 +417,13 @@ class Nested(Field):
...
+        *,
+        default: typing.Any = missing_,
+        exclude: typing.Union[typing.List, typing.Tuple] = tuple(),
```
sloria: Can also be a `set`, right? Same with `only`.

Might want to DRY this with `only` and `exclude` in `Schema`.

src/marshmallow/class_registry.py
```diff
@@ -15,8 +16,11 @@
 # }
 _registry = {}
 
+if typing.TYPE_CHECKING:
```
sloria: Honest question: Is this the accepted way to do things when an import is only used for typing? In practice there should be near-zero cost for this import, but if this is the accepted wisdom, then I guess it's fine.

fuhrysteve: I'm totally with you on micro-optimizing import time....

So this was a rather instinctual attempt to prevent circular imports. At a quick glance I'm not sure whether this is an actual issue or not for `class_registry.py` (still working on my local test environment! Maybe that'll help answer the question)

I'll take it out if it's not necessary

fuhrysteve: In any case, it certainly is an eyesore. If only the python devs could figure out how to make the whole circular import problem just go away for those of us who aren't really using python as a scripting language

fuhrysteve: Looks like my gut was right on this one! Circular import city:

```python
ImportError while loading conftest '/home/fuhrysteve/projects/marshmallow/tests/conftest.py'.
tests/conftest.py:4: in <module>
    from tests.base import User, UserSchema, Blog
tests/base.py:9: in <module>
    from marshmallow import Schema, fields, post_load, validate, missing
.tox/py36/lib/python3.6/site-packages/marshmallow/__init__.py:1: in <module>
    from marshmallow.schema import Schema, SchemaOpts
.tox/py36/lib/python3.6/site-packages/marshmallow/schema.py:13: in <module>
    from marshmallow import base, fields as ma_fields, class_registry
.tox/py36/lib/python3.6/site-packages/marshmallow/fields.py:13: in <module>
    from marshmallow import validate, utils, class_registry
.tox/py36/lib/python3.6/site-packages/marshmallow/class_registry.py:12: in <module>
E   ImportError: cannot import name 'Schema'
```

fuhrysteve:
This seems like a reasonable start!

For better or worse, I dropped support for python 3.5... not sure if that's ok or not, but they improved a lot of the syntax for typing in python 3.6.

Let me know what you guys want to do with this! In the meantime I'll do my best to keep it up to date with the upstream dev branch

sloria:
I'd love to drop support for 3.5 but had assumed it was a non-starter since it's still a maintained version and the default version on a number of Linux distros. 

Here are the numbers:

[Marshmallow 3: Python 3 Usage by OS](https://bigquery.cloud.google.com/savedquery/501466732326:e87466e468a344f6a19c774ff4801f0b)
[Marshmallow 3: Python 3.5 Usage by OS](https://bigquery.cloud.google.com/results/spring-house-88715:US.bquijob_54827b3c_16b84cf142e).

h/t @deckar01 for these queries

Ignoring Linux downloads as mostly from CI (https://github.com/marshmallow-code/marshmallow/issues/1120#issuecomment-482329287):

non-CI marshmallow 3 downloads on Python 3 in the past 30 days: 4728
non-CI marshmallow 3 downloads on Python 3.5 in the past 30 days: 270

So we have 5.7% of ma 3 users on Python 3.5.

On one hand, it seems user-hostile to drop a Python version this late in the RC process (we're sooo close to a stable release). On the other hand, supporting Python 3.5 is a non-trivial maintenance burden, given all the improvements in Python 3.6 we'd miss out on.

@deckar01 @lafrech Thoughts on removing 3.5 support?

lafrech:
> Thoughts on removing 3.5 support?

Not sure.

Let me speak for Debian users. Production environments generally run Debian stable, which is currently Stretch (Python 3.5). The next stable, Buster, should ship in July with Python 3.7 but I doubt all production environments will make the move right away.

We could argue that MA3 is for new projects, therefore new environments and new environments would use new stable. But projects may be stuck on old environments for many reasons.

(From a Debian perspective, we could even drop 3.6 as well and require 3.7...)

> supporting Python 3.5 is a non-trivial maintenance burden, given all the improvements in Python 3.6 we'd miss out on.

Can you elaborate on this?

I've been reading the changelog and I didn't see much revolutionary stuff, apart from async improvements that are irrelevant here, so I don't really see the benefits.

sloria:
> Can you elaborate on this?

The main thing is the stabilization of the `typing` module and the variable annotations. IIRC, many of the typing changes have been backported, but those backports only exist in newer 3.5.x versions. These issues can be a pain to trace down.

That said, I think it wouldn't be too much hassle to keep Python 3.5 support. So I don't really have a strong opinion either way.

fuhrysteve:
Looks like [python 3.5 is scheduled for EOL](https://devguide.python.org/#status-of-python-branches) in 2020-09-13

I'll take a stab at making this py3.5 compatible. It might make for some ugliness (i.e. putting stuff in comments like this: `# typing: List[str]`), but I think we can get it to be functionally equivalent.

tox.ini
```diff
@@ -1,9 +1,12 @@
...
-envlist = lint,py35,py36,py37,docs
+envlist = lint,py36,py37,docs
 
 [testenv]
 extras = tests,reco
-commands = pytest {posargs}
+whitelist_externals = mypy
+commands =
+    pytest {posargs}
+    mypy src/marshmallow
```
sloria: Mypy should be run with pre-commit as part of the `lint` command.

fuhrysteve:
It's on my calendar to take another stab at this tomorrow morning.. thanks for all the feedback so far! :smile:

sloria:
@fuhrysteve Are you still planning to look at this? If not, let us know and someone else can pick it up.

Unrelated: would you be interested in joining the marshmallow Slack? Email me at sloria1 AT gmail.com and I can send you an invite.

fuhrysteve:
Hey @sloria ! I'm behind on things and it's probably going to take me a little while to circle back to this. Feel free to do whatever you think is best and hopefully I'll be able to contribute more once the dust settles a little more over here! I'll shoot you a note about the slack channel

sloria:
No rush on this at all. Tbh, I'm still waffling on what the best approach to adding types are.. let's discuss on Slack =)

sloria:
I'm wondering if we should just wait until we drop Python 3.5 to go full bore on annotations, to avoid having to migrate from type comments. wdyt, @fuhrysteve ?

sloria:
I've updated this with `dev`, restored Python 3.5 compatibility, added `mypy` to pre-commit, and expanded type coverage in `fields.py`. I believe this is good to merge. We can expand type coverage in future PRs.

Issue #1234, opened 2019-06-07
DateTime timezone management rework


lafrech:
There are various issues I'd like to address in `DateTime` timezone management.

- ~Different behaviour depending on whether dateutil is installed (https://github.com/marshmallow-code/marshmallow/issues/497). Explicit would be better than implicit.~ (Fixed in #1265)
- Awareness not enforced (https://github.com/marshmallow-code/marshmallow/issues/520).

I begun work on this in #933 a while ago but I never got this through. I'd like to come up with a good design before getting back to it.

----------------------------------------------------------------------------------------

~AFAIU, the only action for which dateutil is required is deserializing iso formatted aware datetimes with non-utc timezones.~ Nope. dateutil is never needed (See below and #1265).

~Only iso requires dateutil. rfc format has timezones (https://tools.ietf.org/html/rfc822) but those are not as complex as iso timezones so they don't need dateutil.~ We don't use dateutil for RFC datetimes anymore (https://github.com/marshmallow-code/marshmallow/pull/1246 and https://github.com/marshmallow-code/marshmallow/pull/1248).

----------------------------------------------------------------------------------------

I've been thinking of using two field classes: `NaiveDateTime` and `AwareDateTime`.

Here's what they could do when deserializing:

Input | `NaiveDateTime` | `AwareDateTime`
-- | - | -
Naive | Pass | Fail
Aware UTC | Fail | Pass
Aware non UTC | Fail | Pass unless `utc_only` is `True`

We could add a `utc_only` flag to `AwareDateTime`. An exception would be raised at instantiation time if it is `False` and dateutil is not installed. (This may not be at import time if fields are declared programmatically, but it is better than deserialization time.) `utc_only` would default to `False`. Basically, this means the user who do not want to install dateutil would have to opt-in for `utc_only`. If it's too much pain, he could subclass the field to avoid providing the parameter on each instance.

----------------------------------------------------------------------------------------

There may be other aspects to consider:

- Could `NaiveDateTime` accept an aware datetime? Should it drop the TZ? I don't think so. Should it convert it to a specified TZ before dropping the TZ? It could accept a `timezone` argument specifying a target tz. An exception would be raised if target TZ is not UTC and dateutil is not installed.

`timezone` | Deserialization of aware datetimes in `NaiveDateTime`
-- | -
`None` | Fail
`'UTC'` | Convert to UTC then drop TZ
[...]

PR #1265, opened 2019-06-20
Copy ISO8601 deserialization functions from Django

lafrech:
Related to https://github.com/marshmallow-code/marshmallow/issues/1234 and https://github.com/marshmallow-code/marshmallow/pull/1252#issuecomment-503103373.

Using ISO8601 deserialization functions from Django makes python-dateutil dependency useless.

This PR copies the deserialization functions from Django and removes python-dateutil.

I can finish this later on (run black, cleanup, review docstrings) but feedback welcome already.

### DIFF
--- a/docs/upgrading.rst
+++ b/docs/upgrading.rst
@@ -883,6 +883,17 @@ In marshmallow 2, it was possible to have multiple fields with the same ``attrib
     MySchema()
     # No error
 
+
+``python-dateutil`` recommended dependency is removed
+*****************************************************
+
+In marshmallow 2, ``python-dateutil`` was used to deserialize RFC or ISO 8601
+datetimes if it was installed. In marshmallow 3, datetime deserialization is
+done with no additional dependency.
+
+``python-dateutil`` is no longer used by marshmallow.
+
+
 Upgrading to 2.3
 ++++++++++++++++
 
--- a/setup.py
+++ b/setup.py
@@ -3,13 +3,10 @@ import re
 from setuptools import setup, find_packages
 
 EXTRAS_REQUIRE = {
-    "reco": ["simplejson"],
-    "tests": ["pytest", "pytz"],
+    "tests": ["pytest", "pytz", "simplejson"],
     "lint": ["flake8==3.7.7", "flake8-bugbear==19.3.0", "pre-commit==1.17.0"],
 }
-EXTRAS_REQUIRE["dev"] = (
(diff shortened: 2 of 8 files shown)


lafrech:
Blacked. Looks good to me for now.

I'll pursue https://github.com/marshmallow-code/marshmallow/issues/1234 in other PRs.

setup.py
```diff
@@ -3,7 +3,7 @@
 from setuptools import setup, find_packages
 
 EXTRAS_REQUIRE = {
-    "reco": ["python-dateutil>=2.7.0", "simplejson"],
+    "reco": ["simplejson"],
```
sloria: We can probably just remove `"reco"` full stop. `simplejson` is only used in the tests, so you can just move it to the line below.

sloria (approved):
Looks great. I agree that this would simplify things a lot. As an added bonus, this appears to improve datetime deserialization performance by ~2x.

image

Old implementation w/ dateutil:

```
16 µs ± 367 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

Implementation in this PR:

```
8.68 µs ± 187 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

Do you have time to update the changelog and upgrading docs? Once that's done, go ahead and merge.

lafrech:
> We can probably just remove `"reco"` full stop. `simplejson` is only
> used in the tests, so you can just move it to the line below.

Sure. Done.

-- 
Jérôme

docs/upgrading.rst
```diff
@@ -883,6 +883,17 @@ In marshmallow 2, it was possible to have multiple fields with the same ``attrib
...
 
+
+`python-dateutil` recommended dependency is removed
+***************************************************
+
+In marshmallow 2, `python-dateutil` was used to deserialize RFC or ISO 8601
+datetimes if it was installed. In marshmallow 3, datetime deserialization is
+done with no additional dependency.
+
+`python-dateutil` is not a recommended dependency anymore.
```
sloria:
```suggestion
``python-dateutil`` is no longer used by marshmallow.
```

lafrech: I applied this change.

I never know when to use '`' and when to use '``'. I don't think we're consistent through the docs.

sloria: Double-backticks is for simple, pre-formatted text. Single-backticks is used when linking to other objects (e.g. autodocc'd or intersphinx members).

lafrech: Then there are a lot of single-backticks uses with no link in the docs (at least in upgrading.rst) where double-backticks should be used instead.

We can fix this when we get the time.

Issue #1268, opened 2019-06-28
Email and URL call String.__init__ rather than super().__init__


dursk:
It looks like this is a relic of the past, when multiple inheritance was being used (https://github.com/marshmallow-code/marshmallow/commit/f924a6e71612c1babb41a13f024e5790fd1924d8).

It seems like we should be able to do `super().__init__` in both these cases now.

It's currently causing method resolution issues when using multiple inheritance and subclassing these String subclasses, for example:
```python
class CustomBaseField(marshmallow.fields.Field):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         # some custom logic

class CustomEmail(marshmallow.fields.Email, CustomBaseField):
    pass
```
In the above scenario, when instantiating `CustomEmail`, `marshmallow.String.__init__` runs although it should be `CustomBaseField.__init__`

PR #1269, opened 2019-06-28
Update string sublcasses to call super().__init__

dursk:
Fix for issue https://github.com/marshmallow-code/marshmallow/issues/1268


sloria:
Thanks!

PR #1272, opened 2019-06-30
Switch to Azure Pipelines

sloria:
#1272 

- [x] Get builds passing
- [x] Update docs
- [x] Forward-port to `dev`
- [x] Remove travis integration on repo
- [x] Require AP check to pass


sloria:
So far so good. By my rough measurements, AP builds finish in about half the time that Travis takes.

PR #1275, opened 2019-07-02
Drop official support for Python 2.6

### LINKED ISSUE #1274: Drop Python 2.6 in marshmallow 2
Python 2.6 has been EOL since 2013.
The Python 2.6 build is currently broken, and I've not the motivation to fix it.
In the past 30 days, there's been only 1--exactly 1--non-Linux download of [marshmallow 2 on Python 2.6](https://bigquery.cloud.google.com/results/spring-house-88715:US.bquijob_28beb1b1_16badd0a401).

image

I think it's past time we drop official Python 2.6 support in marshmallow 2. We don't need to change the code--just stop testing against it.

Moving forward, I think we should keep the policy of dropping Python versions iff they are past their EOL date and have low usage (this will be a judgment call). 

Any objections, @lafrech @deckar01 ?

sloria:
close #1274

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,13 @@
 Changelog
 ---------
 
+2.20.0 (unreleased)
+*******************
+
+Deprecation/Removal:
+
+- Python 2.6 is no longer officially supported (:issue:`1274`).
+
 2.19.5 (2019-06-18)
 +++++++++++++++++++
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -124,7 +124,7 @@ To run syntax checks: ::
 
     $ tox -e lint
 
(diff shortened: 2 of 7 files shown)


CHANGELOG.rst
```diff
@@ -1,6 +1,15 @@
...
 ---------
 
+2.20.0 (unreleased)
+*******************
+
+Deprecation/Removal:
+
+- Python 2.6 is no longer officially supported (:issue:`1274`).
+  NOTE: Dropping "official support" in this case means we will not test
+  against Python 2.6 in CI, but the code is still expected to work with Python 2.6.
```
lafrech: If it ain't tested, it is broken.

If we don't test anymore, we may introduce some PY2.7+ syntax without realizing it. I'd just remove this note.

But I don't mind.

sloria: Fair enough. I added the note because this case is different from others in that we're not updating the codebase. But `!tested == broken` is truthy, as you said. I've removed the note.

PR #1276, opened 2019-07-02
Call super(), not super(String, self) in UUID

lafrech:
Sort of like #1269.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -746,7 +746,7 @@ class UUID(String):
 
     def _serialize(self, value, attr, obj, **kwargs):
         validated = str(self._validated(value)) if value is not None else None
-        return super(String, self)._serialize(validated, attr, obj, **kwargs)
+        return super()._serialize(validated, attr, obj, **kwargs)
 
     def _deserialize(self, value, attr, data, **kwargs):
         return self._validated(value)


sloria:
Huh, I'm surprised `pyupgrade` didn't catch this one.

src/marshmallow/fields.py
```diff
@@ -746,7 +746,7 @@ def _validated(self, value):
 
     def _serialize(self, value, attr, obj, **kwargs):
         validated = str(self._validated(value)) if value is not None else None
-        return super(String, self)._serialize(validated, attr, obj, **kwargs)
+        return super()._serialize(validated, attr, obj, **kwargs)
```
asottile, line 749: pyupgrade didn't catch this one because the `super(...)` call was wrong -- this change also ~potentially changes behaviour

`super`'s first argument should be the class you want to look upwards from, so before it was "upwards from String" and now it's "upwards from UUID" (might want to double check that this still does the right thing, I didn't look at the code)

I put together a small example and write up on this [here](https://github.com/asottile/t/tree/master/2019-08-18-super-wrong-class)

sloria: Aha. Nice explanation. Looks like this was PEBCAK after all 😆

asottile: looking at the code, the only difference in behaviour I can see is that in python2 it wouldn't have converted to `unicode` 🤷‍♂

PR #1278, opened 2019-07-02
Add NaiveDateTime and AwareDateTime

### LINKED ISSUE #1234: DateTime timezone management rework

lafrech:
Closes #1234.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1064,6 +1064,7 @@ class DateTime(Field):
 
     default_error_messages = {
         "invalid": "Not a valid {obj_type}.",
+        "invalid_awareness": "Not a valid {awareness} {obj_type}.",
         "format": '"{input}" cannot be formatted as a {obj_type}.',
     }
 
@@ -1123,7 +1124,7 @@ class DateTime(Field):
 class NaiveDateTime(DateTime):
     """A formatted naive datetime string."""
 
-    OBJ_TYPE = "naive datetime"
+    AWARENESS = "naive"
 
     def __init__(self, format=None, *, timezone=None, **kwargs):
         super().__init__(format=format, **kwargs)
@@ -1133,7 +1134,11 @@ class NaiveDateTime(DateTime):
         ret = super()._deserialize(value, attr, data, **kwargs)
         if is_aware(ret):
             if self.timezone is None:
-                raise self.fail("invalid", obj_type=self.OBJ_TYPE)
+                raise self.fail(
+                    "invalid_awareness",
+                    awareness=self.AWARENESS,
+                    obj_type=self.OBJ_TYPE,
+                )
             ret = ret.astimezone(self.timezone).replace(tzinfo=None)
         return ret
 
@@ -1141,7 +1146,7 @@ class NaiveDateTime(DateTime):
 class AwareDateTime(DateTime):
     """A formatted aware datetime string."""
 
-    OBJ_TYPE = "aware datetime"
+    AWARENESS = "aware"
 
     def __init__(self, format=None, *, default_timezone=None, **kwargs):
         super().__init__(format=format, **kwargs)
@@ -1151,7 +1156,11 @@ class AwareDateTime(DateTime):
         ret = super()._deserialize(value, attr, data, **kwargs)
         if not is_aware(ret):
             if self.default_timezone is None:
-                raise self.fail("invalid", obj_type=self.OBJ_TYPE)
+                raise self.fail(
+                    "invalid_awareness",
+                    awareness=self.AWARENESS,
+                    obj_type=self.OBJ_TYPE,
+                )
             ret = ret.replace(tzinfo=self.default_timezone)
         return ret
 
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1122,7 +1122,15 @@ class DateTime(Field):
 
 
 class NaiveDateTime(DateTime):
-    """A formatted naive datetime string."""
+    """A formatted naive datetime string.
+
+    :param str format: See :class:`DateTime`.
+    :param timezone timezone: Used on deserialization. If `None`,
+        aware datetimes are rejected. If not `None`, aware datetimes are
+        converted to this timezone before their timezone information is
+        removed.
+    :param kwargs: The same keyword arguments that :class:`Field` receives.
+    """
 
     AWARENESS = "naive"
 
@@ -1144,7 +1152,14 @@ class NaiveDateTime(DateTime):
 
 
 class AwareDateTime(DateTime):
-    """A formatted aware datetime string."""
+    """A formatted aware datetime string.
+
+    :param str format: See :class:`DateTime`.
+    :param timezone default_timezone: Used on deserialization. If `None`, naive
+        datetimes are rejected. If not `None`, naive datetimes are set this
+        timezone.
+    :param kwargs: The same keyword arguments that :class:`Field` receives.
+    """
 
     AWARENESS = "aware"
 
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1136,10 +1136,16 @@ class AwareDateTime(DateTime):
     """A formatted aware datetime string."""
     OBJ_TYPE = "aware datetime"
 
+    def __init__(self, format=None, *, default_timezone=None, **kwargs):
+        super().__init__(format=format, **kwargs)
+        self.default_timezone = default_timezone
+
     def _deserialize(self, value, attr, data, **kwargs):
         ret = super()._deserialize(value, attr, data, **kwargs)
         if not is_aware(ret):
-            raise self.fail("invalid", obj_type=self.OBJ_TYPE)
+            if self.default_timezone is None:
+                raise self.fail("invalid", obj_type=self.OBJ_TYPE)
+            ret = ret.replace(tzinfo=self.default_timezone)
         return ret
 
 
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -11,7 +11,7 @@ from collections.abc import Mapping as _Mapping
 
 from marshmallow import validate, utils, class_registry
 from marshmallow.base import FieldABC, SchemaABC
-from marshmallow.utils import is_collection, missing as missing_, resolve_field_instance
+from marshmallow.utils import is_collection, missing as missing_, resolve_field_instance, is_aware
 from marshmallow.exceptions import (
     ValidationError,
     StringNotCollectionError,
@@ -35,6 +35,8 @@ __all__ = [
     "Boolean",
     "Float",
     "DateTime",
+    "NaiveDateTime",
+    "AwareDateTime",
     "Time",
     "Date",
     "TimeDelta",
@@ -1026,19 +1028,13 @@ class Boolean(Field):
 
 
 class DateTime(Field):
-    """A formatted datetime string in UTC.
+    """A formatted datetime string.
 
     Example: ``'2014-12-22T03:12:58.019077+00:00'``
-
-    Timezone-naive `datetime` objects are converted to
-    UTC (+00:00) by :meth:`Schema.dump <marshmallow.Schema.dump>`.
-    :meth:`Schema.load <marshmallow.Schema.load>` returns `datetime`
-    objects that are timezone-aware.
 
     :param str format: Either ``"rfc"`` (for RFC822), ``"iso"`` (for ISO8601),
         or a date format string. If `None`, defaults to "iso".
     :param kwargs: The same keyword arguments that :class:`Field` receives.
-
     """
 
     SERIALIZATION_FUNCS = {
@@ -1117,6 +1113,28 @@ class DateTime(Field):
     @staticmethod
     def _make_object_from_format(value, data_format):
         return dt.datetime.strptime(value, data_format)
+
+
+class NaiveDateTime(DateTime):
+    """A formatted naive datetime string."""
+    OBJ_TYPE = "naive datetime"
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        ret = super()._deserialize(value, attr, data, **kwargs)
+        if is_aware(ret):
+            raise self.fail("invalid", obj_type=self.OBJ_TYPE)
+        return ret
+
+
+class AwareDateTime(DateTime):
+    """A formatted aware datetime string."""
+    OBJ_TYPE = "aware datetime"
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        ret = super()._deserialize(value, attr, data, **kwargs)
+        if not is_aware(ret):
+            raise self.fail("invalid", obj_type=self.OBJ_TYPE)
+        return ret
 
 
 class Time(Field):
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -11,7 +11,12 @@ from collections.abc import Mapping as _Mapping
 
 from marshmallow import validate, utils, class_registry
 from marshmallow.base import FieldABC, SchemaABC
-from marshmallow.utils import is_collection, missing as missing_, resolve_field_instance, is_aware
+from marshmallow.utils import (
+    is_collection,
+    missing as missing_,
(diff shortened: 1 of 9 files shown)


sloria:
Is `NaiveDateTime` necessary? What's the use case?

lafrech:
In Python, naive and aware datetimes are different objects with different properties and some operations only work on either naive or aware datetimes. I was caught by this already (https://github.com/lepture/flask-oauthlib/issues/294). Being able to enforce the type is therefore a must have IMHO. Otherwise, the user needs to do it manually.

Whether it is done by a specific class or a flag is an implementation choice. I think different classes is better, more explicit. Generally we have a class for a type and the flags are just parameters.

sloria:
It makes sense to have validation that an input datetime includes timezone info, but I'm more curious about cases when you want to enforce naiveness. Or is it more about having symmetry?

lafrech:
Symmetry would be good enough a reason for me, since it is relatively easy to achieve.

[This issue](https://github.com/lepture/flask-oauthlib/issues/294) is an example of a lib that expects a naive datetime and will fail if fed an aware datetime. I don't use it anymore, but by the time, I had to enforce naiveness myself manually.

This issue can be extended to every use case involving date comparisons.

If the API stores datetimes, it most likely wants to enforce the type, if only to allow the user to sort objects by datetime. Perhaps most of the time, the DB driver will uniformize the format, by making a silent choice to force aware or naive, but I think this should be the role of the deserialization layer.

Maybe I'm missing something because once I fell into this, I figured it was something marshmallow absolutely needed to address, but since then, I heard no one complain, so maybe it is not such an issue in practice.

(In any case, I'd still consider this PR relevant, as it comes with a nice cleanup, hopefully no regression, and a code easier to maintain and with better test coverage.)

sloria:
Sure, I'm fine with that reasoning and with going forward with this. Will take a closer look hopefully over the next few days.

deckar01:
> I heard no one complain

I suspect loading datetimes is just not a very common behavior of most apps. Not until I made a webcam timelapse scheduler recently did I ever have to accept a datetime from the user.

src/marshmallow/fields.py
```diff
@@ -1112,15 +1106,26 @@ def _make_object_from_format(value, data_format):
...
-class LocalDateTime(DateTime):
-    """A formatted datetime string in localized time, relative to UTC.
+class NaiveDateTime(DateTime):
+    """A formatted naive datetime string."""
+    OBJ_TYPE = "naive datetime"
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        ret = super()._deserialize(value, attr, data, **kwargs)
+        if is_aware(ret):
+            raise self.fail("invalid", obj_type=self.OBJ_TYPE)
```
sloria: Would there be value in using a different key in `default_error_messages` in order to allow users to have different messages for "invalid datetime" vs "invalid awareness"?

I've no strong opinion either way.

src/marshmallow/utils.py
```diff
@@ -136,6 +136,11 @@ def __str__(self):
 UTC = utc = UTC()  # UTC is a singleton
 
 
+# https://stackoverflow.com/a/27596917
+def is_aware(datetime):
+    return datetime.tzinfo is not None and datetime.tzinfo.utcoffset(datetime) is not None
```
sloria: Even after reading the SO post and relevant Python docs, I still don't understand why utcoffset can be `None` when tzinfo is not `None`. But your implementation is correct nonetheless.

>  If d.tzinfo is None, or if d.tzinfo is not None but d.tzinfo.utcoffset(d) returns None, d is naive.

 ¯\\_(ツ)_/¯

src/marshmallow/utils.py
```diff
@@ -235,16 +232,12 @@ def from_iso_date(value):
...
-    """Return the ISO8601-formatted UTC representation of a datetime object."""
-    if localtime and datetime.tzinfo is not None:
-        localized = datetime
-    else:
-        if datetime.tzinfo is None:
-            localized = UTC.localize(datetime)
-        else:
-            localized = datetime.astimezone(UTC)
-    return localized.isoformat(*args, **kwargs)
+def isoformat(datetime, *args, **kwargs):
```
sloria: `*args` and `**kwargs` can be removed here, as they're never passed

tests/test_deserialization.py
```diff
@@ -410,48 +410,114 @@ def test_invalid_datetime_deserialization(self, in_value):
...
-        assert_datetime_equal(field.deserialize(datestring), dtime)
+        assert field.deserialize(datestring) == dt.datetime(2019, 1, 2, 10, 11, 12, 123456)
+
+        field = fields.NaiveDateTime(format="%H:%M:%S.%f %Y-%m-%d")
+        assert field.deserialize(datestring) == dt.datetime(2019, 1, 2, 10, 11, 12, 123456)
+
+        field = fields.AwareDateTime(format="%H:%M:%S.%f %Y-%m-%d")
+        with pytest.raises(ValidationError) as excinfo:
+            field.deserialize(datestring)
+        assert excinfo.value.args[0] == "Not a valid aware datetime."
```
sloria: Nit: Passing `match` to `pytest.raises` is tad nicer and more consistent with the other tests.

sloria (commented):
I'm basically good with this. My comments are nits. Once this is polished up, should be good to merge. 🎉

src/marshmallow/utils.py
lafrech: I'm with you here. I just applied the docs to be on the safe side.

https://docs.python.org/3/library/datetime.html#available-types

> An object of type time or datetime may be naive or aware. A datetime object d is aware if d.tzinfo is not None and d.tzinfo.utcoffset(d) does not return None. If d.tzinfo is None, or if d.tzinfo is not None but d.tzinfo.utcoffset(d) returns None, d is naive. A time object t is aware if t.tzinfo is not None and t.tzinfo.utcoffset(None) does not return None. Otherwise, t is naive.

lafrech: Ideally, Python core would provide an `aware` property to abstract this.

src/marshmallow/fields.py
lafrech: Done.

lafrech:
Rebased and documented.

Feedback welcome.

I'm open to changes, especially about the new parameter names. And the docs.

src/marshmallow/fields.py
```diff
@@ -1059,9 +1059,9 @@ class DateTime(Field):
 
     SCHEMA_OPTS_VAR_NAME = "datetimeformat"
 
-    localtime = False
     default_error_messages = {
         "invalid": "Not a valid {obj_type}.",
+        "invalid awareness": "Not a valid {awareness} {obj_type}.",
```
sloria:
```suggestion
        "invalid_awareness": "Not a valid {awareness} {obj_type}.",
```

src/marshmallow/fields.py
```diff
@@ -1114,15 +1114,62 @@ def _make_object_from_format(value, data_format):
         return dt.datetime.strptime(value, data_format)
 
 
-class LocalDateTime(DateTime):
-    """A formatted datetime string in localized time, relative to UTC.
+class NaiveDateTime(DateTime):
+    """A formatted naive datetime string.
+
+    :param str format: See :class:`DateTime`.
+    :param timezone timezone: Used on deserialization. If None, aware datetimes
```
sloria:
```suggestion
    :param timezone timezone: Used on deserialization. If `None`, aware datetimes
```

src/marshmallow/fields.py
```diff
@@ -1114,15 +1114,62 @@ def _make_object_from_format(value, data_format):
...
+        are rejected. If not None, aware datetimes are converted to this
```
sloria:
```suggestion
        are rejected. If not `None`, aware datetimes are converted to this
```

src/marshmallow/fields.py
```diff
@@ -1114,15 +1114,62 @@ def _make_object_from_format(value, data_format):
...
+        timezone before their timezone information is removed.
+    :param kwargs: The same keyword arguments that :class:`Field` receives.
+    """
+
+    AWARENESS = "naive"
 
-        ex. ``"Sun, 10 Nov 2013 08:23:45 -0600"``
+    def __init__(self, format=None, timezone=None, **kwargs):
```
sloria:
```suggestion
    def __init__(self, format=None, *, timezone=None, **kwargs):
```

src/marshmallow/fields.py
```diff
@@ -1114,15 +1114,62 @@ def _make_object_from_format(value, data_format):
...
+        super().__init__(format=format, **kwargs)
+        self._timezone = timezone
```
sloria: Why the underscore?

src/marshmallow/fields.py
```diff
@@ -1114,15 +1114,62 @@ def _make_object_from_format(value, data_format):
...
+    :param timezone default_timezone: Used on deserialization. If None, naive
+        datetimes are rejected. If not None, naive datetimes are set this
+        timezone.
+    :param kwargs: The same keyword arguments that :class:`Field` receives.
     """
 
-    localtime = True
+    AWARENESS = "aware"
+
+    def __init__(self, format=None, default_timezone=None, **kwargs):
```
sloria:
```suggestion
    def __init__(self, format=None, *, default_timezone=None, **kwargs):
```

src/marshmallow/fields.py
```diff
@@ -1114,15 +1114,62 @@ def _make_object_from_format(value, data_format):
...
+        super().__init__(format=format, **kwargs)
+        self._default_timezone = default_timezone
```
sloria: Same here: why the underscore?

tests/test_schema.py
```diff
@@ -1636,7 +1631,7 @@ def test_meta_serializer_fields():
     assert result["uppername"] == "JOHN"
     assert result["is_old"] is False
     assert result["created"] == utils.isoformat(u.created)
-    assert result["updated_local"] == utils.isoformat(u.updated, localtime=True)
+    # assert result["updated_local"] == utils.isoformat(u.updated, localtime=True)
```
sloria:
```suggestion
```

tests/test_schema.py
```diff
@@ -1645,7 +1640,7 @@ def test_meta_fields_mapping(user):
     s = UserMetaSchema()
     s.dump(user)  # need to call dump to update fields
     assert type(s.fields["balance"]) == fields.Decimal
-    assert type(s.fields["updated_local"]) == fields.LocalDateTime
+    # assert type(s.fields["updated_local"]) == fields.LocalDateTime
```
sloria:
```suggestion
```

src/marshmallow/fields.py
```diff
@@ -1114,15 +1114,62 @@ def _make_object_from_format(value, data_format):
...
+                )
+            ret = ret.astimezone(self._timezone).replace(tzinfo=None)
+        return ret
+
+
+class AwareDateTime(DateTime):
+    """A formatted aware datetime string.
+
+    :param str format: See :class:`DateTime`.
+    :param timezone default_timezone: Used on deserialization. If None, naive
```
sloria:
```suggestion
    :param timezone default_timezone: Used on deserialization. If `None`, naive
```

src/marshmallow/fields.py
```diff
@@ -1114,15 +1114,62 @@ def _make_object_from_format(value, data_format):
...
+        datetimes are rejected. If not None, naive datetimes are set this
```
sloria:
```suggestion
        datetimes are rejected. If not `None`, naive datetimes are set to this
```

sloria (commented):
Looking awesome. Nice job with the tests. Just nits left.

src/marshmallow/fields.py
```diff
@@ -1114,15 +1114,62 @@ def _make_object_from_format(value, data_format):
...
+
+    def __init__(self, format=None, default_timezone=None, **kwargs):
+        super().__init__(format=format, **kwargs)
+        self._default_timezone = default_timezone
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        ret = super()._deserialize(value, attr, data, **kwargs)
+        if not is_aware(ret):
+            if self._default_timezone is None:
+                raise self.fail(
```
sloria: `raise self.fail` doesn't actually do anything different from `self.fail`. I noticed we do this in other places and made a fix here: #1301

src/marshmallow/fields.py
lafrech: Well. It is related to https://github.com/marshmallow-code/marshmallow/issues/1281. But this has not been settled yet, so let's keep everything not underscored for now, for consistency.

lafrech:
Rebased. Remarks addressed.

sloria (approved):
Very nice. Can you update the changelog then merge?

 🚢 🇮🇹

PR #1290, opened 2019-07-08
Fix contained parent in container fields

lafrech:
I think this is the right fix for https://github.com/marshmallow-code/marshmallow/issues/779#issuecomment-396354987.

I had to modify some tests. Apparently, the former behaviour was intended but I don't see why so I assume it was a mistake.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -585,8 +585,7 @@ class List(Field):
     def _bind_to_schema(self, field_name, schema):
         super()._bind_to_schema(field_name, schema)
         self.inner = copy.deepcopy(self.inner)
-        self.inner.parent = self
-        self.inner.name = field_name
+        self.inner._bind_to_schema(field_name, self.parent)
         if isinstance(self.inner, Nested):
             self.inner.only = self.only
             self.inner.exclude = self.exclude
@@ -663,8 +662,7 @@ class Tuple(Field):
         new_tuple_fields = []
         for field in self.tuple_fields:
             field = copy.deepcopy(field)
-            field.parent = self
-            field.name = field_name
+            field._bind_to_schema(field_name, self.parent)
             new_tuple_fields.append(field)
 
         self.tuple_fields = new_tuple_fields
@@ -1303,15 +1301,13 @@ class Mapping(Field):
         super()._bind_to_schema(field_name, schema)
         if self.value_field:
             self.value_field = copy.deepcopy(self.value_field)
-            self.value_field.parent = self
-            self.value_field.name = field_name
+            self.value_field._bind_to_schema(field_name, self.parent)
         if isinstance(self.value_field, Nested):
             self.value_field.only = self.only
             self.value_field.exclude = self.exclude
         if self.key_field:
             self.key_field = copy.deepcopy(self.key_field)
-            self.key_field.parent = self
-            self.key_field.name = field_name
+            self.key_field._bind_to_schema(field_name, self.parent)
 
     def _serialize(self, value, attr, obj, **kwargs):
         if value is None:
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -107,6 +107,7 @@ class TestParentAndName:
         foo = fields.Field()
         bar = fields.List(fields.Str())
         baz = fields.Tuple([fields.Str(), fields.Int()])
+        bax = fields.Mapping(fields.Str(), fields.Int())
 
     @pytest.fixture()
     def schema(self):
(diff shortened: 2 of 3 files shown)


sloria:
I think the former behavior is actually correct. The `parent` of an inner field is the container field, while its `root` will be the schema. Can the `List(Nested('self'))` case be addressed without changing these semantics?

lafrech:
Yep. I wasn't sure about the meaning or `parent`. Thanks for clarifying.

`root`:

> Reference to the `Schema` that this field belongs to even if it is buried in a container field (e.g. `List`).

IIUC, the proper fix is to use `root` instead of `parent` in `Nested`.

```py
                elif self.nested == "self":
-                   schema_class = self.parent.__class__
+                   schema_class = self.root.__class__
```

Looks like this bugfix should be applied to 2.x as well, right?

sloria:
> IIUC, the proper fix is to use root instead of parent in Nested.

I don't think that will handle deep self-nesting e.g.

```python
class TreeNodeSchema():
    children = List(Nested('self'))

class MySchema(Schema):
    tree = Nested(TreeNodeSchema())
```

sloria:
Btw, I'm tempted to deprecate `'self'` in favor of explicitly passing the schema name.

```python
class TreeNodeSchema():
    children = List(Nested('TreeNodeSchema'))
```

wdyt?

lafrech:
I'm not sure why it wouldn't work. Is it because `root` would be `MySchema` rather than `TreeNodeSchema`?

Looking at it I remembered that I wanted to remove `root` in MA 4 so I thought "what if we just removed `self` and asked for the schema name instead"?

Therefore, +1 for removing `self`.

sloria:
> I'm not sure why it wouldn't work. Is it because root would be MySchema rather than TreeNodeSchema?

Right.

> Therefore, +1 for removing self.

OK, let's go with this, eventually. So unless you find an easy fix for `List(Nested('self'))` that doesn't break `root`/`parent`, I think we can just document this limitation of the List(Nested) form.

lafrech:
I suppose a fix could be to climb up `parent` hierarchy, just like when searching for `root`, but stopping at the first `SchemaABC` instance.

IIUC, your plan is to deprecate `self` in MA3 and remove it in MA4, so we'd live with this ugly fix only until MA4. Fine.

sloria:
> I suppose a fix could be to climb up parent hierarchy, just like when searching for root, but stopping at the first SchemaABC instance.

Yes, that should work.

> IIUC, your plan is to deprecate self in MA3 and remove it in MA4, so we'd live with this ugly fix only until MA4. Fine.

Right.

deckar01:
This change makes sense to me. `Field.parent` is supposed to be a schema, but it was being set to collection field instances. This mostly worked, because the few things that `parent` is used for are also proxied through the fields to their parent schema. The thing that broke was `Nested('self')` expecting `self.parent.__class__` to be a schema. Even if `Nested('self')` was deprecated/removed, this fix would avoid unnecessary indirection on some nested container attributes.

sloria:
Interesting, so I think the original intention was that `parent` was the _immediate_ parent, which could be either a Field or Schema. Without doing a code dive, I'm not sure what use cases would break if we changed parent to always be the nearest ancestor Schema. I'm open to the idea if you both think it's more intuitive.

deckar01:
I can see how allowing `parent` to point to fields might be more useful. ~~The only outlier was `Nested` pointing to its parent instead of itself.~~

I initially thought "climb up parent hierarchy" was referring to `_bind_to_schema()`, but maybe you meant just for resolving `Nested('self')`. That would work.

sloria:
> I initially thought "climb up parent hierarchy" was referring to _bind_to_schema(), but maybe you meant just for fixing Nested('self'). That would work.

Right, we'd only do the climb-to-nearest-schema for the `'self'` use case.

lafrech:
> Without doing a code dive, I'm not sure what use cases would break if we changed parent to always be the nearest ancestor Schema.

I think that's what this PR does and it only breaks the tests that precisely test `parent` in container fields. I didn't see any other side-effects.

It's unclear to me who/what actually uses the `parent` attribute outside of our own code. Is this something other users/libs might be relying upon?

> I can see how allowing parent to point to fields might be more useful.

I have no use case in mind.

The "climb up" patch would solve the `Nested` issue so at least we got this covered. The question about what to do about `parent` remains.

> I'm open to the idea if you both think it's more intuitive.

I have no idea. It seems clearer to you guys than it is to me.

sloria:
On a surface level, `parent` = immediate parent makes more intuitive sense to me. Let's keep the current semantics, if only to avoid introducing an unnecessary breaking change. Then use the climb-up patch here.

lafrech:
Replaced by https://github.com/marshmallow-code/marshmallow/pull/1294.

Issue #1280, opened 2019-07-03
Propose add required to schema constructor


heckad:
For a put request model require of `id` field, but for post request `id` not needed. For disabling in a post request I use `dump_only=["id"]`, but I have not found such a way to do `id` required. For this reason, I suggest adding `required` to the schema constructor.

sloria:
Your use case can be achieved by using a separate schema for POSTs vs PUTs.

lafrech:
In my APIs, I use the same schema but the ID for the PUT is in the URL, so it is injected in the view func (by webargs) as a path parameter.

heckad:
> Your use case can be achieved by using a separate schema for POSTs vs PUTs.

I use marshmallow sqlalchemy. It generates schema by models. How to observe the dry-principles? What is the problem to add?

sloria:
```python
class WidgetSchema(ModelSchema):
    class Meta:
        model = Widget

class CreateWidgetSchema(WidgetSchema):
    class Meta(WidgetSchema.Meta):
        exclude = ('id',)
```

Adding another parameter is not ideal. We try to keep marshmallow's API surface as small as is practical. An added parameter adds combinatorial complexity to the API because it needs to be tested against the existing parameters. There's additional consequences to consider, like whether the parameter needs to also be included in `load`/`dump`, `Meta` options, whether to propagate to nested schemas or not, etc. Then there's the difficulty in naming and documenting to meet user expectations (what does it mean for a schema to be "required"?). This is a maintenance burden we are not willing to take.

heckad:
You have `required` in the `Field` class. It means that the `Raise a ValidationError if the field value is not supplied during deserialization.` I propose to add the ability for specifying it in the constructor of the schema. And, I think it need be added to `load`/`dump` and `Meta` too. If it is possible to specify at the field level, then why should one forget about the possibility of specification elsewhere? If something I am ready to do it.

sloria:
"required" makes intuitive sense at the field level but not at the schema level. It'd be too easy to misundertand the semantics of `MySchema(required=True)`. Furthermore, it's not clear on first glance what this should do:

```python
class MySchema(Schema):
    foo = fields.Str(required=False)
    bar = fields.Str()

MySchema(required=True).load({})
```

Like I said, we're not willing to take the maintenance cost of adding this, especially when there is valid workaround using simple inheritance.

heckad:
[...]

PR #1297, opened 2019-07-10
Implementation required_fields

heckad:
[Propouse](https://github.com/marshmallow-code/marshmallow/issues/1280#issuecomment-509678660)

Motivation:
Sometimes need to specify which field must be load without modifying the field description. For example, schema generated by marshmallow-sqlalchemy or django-rest-marshmallow. This pull request adds this feature. 

For example, you have a session model with fields: id(integer), table_id(integer), user_id(integer), comment(Text), mark(integer). You want that for post requests model create only if a request contains table_id, user_id and excluded id field and for put request data should contain id field. Now for that use case need create to schemas.

```
class CreateSessionSchema(ma.ModelSchema):
    class Meta(BaseMeta):
        model = Session
        exlude = ["id"]
        required_fields= ["table_id", "user_id"]

class UpdateSessionSchema(ma.ModelSchema):
    class Meta(BaseMeta):
        model = Session
        required_fields = ["id"]
```
 
Now if data in a request not contained fields from required_fields will raise the exception: ValidationError.

P.S: Where to need to add documentation and tests for it?

### DIFF
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -665,7 +665,12 @@ class BaseSchema(base.SchemaABC):
                 raw_value = data.get(field_name, missing)
                 if raw_value is missing:
                     if field_name in self.load_necessary:
-                        field_obj.fail("required")
+                        try:
+                            field_obj.fail("required")
+                        except ValidationError as err:
+                            error_store.store_error(err.messages, field_name=field_name, index=index)
+                            continue
+
                     # Ignore missing field if we're allowed to.
                     if partial is True or (
                         partial_is_collection and attr_name in partial
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -362,7 +362,7 @@ class BaseSchema(base.SchemaABC):
         dump_only=(),
         partial=False,
         unknown=None,
-        load_necessary=(),
+        load_necessary=()
     ):
         # Raise error if only or exclude is passed as string, not list of strings
         if only is not None and not is_collection(only):
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -668,7 +668,9 @@ class BaseSchema(base.SchemaABC):
                         try:
                             field_obj.fail("required")
                         except ValidationError as err:
-                            error_store.store_error(err.messages, field_name=field_name, index=index)
+                            error_store.store_error(
+                                err.messages, field_name=field_name, index=index
+                            )
                             continue
 
                     # Ignore missing field if we're allowed to.
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -218,6 +218,7 @@ class SchemaOpts:
         self.dump_only = getattr(meta, "dump_only", ())
         self.unknown = getattr(meta, "unknown", RAISE)
         self.register = getattr(meta, "register", True)
+        self.load_necessary = getattr(meta, "load_necessary", ())
 
 
 class BaseSchema(base.SchemaABC):
@@ -344,6 +345,8 @@ class BaseSchema(base.SchemaABC):
             class registry. Must be `True` if you intend to refer to this `Schema`
             by class name in `Nested` fields. Only set this to `False` when memory
             usage is critical. Defaults to `True`.
+        - ``load_necessary``:  Tuple or list of fields of necessarily include in the
+            deserialization result
         """
 
         pass
@@ -358,7 +361,8 @@ class BaseSchema(base.SchemaABC):
         load_only=(),
         dump_only=(),
         partial=False,
-        unknown=None
+        unknown=None,
+        load_necessary=(),
     ):
         # Raise error if only or exclude is passed as string, not list of strings
         if only is not None and not is_collection(only):
@@ -375,6 +379,7 @@ class BaseSchema(base.SchemaABC):
         self.dump_only = set(dump_only) or set(self.opts.dump_only)
         self.partial = partial
         self.unknown = unknown or self.opts.unknown
+        self.load_necessary = set(self.opts.load_necessary) | set(load_necessary)
         self.context = context or {}
         self._normalize_nested_options()
         #: Dictionary mapping field_names -> :class:`Field` objects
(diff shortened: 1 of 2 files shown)


heckad:
@sloria?

sloria:
Please do not ping so soon after sending your PR. We (the maintainers) are working on this in our free time, so please respect that.

We're currently focused on the remaining few changes for 3.0 stable so we've put a hold on reviewing/merging new features until the final release.

heckad:
@sloria? 3 version has already been released.

sloria:
After giving this some thought, I still think this is unnecessary API surface. marshmallow-sqlalchemy already provides a way to override autogenerated fields.

```python
class SessionSchema(ma.ModelSchema):
    table_id = field_for(Session, "table_id", required=True)
    user_id = field_for(Session, "user_id", required=True)
    
    class Meta(BaseMeta):
        model = Session

create_schema = SessionSchema(exclude=('id', )
```

You can also override these on the ORM models:

```python
class Session(Model):
    table_id = Column(..., info={"marshmallow": {"required": True}})
```

See https://marshmallow-sqlalchemy.readthedocs.io/en/latest/recipes.html#overriding-generated-fields

For updates, you can use partial:

```python
schema = SessionSchema()
# make all fields except 'id' optional
partial_fields = schema.fields.keys() - {'id'}
schema.load(..., partial=partial_fields)
```

heckad:
So with the rest of the fields is the same. Why is there an opportunity for them, but in this case this is already too much? Also, why do you persistently offer to patch a sqlalchemy model? It's terrible. This will not allow changing the validator if necessary. In the first case need to create a special schema for any endpoint, but this can leads to an error. For example register endpoint. Now I add resources using one schema. 
```
    app.api.add_resource(resources_factory("ItemListEndpoints", Item, ItemSchema), "/items")
    app.api.add_resource(resources_factory("StoreListEndpoints", Store, StoreSchema), "/stores")
    app.api.add_resource(resources_factory("CityListEndpoints", City, CitySchema), "/cities")
    app.api.add_resource(resources_factory("CuisineListEndpoints", Cuisine, CuisineSchema), "/cuisines")
    app.api.add_resource(resources_factory("CountryListEndpoints", Country, CountrySchema), "/countries")
    app.api.add_resource(resources_factory("TableListEndpoints", Table, TableSchema), "/tables")
    app.api.add_resource(resources_factory("StoreRoomListEndpoints", StoreRoom, StoreRoomSchema), "/store_rooms")
    app.api.add_resource(resources_factory("OrderListEndpoints", Order, OrderSchema), "/orders")
    app.api.add_resource(resources_factory("OrderItemListEndpoints", OrderItem, OrderItemSchema), "/order_items")
```
In `resources_factory` schema specify for `GET,POST,PUT,DELETE` methods throw constructor, but if use your solution need send `CreateItemSchema, UpdateItemSchema,ResultItemSchema, and this for all endpoints. It lead to mistakes with copy and not modify classes. Because need that this feature is available through the constructor. It simplyfiy code and reduces errors.

sloria:
> Also, why do you persistently offer to patch a sqlalchemy model? It's terrible. This will not allow changing the validator if necessary.

Not sure what you mean; the dictionary just gets passed as keyword arguments to the generated fields, so you could include `validate` if you needed to.

> In resources_factory schema specify for GET,POST,PUT,DELETE methods throw constructor, but if use your solution need send `CreateItemSchema, UpdateItemSchema,ResultItemSchema, and this for all endpoints. It lead to mistakes with copy and not modify classes

It is still possible to use a single schema per resource. In fact, we're doing this at work on a regular basis (using Flask-RESty). Re-usability becomes a bit simpler when you use class-based views. See https://flask-resty.readthedocs.io/en/latest/guide.html#views . The `self.update(id, partial=True)` just passes `partial` to `schema.load`.

heckad:
> Not sure what you mean; the dictionary just gets passed as keyword arguments to the generated fields, so you could include `validate` if you needed to.

The project consists of 5 microservices and the database schema is in the `common` repository. It rummaged between all microservices and moderated the dba team. They don't want to add any code to models only for one service used marshmallow-sqlalchemy.

> It is still possible to use a single schema per resource. In fact, we're doing this at work on a regular basis (using Flask-RESty). Re-usability becomes a bit simpler when you use class-based views. See https://flask-resty.readthedocs.io/en/latest/guide.html#views . The `self.update(id, partial=True)` just passes `partial` to `schema.load`.

Unfortunately, the project has its own custom endpoint generator with its own logic and we only need to add validation for input data. Therefore, this solution does not suit us.  And still, it’s not clear how to say what fields must be in request. (A list of fields is also generated in the factory.)

heckad:
Do you understand why we need this feature? Now we use the fork and we can't easily upgrade to new versions. @sloria, please adds this feature.

sloria:
No need for a fork--you can achieve this feature using a custom base schema and options class:

```python
from marshmallow import Schema as BaseSchema, SchemaOpts as BaseSchemaOpts


class SchemaOpts(BaseSchemaOpts):
    def __init__(self, meta, **kwargs):
        super().__init__(meta, **kwargs)
        self.required_fields = getattr(meta, "required_fields", set())


class Schema(BaseSchema):
    OPTIONS_CLASS = SchemaOpts

    def __init__(self, *, required_fields=set(), **kwargs):
        self.required_fields = required_fields or self.opts.required_fields
        super().__init__(**kwargs)

        super().on_bind_field(field_name, field_obj)
        if field_name in self.required_fields:
            field_obj.required = True


# -----------------------------------------------------------------------------

from marshmallow import fields, ValidationError


class MySchema(Schema):
    id = fields.Str()
    name = fields.Str()

    class Meta:
        required_fields = {'id'}


try:
    MySchema().load({})
except ValidationError as error:
    print(error.messages)  # {'id': ['Missing data for required field.']}

```

heckad:
Goog solution. thanks.  But I don't understand why did you forget about this option (specification required fields throw constructor and Meta)?

sloria:
My guess is YAGNI ("You Ain't Gonna Need It"). Given that `partial` meets the 90% use case, we decided not increase the API surface further than necessary.

PR #1298, opened 2019-07-11
Fix data_key with dotted path

davidlouis:
I was expecting that I would be able to use `data_key` like so:
`bar = fields.Str(data_key="data.bar")`

which would in turn deserialize the following data such that bar == "paths!"
`data = {"foo": "dotted", "data": {"bar": "paths!"}}`

### DIFF
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -1178,6 +1178,18 @@ class TestSchemaDeserialization:
         assert result["email"] == "foo@bar.com"
         assert "years" not in result
 
+    def test_deserialize_dotted_path_with_data_key_param(self):
+        class MySchema(Schema):
+            foo = fields.Str(data_key="foo")
+            bar = fields.Str(data_key="data.bar")
+
+        sch = MySchema(unknown=EXCLUDE)
+        data = {"foo": "dotted", "data": {"bar": "paths!"}}
+
+        ret = sch.load(data)
+        assert ret["foo"] == "dotted"
+        assert ret["bar"] == "paths!"
+
     def test_deserialize_with_dump_only_param(self):
         class AliasingUserSerializer(Schema):
             name = fields.String()
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -1180,15 +1180,29 @@ class TestSchemaDeserialization:
 
     def test_deserialize_dotted_path_with_data_key_param(self):
         class MySchema(Schema):
-            foo = fields.Str(data_key="foo")
-            bar = fields.Str(data_key="data.bar")
+            foo = fields.Str()
+            bar = fields.Str(data_key="nested.bar")
 
         sch = MySchema(unknown=EXCLUDE)
-        data = {"foo": "dotted", "data": {"bar": "paths!"}}
+        data = {"foo": "unnested data", "nested": {"bar": "nested data"}}
 
         ret = sch.load(data)
-        assert ret["foo"] == "dotted"
-        assert ret["bar"] == "paths!"
+        assert ret["foo"] == "unnested data"
+        assert ret["bar"] == "nested data"
+
+    def test_deserialize_dotted_path_validation(self):
+        class MySchema(Schema):
+            bar = fields.Integer(data_key="data.bar")
+
+        sch = MySchema(unknown=EXCLUDE)
+        data = {"data": {"bar": "misplaced string"}}
+
+        with pytest.raises(ValidationError) as excinfo:
+            sch.load(data)
+        errors = excinfo.value.messages
+        # The validation should reference the dotted path location in the
+        # source data, rather than just "bar"
+        assert errors["data.bar"] == ["Not a valid integer."]
 
     def test_deserialize_with_dump_only_param(self):
         class AliasingUserSerializer(Schema):
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -132,3 +132,4 @@ Contributors (chronological)
 - Kostas Konstantopoulos `@kdop <https://github.com/kdop>`_
 - Stephen J. Fuhry `@fuhrysteve <https://github.com/fuhrysteve>`_
 - `@dursk <https://github.com/dursk>`_
+- David Louis `@davidlouis https://github.com/davidlouis`_
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -132,4 +132,4 @@ Contributors (chronological)
 - Kostas Konstantopoulos `@kdop <https://github.com/kdop>`_
 - Stephen J. Fuhry `@fuhrysteve <https://github.com/fuhrysteve>`_
 - `@dursk <https://github.com/dursk>`_
-- David Louis `@davidlouis https://github.com/davidlouis`_
+- David Louis `@davidlouis <https://github.com/davidlouis>`_
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -492,7 +492,7 @@ class BaseSchema(base.SchemaABC):
             ]
             self._pending = False
             return ret
-        items = []
+        ret = dict_class()  # empty dict or OrderedDict
         for attr_name, field_obj in fields_dict.items():
             if getattr(field_obj, "load_only", False):
                 continue
@@ -507,8 +507,9 @@ class BaseSchema(base.SchemaABC):
             )
             if value is missing:
                 continue
-            items.append((key, value))
-        ret = dict_class(items)
+
+            set_value(ret, key, value)
+
         return ret
 
     def dump(self, obj, *, many=None):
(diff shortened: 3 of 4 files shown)


tests/test_deserialization.py
```diff
@@ -1178,6 +1178,18 @@ class AliasingUserSerializer(Schema):
         assert result["email"] == "foo@bar.com"
         assert "years" not in result
 
+    def test_deserialize_dotted_path_with_data_key_param(self):
+        class MySchema(Schema):
+            foo = fields.Str(data_key="foo")
```
lafrech: `data_key` is equal to field name

tests/test_deserialization.py
```diff
@@ -1178,6 +1178,18 @@ class AliasingUserSerializer(Schema):
...
+            bar = fields.Str(data_key="data.bar")
+
+        sch = MySchema(unknown=EXCLUDE)
```
lafrech, line 1186: Why do we need `EXCLUDE`? (Not sure why we need it in test above either.)

tests/test_deserialization.py
```diff
@@ -1178,6 +1178,18 @@ class AliasingUserSerializer(Schema):
...
+        data = {"foo": "dotted", "data": {"bar": "paths!"}}
+
+        ret = sch.load(data)
+        assert ret["foo"] == "dotted"
+        assert ret["bar"] == "paths!"
+
```
lafrech, line 1206: Nitpick: `dotted` is a bit misleading. At first I thought it was meant to suggest that the `foo` field is accessed with dotted notation, while it is the other one. It took me a few reads to realize it is meant to be read as "dotted" + "paths!" = "dotted paths!".

lafrech:
Thanks.

Looks good to me.

Should we also test an error case to check the error is correctly reported?

tests/test_deserialization.py
davidlouis: If I don't use `EXCLUDE` the schema wants to have a `data`  field to load the top level json element "data".
In the real world case where I want to use this I'm happy with using `EXCLUDE` because the top level keys don't mean anything.

I think it would get complicated to support `unknown=RAISE` and schemas reading values from dotted paths, at least from my limited understanding!

tests/test_deserialization.py
davidlouis: Good point it is redundant here, I simply wanted the test to include a key with a dotted path and a  key without. 

I can make a small change her to this data key so it is different, and a change so the non dotted path field doesn't have the word "dotted" in the test case example as per your comment below! I can probably get this done some time tomorrow.

tests/test_deserialization.py
lafrech: Oh, of course. I understand, now. Thanks.

(Answer to myself: in the test above, `EXCLUDE` was added to exclude `years` from the input.)

tests/test_deserialization.py
lafrech: Good.

> I can make a small change her to this data key so it is different

Yes. But this is already tested in the test above. However, maybe you can test a multiple dot path like "data.foo/bar".

davidlouis:
> Thanks.
> 
> Looks good to me.
> 
> Should we also test an error case to check the error is correctly reported?

Can you elaborate on your idea for this additional test? Maybe it's just late but I'm not sure what your line of thinking is there!

lafrech:
Maybe it is useless. I meant trigger an error (e.g. wrong type) in data.bar to check that it is reported as {"data": {"bar": error }} and not as {"data": error}. I hope this is clearer.

deckar01:
This PR currently does not handle dumping dotted keys. One thing to consider is that all the places that dotted delimiters are documented for use are on the schema. Those features are all for referencing nested fields to augment an existing schema. The `Pluck` field should support this use case and handle dumping back to the nested structure.

The `Pluck` field is not showing up in the docs for some reason. Edit: Fixed in https://github.com/marshmallow-code/marshmallow/pull/1300.

https://github.com/marshmallow-code/marshmallow/blob/8b0fe22a/src/marshmallow/fields.py#L516-L527

davidlouis:
Hi @deckar01 - Thanks for the feedback.
I did investigate the use of `Pluck` to achieve this but was unable to get it working. 
The following test errors but should highlight what I was trying to achieve using `Pluck`

  ```
  def test_pluck_deserialize(self):

        class ChildSchema(Schema):
            foo = fields.Integer()

        class MySchema(Schema):
            foo = fields.Pluck(ChildSchema, data_key="data", field_name="foo")

        sch = MySchema()
        data = {
            "data": {"foo": 1}
        }

        ret = sch.load(data)
        assert ret == {"foo": 1}
```

Maybe I'm missing something and using this incorrectly. Ideally I'd also like to be able to avoid specifying the `ChildSchema` at all for such a simple use case of grabbing a single bit of data from a nested json structure. Can you give me any pointers to the way you're intending `Pluck` to be used here?

Finally, I did have a quick look at .dump() and it did seem to correctly serialize the dotted path correctly into a nested json structure. Because this appeared to already be working I left it at that.

**Comment by davidlouis** [comment]
@lafrech - Oh! I understand now. Great idea for another test :+1:

**Comment by deckar01** [comment]
This use case is inverted from what Pluck does. You want to load from a nested structure into a flat structure. Sorry for the confusion.

```py

class MySchema(Schema):
    foo = fields.Integer(data_key='data.foo')

sch = MySchema()
data = {'foo': 1}

ret = sch.dump(data)
assert ret == {'data': {'foo': 1}}
# {'data.foo': 1} != {'data': {'foo': 1}}
```

I don't think outputting a dotted key is desirable, because it can't be loaded again.

davidlouis:
You're right @deckar01, the serialization behaviour of the dotted path data key is inconsistent with the change i made for deserialization. I'm not sure why I thought it was working already

I've made an extra commit that changed this behaviour and makes (what is essentially) the test you wrote above pass.

I think with what we've got now we can use a Schema field like this and get sensible, consistent serialization and deserialization

sloria:
Are there valid cases where you would want the input or output key to be a have a "."? As @deckar01 pointed out, we currently only use dot-delimited strings on the schema, not fields. I'm a bit nervous we might break legitimate use cases here.

IIRC, Some config formats like .ini and .toml formats use dot-delimited keys.

sloria:
You can achieve the desired behavior with a custom field.

```python
from marshmallow.utils import get_value, set_value


class Reach(fields.Field):
    def __init__(self, inner, path, **kwargs):
        super().__init__(**kwargs)
        self.inner = inner
        self.path = path

        val = get_value(value, self.path)
        return self.inner.deserialize(val, **kwargs)

        val = self.inner._serialize(value, attr, obj, **kwargs)
        ret = {}
        set_value(ret, self.path, val)
        return ret


class MySchema(Schema):
    foo = fields.Str()
    bar = Reach(fields.Int(), data_key="data", path="bar.baz")


data = {"foo": "dotted", "data": {"bar": {"baz": "42"}}}
schema = MySchema()
loaded = schema.load(data)
print("loaded: ", loaded)
print("dumped: ", schema.dump(loaded))

```

```
loaded:  {'foo': 'dotted', 'bar': 42}
dumped:  {'foo': 'dotted', 'data': {'bar': {'baz': 42}}}
```

deckar01:
My confusion with the behavior of `Pluck` concerns me a little. Currently `Pluck = flat -> nested -> flat`, but it is easily confused with this use case `nested -> flat -> nested`. `Pluck` describes the operation during serialization which is similar to `Function` and `Method`, but it is really the opposite operation during deserialization. Would it make sense to provide a field that performs the opposite operation? `Nest`?

---

Side note: (way off topic) Some of the recent discussion about multiple schemas per model have got me thinking about what it would look like to decouple field mapping from schema declaration. You only ever need one deserialized data structure and one way to validate it, but there are potentially multiple ways you need to load into or dump from it (API versioning, nested views, etc).

<details>
<summary>Example</summary>

```py
class AuthorSchema(Schema):
    name = fields.Str()

class BookSchema(Schema):
    title = fields.Str()
    length = fields.Int()
    author = fields.Nested(AuthorSchema)

class BookRequest(Interface):
    length = Pluck('stats', 'length')
    author = {'name': 'author'}

schema = UserSchema()
interface = UserRequest()
data = {'title': 'Example', 'author': 'Arthur', 'stats': {'length': 256}}
schema.load({'username': 'example'}, interface)
```
</details>

sloria:
Closing for now, because I don't think `data_key` is the right approach for the use case. Let's continue discussion on #1315

davidlouis:
@deckar01 - I agree! I originally tried for sometime to use `Pluck` because it sounded like precisely the type of behaviour I was after here. I also looked at `Nested` (again sounded promising) but wasn't actually what I was after!

@sloria - It's a very good point about the use case where you have dots in your actual keys. Thanks for the custom field ID. I had a not so elegant backup solution similar to this. I think a use case like this would be nice to have 'out of the box', using the base field classes directly. I'll use this custom field for now to give me a feel on this looks and feels using this on my models

I'll keep a keen eye on the discussion in the follow up. :+1:

PR #1301, opened 2019-07-14
Add Field.make_error to fix `raise from` use cases

sloria:
`raise self.fail(...) from exc` was not working as expected.
This adds a `make_error` method for cases where you need to
`raise from`.

Confirmed this is working with:

```python

field = fields.DateTime()

field.deserialize('foo')
```

Before:

```
  File "/Users/sloria/projects/marshmallow-code/marshmallow/src/marshmallow/fields.py", line 1103, in _deserialize
    return func(value)
  File "/Users/sloria/projects/marshmallow-code/marshmallow/src/marshmallow/utils.py", line 198, in from_iso_datetime
    raise ValueError("Not a valid ISO8601-formatted datetime string")
ValueError: Not a valid ISO8601-formatted datetime string

During handling of the above exception, another exception occurred:

  File "t.py", line 5, in <module>
    field.deserialize('foo')
  File "/Users/sloria/projects/marshmallow-code/marshmallow/src/marshmallow/fields.py", line 301, in deserialize
    output = self._deserialize(value, attr, data, **kwargs)
  File "/Users/sloria/projects/marshmallow-code/marshmallow/src/marshmallow/fields.py", line 1105, in _deserialize
    raise self.fail("invalid", input=value, obj_type=self.OBJ_TYPE) from exc
  File "/Users/sloria/projects/marshmallow-code/marshmallow/src/marshmallow/fields.py", line 249, in fail
    raise ValidationError(msg)
marshmallow.exceptions.ValidationError: Not a valid datetime.
```

After:

```
  File "/Users/sloria/projects/marshmallow-code/marshmallow/src/marshmallow/fields.py", line 1106, in _deserialize
    return func(value)
  File "/Users/sloria/projects/marshmallow-code/marshmallow/src/marshmallow/utils.py", line 198, in from_iso_datetime
    raise ValueError("Not a valid ISO8601-formatted datetime string")
ValueError: Not a valid ISO8601-formatted datetime string

The above exception was the direct cause of the following exception:

  File "t.py", line 5, in <module>
    field.deserialize('foo')
  File "/Users/sloria/projects/marshmallow-code/marshmallow/src/marshmallow/fields.py", line 304, in deserialize
    output = self._deserialize(value, attr, data, **kwargs)
  File "/Users/sloria/projects/marshmallow-code/marshmallow/src/marshmallow/fields.py", line 1110, in _deserialize
    ) from exc
marshmallow.exceptions.ValidationError: Not a valid datetime.
```

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -234,10 +234,7 @@ class Field(FieldABC):
         if errors:
             raise ValidationError(errors, **kwargs)
 
-    # Hat tip to django-rest-framework.
-    def fail(self, key, **kwargs):
-        """A helper method that simply raises a `ValidationError`.
-        """
+    def make_error(self, key: str, **kwargs) -> ValidationError:
         try:
             msg = self.error_messages[key]
         except KeyError as exc:
@@ -246,7 +243,13 @@ class Field(FieldABC):
             raise AssertionError(msg) from exc
         if isinstance(msg, (str, bytes)):
             msg = msg.format(**kwargs)
-        raise ValidationError(msg)
+        return ValidationError(msg)
+
+    # Hat tip to django-rest-framework.
+    def fail(self, key: str, **kwargs):
+        """A helper method that simply raises a `ValidationError`.
+        """
+        raise self.make_error(key=key, **kwargs)
 
     def _validate_missing(self, value):
         """Validate missing values. Raise a :exc:`ValidationError` if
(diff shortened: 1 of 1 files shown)


lafrech:
At this point, `fail` is almost useless, isn't it? Calling `raise self.make_error` is not much more trouble than calling `self.fail` and we'd get a more unified interface by removing `fail`, better than two different interfaces depending on whether we want to use `from`. But it would break all codes calling `fail`...

Another option would be to add a second positional arg in `fail` to pass an exception. I don't find this option fantastic either. Just an idea.

```py
    # Hat tip to django-rest-framework.
    def fail(self, key: str, exc=None, **kwargs):
        """A helper method that simply raises a `ValidationError`.
        """
        [...]
        if exc:
            raise ValidationError(msg) from exc
        else:
            raise ValidationError(msg)
```

(Also, the `# Hat tip to django-rest-framework.` ends up on the almost empty `fail` method in your PR. Maybe it deserves to be on `make_error`. Or maybe you put `fail` above `make_error` so the comment applies implicitly to both.)

sloria:
> At this point, fail is almost useless, isn't it? 

Maybe. It's part of the public API, and it doesn't seem worth the breaking change to remove it.

> Another option would be to add a second positional arg in fail to pass an exception. 

I actually went that way first, but it doesn't seem better. Without looking at the implementation of `.fail` it's not intuitive what passing `exc` will do.

> Also, the # Hat tip to django-rest-framework.

I guess that "hat tip" was for the `.fail` API and its name, not necessarily the implementation (though the implementation is near identical), which is why I left it where it is.

lafrech:
> I actually went that first, but it doesn't seem better. Without looking at the implementation of .fail it's not intuitive what passing exc does.

I never heard about the `from` syntax before it was added here. Documenting the `exc` parameter in `fail` docstring would make it clear what purpose it serves (we could call it `from_exc` if it helps) and would allow people to discover the feature more easily.

In this PR's implementation, it may not be clear at first why there are both `fail` and `make_error`, until you find examples of `make_error` calls that illustrate what it is here for.

Granted, the role of marshmallow is not to help users discover Python purposes. And yes, I think I understand what you mean: the `fail("message", exception)` syntax does not mimic `raise MyException("message") from exception`.

I don't mind either way. I find no solution obviously better and really satisfying.

sloria:
Mostly, using a separate method is more flexible/future-proof. Aside from enabling `raise from`, it allows custom field creators to modify the `ValidationError` if they want to.

If we want, we can deprecate `self.fail` in the future and just use `make_error`. 

I've no strong preference either way, either 😅

lafrech (approved):
Alright, let's go with this.

PR #1306, opened 2019-07-17
Fast List(Nested)

deckar01:
Allow `List(Nested)` to hit the `schema.dump(many)`/`schema.load(many)` fast paths that `Nested(many)` uses.

Helps #779


src/marshmallow/fields.py
```diff
@@ -606,9 +607,12 @@ def _bind_to_schema(self, field_name, schema):
     def _serialize(self, value, attr, obj, **kwargs):
         if value is None:
             return None
-        if utils.is_collection(value):
-            return [self.inner._serialize(each, attr, obj, **kwargs) for each in value]
-        return [self.inner._serialize(value, attr, obj, **kwargs)]
+        if not utils.is_collection(value):
+            value = [value]
```
lafrech: Unrelated to this PR but I'm wondering the point of supporting this.

Removing line 611 in original code only breaks the test testing specifically this:

```py
class DateTimeList:
    def __init__(self, dtimes):
        self.dtimes = dtimes

def test_datetime_list_serialize_single_value(self):
    obj = DateTimeList(dt.datetime.utcnow())
    field = fields.List(fields.DateTime)
    result = field.serialize("dtimes", obj)
    assert len(result) == 1
    assert type(result[0]) == str
```

(Removal PR: https://github.com/marshmallow-code/marshmallow/pull/1307)

src/marshmallow/fields.py
```diff
@@ -487,29 +487,34 @@ def _nested_normalized_option(self, option_name):
...
+        if many and not utils.is_collection(value):
             self.fail("type", input=value, type=value.__class__.__name__)
 
-    def _load(self, value, data, partial=None):
+    def _load(self, value, data, partial=None, many=False):
         try:
-            valid_data = self.schema.load(value, unknown=self.unknown, partial=partial)
+            many = self.many or many
+            valid_data = self.schema.load(
+                value, unknown=self.unknown, partial=partial, many=many
```
deckar01: I found it odd that `load()` wasn't actually passing `many` to `schema.load()` before, but `dump()` was. It turns out `Nested` sets `schema.many` when it builds its own schema instance, so the `many` in `dump()` was redundant before, but it is useful on both now.

sloria (approved):

PR #1307, opened 2019-07-17
Don't let List serialize a single value as a list

lafrech:
No big deal but I don't see the point of supporting this.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,8 +7,11 @@ Changelog
 Features:
 
 - *Backwards-incompatible*: ``DateTime`` does not affect timezone information
-  on serialization and deserialization.
-- Add ``NaiveDateTime`` and ``AwareDateTime`` to enforce timezone awareness.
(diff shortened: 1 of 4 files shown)


deckar01:
I can't find any place that it's documented.

sloria (approved):
I'm good with the removal. My guess is this was old copy-pasta from [flask-restful's behavior](https://github.com/flask-restful/flask-restful/blob/f6921bbcc3a392cd92a1381357b5efa069cc55ea/flask_restful/fields.py#L205). DRF [doesn't do it, though](https://github.com/encode/django-rest-framework/blob/044252af1c09b0da6899ffc3e88f192721cd38e3/rest_framework/fields.py#L1660).

Doesn't seem like good behavior, in any case.  IMO we should be trying to remove as many unnecessary type-checks during serialization as possible.

@lafrech Can you document in `updating` and `CHANGELOG` then merge?

🚢 🚢 🚢 🇮🇹

PR #1308, opened 2019-07-17
Adjusted documentation for Regexp validator method

macdonaldezra:
I added a brief comment indicating that the Regexp method uses re's match() function. Knowing that match() is used for the validator could help newcomers write better regular expression validators.


sloria:
Thanks!

PR #1309, opened 2019-07-18
Small perf improvements

sloria:
Just a few free micro-optimizations. Results in ~3% improvement in dump speed, using `tox -e benchmark`.

* Memoize _has_processors
* Don't make intermediate list
* Duck-type iterables


lafrech:
LGTM.

TIL about ``functools.lru_cache``.

PR #1316, opened 2019-07-20
Validate missing field values when missing option is set

kissgyorgy:
When the `missing` `Field` option was set, it's value (or it's return value in
case it was a callable) was not validated. This made mistakes possible when the
implementation of the callable returned a wrong value or a wrong value for the
`missing` option was defined.

This is useful when using an external or third-party package function for the
missing option, and using a validator at the same time.  
For example, using a function for the missing value which returns an
environment variable makes it possible to validate environment variables.

As the readability of this function is already difficult, it has been optimized 
for speed with the nested `if` instead of more `elif`-s.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -302,12 +302,16 @@ class Field(FieldABC):
         # Validate required fields, deserialize, then validate
         # deserialized value
         self._validate_missing(value)
+
         if value is missing_:
-            _miss = self.missing
-            return _miss() if callable(_miss) else _miss
-        if getattr(self, "allow_none", False) is True and value is None:
+            if self.missing is missing_:
+                return missing_
+            output = self.missing() if callable(self.missing) else self.missing
+        elif value is None and getattr(self, "allow_none", False):
             return None
-        output = self._deserialize(value, attr, data, **kwargs)
+        else:
+            output = self._deserialize(value, attr, data, **kwargs)
+
         self._validate(output)
         return output
 
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -1321,6 +1321,23 @@ class TestSchemaDeserialization:
         assert result["name"] == "Mick"
         assert result["birthdate"] == bdate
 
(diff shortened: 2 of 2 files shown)


sloria:
I'm not sure about this. Validating a value that is entered by the developer--not by clients--feels a bit odd. I don't think it's worth the performance penalty to validate something that should be validated by the developer.

sloria:
Closing for now, but feel free to continue discussion on here.

kissgyorgy:
> I'm not sure about this. Validating a value that is entered by the developer--not by clients--feels a bit odd. I don't think it's worth the performance penalty to validate something that should be validated by the developer.

I agree, but there are some use cases which make sense, for example, getting a value from a dynamic location, when that value come from a user input like a database, or an environment variable.

The other alternative would be to make it optional, for example with a `validate_missing` argument.

Here is what I'm currently doing to work around this:
```python

from flask.helpers import get_env
from marshmallow.utils import missing

class ValidateMissingField(fields.Field):
    def __init__(self, field_cls, *, missing, **kwargs):
        super().__init__(**kwargs)
        self._wrapped_field = field_cls(**kwargs)
        self._missing = missing

    def serialize(self, *args, **kwargs):
        return self._wrapped_field.serialize(*args, **kwargs)

    def deserialize(self, *args, **kwargs):
        output = self._wrapped_field.deserialize(*args, **kwargs)
        if output is missing:
            output = self._missing() if callable(self._missing) else self._missing
            self._validate(output)
        # if not missing, it has been already validated
        return output


class ConfigSchema(Schema):
    env = ValidateMissingField(
        fields.String,
        attribute="ENV",
        validate=OneOf(["production", "development"]),
        missing=get_env,
    )
```

This is a lot of code for such a simple use case.

Also, I think what the following code does is not intuitive at all:
```python
    env = ValidateMissingStringField(
        attribute="ENV", missing=get_env, validate=OneOf(["production", "development"])
    )
```
because the `OneOf` validator is simply ignored. I expected it to work.

sloria:
Let's continue discussion of this in #1318 , for visibility.

PR #1317, opened 2019-07-21
Add `require_tld` parameter to `Url` field doc

atmo:
Docstring for `Url` field is missing description for its `require_tld` parameter.


sloria:
Good catch. Thanks!

PR #1322, opened 2019-07-21
Add Schema.from_dict for generating schemas from dicts

### LINKED ISSUE #1312: Add a way to generate a Schema from a dictionary
Generating schemas at runtime is a fairly common use case that I've run into a handful of times at this point ([webargs](https://github.com/marshmallow-code/webargs/blob/de061e037285fd08a42d73be95bc779f2a4e3c47/src/webargs/core.py#L50-L63), [environs](https://github.com/sloria/environs/blob/91b32a0e4bdf40281fab3978368164b27ce65e39/environs.py#L83-L95) , form builders...). 

It might make sense to incorporate this into marshmallow core.

## Proposed API

```python
>>> MyDynamicSchema = Schema.from_dict({"foo": fields.Str()})
>>> MyDynamicSchema
<class 'marshmallow.schema.GeneratedSchema'>
```

```python
class AlbumSchema(Schema):
     artist = fields.Nested(Schema.from_dict({"name": fields.Str()})
```

sloria:
close #1312

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,7 @@ Features:
 
 - Optimize ``List(Nested(...))`` (:issue:`779`).
 - Minor performance improvements and cleanup (:pr:`1328`).
+- Add ``Schema.from_dict`` (:issue:`1312`).
 
 Deprecations/Removals:
 
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -41,6 +41,20 @@ Create a schema by defining a class with variables mapping attribute names to :c
 
     For a full reference on the available field classes, see the :ref:`API Docs <api_fields>`.
 
+Creating Schemas From Dictionaries
+----------------------------------
+
+You can create a schema from a dictionary of fields using the `from_dict <marshmallow.Schema.from_dict>` method.
+
+.. code-block:: python
+
+    from marshmallow import Schema, fields
+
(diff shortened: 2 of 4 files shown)


lafrech (approved):
LGTM.

Maybe another argument could be added to pass meta options in the call to `from_dict` but it can be added later on if someone needs it.

sloria:
I wanted to keep the API surface to a minimum here. If a user needs to customize `Meta`, they can declare a class.

lafrech:
Note: `from_dict` only generates fields for top-level fields in the dict. It does not create `Nested` fields automatically: https://stackoverflow.com/questions/58190523/how-to-use-schema-from-dict-for-nested-dictionaries.

sloria:
It was not the intention to automagically create `Nested` and `List` fields for dicts and lists. I don't think it's necessary nor intuitive, but I'm not closed off to the idea if others think it's the way to go.

hardikkat24:
@sloria Wouldn't it be great for `fields.Nested()` to accept both dictionary and Schema? Thanks!

sloria:
@hardikkat24 That can be achieved quite easily with a subclass of `Nested`. See https://github.com/marshmallow-code/webargs/blob/0357a525118acbc19bc5a1981f6161ee79110579/src/webargs/fields.py#L26-L40

Issue #1303, opened 2019-07-17
Wrong message on type validation errors


cristi23:
## Before going too deep

I'm using marshmallow==2.19.5 and I noticed that there's a release candidate for marshmallow 3 so if this issue has somehow been solved, please ignore the rest. However, the only issue I noticed which seemed like it might have some relevance is this: https://github.com/marshmallow-code/marshmallow/issues/852, but the PR for this doesn't seem to address the problem that I'm having. (PR: https://github.com/marshmallow-code/marshmallow/pull/1067/files).

## Reproducing the issue

Let's consider the following code with 3 schema definitions:

```python
from marshmallow import Schema, fields

class TwoLevelsNestedResource(Schema):
    some_irrelevant_field = fields.String(required=True)


class OneLevelNestedResource(Schema):
    field_a = fields.String(required=True)
    field_b = fields.Integer(required=True)
    field_c = fields.Nested(TwoLevelsNestedResource, required=True)


class MainResource(Schema):
    main_field_a = fields.String(required=True)
    main_field_b = fields.Integer(required=True)
    nested_field = fields.Nested(OneLevelNestedResource, required=True, many=False)
```

Running the following function would result in the data being validated without any errors which is what we want:
```python
def load_correct_data():
    print(MainResource().load({
        'main_field_a': '',
        'main_field_b': 0,
        'nested_field': {
            'field_a': '',
            'field_b': 0,
            'field_c': {
                'some_irrelevant_field': ''
            }
        }
    }).errors)

load_correct_data()
```

Running the next function should print a validation error because the data which is inputted is not correct.

```python
def load_wrong_data_a():
    print(MainResource().load('not what we want').errors)

load_wrong_data_a()
```

As the root level object we should send a dictionary, not a string. Therefore, we should get a type validation error. The problem is that we should get whatever is defined as the default error message for the `fields.Nested` type which is `"Invalid type."`. However, any of the following 2 error messages might be returned at different calls of the same function.

```python
{'_schema': ['Invalid input type.']}
```
```python
{'_schema': ['Invalid type.']}
```
Furthermore, the following code has the same issue, one level deeper in the schema:

```python
def load_wrong_data_b():
    print(MainResource().load(dict(
        main_field_a='',
        main_field_b=0,
        nested_field='again, not what we want',
    )).errors)

load_wrong_data_b()
```
The two possible outputs, alternatively are:

```python
{'nested_field': {'_schema': ['Invalid input type.']}}
```
```python
[...]

PR #1323, opened 2019-07-22
ported fix made by lafrech to correctly handle type errors in 2.x

cristi23:
This should fix https://github.com/marshmallow-code/marshmallow/issues/1303

The solution was initially designed for addressing a different (but similar) issue on version 3.x

The code is therefore very similar to https://github.com/marshmallow-code/marshmallow/pull/857

The tests are all passing.

### DIFF
--- a/src/marshmallow/marshalling.py
+++ b/src/marshmallow/marshalling.py
@@ -10,6 +10,8 @@ and from primitive types.
 
 from __future__ import unicode_literals
 
+import collections
+
 from marshmallow.utils import is_collection, missing, set_value
 from marshmallow.compat import text_type, iteritems
 from marshmallow.exceptions import (
@@ -23,6 +25,7 @@ __all__ = [
 
 # Key used for field-level validation errors on nested fields
 FIELD = '_field'
+
 
 class ErrorStore(object):
 
@@ -76,6 +79,7 @@ class ErrorStore(object):
             # on the ValidationError's data attribute
             value = err.data or missing
         return value
+
 
 class Marshaller(ErrorStore):
     """Callable class responsible for serializing data and storing errors.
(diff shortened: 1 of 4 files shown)


sloria:
Thanks for the PR! I'll take a closer look at this later. In the meatime, would you mind adding a regression test for this?

cristi23:
Sure thing!

cristi23:
Just wanted to let you know that I haven't forgotten about this, but I've been a bit busy and couldn't get to it. I hope I'll manage to do it this week. Sorry for the delay.

cristi23:
I added a couple of tests to cover the issues reported in https://github.com/marshmallow-code/marshmallow/issues/1303. Sorry for the delay!

I should point out that in this form the code will not support overrideable error_messages for type such as in this example:
```python

class Test(Schema):
    foo = fields.String(required=True, error_messages={'type': 'foo must be string'})
```

The output error message of
```python
Test().load({
    'foo': ''
})
```
will always be "Invalid input type." because of how the code hard codes the error message for this particular scenario.

src/marshmallow/marshalling.py
```diff
@@ -242,25 +247,22 @@ def deserialize(self, data, fields_dict, many=False, partial=False,
                         data=ret,
                     )
             return ret
-        if data is not None:
-            partial_is_collection = is_collection(partial)
-            ret = dict_class()
+
+        partial_is_collection = is_collection(partial)
```
sloria: This can be moved to the `else` below; no need to compute if it won't be used.

tests/test_marshalling.py
```diff
@@ -260,3 +260,27 @@ def test_deserialize_fields_with_dump_only_param(self, unmarshal):
         assert 'years' not in result
 
         assert 'always_invalid' not in unmarshal.errors
+
+    def test_deserialize_wrong_type_root_data(self, unmarshal):
+        data = ''
+        fields_dict = {
```
sloria: Nitpick: can just be

```suggestion
        fields_dict = {}
```

sloria (commented):
Thanks for adding the tests. Re: the overrideable error_messages--I don't think that usage was supported even before this change. Or am I missing something?

PR #1351, opened 2019-08-19
pre-commit autoupdate && pre-commit run --all-files

asottile:
mostly just some minor new tricks that `pyupgrade` has learned :)


sloria:
Very cool. Thanks for the heads up. 

I should probably all-repos this =)

PR #1354, opened 2019-08-19
Avoid DeprecationWarning about ABC in Python 3.x

nicktimko:
Python 3.x complains about the `collections.Mapping` vs `collections.abc.Mapping`. There is already a shim for it in `marshmallow.compat`, so switched to using that.

> `/usr/lib/python3.7/site-packages/marshmallow/marshalling.py:253`: `DeprecationWarning`: Using or importing the ABCs from `collections` instead of from `collections.abc` is deprecated, and in 3.8 it will stop working
> ```
>     if not isinstance(data, collections.Mapping):
> ```

Though I poked at the `collections.Mapping` in 3.8 and it said "...is deprecated since Python 3.3,and in 3.9 it will stop working", so it's maybe not as critical, but it removes nagging.  See https://github.com/python/cpython/pull/13409

### DIFF
--- a/src/marshmallow/marshalling.py
+++ b/src/marshmallow/marshalling.py
@@ -10,10 +10,8 @@ and from primitive types.
 
 from __future__ import unicode_literals
 
-import collections
-
 from marshmallow.utils import is_collection, missing, set_value
-from marshmallow.compat import text_type, iteritems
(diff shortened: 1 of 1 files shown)


sloria:
Thanks!

Issue #1286, opened 2019-07-05
_deserialize() got an unexpected keyword argument 'partial'


palango:
I recently updated to 3.0.0rc8 (see https://github.com/raiden-network/raiden/pull/4334), and now run into a stacktrace in one test that I cannot make sense of.

i'm posting this here in the hope that someone can give me a hint on how to proceed in finding the problem.

The stacktrace is the following:
```
raiden/tests/integration/long_running/test_settlement.py::test_batch_unlock[matrix-False-2] Traceback (most recent call last):
  File "src/gevent/greenlet.py", line 716, in gevent._greenlet.Greenlet.run
  File "/Users/paul/Projects/brainbot/raiden/raiden/tests/integration/long_running/test_settlement.py", line 344, in run_test_batch_unlock
    state_change_identifier=HIGH_STATECHANGE_ULID,
  File "/Users/paul/Projects/brainbot/raiden/raiden/storage/restore.py", line 42, in channel_state_until_state_change
    state_change_identifier=state_change_identifier,
  File "/Users/paul/Projects/brainbot/raiden/raiden/storage/wal.py", line 25, in restore_to_state_change
    state_change_identifier=state_change_identifier
  File "/Users/paul/Projects/brainbot/raiden/raiden/storage/sqlite.py", line 788, in get_snapshot_before_state_change
    row.identifier, row.state_change_identifier, self.serializer.deserialize(row.data)
  File "/Users/paul/Projects/brainbot/raiden/raiden/storage/serialization/serializer.py", line 68, in deserialize
    data = DictSerializer.deserialize(json.loads(data))
  File "/Users/paul/Projects/brainbot/raiden/raiden/storage/serialization/serializer.py", line 56, in deserialize
    return schema.load(deepcopy(data))
  File "/Users/paul/Projects/brainbot/raiden/venv/lib/python3.7/site-packages/marshmallow/schema.py", line 732, in load
    data, many=many, partial=partial, unknown=unknown, postprocess=True
  File "/Users/paul/Projects/brainbot/raiden/venv/lib/python3.7/site-packages/marshmallow/schema.py", line 851, in _do_load
    index_errors=self.opts.index_errors,
  File "/Users/paul/Projects/brainbot/raiden/venv/lib/python3.7/site-packages/marshmallow/schema.py", line 686, in _deserialize
    index=index,
[...]

PR #1364, opened 2019-08-26
Add _deserialize v3 upgrading guide

charlax:
As per comment on #1286

### DIFF
--- a/docs/upgrading.rst
+++ b/docs/upgrading.rst
@@ -1041,6 +1041,29 @@ done with no additional dependency.
 ``python-dateutil`` is no longer used by marshmallow.
 
 
+Custom Fields
+*************
+
+To make your custom fields compatible with marshmallow 3, ``_deserialize``
+should accept ``**kwargs``:
+
+.. code-block:: python
+
+    from marshmallow import fields, ValidationError
+    from packaging import version
+
+    # 2.x
+    class MyCustomField(fields.Field):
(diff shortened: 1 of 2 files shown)


docs/upgrading.rst
```diff
@@ -1041,6 +1041,31 @@ done with no additional dependency.
 ``python-dateutil`` is no longer used by marshmallow.
 
 
+Custom fields
+*************
+
+To make your custom fields compatible with marshmallow 3, ``deserialize``
```
lafrech: s/deserialize/_deserialize/

charlax: indeed 🤦‍♂

lafrech:
It could be worth also mentioning that `Nested` subclasses now receive `partial` keyword attribute.

For completeness sake, we should add a note about `**kwargs` in the *3.0.0b19 (2018-10-24)* section of the CHANGELOG. Nice if you do it but not necessary for this PR.

charlax:
Not sure to understand re. `Nested`... would you mind writing the change? I'm not using it that much...

lafrech:
On second thought, it is not really an upgrade issue as it won't break anything. It could be mentioned somewhere in the docs but it is probably an edge case. Let's forget it for now.

I added the note about **kwargs to the CHANGELOG.

lafrech:
Thanks.

charlax:
You're welcome!

Issue #1368, opened 2019-08-28
many is passed through to pre/post_dump/load method even when pass_many=False


sloria:
>Here's a sample project showing my issue:
https://github.com/phrfpeixoto/marshmallow_bug/tree/master

> It works ok o 2.x Branch. I'm holding to that version for now.

_Originally posted by @phrfpeixoto in https://github.com/marshmallow-code/marshmallow/issues/1150#issuecomment-523160159_

sloria:
I'm not sure that this is a bug. In v3, `many` *always* gets passed to `post_dump` methods, and it will be equal to the value of `many` upon calling `dump`.

`pass_many` controls whether your post_dump method receives the original input list or if it handles single items. It does *not* control whether or not the `many` argument passed.

https://marshmallow.readthedocs.io/en/latest/extending.html#passing-many

@phrfpeixoto What use case is broken for you? Is it fixed by adding `**kwargs` to your post_dump method, as suggested [here](https://marshmallow.readthedocs.io/en/latest/upgrading.html#decorated-methods-and-handle-error-receive-many-and-partial)?

---

Sidenote: perhaps `pass_collection` would be a better name than `pass_many`. We should consider this for 4.0.

lafrech:
I'm not sure I have the whole picture.

In the diff I propose in https://github.com/marshmallow-code/marshmallow/issues/1150#issuecomment-525496932, my idea was that since we iterate on the input to pass it to the decorated processor one item at a time, it made sense, intuitively, to pass many=False on each call.

I tried it and it broke no test but I didn't investigate any further.

sloria:
It's debatable whether `many` should equal the "dump-time" value of `many` or always be `False` when `pass_many=False`. I think the current behavior is fine, tbh.

I think the confusion in the OP was that `pass_many=False` would cause `many` to not be passed to the method (which was the 2.x behavior).

lafrech:
OK, I don't know and don't mind either way. I have no example of case where the decorated processor actually uses `many`.

Since it is swallowed by `**kwargs`, it shouldn't make any difference.

The issue is people who upgrade without reading the upgrading docs first (like my colleague just did, just like I would have done myself as well) get the error about `many` and get confused.

sloria:
Yeah. I think renaming `pass_many` to `pass_collection` might clarify things a bit.

phrfpeixoto:
Thanks for replying, both of you.

[...]

PR #1370, opened 2019-08-28
Improving pre- and post-processors documentation

phrfpeixoto:
As discussed on #1368, the documentation for `pass_many` and `many` parameters on the pre- and post-processors is not clear about the behaviour of these parameters.

This PR aims to improve the documentation.

Could not find a 3.0-line branch, so I've branched from the 3.0.1 tag.

### DIFF
--- a/docs/extending.rst
+++ b/docs/extending.rst
@@ -28,18 +28,14 @@ Data pre-processing and post-processing methods can be registered using the `pre
     result = schema.load({"name": "Steve", "slug": "Steve Loria "})
     result["slug"]  # => 'steve-loria'
 
-
 Passing "many"
 ++++++++++++++
 
-By default, despite of the ``many`` parameter value, pre- and post-processing methods receive one object/datum at a time, transparently handling the ``many`` parameter passed to the ``Schema``'s :func:`~marshmallow.Schema.dump`/:func:`~marshmallow.Schema.load` method at runtime.
+By default, pre- and post-processing methods receive one object/datum at a time, transparently handling the ``many`` parameter passed to the ``Schema``'s :func:`~marshmallow.Schema.dump`/:func:`~marshmallow.Schema.load` method at runtime.
 
 In cases where your pre- and post-processing methods needs to handle the input collection when processing multiple objects, add ``pass_many=True`` to the method decorators.
 
-Your method will then receive the input data (which may be a single datum or a collection, depending ont the dump/load call) along with a relevant ``many`` value
-
-.. versionchanged:: 3.0.0
-    ``many`` is always passed as a keyword arguments to the pre- and post-processing methods.
+Your method will then receive the input data (which may be a single datum or a collection, depending on the dump/load call).
 
 Example: Enveloping
 +++++++++++++++++++
--- a/docs/extending.rst
+++ b/docs/extending.rst
@@ -32,10 +32,14 @@ Data pre-processing and post-processing methods can be registered using the `pre
 Passing "many"
 ++++++++++++++
 
-By default, pre- and post-processing methods receive one object/datum at a time, transparently handling the ``many`` parameter passed to the schema at runtime.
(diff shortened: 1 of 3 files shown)


docs/extending.rst
```diff
@@ -32,10 +32,14 @@ Data pre-processing and post-processing methods can be registered using the `pre
...
 
-By default, pre- and post-processing methods receive one object/datum at a time, transparently handling the ``many`` parameter passed to the schema at runtime.
+By default, despite of the ``many`` parameter value, pre- and post-processing methods receive one object/datum at a time, transparently handling the ``many`` parameter passed to the ``Schema``'s :func:`~marshmallow.Schema.dump`/:func:`~marshmallow.Schema.load` method at runtime.
 
-In cases where your pre- and post-processing methods need to receive the input collection  when ``many=True``, add ``pass_many=True`` to the method decorators. The method will receive the input data (which may be a single datum or a collection) and the boolean value of ``many``.
+In cases where your pre- and post-processing methods needs to handle the input collection when processing multiple objects, add ``pass_many=True`` to the method decorators.
 
+Your method will then receive the input data (which may be a single datum or a collection, depending ont the dump/load call) along with a relevant ``many`` value
+
+.. versionchanged:: 3.0.0
```
sloria: We don't include these in the narrative docs.

docs/extending.rst
```diff
@@ -32,10 +32,14 @@ Data pre-processing and post-processing methods can be registered using the `pre
 Passing "many"
 ++++++++++++++
 
-By default, pre- and post-processing methods receive one object/datum at a time, transparently handling the ``many`` parameter passed to the schema at runtime.
+By default, despite of the ``many`` parameter value, pre- and post-processing methods receive one object/datum at a time, transparently handling the ``many`` parameter passed to the ``Schema``'s :func:`~marshmallow.Schema.dump`/:func:`~marshmallow.Schema.load` method at runtime.
 
-In cases where your pre- and post-processing methods need to receive the input collection  when ``many=True``, add ``pass_many=True`` to the method decorators. The method will receive the input data (which may be a single datum or a collection) and the boolean value of ``many``.
+In cases where your pre- and post-processing methods needs to handle the input collection when processing multiple objects, add ``pass_many=True`` to the method decorators.
 
+Your method will then receive the input data (which may be a single datum or a collection, depending ont the dump/load call) along with a relevant ``many`` value
```
sloria: This might still mislead people to think that ``many`` is only passed when ``pass_many=True``. I'll revise.

src/marshmallow/decorators.py
```diff
@@ -14,12 +16,12 @@ class UserSchema(Schema):
         age = fields.Integer(required=True)
 
         @post_load
-        def lowerstrip_email(self, item, **kwargs):
+        def lowerstrip_email(self, item, partial, many, **kwargs):
```
sloria: I don't think we should exemplify including kwargs that are unused.

Instead, we can add a note that all decorated methods recieve `many` and `partial`. I'll revise.

src/marshmallow/decorators.py
```diff
@@ -71,9 +73,10 @@ def validates_schema(
 ):
     """Register a schema-level validator.
 
-    By default, receives a single object at a time, regardless of whether ``many=True``
-    is passed to the `Schema`. If ``pass_many=True``, the raw data (which may be a collection)
-    and the value for ``many`` is passed.
+    By default it receives a single object at a time, transparently handling the ``many``
+    argument passed to the `Schema`'s :func:`~marshmallow.Schema.validate` call.
+    If ``pass_many=True``, the raw data (which may be a collection) is passed
+    to the decorated method, along with a relevant ``many`` value.
```
sloria: Same here: this might mislead a user to think that `many` is only passed when `pass_many=True`.

sloria (commented):
Thanks for the PR. I'll make some revisions and merge.

PR #1375, opened 2019-09-04
Don't swallow TypeErrors raised by _bind_to_schema or on_bind_field

sloria:
Considering this a bug fix because it was not intentional to swallow TypeErrors here. I can't imagine anyone depending on this behavior.

I discovered this while writing a custom field that intentionally raised a `TypeError`, only to find that it would never get raised.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,8 @@ Changelog
 Bug fixes:
 
 - Fix incorrect ``super()`` call in ``SchemaMeta.__init__``.
+- Don't swallow ``TypeError`` exceptions raised by ``Field._bind_to_schema``
+  or ``Schema.on_bind_field``.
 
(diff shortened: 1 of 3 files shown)


lafrech (approved):
The whole try/except is meant to provide a user-friendly warning about a common mistake. Nobody is supposed to actually rely on this. I mean writing code that would catch that `TypeError` we reraise here and act upon it would be terrible.

My point is, no problem with this not being a breaking change.

sloria:
Yeah, the problem here was that we weren't re-raising the `TypeError` when it not caused by the expected user mistake.

This bug is also on v2. I'll remake the PR against 2.x-line and forward-port.

lafrech:
> Yeah, the problem here was that we weren't re-raising the TypeError when it not caused by the expected user mistake.

Oh, right. I didn't totally grab that.

Issue #229, opened 2015-06-28
skip_missing is gone, how to skip None elements ?


touilleMan:
I saw that v2.0.0b removed the `skip_missing` option, claiming it skips the missing entries by default.
However in my case (using mongoengine's Document) a field is never missing: it returns `None` which is, strictly speaking, _something_...

I can see from the commit (https://github.com/marshmallow-code/marshmallow/commit/a3908d36967ac30764ceb66c52c99357425fbdec) that previous behavior of skipping empty values (like None or empty list/tuple) has been obliterated. Is there now a new way to do this ?

I'm thinking for now to overload the `_serialize` method of each field to check for the empty value and return the `missing` singleton in such a case... but this sound really cumbersome to me, there must be a better way !

ghost:
Same problem with nested Schema representing lazy loaded SqlAlchemy entities during serialization.

mcquay239:
I beleive this is a relevant test:

```
assert not Marshaller()({"v": None}, {"v": fields.Integer()})
```

VladimirPal:
@sloria Tell us please how can we skip None elements?
I've created this PR https://github.com/marshmallow-code/marshmallow/pull/254 and it works for me, it will skip None elements, but it breaks a lot of tests, looks like i don't understand something

VladimirPal:
I think we need to have skip_if argument:

``` python
    name = fields.Str(skip_if=None)
```

@sloria If it's ok i can create PR

sloria:
I'm not sure that a user would want to configure skippable values at the individual field level; it might make more sense to configure at the field _class_ level.

Something like:

``` python
class SkippableString(fields.Str):
    skip_values = [None, '']

# Or, if you prefer
fields.Str.skip_values = [None, '']
```

Thoughts?

VladimirPal:
What if i need all my fields with None value have been ignored in schema? I need to create some Meta class?
Here is my case:
In frontend i have a backbone model and it fetch data from backend, backend returns this None values and when i save model, i get errors - "Field may not be null", because i sent the data that i got from backend, i think it's a weird behaviour

touilleMan:
I solve this trouble in my library (marshmallow intregration with mongoengine) by creating a mixin class which overload the `_serialize` method

``` python
class SkipEmptyClass(object):
    def __init__(self, *args, **kwargs):
        skip_empty = kwargs.pop('skip_empty', True)
[...]

PR #1381, opened 2019-09-08
Add missing_values parameter to Field

sloria:
Proves the concept proposed in https://github.com/marshmallow-code/marshmallow/issues/713#issuecomment-529154890

Still needs docs and details, but this is ready for review.

TODO:

- [x] ~Respect `missing_values` on serialization?~ see discussion
- [ ] Update `Field.__repr__` to include `missing_values`
- [ ] Docs

### DIFF
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -25,7 +25,7 @@ class TestDeserializingNone:
         with pytest.raises(ValidationError, match="Field may not be null."):
             field.deserialize(None)
 
-    def test_allow_none_is_true_if_missing_is_true(self):
+    def test_allow_none_is_true_if_missing_is_none(self):
         field = fields.Field(missing=None)
         assert field.allow_none is True
         field.deserialize(None) is None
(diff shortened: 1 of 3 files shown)


tests/test_deserialization.py
```diff
@@ -1338,6 +1338,50 @@ class AliasingUserSerializer(Schema):
...
+    def test_deserialize_required_field_with_custom_missing_values(self):
+        class ArtistSchema(Schema):
+            album_names = fields.List(
+                fields.Str(), required=True, missing_values=([], ())
+            )
+
+        with pytest.raises(ValidationError, match="required"):
+            ArtistSchema().load({"album_names": []})
+
+    def test_setting_default_missing_values(self, monkeypatch):
```
sloria, line 1372: This test might not be necessary--it's sorta just testing the Python language. Just wanted to make sure [this use case was met](https://github.com/marshmallow-code/marshmallow/issues/713#issuecomment-529205297).

lafrech:
LGTM.

> Respect `missing_values` on serialization?

What would that do? I'm tempted to say no. Do we do anything specific with `missing` already?

sloria:
It would make `default` value get returned for missing values on serialization.

```python


class ArtistSchema(Schema):
    name = fields.Str(default="---", missing_values=("",))


sch = ArtistSchema()
print(sch.dump({}))   # =>  {'name': '---'}
print(sch.dump({"name": ""})) # => {'name': ''} now, but maybe should be same as above
```

Sure, it's not a common use case, but it seems inconsistent to not handle it.

lafrech:
OK, I get it.

Maybe it shouldn't be the same parameter.

IIUC, the primary use case for this is an API for which it is complicated to enforce removal of empty-ish values. There is no reason the same constraint would apply to the serialized form of the data.

But the database may have its own constraint. When using SQL, I already felt the need to treat `None` as missing to avoid returning `null` (https://github.com/marshmallow-code/marshmallow/issues/229#issuecomment-363250656) and your proposal would solve that. But I didn't want to treat `null` as missing when deserializing.

Symmetry would recommend `default_values`, although the name is a bit misleading.

sloria:
That's a good point; the client inputs for "no data" may be different from the database's representation. 

`default_values` is indeed misleading. If we go with https://github.com/marshmallow-code/marshmallow/issues/778 , we could do `load_default_values` and `dump_default_values`, I guess. Maybe we defer the serialization param for now.

lafrech:
> `default_values` is indeed misleading. If we go with https://github.com/marshmallow-code/marshmallow/issues/778 , we could do `load_default_values` and `dump_default_values`, I guess. Maybe we defer the serialization param for now.

#1742 is about to close #778.

We could then change the names here. And do the serialization part in here or another PR.

Issue #1160, opened 2019-02-26
Nested schema doesn't load nested field if many is not set in both Schema and field constructors


Kamforka:
I found a strange behaviour of the `Nested` field when I use instantiated schemas for its `nested` parameter. To reproduce one can use the following snippet:

```python
from marshmallow import Schema, fields


class BookSchema(Schema):
    id = fields.Int()
    title = fields.String()

    class Meta:
        strict = True


class UserSchema(Schema):
    id = fields.Int()
    name = fields.String()
    books = fields.Nested(BookSchema(many=True)) 

    class Meta:
        strict = True


books = [{'id': 1, 'title': 'First book'}, {'id': 2, 'title': 'Second book'}]
user = {'id': 1, 'name': 'Peter', 'books': books}

user_dump = UserSchema().dump(user).data
# output: {'id': 1, 'books': {}, 'name': 'Peter'}

user_load = UserSchema().load(user_dump).data
# output: marshmallow.exceptions.ValidationError: {'books': {'_schema': ['Invalid input type.']}}
```

I'd expect it to return a list of books since I set the `many=True` parameter of the schema constructor.

A workaround is to set `many` in both the schema and the field like this:

```python
books = fields.Nested(BookSchema(many=True), many=True)
```

And the outputs are as expected:
```python
user_dump = UserSchema().dump(user).data
# output: {'name': 'Peter', 'books': [{'title': 'First book', 'id': 1}, {'title': 'Second book', 'id': 2}], 'id': 1}
user_load = UserSchema().load(user_dump).data
# output: {'name': 'Peter', 'books': [{'title': 'First book', 'id': 1}, {'title': 'Second book', 'id': 2}], 'id': 1}
```

> Note:
I know I can setup the schema this way and it works:
`books = fields.Nested(BookSchema, many=True)`
However my initial intention was to use it this way:
`books = fields.Nested(BookSchema(custom_arg='A custom arg to be set', many=True))`

I'm using `marshmallow==2.16.3`.

deckar01:
I imagine this will also be fixed by https://github.com/marshmallow-code/marshmallow/pull/1151.
 We may want to consider dropping support for schema instances in the `Nested` arg since we reconstruct it and loose custom parameters.

Kamforka:
@deckar01 This means that in the future one may only use nested fields by passing a schema class or a name reference to the `nested` parameter of `fields.Nested`?

deckar01:
[...]

PR #1385, opened 2019-09-08
Fix propagating `only` and `exclude` to nested field instances

### LINKED ISSUE #1384: Dotted `only` and `exclude` not working for nested schema instances
#1229 fixed propagation of `only` and `exclude` for nested schema classes but no instances.
Dotted strings passed to `only` and `exclude` are not respected when a schema instance is nested.

```python


class Child(Schema):
    name = fields.String()
    age = fields.Integer()


class Parent(Schema):
    favorite = fields.Nested(Child, only=("name", "age"))
    children = fields.List(fields.Nested(Child, only=("name", "age")))


class Parent2(Schema):
    favorite = fields.Nested(Child(only=("name", "age")))
    children = fields.List(fields.Nested(Child(only=("name", "age"))))


family = {
    "favorite": {"name": "Lily", "age": 15},
    "children": [{"name": "Tommy", "age": 12}, {"name": "Lily", "age": 15}],
}
# should be the same
print(Parent(only=("favorite.age", "children.name")).dump(family))
print(Parent2(only=("favorite.age", "children.name")).dump(family))
```

Expected output:

```
{'children': [{'name': 'Tommy'}, {'name': 'Lily'}], 'favorite': {'age': 15}}
{'children': [{'name': 'Tommy'}, {'name': 'Lily'}], 'favorite': {'age': 15}}
```

Actual output:

```
{'favorite': {'age': 15}, 'children': [{'name': 'Tommy'}, {'name': 'Lily'}]}
{'favorite': {'age': 15, 'name': 'Lily'}, 'children': [{'age': 12, 'name': 'Tommy'}, {'age': 15, 'name': 'Lily'}]}
```

sloria:
Fix #1384

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -468,7 +468,7 @@ class Nested(Field):
             # Inherit context from parent.
             context = getattr(self.parent, "context", {})
             if isinstance(self.nested, SchemaABC):
-                self._schema = self.nested
+                self._schema = copy.deepcopy(self.nested)
                 self._schema.context.update(context)
                 # Respect only and exclude passed from parent and re-initialize fields
                 set_class = self._schema.set_class
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -470,6 +470,23 @@ class Nested(Field):
             if isinstance(self.nested, SchemaABC):
                 self._schema = self.nested
                 self._schema.context.update(context)
+                all_fields = self._schema.fields.keys()
+                if self.only:
+                    field_names = self.only
+                else:
+                    field_names = all_fields
+                if self.exclude:
+                    field_names = field_names - self.exclude
+                fields_to_exclude = all_fields - field_names
+                for field_name in fields_to_exclude:
+                    try:
+                        del self._schema.fields[field_name]
+                        del self._schema.dump_fields[field_name]
+                        del self._schema.load_fields[field_name]
+                    except KeyError:
+                        pass
+                self._schema.only = self.only
+                self._schema.exclude = self.exclude
             else:
                 if isinstance(self.nested, type) and issubclass(self.nested, SchemaABC):
                     schema_class = self.nested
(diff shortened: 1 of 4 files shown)


src/marshmallow/schema.py
```diff
@@ -878,7 +875,9 @@ def __apply_nested_option(self, option_name, field_names, set_operation):
             setattr(self.declared_fields[key], option_name, new_options)
 
     def _init_fields(self):
-        """Update fields based on schema options."""
+        """Update self.fields, self.load_fields, and self.dump_fields based on schema options.
+        This method is private API.
```
sloria, line 879: This was always private even though we don't actually use `__`, so I don't consider this a breaking change.

sloria:
Merging this because I don't want this bug and the bug in #1160 in the wild for too long.

src/marshmallow/fields.py
```diff
@@ -468,8 +468,17 @@ def schema(self):
             # Inherit context from parent.
             context = getattr(self.parent, "context", {})
             if isinstance(self.nested, SchemaABC):
-                self._schema = self.nested
+                self._schema = copy.deepcopy(self.nested)
```
metheoryt, line 471: I have nested schema instance which context contains non-pickling object. In my case it is `Crypto.Cipher.AES` object which i tried to deepcopy in console and got an error. Should I start an issue?

metheoryt:
Hi guys! I've leaved a comment on a diff page, in general now I cannot use non-deepcopyable objects in nested schema context. I use `Crypto.Cipher.AES` from `pycryptodome` package

src/marshmallow/fields.py
sloria: Does it fix it if you use `copy.copy` instead?

```suggestion
                self._schema = copy.copy(self.nested)
```

Now that I think of it, I'm not sure we even need to deepcopy here. Let me know if the above fixes your issue.

metheoryt: @sloria yep, copy works just fine! waiting for release, thanks!

sloria: Thanks for checking. The fix is released in 3.1.1.

PR #1397, opened 2019-09-13
Add a way to trim whitespace characters.

245967906:


sloria:
Thanks for the PR.

If we add this, we might want to add it to `Field` instead of `String`, since there are other non-String subclasses (e.g. Date, DateTime) that this would be relevant for. I'm not completely sure, though--would have to give this more thought. If you have time, you could evaluate what Django, DRF, typesystem, colander, and other similar libs do.

245967906:
@sloria hi, I understand what you mean. I will compare the implementation of similar libraries and make changes later if I find a better solution.

245967906:
@sloria  hi,  I go to see the implementation of DRF and typesystem. summarized as follows:

- DRF's CharField and DatetimeField inherit from the same base class Field, and the trim_whitespace option only works in Charfield.
- In typesystem, Datetime inherits from String, so they can all be affected by trim_whitespace which define in String.

But in marshmallow, things is different from the two implementations above. 

If you implement strip_whitespace in Field according to your idea, for the field like Number, providing trim_whitespace parameter is not semantics. 

So now it is a reference to DRF that ignores DatetimeField, or let Datetime inherit from String like typesystem? i can't make a choice and I think this may require you to make a decision.

This is a more troublesome issue and may take a lot of your time. Anyway, if you have any decisions or ideas, please tell me, I also hope to help this feature complete as soon as possible.

thank you!

sloria:
Thanks for doing that evaluation. Let's close this PR for now and continue discussion on #1397. I'd like to think this through and also get more feedback from other users before landing this.

PR #1429, opened 2019-10-18
Fix load parameter *unknow* propagation

### LINKED ISSUE #1428: unknown parameter passed to load method is not popagated to nested schema field.
## Current behavior
When we deserialize a data structure containing a nested data structure by using the `load` method, the `unknown` parameter is without effect on the nested data structure and a error is raised if an unknown field is present into the nested data structure 

## Expected behavior:
 `unknown` parameter  into load method is propagated to the the load of all nested data structure.

## Functional context:
when processing a REST call in a server, it is often useful to be able to ignore unknown fields in the json provided during the deserialization process. This does not mean that we want these fields to be ignored in all contexts. The use of the `unknown` parameter of the `load` method allows this use case. Unfortunately, it does not apply to the nested field.

lmignon:
When deserializing a data structure with the load method, the *unknown* was not propagated to the loading of nested data structures. As result, if a unknown field was present into a nested data structure a ValidationError was raised even if the load methd was called with *unknown=EXCLUDE*. This commit ensures that this parameter is now propagated also to the loading of nested data structures.
fixes #1428

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -300,8 +300,14 @@ def test_load_unknown(val):
     class Inner(Schema):
         name = fields.String()
 
+        class Meta:
+            unknown = RAISE
+
     class Outer(Schema):
         inner = fields.Nested(Inner)
+
+        class Meta:
+            unknown = RAISE
 
     assert Outer().load(val, unknown=EXCLUDE) == {"inner": {"name": "name"}}
 
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -289,6 +289,23 @@ def test_load_many_in_nested_empty_collection(val):
     assert Outer().load({"list1": val, "list2": val}) == {"list1": [], "list2": []}
 
 
+@pytest.mark.parametrize(
+    "val",
+    (
+        {"inner": {"name": "name"}, "unknown": 1},
+        {"inner": {"name": "name", "unknown_nested": 1}, "unknown": 1},
+    ),
+)
+def test_load_unknown(val):
+    class Inner(Schema):
+        name = fields.String()
+
+    class Outer(Schema):
+        inner = fields.Nested(Inner)
+
+    assert Outer().load(val, unknown=EXCLUDE) == {"inner": {"name": "name"}}
+
+
 def test_loads_returns_a_user():
     s = UserSchema()
     result = s.loads(json.dumps({"name": "Monty"}))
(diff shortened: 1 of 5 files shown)


lmignon:
@lafrech @sloria What do you thing about this proposed fix? I tried to provide the required documentation to make this behavior explicit.

tests/test_schema.py
```diff
@@ -289,6 +289,23 @@ class Outer(Schema):
...
 
+@pytest.mark.parametrize(
+    "val",
+    (
+        {"inner": {"name": "name"}, "unknown": 1},
+        {"inner": {"name": "name", "unknown_nested": 1}, "unknown": 1},
+    ),
+)
+def test_load_unknown(val):
+    class Inner(Schema):
```
SVilgelm, line 300: I think it would be better to explicitly set:
```python
class Meta:
    unknown = RAISE
```
for both classes here

SVilgelm (approved):
Looks good to me, i have just one nit comment

tests/test_schema.py
lmignon: Thank your for the review @SVilgelm

Code updated

SVilgelm (approved):
Thanks, LGTM

Sytten (approved):

deckar01:
`Nested.load()` calls `Schema.load()` internally, so when an `unknown` argument is supplied to the `Nested` constructor, this PR causes all of the nested `unknown` options to be recursively overridden.

```py
from marshmallow import fields, Schema, RAISE, INCLUDE, EXCLUDE


class C(Schema):
    foo = fields.Str()
    class Meta:
        unknown = EXCLUDE

class B(Schema):
    c = fields.Nested(C)

class A(Schema):
    b = fields.Nested(B, unknown=INCLUDE)
    class Meta:
        unknown = RAISE

schema = A()
print(schema.load({
    'b': {
        'extra': 'good',
        'c': {
            'foo': 'bar',
            'unexpected': 'bad'
        }
    }
}))
```

Current behavior (3.2.2):

```py
{'b': {'c': {'foo': 'bar'}, 'extra': 'good'}}
```

Proposed behavior:

```py
{'b': {'c': {'foo': 'bar', 'unexpected': 'bad'}, 'extra': 'good'}}
```

bhperry:
Any chance of this merging?

PR #1448, opened 2019-11-04
Fix logic for field both load_only and dump_only

lafrech:
I just stumbled upon this.

I don't really see the use case for a field both `dump_only` and `load_only` but I don't think it is forbidden, is it?

Without the fix, due to the `elif`, the `dump_only` attribute is ignored.


sloria (approved):
Good catch. I wouldn't expect this to be a common use case, but Flask-RESTy uses load-only+dump-only fields for input fields that are only used for filtering: https://flask-resty.readthedocs.io/en/latest/special_topics.html#filter-only-fields .

I'll merge and release this now.

Issue #1452, opened 2019-11-09
utils.get_value swallows AttributeError thrown elsewhere in the call chain


billyrrr:
Hi, I noticed that some ```AttributeError``` is swallowed by ```utils.get_value```. I have written a test case that reproduces this error. 

```python

class Square:

    def __init__(self, p1: tuple, side_length):
        self.p1 = p1
        self.side_length = side_length

    @property
    def p2_fail(self):
        return self.p1 + self.non_existent_property

    # # p3 is not defined
    # @property
    # def p3_not_defined(self):
    #     pass

    @property
    def favorite_food(self):
        raise Exception

    @property
    def p4(self):
        return self.p1[0]+self.side_length, self.p1[1]+self.side_length


def test_get_property_value_from_object():

    # Will succeed
    sq = Square(p1=(0, 0), side_length=1)
    assert utils.get_value(sq, "p4") == (1, 1)
    assert utils.get_value(sq, "p3_not_defined") == utils.missing
    with pytest.raises(Exception):
        utils.get_value(sq, "favorite_food")

    # Will fail
    with pytest.raises(AttributeError):
        utils.get_value(sq, "p2_fail")

```

(Marshmallow is built using the current master branch. Test is run in Python 3.7 venv)

Expected behavior: ```utils.get_value(sq, "p2_fail")``` raises ```AttributeError``` as ```p2_fail``` is a valid attribute in ```sq```, but the error is thrown in the scope of ```p2_fail```. 

Current behavior: ```utils.get_value(sq, "p2_fail")``` returns ```utils.missing``` rather than throwing an error. 

This may have been caused by ```getattr``` swallowing ```AttributeError``` when ```default``` is set, but since it is part of the standard library we have to make changes in ```utils.get_value```

The current design results in critical error messages being swallowed in runtime. I am about to submit a pull request that includes tests written above. 

Although this error can be fixed by specifying a ```get_attribute``` function, it is nevertheless unexpected that the default behavior does not differentiate between "p2_fail" case and "p3_not_defined" case. I think that this is a defect. 

Thanks!

lafrech:
> This may have been caused by getattr swallowing AttributeError when default is set, but since it is part of the standard library we have to make changes in utils.get_value

Exactly.

I get your point, but I'd find it consistent with standard getattr to keep things as they are.

billyrrr:
How about adding a line to warn the coders in the documentation of utils.get_value?

lafrech:
[...]

PR #1453, opened 2019-11-09
Add test case for #1452 utils.get_value swallows AttributeError

billyrrr:
See #1452


lafrech:
Closing as we settled in #1475 on keeping things as they are.

PR #1466, opened 2019-12-05
Make Number._format_num return typing.Any

hukkin:
This allows a subclass that overrides/slash extends the method to set a more specific return type annotation.


sloria:
For reference: https://github.com/sloria/environs/blob/67a9433db88b4df7f8e621a40d69a75a09d4ab21/environs.py#L198-L201

PR #1467, opened 2019-12-05
Mypy strictness

hukkin:
- Add mypy strictness flags:
    ```
    warn_unreachable = true
    warn_unused_ignores = true
    warn_redundant_casts = true
    ```
- Fix issues found by mypy

### DIFF
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -373,7 +373,7 @@ class Schema(base.SchemaABC, metaclass=SchemaMeta):
         # Raise error if only or exclude is passed as string, not list of strings
         if only is not None and not is_collection(only):
             raise StringNotCollectionError('"only" should be a list of strings')
-        if exclude is not None and not is_collection(exclude):
+        if not is_collection(exclude):
             raise StringNotCollectionError('"exclude" should be a list of strings')
         # copy declared fields from metaclass
         self.declared_fields = copy.deepcopy(self._declared_fields)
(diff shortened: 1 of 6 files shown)


src/marshmallow/schema.py
```diff
@@ -1025,9 +1025,13 @@ def _bind_field(self, field_name: str, field_obj: ma_fields.Field) -> None:
...
         except TypeError as error:
-            # field declared as a class, not an instance
-            if isinstance(field_obj, type) and issubclass(field_obj, base.FieldABC):
-                msg = (
+            # Field declared as a class, not an instance. Type ignore because
+            # we handle unsupported arg types -> this is dead code from
+            # typechecker's perspective.
+            if isinstance(field_obj, type) and issubclass(  # type: ignore
+                field_obj, base.FieldABC
+            ):
```
hukkin, line 1033: The formatting here is awkward, but the only formatting that passes both flake8 line length and black checks. Another option is to keep the `if` as a one-liner and add `# noqa: E501` after the `# type: ignore` to disable flake8 line length check.

hukkin: Also, I don't know what the importance of the custom error message here is, but perhaps we could consider not handling this TypeError at all and just let the original exception bubble up always?

sloria: It makes debugging a bit easier. The original exception message doesn't really pinpoint what the issue is.

sloria (approved):
Thank you!

PR #1480, opened 2019-12-19
Add missing type annotations for fields init

### LINKED ISSUE #1479: Improve fields type annotations a little bit.
I am using `mypy` with an option `--disallow-untyped-calls`

From mypy docs
```
--disallow-untyped-calls
This flag reports an error whenever a function with type annotations calls a function defined without annotations.
```

Running command
```
mypy --disallow-untyped-calls .
```
on following code

```python
import marshmallow as ma
class MySchema(ma.Schema):
    arb = ma.fields.Float()
```
gives me output

``` 
error: Call to untyped function "Float" in typed context
```
The same for `Pluck`, `Method`, `Constant`
For other `ma.fields` seem to be fine.

Reskov:
Resolves #1479


sloria:
Thanks!

Issue #1415, opened 2019-09-19
Feature request: Validate should accept deserialized values as valid for all fields


bitdancer:
One of our big motivations for using Marshmallow is to validate that the data created by our programs actually matches the schema of the data we want to transmit on the wire.  There appears to be no way to do this with Marshmallow 3, since validation was removed from 'dump'.  (The the other motivation is to validate received data, of course, but that is handled just fine.)

This has been discussed in three closed issues that I found, but no solution has been suggested  using existing Marshmallow 3 features.

In #656 the case I'm immediately facing is discussed: if I pass an actual datetime object to validate, I get an error that it is not a valid datetime.  This is because validate is only handling string input.  A poster in that issue suggested that it would be good if validate accepted the deserialized values for all fields as valid.  lafrech responded that if someone was interested in this as a feature to open a new issue with a feature request.  This is that issue.

As additional context, in issue #1372 we are told that "Validation is only meant to be applied to serialized data, during a load", and that "Marshmallow does not validate object data before dump".  The porting to Marshmallow 3 docs repeat the second part of that, but don't mention the first part.  Further, the porting docs say "You must validate your data before serializing it".  However, you *can't* validate the data before dumping it if it isn't strings.  So that's a bit of a conundrum.  Further, you *can't* dump it first, because of the reason the porting docs say you must validate first: if you try to dump invalid data you will get various exceptions (for example, if you mistakenly set a fields.DateTime field to the string version of a datetime, on dump you will get an attribute error on 'isoformat').

The third closed issue on this topic that I found is #1393, which was closed with "Validating deserialized objects isn't supported out-of-box. Serializing your object before passing it to schema.validate or schema.load is your best option."  But, as discussed above, this is not a viable solution since you will get exceptions on certain kinds of invalid data when serializing, which makes it difficult to figure out where the actual problem is (which field's value is causing the problem).

It seems to me that it is obviously desirable to validate data before putting it on the wire.  The programmer is going to make mistakes, some of which won't be found until runtime.

So, in summary, I think the ability to validate deserialized data is a major missing feature of Marshmallow 3.  It doesn't look like we'll be able to port our project to Marshmallow 3 until this feature is available.

(Aside: I don't think the validation we are using with Marshmallow 2 is actually doing all of the job we thought it was doing, but that is a different problem.)

sloria:
Thanks for the detailed report. Seems like there's two main asks here:

[...]

PR #1485, opened 2020-01-03
Add IP address field type.

mgetka:
This pull request adds IPv4/IPv6 fields serialization and deserialization w/ tests included. To avoid confusion I am excluding the possibility to specify IP addresses as an integer or byte values.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1635,14 +1635,7 @@ class IP(Field):
         if value is None:
             return None
         try:
-            if not isinstance(value, str):
-                # ip_address function is flexible in the terms of input value. In the case of
-                # marshalling, integer and binary address representation parsing may lead to
-                # confusion.
-                raise TypeError(
-                    "Only dot-decimal and hexadecimal groups notations are supported."
-                )
-            return ipaddress.ip_address(value)
+            return ipaddress.ip_address(utils.ensure_text_type(value))
         except (ValueError, TypeError) as error:
             raise self.make_error("invalid_ip") from error
 
@@ -1658,9 +1651,7 @@ class IPv4(IP):
         if value is None:
             return None
         try:
-            if not isinstance(value, str):
-                raise TypeError("Only dot-decimal notation is supported.")
-            return ipaddress.IPv4Address(value)
+            return ipaddress.IPv4Address(utils.ensure_text_type(value))
         except (ValueError, TypeError) as error:
             raise self.make_error("invalid_ip") from error
 
@@ -1676,9 +1667,7 @@ class IPv6(IP):
         if value is None:
             return None
         try:
-            if not isinstance(value, str):
-                raise TypeError("Only hexadecimal groups notation is supported.")
-            return ipaddress.IPv6Address(value)
+            return ipaddress.IPv6Address(utils.ensure_text_type(value))
         except (ValueError, TypeError) as error:
             raise self.make_error("invalid_ip") from error
 
@@ -1767,7 +1756,7 @@ class Function(Field):
             typing.Callable[[typing.Any], typing.Any],
             typing.Callable[[typing.Any, typing.Dict], typing.Any],
         ] = None,
-        **kwargs
+        **kwargs,
     ):
         # Set dump_only and load_only based on arguments
         kwargs["dump_only"] = bool(serialize) and not bool(deserialize)
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -52,6 +52,8 @@ __all__ = [
     "URL",
     "Email",
     "IP",
+    "IPv4",
+    "IPv6",
     "Method",
     "Function",
     "Str",
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1642,8 +1642,6 @@ class IP(String):
         """Format the value or raise a :exc:`ValidationError` if an error occurs."""
         if value is None:
             return None
-        if isinstance(value, (ipaddress.IPv4Address, ipaddress.IPv6Address)):
-            return value
         try:
             if isinstance(value, (int, bytes)):
                 # ip_address function is flexible in the terms of input value. In the case of
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1645,7 +1645,7 @@ class IP(Field):
         if value is None:
             return None
         try:
-            if isinstance(value, (int, bytes)):
+            if not isinstance(value, str):
                 # ip_address function is flexible in the terms of input value. In the case of
                 # marshalling, integer and binary address representation parsing may lead to
                 # confusion.
@@ -1668,7 +1668,7 @@ class IPv4(IP):
         if value is None:
             return None
         try:
-            if isinstance(value, (int, bytes)):
+            if not isinstance(value, str):
                 raise TypeError("Only dot-decimal notation is supported.")
             return ipaddress.IPv4Address(value)
         except (ValueError, TypeError) as error:
@@ -1686,7 +1686,7 @@ class IPv6(IP):
         if value is None:
             return None
         try:
-            if isinstance(value, (int, bytes)):
+            if not isinstance(value, str):
                 raise TypeError("Only hexadecimal groups notation is supported.")
             return ipaddress.IPv6Address(value)
         except (ValueError, TypeError) as error:
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1646,10 +1646,16 @@ class IP(Field):
         super().__init__(*args, **kwargs)
         self.exploded = exploded
 
-    def _validated(
-        self, value
+    def _serialize(self, value, attr, obj, **kwargs) -> typing.Optional[str]:
+        if value is None:
+            return None
+        if self.exploded:
+            return value.exploded
+        return value.compressed
+
+    def _deserialize(
+        self, value, attr, data, **kwargs
     ) -> typing.Optional[typing.Union[ipaddress.IPv4Address, ipaddress.IPv6Address]]:
-        """Format the value or raise a :exc:`ValidationError` if an error occurs."""
         if value is None:
             return None
         try:
@@ -1664,26 +1670,15 @@ class IP(Field):
         except (ValueError, TypeError) as error:
             raise self.make_error("invalid_ip") from error
 
-    def _serialize(self, value, attr, obj, **kwargs) -> typing.Optional[str]:
-        if value is None:
-            return None
-        if self.exploded:
-            return value.exploded
-        return value.compressed
-
-    def _deserialize(
-        self, value, attr, data, **kwargs
-    ) -> typing.Optional[typing.Union[ipaddress.IPv4Address, ipaddress.IPv6Address]]:
-        return self._validated(value)
-
 
 class IPv4(IP):
     """A IPv4 address field."""
 
     default_error_messages = {"invalid_ip": "Not a valid IPv4 address."}
 
-    def _validated(self, value) -> typing.Optional[ipaddress.IPv4Address]:
-        """Format the value or raise a :exc:`ValidationError` if an error occurs."""
+    def _deserialize(
+        self, value, attr, data, **kwargs
+    ) -> typing.Optional[ipaddress.IPv4Address]:
         if value is None:
             return None
         try:
@@ -1693,19 +1688,15 @@ class IPv4(IP):
         except (ValueError, TypeError) as error:
             raise self.make_error("invalid_ip") from error
 
-    def _deserialize(
-        self, value, attr, data, **kwargs
-    ) -> typing.Optional[ipaddress.IPv4Address]:
-        return self._validated(value)
-
 
 class IPv6(IP):
     """A IPv6 address field."""
 
     default_error_messages = {"invalid_ip": "Not a valid IPv6 address."}
 
-    def _validated(self, value) -> typing.Optional[ipaddress.IPv6Address]:
-        """Format the value or raise a :exc:`ValidationError` if an error occurs."""
+    def _deserialize(
+        self, value, attr, data, **kwargs
+    ) -> typing.Optional[ipaddress.IPv6Address]:
         if value is None:
             return None
         try:
@@ -1714,11 +1705,6 @@ class IPv6(IP):
             return ipaddress.IPv6Address(value)
         except (ValueError, TypeError) as error:
             raise self.make_error("invalid_ip") from error
-
-    def _deserialize(
-        self, value, attr, data, **kwargs
-    ) -> typing.Optional[ipaddress.IPv6Address]:
-        return self._validated(value)
 
 
 class Method(Field):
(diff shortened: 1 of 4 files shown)


src/marshmallow/fields.py
```diff
@@ -1629,6 +1631,42 @@ def __init__(self, *args, **kwargs):
...
+        if isinstance(value, (ipaddress.IPv4Address, ipaddress.IPv6Address)):
+            return value
+        try:
+            if isinstance(value, (int, bytes)):
+                # ip_address function is flexible in the terms of input value. In the case of
+                # marshalling, integer and binary address representation parsing may lead to
+                # confusion.
+                raise TypeError(
+                    "Only dot-decimal and hexadecimal groups notations are supported. "
+                    "Got %s."
```
lafrech: I can't find a source to back this up but I think we have a consensus on not returning input values in error messages.

mgetka: Fixed.

src/marshmallow/fields.py
```diff
@@ -1629,6 +1631,41 @@ def __init__(self, *args, **kwargs):
...
+
+    default_error_messages = {"invalid_ip": "Not a valid IP address."}
+
+    def _validated(
+        self, value
+    ) -> typing.Optional[typing.Union[ipaddress.IPv4Address, ipaddress.IPv6Address]]:
+        """Format the value or raise a :exc:`ValidationError` if an error occurs."""
+        if value is None:
+            return None
+        if isinstance(value, (ipaddress.IPv4Address, ipaddress.IPv6Address)):
```
lafrech: Currently, marshmallow does not accept deseralized form when deserializing. This is discussed in #1415.

lafrech (commented):
Should we add a parameter to only accept IPv4 or IPv6 rather than accepting any?

Should we also provide `IPv4` and `IPv6` fields as a shortcut?

Also, I don't see the point of the `_serialize` method. Apparently you copied it from an already existing field (`UUID`) so it's nothing new. But this seems a bit dodgy to me. Let's discuss this in #1489.

src/marshmallow/fields.py
mgetka: Dropping lines 1645-1646 will actually have no effect on the interface level since `ipaddress.ip_address` function accepts `ipaddress.IPv4Address` and `ipaddress.IPv6Address` instances, and returns their copies.

Are you suggesting to add explicit type checking and accept only `str` values?

mgetka:
> Also, I don't see the point of the _serialize method. Apparently you copied it from an already existing field (UUID) so it's nothing new. But this seems a bit dodgy to me.

That is correct, I've overridden the `_serialize` based on insights collected from `UUID` field implementation. After inspecting the implementation of `String` base class, I can no longer see any purpose of `_serialize` method.

src/marshmallow/fields.py
lafrech: I see no need to explicitly exclude deserialized form. But that could be the reason I'm in favour of #1415. I didn't check but I'd bet other fields already accept deserialized data.

In any case, I see no need for this check. As you said, the behaviour will stay the same.

mgetka: In that case, the issue is resolved via 985f2ba

src/marshmallow/fields.py
```diff
@@ -1634,6 +1638,91 @@ def __init__(self, *args, **kwargs):
...
+
+    default_error_messages = {"invalid_ip": "Not a valid IPv4 address."}
+
+    def _validated(self, value) -> typing.Optional[ipaddress.IPv4Address]:
+        """Format the value or raise a :exc:`ValidationError` if an error occurs."""
+        if value is None:
+            return None
+        try:
+            if isinstance(value, (int, bytes)):
+                raise TypeError("Only dot-decimal notation is supported.")
```
lafrech: What about hexadecimal groups as above in generic IP field?

mgetka: IPv6 error message mistake. Dot decimal notation is exclusive for IPv4 and hexadecimal groups are appropriate for IPv6. Fixed.

mgetka:
Also please note that _exploded_ notation is specific for IPv6 addresses, but the argument is also present in IPv4 constructor. This feature is sourced from [`ipaddress` python standard library](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.exploded), where exploded notation can be applied to both IPv4 and IPv6 for consistency.

sloria:
Would it make sense to implement these as validators? That would allow combining with other validators.

mgetka:
I have to admit that maybe I was too inspired by the implementation of the `UUID` field, and implemented the functionality without more complete understanding of fields module structure. As I can see, fields inheriting from `String`, are in general deserialized from strings to strings with additional input value format validation. We can see it in the `Url` and `Email` fields cases. Exception from this rule is the `UUID` field, which deserializes from string into UUID object representation. Therefore, it is not clear for me why `UUID` inherits from `String` and not `Field`, like in the case of `DateTime` or `TimeDelta` which also deserializes to object representation.

`ipaddress` library utilities provides with IP address object representations with validation occurring on object construction. In my opinion creation of the object representation for the purpose of validation, and then discarding it would be an overkill. Also I can see clear contract between _validators_ based implementation of `Url` and `Email` fields and  `String` base class features. But IP addresses are not the case - they are more like `DateTime` or `TimeDelta` fields. Therefore, IP field now inherits from `Field` and not `String`.

Also, I can see, that `_validated` method is utilized for chained validation in inheriting fields classes.
For IP fields it is not the case, since IPv4 and IPv6 Fields implements its own independent validation procedures. I have moved validation/object creation into `_deserialize` methods.

Returning to the case of the UUID field. Maybe it should be discussed whether it should inherit from `String` or rather `Field` to get a more transparent structure of the module?

I am also wondering whether IP specific validators are welcome? Based on `ipaddress` functionalities it is straightforward to implement validators like: `PrivateIPAddress`, `GlobalIPAddress` or one that checks whether IP address belongs to specified subnetwork.

tests/test_deserialization.py
```diff
@@ -841,6 +842,83 @@ def test_invalid_uuid_deserialization(self, in_value):
...
+        ipv4_str = "140.82.118.3"
+        result = field.deserialize(ipv4_str)
+        assert isinstance(result, ipaddress.IPv4Address)
+        assert str(result) == ipv4_str
+
+        ipv4 = ipaddress.ip_address("172.217.16.206")
+        result = field.deserialize(ipv4)
+        assert isinstance(result, ipaddress.IPv4Address)
+        assert result == ipv4
+
```
lafrech: Do we have to test this?

In #1415, we discuss accepting deserialized form as input to deserialization methods. For now it only works if the deserialization method happens to accept deserialized form, like this one does, but it is more a happy accident than a design choice.

I don't have a strong opinion about this test. But until we go on with #1415, I'm lukewarm to guarantee this behaviour for a given field.

Are there other field for which we test this?

lafrech (commented):
Looks good to me.

Just a question/comment about a test. I wouldn't mind another opinion here.

Your remarks about UUID field make sense. Reworking it could be done in another PR, and perhaps in marshmallow 4 if it is a breaking change.

IP specific validation welcome. Should those be field constructor arguments instead like exploded?

src/marshmallow/fields.py
```diff
@@ -1616,6 +1620,79 @@ def __init__(self, *args, **kwargs):
...
+            return value.exploded
+        return value.compressed
+
+    def _deserialize(
+        self, value, attr, data, **kwargs
+    ) -> typing.Optional[typing.Union[ipaddress.IPv4Address, ipaddress.IPv6Address]]:
+        if value is None:
+            return None
+        try:
+            if isinstance(value, (int, bytes)):
```
deckar01: Fields that inherit from `String` don't have to store their deserialized value internally as a string. Because fields are both dumpers and loaders, it is better to think of `String` as `String -> Any -> String`. Subclassing `String` allows you to specialize `Any` to a more concrete type.

IMO explicitly checking for ints and bytes is unnecessary. If you inherit form `String`, `super()._deserialize()` will provide all the type validation that is needed. `ipaddress` takes care of validating the format.

tests/test_deserialization.py
mgetka: And again, such case is present in UUID field tests. For the IP fields, I have removed tests on unintended features like this.

src/marshmallow/fields.py
mgetka: `ipaddress` utilities are flexible in the terms of input ip addresses representations. Beside classic string representations it also accepts binary and integer representations. Explicit type checking was my arbitrary choice in order to narrow down the original `ipaddress` functionality. In my opinion deserialization of following JSON
```
{"ip": 1}
```
into this object representation
```
{"ip": IPv4Address('0.0.0.1')}
```
may be unexpected. But as I can see it leads to confusion on the implementation level, thus I have added constructor parameter determining whether decimal representations should be accepted. The parameter is `False` by default.

deckar01: Only passing strings to `ipaddress` seems reasonable. Excluding a subset of the types that `ipaddress` accepts seems fragile. If `ipaddress` starts accepting new types in the future this would let other types through. Allowing only strings would make the type checking explicit and resilient to upstream changes. The `String` field is a nice way to inherit this behavior declaratively.

mgetka: Accepting only strings when deserializing IP fields is something that I can agree on. But achieving this via inheriting from `String` is something that is not so clear for me. From the sources and the discussion I can see two approaches on the `String` base class.

First approach is the one I described in previous comment, where `String` based fields deserialize into `str`, and have specific validator attached. Examples of such fields are `URL` and `Email` which are just string values with additional format constraints. If so, `UUID` filed nor `IP` field should inherit from `String`, since they are not based on a validator and doesn't serialize into string.

The second approach, is the one where `String` based fields are like `String -> Any -> String`. So, if I understand it correctly, all fields that serialize to strings, regardless of deserialized type, should inherit from `String`. Then, `UUID` and `IP` fields should inherit from `String`, as well as all `DateTime` and `Time` fields, which currently inherit from `Field`.

mgetka: Since there is no feedback on this matter, I decided on an approach that better suits my understanding of the structure of the module.

deckar01: The only thing `String._deserialize()` would do differently is decode bytes into a UTF-8 string using `utils.ensure_text_type()`. That has been the default behavior for all of the string fields for a long time, but it's not clear what the use case was from the original commit.

@sloria Do you remember why byte strings are assumed to be UTF-8 encoded and decoded into strings? Should we do that for IP address strings also?

Original commit: https://github.com/marshmallow-code/marshmallow/commit/ff18e782be9d04557f027116b147f16bcecb9999

mgetka: I suppose that it may be related to the fact that `Schema` objects can have `render_module` altered to some unusual one that does not perform data decoding? If it's so, it is correct to utilize `utils.ensure_text_type()` as well in IP fields, since `ipaddress` utilities will not treat `b"192.168.0.1"` the same way as `"192.168.0.1"`. I'll wait for additional feedback on that matter.

mgetka: To be compatible with other field interfaces I've utilised `utils.ensure_text_type()` helper function. From the IP fields perspective it is acceptable - all expected behaviours are consistent with original approach. Numeric (ie. `int(1) -> 0.0.0.1`) and binary (ie. `b"\x00\x00\x00\x01" -> 0.0.0.1`) inputs will not be interpreted by `ip_address` library utilities. 

If the usage of `utils.ensure_text_type()` is obsolete or somehow stems from no longer applicable interface requirements, I belive it shpould be discussed in another issue, and appropriate patch should be applied on all fields utilizing the function.

sloria:
@lafrech @deckar01 Can one of you do a final CR pass on this? Looks like it's close to--if not completely--ready to merge.

lafrech (commented):
LGTM. I'm just concerned about the code duplication. Could `_deserialize` be factorized, using class attributes for the object class, like what is done for the error message? (The typing would be a little less strict, but oh well.)

mgetka:
> LGTM. I'm just concerned about the code duplication. Could `_deserialize` be factorized, using class attributes for the object class, like what is done for the error message? (The typing would be a little less strict, but oh well.)

I've tried to define class variable typed as `typing.Callable[[str], typing.Union[ipaddress.IPv4Address, ipaddress.IPv6Address]]` which holds `ipaddress.ip_address` for generic IP field, and `ipaddress.IPv4Address` and `ipaddress.IPv6Address`  classes for IPv4 and IPv6 fields respectively, but there are issues with `ipaddress.ip_address` becoming a method. The issue can be suppressed by decorating the `ipaddress.ip_address` with `staticmethod` like
```
DESERIALIZATION_CALLABLE = staticmethod(ipaddress.ip_address)
```
but I'm not entirely comfortable with such construct. Also, mypy doesn't like it as well since there is unresolved issue [#3482](https://github.com/python/mypy/issues/3482) which is of relevance for class methods and static methods defined this way.

src/marshmallow/fields.py
```diff
@@ -1625,6 +1629,56 @@ def __init__(self, *args, **kwargs):
...
 
+class IP(Field):
+    """A IP address field.
+
+    :param bool exploded: If `True`, serialize ipv6 address in long form, ie. with groups
+        consisting entirely of zeros included."""
+
+    default_error_messages = {"invalid_ip": "Not a valid IP address."}
+
+    DESERIALIZATION_CLASS = None  # type: typing.Optional[typing.Type]
```
lafrech, line 1640: I'd write

    DESERIALIZATION_CLASS = ipaddress.ip_address

Then in `_deserialize`

            return self.DESERIALIZATION_CLASS(utils.ensure_text_type(value))

Unless there is a reason not to do that?

mgetka: I've presented rationale for this in the comment to the pull request. Long story short, `ip_address` is a function and eagers to become a method when assigned to a class attribute.

lafrech: Sorry, I did not understand your PR comment.

As you say, it works if we use `staticmethod`:

```py 
DESERIALIZATION_FUNC = staticmethod(ipaddress.ip_address)
```

mypy doesn't complain if I write


```py 
    DESERIALIZATION_FUNC = staticmethod(
        ipaddress.ip_address
    )  # type: typing.Union[staticmethod, ipaddress.IPv4Address, ipaddress.IPv6Address]
```

but I have little experience with Typing so I can't tell if it is the right way to go.

Type checking aside, I found the staticmethod alternative better but I'm starting to understand your rationale.

(Slight performance advantage for the staticmethod alternative because the check is not performed every load, but that's peanuts.)

lafrech (approved):
Sorry for the delay and thanks for your work on this.

lafrech:
@sloria, @deckar01, any objection to merging this?.

sloria:
@lafrech I didn't to a fine-toothed review of this, but it looked good last time I looked. I defer to you on merging this.

lafrech:
Merged and released.

Thanks again @mgetka.

PR #1488, opened 2020-01-12
FIX: fix schema.load(data) typing for many=True case

dfirst:
Previous typing expect data arg only as 'Mapping' but in case of many=True we can pass 'Iterable'.


sloria:
Thanks!

Issue #1497, opened 2020-01-30
Regression on list of nullable nested fields


Meallia:
Hello,

While porting some code from 2.20 to 3.3, I stumbled upon this change in behavior regarding list of nullable fields.

in 2.20 the following code works:
```python
from marshmallow import fields, Schema

class S(Schema):
    f = fields.List(fields.Nested(Schema, allow_none=True))

assert S().dump(dict(f=[None])).data == dict(f=[None])
```
whereas the same code in 3.3 fails with
`AssertionError: assert {'f': [{}]} == {'f': [None]}`
```python
from marshmallow import fields, Schema

class S(Schema):
    f = fields.List(fields.Nested(Schema, allow_none=True))

assert S().dump(dict(f=[None])) == dict(f=[None])
```

I could not find anything related to this in the upgrade instructions or the documentation.
I'm fully aware that this is an edge case and I don't mind looking into it if it's actually considered as a bug.

here are two unit tests that can be used to reproduce this issue:
```python
def test_list_of_nested_nullable_serialize_none_to_none(self):
    res = fields.List(fields.Nested(Schema, allow_none=True)).serialize("foo", {"foo": [None]})
    assert res == [None]

def test_nested_multiple_nullable_serialize_none_to_none(self):
    res = fields.Nested(Schema, allow_none=True, many=True).serialize("foo", {"foo": [None]})
    assert res == [None]
```

deckar01:
I was able to track this change down to #1306. This behavior was inherited from `Nested(many)` when optimizing `List(Nested)` to use the same fast `Schema.dump(many)` call. This was not an expected side effect, but the behavior was already present in a closely related field.

Using 2.x you can reproduce this behavior using `Nested(many)`:

```py
class S(Schema):
    f = fields.Nested(Schema, allow_none=True, many=True)

S().dump(dict(f=[None]))
# MarshalResult(data={u'f': [{}]}, errors={})
```

`allow_none` doesn't actually have any effect during `dump`, and neither of these examples behave differently when it is omitted or set to `False`. `None` will generally be accepted during serialization, because we don't validate during serialization (anymore, we used to a little).

When the value of a field is `None`, all of the field dump methods, including `Nested`, explicitly check for `None` and return `None`. #1306 lets `Schema.dump` handle everything and it does not have any special `None` handling. `Schema().dump(None)` is `{}`.

@sloria @lafrech Should `Schema().dump(None)` return `None` instead of `{}`?

[...]

PR #1498, opened 2020-01-31
Modify Nested field to handle the combination of many=True and allow_…

Meallia:
Fix for #1497 
Hi,

I tried to fix the issue by modifying `Schema` as suggested in the issue discussion but it would fail to detect validation errors when `[None]` is passed to a `Nested(many=True, allow_none=False)` field as the schema is not aware of `allow_none`. 

Instead, I made the `Nested` field check when (de-)serializing if `many` and `allow_none` are both set to `True`. in that case, we simply pass the non-`None` values to the underlying schema then insert `None` values back in the resulting list.

### DIFF
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -38,6 +38,24 @@ class TestDeserializingNone:
         field = fields.Tuple([fields.String()], allow_none=True)
         assert field.deserialize(None) is None
 
+    def test_list_of_nested_allow_none_deserialize_none_to_none(self):
+        field = fields.List(fields.Nested(Schema(), allow_none=True))
+        assert field.deserialize([None, {}]) == [None, {}]
+
+    def test_nested_multiple_allow_none_deserialize_none_to_none(self):
+        field = fields.Nested(Schema(), allow_none=True, many=True)
+        assert field.deserialize([None, {}]) == [None, {}]
+
+    def test_list_of_nested_non_allow_none_deserialize_none_to_validation_error(self):
+        field = fields.List(fields.Nested(Schema(), allow_none=False))
+        with pytest.raises(ValidationError):
+            field.deserialize([None])
+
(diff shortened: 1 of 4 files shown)


deckar01:
> it would fail to detect validation errors when `[None]` is passed to a `Nested(many=True, allow_none=False)`

That is existing behavior and orthogonal to #1497.

tests/test_deserialization.py
```diff
@@ -38,6 +38,24 @@ def test_tuple_field_deserialize_none_to_none(self):
         field = fields.Tuple([fields.String()], allow_none=True)
         assert field.deserialize(None) is None
 
+    def test_list_of_nested_allow_none_deserialize_none_to_none(self):
+        field = fields.List(fields.Nested(Schema(), allow_none=True))
+        assert field.deserialize([None, {}]) == [None, {}]
+
+    def test_nested_multiple_allow_none_deserialize_none_to_none(self):
+        field = fields.Nested(Schema(), allow_none=True, many=True)
+        assert field.deserialize([None, {}]) == [None, {}]
```
deckar01, line 47: The `allow_none` argument on `Nested` applies to the `Nested` container, not its children. Depending on `many` this is either a `List`, or a pass through container.

https://marshmallow.readthedocs.io/en/stable/api_reference.html#marshmallow.fields.Nested

Use `List(Nested(allow_none))` instead. #779 is an RFC that is amassing reasons `Nested(many)` should be deprecated. I will add this to the list.

deckar01 (changes requested):

Meallia:
Oh, yes, you're absolutely right about allow_none for nested attributes. So I guess we only need to revert the optimisation on lists of nested attributes.

Is this merge request still relevant then ?

Sorry for the delay, I could not find the time to follow up on the related issue but I will I'm the coming days.

lafrech:
#1497 was fixed by #1503.

PR #1500, opened 2020-02-01
Remove unnecessary typecasts

hukkin:
- Fix the issue in comment `# FIXME: Why doesn't mypy think validate.URL is a callable here?`
    - The problem was that we try to do `List[SubClass] + List[SuperClass]`. This is not supported by list's `+` operand, but `List[SuperClass] + List[SubClass]` is.
- Remove unnecessary typecasts
- Use `extend-ignore` instead of `ignore` in flake8 configuration to not override default ignores.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,7 +5,7 @@ license_files = LICENSE
 universal = 1
 
 [flake8]
-ignore = E203, E266, E501, W503, E731, B903
+extend-ignore = E203, E266, E501, E731, B903
 max-line-length = 90
 max-complexity = 18
 select = B,C,E,F,W,T4,B9
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -1428,7 +1428,7 @@ class TestSchemaDeserialization:
             MySchema().load({"email": "foo"})
         errors = excinfo.value.messages
         assert len(errors["email"]) == 2
-        assert "Not a valid email address." in errors["email"][0]
+        assert any("Not a valid email address." in e for e in errors["email"])
 
     def test_multiple_errors_can_be_stored_for_a_url_field(self):
         def validate_with_bool(val):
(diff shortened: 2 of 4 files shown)


tests/test_deserialization.py
```diff
@@ -1428,7 +1428,7 @@ class MySchema(Schema):
             MySchema().load({"email": "foo"})
         errors = excinfo.value.messages
         assert len(errors["email"]) == 2
-        assert "Not a valid email address." in errors["email"][0]
+        assert any("Not a valid email address." in e for e in errors["email"])
```
hukkin: I made the tests not care about error message ordering. Is the order relevant? If yes, I guess we can revert the test changes and change the `self.validators.append(validator)` lines to `self.validators.insert(0, validator)`

setup.cfg
```diff
@@ -5,7 +5,7 @@ license_files = LICENSE
 universal = 1
 
 [flake8]
-ignore = E203, E266, E501, W503, E731, B903
```
sloria, line 8: This flake8 config is meant to match the recommended config for black: https://github.com/psf/black/blob/master/.flake8 (except with the max-line-length bumped up a bit). So `ignore` should not be edited. Can you revert this?

tests/test_deserialization.py
sloria: I believe it was intentional for the `Email` and `URL` validators to run first, which is why the code did `.insert` rather than `.append`. Can you revert the change?

sloria (changes requested):
Nice cleanup. Thanks!

Just a couple small reversions and this is good to go.

setup.cfg
hukkin: I believe your link is not black's recommendation, it is simply what black project itself happens to use. The minimal black complete config can be found here https://github.com/psf/black#line-length and is simply `extend-ignore = E203`. W503 is disabled by default in flake8 so when using `extend-ignore`, only E203 needs to be ignored (plus any non-black-related ignores specific to marshmallow).

I can of course still revert if you like that better.

tests/test_deserialization.py
hukkin: :+1: reverted

setup.cfg
sloria: Ah, you're right. Thanks for clarifying

sloria (approved):

PR #1503, opened 2020-02-03
Revert Fast List(Nested)

### LINKED ISSUE #1424: List(Pluck()) raises while Pluck(many=True) works

### LINKED ISSUE #1497: Regression on list of nullable nested fields

deckar01:
`Nested(many)` does not handle `allow_none` properly due to using `Schema.load(many)`. Optimizing `List(Nested)` by reusing this code caused it to inherit this behavior, which was a regression. A regression test was added to avoid this when optimizing `List(Nested)` in the future.

Fixes #1497
Fixes #1424

Edit by @lafrech: Reverts #1306

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -576,19 +576,16 @@ class Nested(Field):
         if many and not utils.is_collection(value):
             raise self.make_error("type", input=value, type=value.__class__.__name__)
 
-    def _load(self, value, data, partial=None, many=False):
-        many = self.schema.many or self.many or many
+    def _load(self, value, data, partial=None):
         try:
-            valid_data = self.schema.load(
-                value, unknown=self.unknown, partial=partial, many=many
-            )
+            valid_data = self.schema.load(value, unknown=self.unknown, partial=partial)
         except ValidationError as error:
             raise ValidationError(
                 error.messages, valid_data=error.valid_data
             ) from error
         return valid_data
 
-    def _deserialize(self, value, attr, data, partial=None, many=False, **kwargs):
+    def _deserialize(self, value, attr, data, partial=None, **kwargs):
         """Same as :meth:`Field._deserialize` with additional ``partial`` argument.
 
         :param bool|tuple partial: For nested schemas, the ``partial``
@@ -597,8 +594,8 @@ class Nested(Field):
         .. versionchanged:: 3.0.0
             Add ``partial`` parameter.
         """
-        self._test_collection(value, many=many)
-        return self._load(value, data, partial=partial, many=many)
+        self._test_collection(value)
+        return self._load(value, data, partial=partial)
 
 
 class Pluck(Nested):
@@ -709,9 +706,6 @@ class List(Field):
     def _deserialize(self, value, attr, data, **kwargs) -> typing.List[typing.Any]:
         if not utils.is_collection(value):
             raise self.make_error("invalid")
-        # Optimize loading a list of Nested objects by calling load(many=True)
-        if isinstance(self.inner, Nested) and not self.inner.many:
-            return self.inner.deserialize(value, many=True, **kwargs)
 
         result = []
         errors = {}
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -562,17 +562,17 @@ class Nested(Field):
             if field.startswith(nested_field)
         ]
 
-    def _serialize(self, nested_obj, attr, obj, many=False, **kwargs):
+    def _serialize(self, nested_obj, attr, obj, **kwargs):
         # Load up the schema first. This allows a RegistryError to be raised
         # if an invalid schema name was passed
         schema = self.schema
         if nested_obj is None:
             return None
-        many = schema.many or self.many or many
-        return schema.dump(nested_obj, many=self.many or many)
+        many = schema.many or self.many
+        return schema.dump(nested_obj, many=many)
 
-    def _test_collection(self, value, many=False):
-        many = self.schema.many or self.many or many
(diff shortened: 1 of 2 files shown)


sloria:
Sorry, I haven't been following the related issues. On a cursory glance this seems like the right course, but it's unfortunate that it will introduce a performance regression. I defer to your and @lafrech 's judgment for merging and releasing this.

src/marshmallow/fields.py
```diff
@@ -562,33 +562,33 @@ def _nested_normalized_option(self, option_name: str) -> typing.List[str]:
...
         schema = self.schema
         if nested_obj is None:
             return None
-        many = schema.many or self.many or many
-        return schema.dump(nested_obj, many=self.many or many)
+        many = schema.many or self.many
+        try:
+            return schema.dump(nested_obj, many=many)
+        except ValidationError as exc:
+            raise ValidationError(exc.messages, valid_data=exc.valid_data) from exc
```
lafrech: I don't understand the need for this try/except block.

lafrech (approved):
This perf optim has issues. Reverting it is the right choice, functionaly.

Of course, it would be fantastic if we managed to minimize load/dump time another way. But for now, let's solve the functional issues.

Just one question, but otherwise, approved.

Approved

src/marshmallow/fields.py
deckar01: I don't remember having to manually resolve a merge conflict for this line. It wasn't in the original diff, but it was part of the upstream merge that was performed to resolve merge conflicts. I only reverted the 2 commits I added, but it looks like git reverted b922a7c063 also.

I will fix this.

lafrech:
Great.

Should we release as bugfix or minor?

This is a bugfix, but it has a perf impact on the List(Nested()) case.

lafrech:
I'm tempted to label this *not* a breaking change because there is no functional drawback. However, we could make it a minor (not patch) release because of the performance regression.

CHANGELOG proposal:

```
3.5.0 (2020-02-18)
******************

Bug fixes:

- Fix list of nullable nested fields ``List(Nested(Field, allow_none=True)`` (:issue:`1497`).
  Because this fix reverts an optimization introduced to speed-up serialization and
  deserialization of lists of nested fields, a negative impact on performance in this
  specific case is expected.
```

@sloria @deckar01

sloria:
Yep, sounds good to me

Issue #805, opened 2018-05-14
Investigate Serialization Performance


deckar01:
Profile schema serialization. Compare the JIT hooks proposed in #629 with other optimization techniques.

deckar01:
Any profiling results that are shared here should be accompanied by the code used to generate them so that they can be verified and tested on other systems.

taion:
I did some casual profiling a while ago... it looks like you need to strip away a ridiculous amount of logic to get to toasted-marshmallow-level performance. Just stuff like inlining the serialization loop and replacing the generic "get value" function with just `getattr` only made up a relatively small portion of the gap. Even doing definitely-wrong things like dropping error handling didn't make up much of the gap. I'll see if I still have the code around when I get back to my work computer on Monday.

I haven't had a chance to read through the toasted-marshmallow code, but I'm even more surprised now over how fast it is.

That said, it's pretty complicated, and it's hard to tell if it's in active use in many places, so I'm a bit scared to pull it on as a dependency for my code. But it speeds things up by a shocking amount.

cc @rowillia

deckar01:
For my testing I modified the benchmark to dump single objects, rather than `many`, since this more closely reflects the use case of a server dumping on request

Source: https://github.com/marshmallow-code/marshmallow/compare/dev...f35f908

### Marshmallow (dev)

<img width="1058" alt="Screen Shot 2019-09-11 at 10 35 14 AM" src="https://user-images.githubusercontent.com/3108007/64711852-e4702280-d47f-11e9-8d0d-768e67c7a31d.png">

### Toasted Marshmallow (latest)

<img width="1020" alt="Screen Shot 2019-09-11 at 10 28 57 AM" src="https://user-images.githubusercontent.com/3108007/64711744-bdb1ec00-d47f-11e9-9806-cd821a4e56b9.png">

---

toasted isn't just Jiting marshmallow. The readme states "Even PyPy benefits from toastedmarshmallow!". If it was just Jit, PyPy wouldn't show any significant improvements.

After inspecting the jit source, I identified a non-jit optimization that provides a significant speed up and should be reasonable to maintain. `get_value`/`get_attribute` appears to represent the entire time difference in these two profiles. toasted avoids this overhead by probing the type of the data container once, then chooses a specialized module that plucks those attributes. This would likely only require adding 2-3 small methods and not require any interface changes.

[...]

PR #1557, opened 2020-04-05
Fix accessing builtin dict's attrs for a missing field

mahenzon:
I've faced this issue, when a schema has a field with name same as builtin dict's has attribute:

```python


class MyNestedSchema(Schema):
    update = fields.Boolean(required=False)


class MySchema(Schema):
    data = fields.Nested(MyNestedSchema)


class MyObj:
    def __init__(self):
        self.data = {
            # NOT passing a value, so it's missing
            # 'update': True,
        }


def main():
    obj = MyObj()
    data = MySchema().dump(obj)
    print(data)


if __name__ == '__main__':
    main()
```

Fails with this:

```
  File "/home/khorenyan/.PyCharm2019.2/config/scratches/scratch_42.py", line 40, in <module>
    main()
  File "/home/khorenyan/.PyCharm2019.2/config/scratches/scratch_42.py", line 35, in main
    data = MySchema().dump(obj)
  File "/home/khorenyan/projs/marshmallow/src/marshmallow/schema.py", line 556, in dump
  File "/home/khorenyan/projs/marshmallow/src/marshmallow/schema.py", line 520, in _serialize
  File "/home/khorenyan/projs/marshmallow/src/marshmallow/fields.py", line 311, in serialize
  File "/home/khorenyan/projs/marshmallow/src/marshmallow/fields.py", line 566, in _serialize
    return schema.dump(nested_obj, many=many)
  File "/home/khorenyan/projs/marshmallow/src/marshmallow/schema.py", line 556, in dump
  File "/home/khorenyan/projs/marshmallow/src/marshmallow/schema.py", line 520, in _serialize
  File "/home/khorenyan/projs/marshmallow/src/marshmallow/fields.py", line 311, in serialize
  File "/home/khorenyan/projs/marshmallow/src/marshmallow/fields.py", line 1103, in _serialize
    elif value in self.truthy:
TypeError: unhashable type: 'dict'
```

It happens this way:
- When being serialized, object is passed to the `marshmallow.utils._get_value_for_key`
- because of the missing `update` field in data on the line `obj[key]` the `KeyError` exception is raised
- and on the next line the `getattr(obj, key, default)` is returned

If the `key` is not found here, it returns default, so if my field is `'update_'` or `'qwerty'`, it works as expected -- the result is returned and it's `{'data': {}}`

But the `'update'` matches the builtin dict's attribute (method) so it's returned. And that's not expected at all.

Of course, I could override the `get_attribute` method in `MyNestedSchema` this way:

```python
class MyNestedSchema(Schema):
    def get_attribute(self, obj: dict, attr: str, default: Any):
        try:
            return obj[attr]
        except KeyError:
            return default

    update = fields.Boolean(required=False)
```

But I think, that needs to be fixed.

I thought about checking if the `getattr(obj, key, default)` expression returns a method of a builtin's object, or checking a key to be one of the builtin's attrs, for example:

```python
>>> dir(dict)
['__class__', '__contains__', '__delattr__', '__delitem__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getitem__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__iter__', '__le__', '__len__', '__lt__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__setitem__', '__sizeof__', '__str__', '__subclasshook__', 'clear', 'copy', 'fromkeys', 'get', 'items', 'keys', 'pop', 'popitem', 'setdefault', 'update', 'values']
```

I'm not sure if that's a proper way to do this check. Maybe we need to check even for `isinstance` instead of the object's type, but here's my current solution.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -99,7 +99,7 @@ Contributors (chronological)
 - Harlov Nikita `@harlov <https://github.com/harlov>`_
 - `@stj <https://github.com/stj>`_
 - Tomasz Magulski `@magul <https://github.com/magul>`_
-- Suren Khorenyan `@surik00 <https://github.com/surik00>`_
+- Suren Khorenyan `@mahenzon <https://github.com/mahenzon>`_
 - Jeffrey Berger `@JeffBerger <https://github.com/JeffBerger>`_
 - Felix Yan `@felixonmars <https://github.com/felixonmars>`_
 - Prasanjit Prakash `@ikilledthecat <https://github.com/ikilledthecat>`_
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -242,7 +242,9 @@ def _get_value_for_key(obj, key, default):
 
     try:
         return obj[key]
-    except (KeyError, IndexError, TypeError, AttributeError):
+    except (KeyError, IndexError, TypeError, AttributeError) as e:
+        if isinstance(e, KeyError) and type(obj) is dict:
+            return default
         return getattr(obj, key, default)
 
 
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -800,3 +800,23 @@ def test_nested_field_many_serializing_generator():
     data = OtherSchema().dump(obj)
 
     assert data.get("objects") == [{"name": "foo"}, {"name": "bar"}]
+
+
+def test_missing_update_field_in_dict_processed_properly():
+    """
+    Checking case when a schema has 'update' field
+    and the object to be dumped is a dict
+    and the value for this field is missing
+
+    https://github.com/marshmallow-code/marshmallow/pull/1557
+    :return:
+    """
+
+    class MySchema(Schema):
+        update = fields.Boolean(required=False)
+
+    schema = MySchema()
+    data = schema.dump({})
+    assert data == {}
+    data = schema.dump({"update": True})
+    assert data == {"update": True}


AUTHORS.rst
```diff
@@ -99,7 +99,7 @@ Contributors (chronological)
 - Harlov Nikita `@harlov <https://github.com/harlov>`_
 - `@stj <https://github.com/stj>`_
 - Tomasz Magulski `@magul <https://github.com/magul>`_
-- Suren Khorenyan `@surik00 <https://github.com/surik00>`_
+- Suren Khorenyan `@mahenzon <https://github.com/mahenzon>`_
```
mahenzon, line 102: I've changed my username some time ago :smile:

src/marshmallow/utils.py
```diff
@@ -242,7 +242,9 @@ def _get_value_for_key(obj, key, default):
 
     try:
         return obj[key]
-    except (KeyError, IndexError, TypeError, AttributeError):
+    except (KeyError, IndexError, TypeError, AttributeError) as e:
+        if isinstance(e, KeyError) and type(obj) is dict:
```
deckar01, line 246: I'm not a fan of carving out an exception for `dict` from this dict-like class handling logic. [This method is a known hot path for serialization](https://github.com/marshmallow-code/marshmallow/issues/805#issuecomment-531019570). Would you consider creating an issue for this? I think there are solutions to this problem that would give users more direct control over this behavior and provide fast paths for serialization at the same time. Maybe something like `schema.dump(obj, from=dict)`.

mahenzon: Opened an issue #1559

mahenzon:
@deckar01 any ideas? it's been a long time already. If we need any changes, let's discuss them. I'm not sure about explicit types to be passed to the schemas' methods

PR #1574, opened 2020-04-21
Fix typing in class_registry

mahenzon:
It processes Schema Type, not Schema instance
https://docs.python.org/3/library/typing.html#typing.Type

Caused by #1259

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -99,7 +99,7 @@ Contributors (chronological)
 - Harlov Nikita `@harlov <https://github.com/harlov>`_
 - `@stj <https://github.com/stj>`_
 - Tomasz Magulski `@magul <https://github.com/magul>`_
-- Suren Khorenyan `@surik00 <https://github.com/surik00>`_
+- Suren Khorenyan `@mahenzon <https://github.com/mahenzon>`_
 - Jeffrey Berger `@JeffBerger <https://github.com/JeffBerger>`_
 - Felix Yan `@felixonmars <https://github.com/felixonmars>`_
 - Prasanjit Prakash `@ikilledthecat <https://github.com/ikilledthecat>`_
--- a/src/marshmallow/class_registry.py
+++ b/src/marshmallow/class_registry.py
@@ -13,14 +13,16 @@ from marshmallow.exceptions import RegistryError
 if typing.TYPE_CHECKING:
     from marshmallow import Schema
 
+    SchemaType = typing.Type[Schema]
+
 # {
 #   <class_name>: <list of class objects>
 #   <module_path_to_class>: <list of class objects>
 # }
-_registry = {}  # type: typing.Dict[str, typing.List["Schema"]]
+_registry = {}  # type: typing.Dict[str, typing.List["SchemaType"]]
 
 
-def register(classname: str, cls: "Schema") -> None:
+def register(classname: str, cls: "SchemaType") -> None:
     """Add a class to the registry of serializer classes. When a class is
     registered, an entry for both its classname and its full, module-qualified
     path are added to the registry.
@@ -64,7 +66,7 @@ def register(classname: str, cls: "Schema") -> None:
 
 def get_class(
     classname: str, all: bool = False
-) -> typing.Union[typing.List["Schema"], "Schema"]:
+) -> typing.Union[typing.List["SchemaType"], "SchemaType"]:
     """Retrieve a class from the registry.
 
     :raises: marshmallow.exceptions.RegistryError if the class cannot be found
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,13 @@
 Changelog
 ---------
 
+3.5.2 (unreleased)
+******************
+
+Bug fixes:
+
+- Fix typing in ``class_registry`` (:pr:`1574`). Thanks :user:`mahenzon`.
+
 3.5.1 (2020-03-05)
 ******************
 
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -2880,3 +2880,13 @@ class TestFromDict:
         dumped = OSchema().dump({"foo": 42, "bar": 24})
         assert isinstance(dumped, OrderedDict)
         assert "bar" not in dumped
+
+
+def test_class_registry_returns_schema_type():
+    class DefinitelyUniqueSchema(Schema):
+        """
+        Just a schema
+        """
+
+    SchemaClass = class_registry.get_class(DefinitelyUniqueSchema.__name__)
+    assert SchemaClass is DefinitelyUniqueSchema


AUTHORS.rst
```diff
@@ -99,7 +99,7 @@ Contributors (chronological)
 - Harlov Nikita `@harlov <https://github.com/harlov>`_
 - `@stj <https://github.com/stj>`_
 - Tomasz Magulski `@magul <https://github.com/magul>`_
-- Suren Khorenyan `@surik00 <https://github.com/surik00>`_
+- Suren Khorenyan `@mahenzon <https://github.com/mahenzon>`_
```
mahenzon, line 102: My username has changed since then :smiley:

src/marshmallow/class_registry.py
```diff
@@ -11,16 +11,19 @@
 from marshmallow.exceptions import RegistryError
 
 if typing.TYPE_CHECKING:
-    from marshmallow import Schema
+    # noinspection PyUnresolvedReferences
+    from marshmallow import Schema  # noqa: F401
```
mahenzon: linters 'think' that `Schema` is unused because I created a separate `SchemaType` annotation variable, so `Schema` is not used to annotate variables

src/marshmallow/class_registry.py
```diff
@@ -11,16 +11,19 @@
 from marshmallow.exceptions import RegistryError
 
 if typing.TYPE_CHECKING:
-    from marshmallow import Schema
+    # noinspection PyUnresolvedReferences
```
sloria: What is this for? Can we live without it? Not a big deal, but I'd rather not add pragmas for tools we're not using.

sloria (commented):
Thanks for the PR. Just a nit/question on there. Otherwise looks good

src/marshmallow/class_registry.py
mahenzon: This is for PyCharm (a JetBrains Python IDE). Otherwise it marks this line as unused and recommends to remove it

lafrech: Is this a check specific to PyCharm or is PyCharm using another tool underneath (pylint)?

mahenzon: Seems like it's PyCharm's specific tool: https://www.jetbrains.com/help/pycharm/disabling-and-enabling-inspections.html#comments-ref

Here's a question about connecting it with pylint: https://stackoverflow.com/questions/41979763/suppress-warning-for-both-pycharm-and-pylint
So I think that's specific

mahenzon:
So, this PR is stuck only because there's no place for some comments in marshmallow? Or what's the issue here?

sloria:
Sorry for the delay. No blockers other than maintainers' time, I'm afraid. I'll get this merged and released soon.

lafrech:
@sloria beat me to it.

I guess the PR is stuck because we lack free time to take care of it (review, merge, release).

Regarding the PyCharm comment, we don't want to open the door to all kinds of comments or .gitignore stuff for specific IDEs. I wouldn't mind introducing pylint pragmas, as pylint is a more generic tool, but even this was never discussed and would be totally arguable.

mahenzon:
> to all kinds of comments or .gitignore stuff for specific IDEs

There's already a line in .gitignore for JetBrains IDEs:
https://github.com/marshmallow-code/marshmallow/blob/784d20dbdf21891f8e3b85422954a9dedef36757/.gitignore#L42

What's the point to keep such stuff away? Developers prefer different IDEs, so why not to make contributing comfortable for everyone? It doesn't affect performance or anything else

deckar01:
https://www.python.org/dev/peps/pep-0484/#runtime-or-type-checking

According to PEP 484, importing in a `TYPE_CHECKING` block and using forward references is all that is needed. What tool is unable to handle this?

mahenzon:
This was for escaping copy-paste. Let me repeat myself (https://github.com/marshmallow-code/marshmallow/pull/1574#discussion_r412495385)

linters 'think' that `Schema` is unused because I created a separate `SchemaType` annotation variable, so `Schema` is not used to annotate variables

I could just add `typing.Type["Schema"]` in each type annotation, but for me it's bulky, better with a separate `SchemaType`

deckar01:
I don't see a description for how to reproduce the problem you were having. Without that, it is difficult to understand if this is the right way to fix it.

mahenzon:
What's wrong with my comments?
OK, take a look at this. Is it better with copy-pasted `typing.Type["Schema"]`?
But with this option we don't need to add `noqa` comments to the import statement

```python
"""A registry of :class:`Schema <marshmallow.Schema>` classes. This allows for string
lookup of schemas, which may be used with
class:`fields.Nested <marshmallow.fields.Nested>`.

.. warning::

    This module is treated as private API.
    Users should not need to use this module directly.
"""
import typing
from marshmallow.exceptions import RegistryError

if typing.TYPE_CHECKING:


# {
#   <class_name>: <list of class objects>
#   <module_path_to_class>: <list of class objects>
# }
_registry = {}  # type: typing.Dict[str, typing.List[typing.Type["Schema"]]]


def register(classname: str, cls: typing.Type["Schema"]) -> None:
    """Add a class to the registry of serializer classes. When a class is
    registered, an entry for both its classname and its full, module-qualified
    path are added to the registry.

    Example: ::

        class MyClass:
            pass

        register('MyClass', MyClass)
        # Registry:
        # {
        #   'MyClass': [path.to.MyClass],
        #   'path.to.MyClass': [path.to.MyClass],
        # }

    """
    # Module where the class is located
    module = cls.__module__
    # Full module path to the class
    # e.g. user.schemas.UserSchema
    fullpath = ".".join([module, classname])
    # If the class is already registered; need to check if the entries are
    # in the same module as cls to avoid having multiple instances of the same
    # class in the registry
    if classname in _registry and not any(
        each.__module__ == module for each in _registry[classname]
    ):
        _registry[classname].append(cls)
    elif classname not in _registry:
        _registry[classname] = [cls]

    # Also register the full path
    if fullpath not in _registry:
        _registry.setdefault(fullpath, []).append(cls)
    else:
        # If fullpath does exist, replace existing entry
        _registry[fullpath] = [cls]
    return None


def get_class(
    classname: str, all: bool = False
) -> typing.Union[typing.List[typing.Type["Schema"]], typing.Type["Schema"]]:
    """Retrieve a class from the registry.

    :raises: marshmallow.exceptions.RegistryError if the class cannot be found
        or if there are multiple entries for the given class name.
    """
    try:
        classes = _registry[classname]
    except KeyError as error:
        raise RegistryError(
            "Class with name {!r} was not found. You may need "
            "to import the class.".format(classname)
        ) from error
    if len(classes) > 1:
        if all:
            return _registry[classname]
        raise RegistryError(
            "Multiple classes with name {!r} "
            "were found. Please use the full, "
            "module-qualified path.".format(classname)
        )
    else:
        return _registry[classname][0]
```

lafrech:
Indeed, there's IDE-specific stuff in gitignore. We tend to avoid this but I guess it slipped through. In fact it dates back to 5 years ago. I'd be tempted to replace current gitignore with https://github.com/github/gitignore/blob/master/Python.gitignore.

Still, putting such lines in gitignore is not an such an annoyance. It it much more of an issue if we scatter the codebase with pragmas for every code checker.

mahenzon:
That's why I created a separate variable `SchemaType = typing.Type["Schema"]`. In this case IDE AND other linters (also that one included in the tox config) are warning about an unused import. We can refer to the PEP as much as we want, but we have 3 options here:
1. copy-paste for annotation
2. ignore linting error
3. PR to the linters / open issues for PyCharm and etc and ask for support for such cases

maybe there's another option, for example add one annotation using this import, then it'll be ok, but it's not suitable here

deckar01:
Ah, so https://github.com/marshmallow-code/marshmallow/pull/1574#discussion_r412495385 is the explanation of what the problem is. This is good information to put in the description.

Could we use `SchemaMeta` instead of `typing.Type["Schema"]`?

mahenzon:
Yeah, seems pretty suitable here. Give me a minute, I didn't think about it. Gonna push new changes

mahenzon:
But I have to note that it won't be equal. Because when we annotate with `Schema` we mean that we are dealing with an instance of `Schema`, when we annotate with `Type[Schema]`, we are dealing with the `Schema` class, not its type

mahenzon:
on a second thought.. I think `SchemaMeta` not suitable here

deckar01:
I am good with using `typing.Type["Schema"]` to keep it as strict as possible. Creating an intermediate type to reduce duplication looks good too. I would move the `SchemaType` definition into the `TYPE_CHECKING` block and keep the forward references as recommended by [PEP 484](https://www.python.org/dev/peps/pep-0484/#runtime-or-type-checking). That ensures we don't perform any unnecessary operations at runtime.

deckar01:
Once it's in the `TYPE_CHECKING` block you should be able to use `typing.Type[Schema]` and avoid the need for the pragma and ignore.

mahenzon:
Nope, it does not work this way. Linters are looking to the `Schema` to be referenced in the type annotation, not in any assignment. So even moving intermediate variable creation in the `TYPE_CHECKING` block would not help. In this case it's OR:
- intermediate variable for annotation AND comments for linters to ignore import
- copy-pasting `typing.Type["Schema"]` in all annotations, no comments for linters

lafrech:
Linter comments aside, I don't think it is _that_ bad to duplicate typing.Type["Schema"] in 3 or 4 places. In fact it may even be easier to read with the duplication.

Typing generally tends to be verbose and the eye gets used to ignoring it. And we don't use typing.Type so often (as we generally work with instances, not classes).

I'm not familiar with typing so I'll leave it to you guys. Just saying don't bother too much about the duplication.

mahenzon:
ok, this seems like a good thought

mahenzon:
OK, in this case PyCharm is OK with this format, but flake8 still thinks that it's an unused import. Still have to add a `noqa` comment

deckar01:
Defining `SchemaType` silences the flake8 unused import warning. I guess it doesn't care about unused variable definitions. https://github.com/deckar01/marshmallow/commit/8eecbaecc97cadde348926e607bb81a50e8ef23e

mahenzon:
hmm your solution seems pretty nice!

mahenzon:
Changed it. It's not an unused variable definition, because this variable is used in the annotation

deckar01:
Flake8 doesn't check forward annotations. I think it is just assuming you wanted to conditionally export it. It assumes imports aren't meant to be exported.

deckar01 (approved):

PR #1575, opened 2020-04-23
Propagate unknown parameter to nested fields

### LINKED ISSUE #1490: "unknown" option should apply to nested fields so that "exclude" works
The fact that `exclude` takes dotted notation for nested fields, but `unknown` does not apply to nested fields, means that `exclude` doesn't work for nested fields in some common use cases unless the nested schema(s) have the desired `Meta.unknown` setting, and then they are not runtime-flexible, which is the great benefit of Schema and `load` taking `exclude`.  Somewhat new to marshmallow so maybe there's a way to accomplish this (setting `exclude` on nested fields at load time) in a different way?

Happy to contribute a PR if you guys think this is a good idea.

mahenzon:
Closes #1490

For backwards-compatibility propagation of the `unknown` happens only when it was passed explicitly

I tried to propagate parent's `unknown` option to the nested, but the tests began to fail.
So, for the sake of backwards compatibility and for explicit behavior it will work only when passing the `unknown` keyword to the schema's `.load()`

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -570,16 +570,18 @@ class Nested(Field):
         if many and not utils.is_collection(value):
             raise self.make_error("type", input=value, type=value.__class__.__name__)
 
-    def _load(self, value, data, partial=None):
+    def _load(self, value, data, partial=None, unknown=None):
         try:
-            valid_data = self.schema.load(value, unknown=self.unknown, partial=partial)
+            valid_data = self.schema.load(
+                value, unknown=unknown or self.unknown, partial=partial,
+            )
         except ValidationError as error:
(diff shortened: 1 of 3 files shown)


MinuMichaelOlassayil:
When is this getting merged? Waiting to use.

deckar01:
There are a few things blocking this PR:

1. `context` is intended for user defined state, not library defined state.
2. This changes behavior, which would require a major version release. `dev` would not be the correct target branch.
3. An instance method argument is not the only way to define schema meta options in bulk. Practical solutions were linked to in #1490. Maximizing convenience does not seem to be a strong reason to bake this functionality into the API since the work arounds are simple and the use case is obscure.

PR #1581, opened 2020-05-04
Make tests not depend on execution order

hukkin:
### Issue description
Two test cases in `test_registry`, namely `test_two_way_nesting` and `test_nesting_with_class_name_many` fail if `test_serialization` test module runs before them in the same pytest invocation. This is due to `test_serialization` declaring a schema class called `ASchema` and the tests in `test_registry` assuming that a class with this name only exists in `test_registry` module.

Unit tests usually should not rely on a certain execution order.

### The fix
Make the execution order dependent tests look for `tests.test_registry.ASchema` with the full path instead of just `ASchema`.

Also add `pytest-randomly` plugin to catch these sorts of flaws in tests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ import re
 from setuptools import setup, find_packages
 
 EXTRAS_REQUIRE = {
-    "tests": ["pytest", "pytest-randomly", "pytz", "simplejson"],
+    "tests": ["pytest", "pytz", "simplejson"],
     "lint": [
         "mypy==0.770",
         "flake8==3.7.9",
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ import re
 from setuptools import setup, find_packages
 
 EXTRAS_REQUIRE = {
-    "tests": ["pytest", "pytz", "simplejson"],
+    "tests": ["pytest", "pytest-randomly", "pytz", "simplejson"],
     "lint": [
         "mypy==0.770",
         "flake8==3.7.9",
--- a/tests/test_registry.py
+++ b/tests/test_registry.py
@@ -127,17 +127,17 @@ class C:
 
 class ASchema(Schema):
     id = fields.Integer()
-    b = fields.Nested("BSchema", exclude=("a",))
+    b = fields.Nested("tests.test_registry.BSchema", exclude=("a",))
 
 
 class BSchema(Schema):
     id = fields.Integer()
-    a = fields.Nested("ASchema")
+    a = fields.Nested("tests.test_registry.ASchema")
 
 
 class CSchema(Schema):
     id = fields.Integer()
-    bs = fields.Nested("BSchema", many=True)
+    bs = fields.Nested("tests.test_registry.BSchema", many=True)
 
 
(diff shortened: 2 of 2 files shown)


sloria (approved):
Wow, good catch. Confirmed that this is an issue

lafrech:
Just wondering if pytest-random could be an annoyance sometimes.

I often find myself using `pytest -sqk test_function_name_substring` to select a test by a partial match. This sometimes catches several tests, like 4 tests while I aimed at only 1, but it is fine as they run fast and in a deterministic order so it is easy to find what I need in the output.

Randomizing test order might break this.

I totally confess my laziness. I know there are ways to pick a specific test from the command line but each time I tried to do it I found it too complicated for the benefit and I went back to the method above.

Perhaps "knowing" that the tests are independent is enough an,d we don't need to enforce it.

Or perhaps we could enforce it only in CI tests, not make it a dev dependency.

Don't bother too much about it, though. I can uninstall it locally if I find it annoying.

hukkin:
@lafrech if there truly are no inter test dependencies, then your use case should work just fine. Each test will pass regardless of whether they run alone or together with others in any combination or order. Using pytest-randomly will make sure that these dependencies get caught earlier rather than later.

I would in fact imagine that pytest-randomly makes the test suite more robust for your use case. It is possible that there is a test dependency that makes one of the 4 tests you select fail, even if they run in the same order as in the whole test suite, simply because some excluded tests dont run before or in between. Randomizing execution order will help catch these issues.

hukkin:
Ok sorry, maybe I misunderstood? Reading your message again, I think what you mean is you want to get the output always in the same order and are not worried about test failures? Then yes uninstalling or running something like `pytest -p no:randomly` or `pytest --randomly-dont-reorganize` would help.

sloria:
As a minimal first step, maybe we merge this without `pytest-randomly`? Thankfully, pytest plugins are easy to opt in to: just install them locally and they're used automatically. I do this for "quality-of-life" plugins like pytest-icdiff and pytest-notifier.

lafrech:
Yes @hukkinj1. Typically, when developing, I add prints to the code and lauch a test. I don't mind being polluted by the print outputs from a few other tests as long as the order is always the same.

    pytest -sqk string

is limited. For instance when there are several tests with a name containing the name of the test I'm running. E.g.

    test_nested_field
    test_nested_field_embedded
    test_nested_raises_error

Assuming I want to run the first test,

    pytest -sqk nested_field

runs all test.

I admit this is a user issue and I'm free to use pytest properly to select tests or disable the plugin and it makes total sense to catch those errors asap. It is a matter of cost/benefit.

As @sloria says, we should at least merge the fix. I'm lukewarm to add it as a test dep, but I wouldn't object to adding it to tox config so that it runs on CI.

hukkin:
:+1: thanks for clarification. I removed the dependency now. If we think its a good idea, can look into adding it to CI later, but right now am in a bit of a hurry.

sloria:
Thank you for working on this @hukkinj1

PR #1586, opened 2020-05-11
Use pprint from stdlib in docs

sloria:
`marshmallow.pprint` hasn't been useful since we stopped serializing to `OrderedDict`s by default (which was in 1.0!). Worse, it's not even correct--`bool`s won't be displayed correctly when printing `OrderedDict`.

We should consider deprecating/removing it. For now, I'm just removing it from the docs.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,7 +24,9 @@ marshmallow: simplified object serialization
 .. code-block:: python
 
     from datetime import date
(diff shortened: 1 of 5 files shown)


lafrech (approved):
Good.

Yeah, let's remove it in MA4 (if we drop PY5 and PY6 in MA4, all dicts will be ordered anyway).

PR #1587, opened 2020-05-11
Split-up API documentation

EpicWink:
Split API reference documentation up into multiple pages, and include member summary table using `autosummary`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,6 +14,7 @@ EXTRAS_REQUIRE = {
         "sphinx-issues==1.2.0",
         "alabaster==0.7.12",
         "sphinx-version-warning==1.1.2",
+        "autodocsumm==0.1.13",
     ],
 }
 EXTRAS_REQUIRE["dev"] = EXTRAS_REQUIRE["tests"] + EXTRAS_REQUIRE["lint"] + ["tox"]
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -147,3 +147,4 @@ Contributors (chronological)
 - Tim Gates `@timgates42 <https://github.com/timgates42>`_
 - Nathan `@nbanmp <https://github.com/nbanmp>`_
 - Ronan Murphy `@Resinderate <https://github.com/Resinderate>`_
+- Laurie Opperman `@EpicWink <https://github.com/EpicWink>`_
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,8 @@ Support:
 - Documentation: improve custom fields example (:issue:`1538`).
   Thanks :user:`pablospizzamiglio` for reporting the problem with the
   old example and thanks :user:`Resinderate` for the PR.
+- Documentation: Split up API reference into multiple pages and 
+  add summary tables (:pr:`1587`). Thanks :user:`EpicWink` for the PR.
 
 3.6.0 (2020-05-08)
 ******************
--- a/docs/api_reference.rst
+++ b/docs/api_reference.rst
@@ -6,59 +6,12 @@ API Reference
 
 .. module:: marshmallow
 
-Schema
-======
-
-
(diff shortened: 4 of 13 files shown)


setup.py
```diff
@@ -14,6 +14,7 @@
         "sphinx-issues==1.2.0",
         "alabaster==0.7.12",
         "sphinx-version-warning==1.1.2",
+        "autodocsumm==0.1.13",
     ],
```
lafrech, line 18: Do we need to add a dependency to https://github.com/Chilipp/autodocsumm ?

What do we need it for that can't be achieved with sphinx autosummary (https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html)?

lafrech (commented):
Looks good. I didn't checkout/build to test, though.

One question about the added dependency.

setup.py
EpicWink: With `sphinx.ext.autosummary`, you need to enumerate every item that you want to include in the documentation (until Sphinx 3.1 is released, where the `recursive` parameter is introduced). This means any new methods, classes and attributes need to be added/removed manually in the documentation as well.

`autodocsumm` provides a way to quickly add a summary table at the top of any [autodoc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) directive. Seeing as there are already docs-building requirements anyway (sphinx etc), another small dependency won't affect CI or the development workflow.

If you don't want the summary table added, it's easy for me to revert the addition of the summary table.

sloria:
Will this break existing links to the docs?

EpicWink:
Yes, it will. I hadn't thought of that. Any links to any of the members that aren't exposed at the top-level will now go straight to the `api_reference` page, so the user will have to look through the submodule pages.

This won't cause any 404s.

sloria:
The autosummary looks nice 👍 . But what's the purpose of splitting up the files? It would be nice if we didn't break existing links. Not the biggest deal if we do, but I'd like to weigh the benefits first.

EpicWink:
Splitting up the files makes the documentation pages a lot shorter, and so easier to navigate. If you look at the [current documentation](https://marshmallow.readthedocs.io/en/stable/api_reference.html), you'll see there is no separation of the documented items. This proposal splits them up by submodule.

sloria:
I think it's not so bad to navigate so long as there's the links in the sidebar to navigate to a given module

Image 2020-05-14 at 10 02 42 AM

Also, it's kinda nice to be able to Cmd+F the entire API reference.

FWIW, having a single page for API reference isn't uncommon: see [Flask](https://flask.palletsprojects.com/en/1.1.x/api/) and [attrs](https://www.attrs.org/en/stable/api.html).

What do you think about just adding the autosummaries for now? That seems like the biggest win here.

EpicWink:
With the Alabaster theme, that sidebar is only visible at the top of the page. I know the common theme for Python web dev is to use the Alabaster theme, but I think ReadTheDocs's theme is much better.

I would say that it is fairly uncommon to have single-page API reference is fairly uncommon, just a couple of popular web-dev libraries use it. I'm not arguing that you should or shouldn't have single-page API reference based on what other projects do, however; that decision should be made with other factors.

The problem with the `autodocsumm` summary tables is that they can only be made for each `autodoc` directive, ie for each module, at the start of that module's documentation. That means that the summary tables will be part of the way down the page. If you're fine with that, I'll create a new pull-request for just adding the summary tables.

sloria (approved):
> With the Alabaster theme, that sidebar is only visible at the top of the page

Even worse on mobile..it shows up on the _bottom_ of the page 😞 

You make valid points. I just didn't want to break existing links if we could avoid it. But we don't need to worry too much about the hash segments (`#somefunction`). I'm OK merging this as is, since it's arguably better than the status quo.

sloria:
@EpicWink Thanks for your work on this!

PR #1619, opened 2020-06-23
Include `default_error_messages` for field classes

weeix:
As a newbie who was finding a way to replace 'Not a valid email address.' with a custom error message and tried guessing the correct dict key like 'validator_failed', 'Email', 'email' (the correct one is 'invalid').

I think it would be great to include `default_error_messages` for each field classes in the API Reference.


sloria:
Thanks for the PR! Good idea to include this. Is it actually necessary to document it on every field, or does it suffice to do it on Field only?

weeix:
Documenting it on every field would be better than aggregating all
possible dict key on Field because reader won't have to guess which dict
key belong to which field. (some dict key, like 'invalid', reside in
multiple subclass so it must be generalized and reader might have to do
some guess work)

sloria:
Right, I forgot we don't use `:inherited-members:` for field classes. Fixed up and merged.

PR #1627, opened 2020-07-08
Fix usage of intersection/union for OrderedSet in Nested field

juannorris:
Fix a bug where a `Nested` field initialized with an instance of a schema (instead of a class) would fail to set the `only` and/or `exclude` properties when marshaling, due to trying to use methods that do not exist in `OrderedSet`.


juannorris:
Hi @sloria. First time contributing to the project, please let me know if I also need to create an issue with details.

I could have changed the code for the `Nested` field where it does `set_class(self.only).intersection(original)` for `set_class(self.only) & original`, but I thought mimicking the basic `set` operations methods would be a better approach. Please let me know if you think otherwise.

Thanks for this great family of libraries btw!

sloria:
Thanks for the PR!

`OrderedSet` is vendorized from Raymond Hettinger's implementation. I hesitate to have it diverge from the original implementation, as I'm not sure whether the omission of those methods was intentional or not. Can we get away without modifying `orderedset.py`?

Alternatively, you could do some research on OrderedSet's implementation to see if others have `union`, etc. implemented. I'm open to swapping out the vendorized module if improved implementations exist.

juannorris:
Thanks for the quick reply @sloria.

Yes, we can get by without modifying `orderedset.py`, by changing these lines:

https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/fields.py#L525
https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/fields.py#L528

to use the `&` and `|` set operators.

This OrderedSet library (https://github.com/simonpercivall/orderedset/blob/master/lib/orderedset/_orderedset.pyx) has these methods already implemented and is based on that same recipe, but I'm guessing you prefer not to have a dependency.

Please let me know how should I proceed, I can quickly change the code in `fields.py` and revert the changes in `orderedset.py` if that's the way to go.

src/marshmallow/orderedset.py
```diff
@@ -70,6 +70,15 @@ def pop(self, last=True):
         self.discard(key)
         return key
 
+    def union(self, other):
+        return self | other
```
sloria: This isn't quite correct, or at least it's inconsistent with regular `set`s because it doesn't cast `other`.

https://til.sloria.com/#2018-06-13-python-sets-difference-between-a-or-b-and-aunionb

sloria (commented):
I think the lowest-impact change is to use `&` and `|` in the lines you linked, and cast `original` to `set_class`. What do you think?

src/marshmallow/orderedset.py
juannorris: Oh you're right.

juannorris:
> I think the lowest-impact change is to use `&` and `|` in the lines you linked, and cast `original` to `set_class`. What do you think?

Yes, I agree with that, I'll update the PR.

sloria:
Thanks--looks good. I'll merge this when I have a moment to give it a closer look and update the changelog and such.

juannorris-santex:
Thanks, appreciate it!

juannorris:
Thanks, I appreciate it!

I've updated the title and description after the CR changes.

Issue #1281, opened 2019-07-03
Move some fields (exclude, only, etc) to private API in  `Schema.__init__`


fuhrysteve:
I'd make `exclude` and `only` private API because they are used in `_normalize_nested_options` which is called only at `__init__`.

I'd also make `load_only` and `dump_only` private because setting them after `__init__` conflicts with

```py
        self.load_only = set(load_only) or set(self.opts.load_only)
        self.dump_only = set(dump_only) or set(self.opts.dump_only)
```

Likewise with `unknown` and `context` due to

```py
        self.unknown = unknown or self.opts.unknown
        self.context = context or {}
```

(setting them to `None` after `__init__` wouldn't be the same as in `__init__`)

At this point, I'm tempted to make everything private. I mean the attributes above plus `many`, `ordered` and `partial`.

Adding the underscore would break libs using marshmallow (like apispec) but that's life. Or is there another way to signify those are private? Like, just a comment in the docstring?

_Originally posted by @lafrech in https://github.com/marshmallow-code/marshmallow/pull/1258#issuecomment-508262389_

lafrech:
Would it make sense to make them read-only rather than strictly private?

It can be useful to access them after init to sort of clone a schema. This is what we do in umongo to generate a marshmallow schema from a (marshmallow-based) model schema.

sirosen:
> Would it make sense to make them read-only rather than strictly private?

I think that would probably make everyone's life easier? Both internally and externally, there would be fewer updates to code that really just wants to read these attributes.

I think all that's needed is
```python
def __init__(self, ...):
    ...
    self._unknown = <init_value>

@property
def unknown(self):
    return self._unknown
```

And since that will presumably happen a bunch of times, it might be worth doing something like
```python
def __init__(self, ...):
    ...
    readonly_property(self, "unknown", <init_value>)

def readonly_property(...): ...  # something to figure out
```

PR #1631, opened 2020-07-15
Rework Schema._get_fields

### LINKED ISSUE #1628: Schema class fields deletion control
I see at `_get_fields` function there is a `pop` parameter that defines what to do with field attributes of schema class:
https://github.com/marshmallow-code/marshmallow/blob/09c5b26bee77be5e23927ddeb472d5ea56733266/src/marshmallow/schema.py#L54-L56
There is also `ordered` parameter here but unlike him `pop` parameter isn't accessible through the API. I think it would be nice to have the corresponding meta parameter that control schema class fields deletion.

lafrech:
Closes #1628.

- Don't pop fields from schema class
- Hardcode base schema class to `base.FieldABC`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,10 +4,14 @@ Changelog
 3.12.0 (unreleased)
 *******************
 
+Features:
+
+- Let ``Field``s be accessed by name as ``Schema`` attributes (:pr:`1631`).
+
 Other changes:
 
-- Improve types in ``marshmallow.validate``.
-- Make `marshmallow.validate.Validator` an abstract base class.
+- Improve types in ``marshmallow.validate`` (:pr:`1786`).
+- Make `marshmallow.validate.Validator` an abstract base class (:pr:`1786`).
 - Remove unnecessary list cast (:pr:`1785`).
 
 3.11.1 (2021-03-29)
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -38,21 +38,17 @@ from marshmallow.warnings import RemovedInMarshmallow4Warning
 _T = typing.TypeVar("_T")
 
 
-def _get_fields(attrs, field_class, pop=False, ordered=False):
+def _get_fields(attrs, field_class, ordered=False):
     """Get fields from a class. If ordered=True, fields will sorted by creation index.
 
     :param attrs: Mapping of class attributes
     :param type field_class: Base field class
-    :param bool pop: Remove matching fields
     """
     fields = [
         (field_name, field_value)
         for field_name, field_value in attrs.items()
         if is_instance_or_subclass(field_value, field_class)
     ]
-    if pop:
-        for field_name, _ in fields:
-            del attrs[field_name]
     if ordered:
         fields.sort(key=lambda pair: pair[1]._creation_index)
     return fields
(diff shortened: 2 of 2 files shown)


sloria:
I believe the existing behavior was taken from DRF: https://github.com/encode/django-rest-framework/blob/599e2b183db846a632b1efd148e6bc97d926ee5c/rest_framework/serializers.py#L284-L287 .

I'm not necessarily opposed to changing it, but we should first understand the rationale for DRF's decision. @lafrech @decaz care to do some research?

lafrech:
I asked on DRF bug tracker: https://github.com/encode/django-rest-framework/issues/7428.

Removing the pop does not break tests on their codebase either.

lafrech:
The rationale in DRF is that the objects are self-serializing: the business object class itself is the schema. In this case, accessing an attribute, the user expect the attribute value, not the field.

This does not apply in marshmallow where the schema is separated from the object.

I suppose this PR could break a lib that would use marshmallow to build self-serializable objects by inheriting schemas. I'm reluctant to label this a breaking change, though. I'm not sure this use case is even possible. umongo, for instance, hides the schema in a class attribute without direct inheritance.

lafrech:
Should we merge this or do we want to wait for v4?

sloria (approved):
Ah sorry this fell off my radar. Thanks for investigating this. I agree that the rationale for popping in DRF does not apply here, so this is good to merge.

Please also update the changelog.

lafrech:
I'm afraid this introduces a conflict between field names and schema attributes. See #1808.

We shall check / list / document all attributes and methods. At least those that are not underscored.

Edit: #1808 is a bit specific because the name shouldn't be accessed in the schema in the first place. It is being addressed. Still, there may be conflicts between field names and schema attributes.

lafrech:
There's a fair amount of public attributes in `Schema`. This sort of contradicts this "feature".

Maybe accessing them through a `fields` attribute was better after all.

We may want to reconsider this for marshmallow 4.

Note that we also planned to make more attributes private (#1281).

Edit: Well, not really if we keep them as read-only, but we could namespace those attributes into an "attributes" structure rather than the fields in a "fields" structure.

Overall, I'm really not sure this was the best move.

AFAIU, though, this doesn't break anything but expectations. I mean fields don't shadow attributes. The only issues is that:

- Attributes shadow fields. A user may call a field one of the many attribute names (e.g. "partial") and be surprised not to get a field when accessing it as attribute. This is a doc issue, but it is also a design smell.
- Code relying on the non-existence of an attribute may fail. This should be quite marginal (#1808, https://github.com/marshmallow-code/apispec/issues/673).

lafrech:
Actually, field names don't shadow attributes set at init, but they do shadow class attributes, which includes methods such as load, dump, etc.

```py
import marshmallow as ma


class Schema(ma.Schema):

    i = ma.fields.Int()
    load = ma.fields.Int()
    fields = ma.fields.Int()
    exclude = ma.fields.Int()


s = Schema()

print(s.i)
print(s.load)
print(s.fields)
print(s.exclude)

print(Schema.i)
print(Schema.load)
print(Schema.fields)
print(Schema.exclude)
```

This tests shows that when access attributes from `Schema` we always get the field, and when accessing from `s` (instance), we get a field for i and load and the real attribute for fields and exclude.

We could come up with nasty ways to work around this, like setting class attributes as instance attributes at init, but at this point, I'm leaning towards reverting this change (only the commit exposing fields by name, not the refactor one).

PR #1632, opened 2020-07-16
Pass unknown parameter to pre_load, post_load, and validates_schema methods

jforand:
When extending a schema, the value of self.unknown is not necessarily correct in pre_load methods.

```
class MySchema(Schema):

    ...

    @pre_load
    def this_is_wrong(self, data, **kwargs):
        print(self.unknown)
        return data


...
MySchema().load(data, unknown=INCLUDE)
```

will print 'raise' rather than 'include'.


sloria:
sorry it took so long to review this. i agree that `unknown` should be passed to the load processors, for consistency with `many` and `partial`. i've updated the PR to also pass unknown to validate_schema methods.

sloria (approved):

PR #1633, opened 2020-07-16
Boolean can handle unhashable type

jun0jang:
Excepted: 
```python
class BrandSchema():
    has_new_product = fields.Boolean(attribute="new_products")

res = BrandSchema().dump({"new_products": [Product(), ...])
# {"has_new_product": True}
```

but it raise unhashable type exception because `value in self.truthy`.
so add defense code. thank you.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -151,3 +151,4 @@ Contributors (chronological)
 - Ram Rachum `@cool-RR <https://github.com/cool-RR>`_
 - `@weeix <https://github.com/weeix>`_
 - Juan Norris `@juannorris <https://github.com/juannorris>`_
+- 장준영 `@jun0jang <https://github.com/jun0jang>`_
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,14 @@
 Changelog
 ---------
 
+3.7.1 (unreleased)
+******************
+
+Bug fixes:
+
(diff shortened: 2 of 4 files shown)


sloria:
We try to avoid isinstance checks when serializing as they can impact performance when serializing many objects. For now I'd suggest something along the lines of

```python
class BrandSchema():
    has_new_product = fields.Function(lambda obj: bool(obj['new_projects']))
```

Will that address your use case?

jun0jang:
actually now i did same using method field.  but i think function field is for much more complex thing.
primitive type field is readability and maintenance than function field.

i agree performance is very important. for libraries, it's even more important.

jun0jang:
actually now i did same using method field.  but i think function field is for much more complex thing.
primitive type field is readability and maintenance than function field.

i agree performance is very important. but it's also important to work well.
so how did you think about doing this?

```python
try:
    if value in self.truthy:
        return True
    elif value in self.falsy:
        return False
except TypeError:
    pass

return bool(value)
```

sloria:
I think that solution makes sense. We use a similar approach for `validate.OneOf` and `validate.NoneOf`.

sloria:
I'm unsure if this should be considered as a breaking change. Previously `Boolean` would reject lists and dicts (albeit ungracefully). This changes accepts those. Perhaps this should go into 4.0?

@lafrech @deckar01 wdyt?

lafrech:
I say non-breaking because

- by API contract, we don't validate on dump 
- it used to raise an exception

Issue #1549, opened 2020-03-25
Implement a field.Set?


mikegrima:
I was curious why Marshmallow didn't have a `field.Set` for Python sets.

I would imagine the code would be mirror identical to `List` (https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/fields.py#L656) only with a `list` being serialized into a `set`, and deserialized back into a `list`.

Would this be easy to implement? Any pitfalls I should be made aware of?

sloria:
I'm not opposed to adding a `Set` field, though we should probably do something along the lines of what @deckar01 has proposed here: https://github.com/marshmallow-code/marshmallow/issues/1263#issuecomment-503755620 . That would make it easier to implement any custom iterable field in userland.

lafrech:
We generally have fields matching the deserialized type, with parameters to customize the serialized form.

I guess there has not been much incentive to serialize lists as sets because

- you lose the order, so it does not round-trip
- users often serialize as json, which means lists become arrays

Unless you meant the opposite.

A `Set` field could indeed be added. It would **serialize** as list and **deserialize** as set.

It could be based on an `Iterable` base field. See #1263.

mikegrima:
@lafrech It would be de-serializing from JSON. The implementation would be similar to Pydantic's.

PR #1643, opened 2020-08-09
Add fields.Iterable as a base class for collection field types

### LINKED ISSUE #1263: Add fields.Iterable as a base class for collection field types
I'd like to deserialize into some types that aren't json serializable, for example a frozenset of integers, `frozenset({1,2,3})`.

I can think of a few options in the current API:

- define a custom `FrozensetField` type and define its `_serialize` and `_deserialize` methods,
- use `fields.Function(deserialize=frozenset, serialize=list)`,
- define a custom replacement for the json module.

Is `fields.Function` the recommended strategy here? I'm unsure because

- it doesn't take a `cls_or_instance` parameter like `fields.List` does, so it's not obvious how to make a `set` *of* some other objects. For example, how do I define a frozenset of `fields.Nested(OtherSchema)`?
- it makes  `Function` the main focus of attention, instead of the data.

Alternatively, an API that comes to mind is something like

```python
class SomeSchema(marshmallow.Schema):
    x = fields.Integer()
    y = fields.String()

class MoreSchema(marshmallow.Schema):
    values = fields.List(fields.Nested(SomeSchema), deserialize=frozenset)
    data = fields.Mapping(
        keys=fields.String(), 
        values=fields.Integer(), 
        deserialize=collections.OrderedDict
    )
```

Like `fields.Function`, these might also take a `serialize` argument, defaulting to `list` and `dict`, respectively. 

This would generalize the existing idea of [`Mapping.mapping_type`](https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/fields.py#L1272) , and would make it an instance attribute rather than a class attribute.

What is the recommended approach for deserializing to `Frozenset[int]` or `Frozenset[MyObject]`? Do you think an API like the idea above would be useful?

Fongshway:
Resolves https://github.com/marshmallow-code/marshmallow/issues/1263, based on what @deckar01 proposed in https://github.com/marshmallow-code/marshmallow/issues/1263#issuecomment-503755620. This PR also helps address https://github.com/marshmallow-code/marshmallow/issues/1549.

- `fields.List` is updated to inherent from `fields.Iterable`, since the majority of the `fields.List` code could be reused.
- Example of creating custom iterable field classes:

```python

class FrozenSet(fields.Iterable):
    iterable_type = frozenset

class ImmutableList(fields.Iterable):
    iterable_type = tuple

```

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -681,8 +681,8 @@ class Iterable(Field):
     .. versionadded:: 3.12.0
     """
 
-    serialization_type = list
-    deserialization_type = list
+    serialization_type: typing.Type[typing.Iterable] = list
+    deserialization_type: typing.Type[typing.Iterable] = list
 
     #: Default error messages.
     default_error_messages = {
@@ -712,17 +712,17 @@ class Iterable(Field):
 
     def _serialize(
         self, value, attr, obj, **kwargs
-    ) -> typing.Optional[typing.List[typing.Any]]:
+    ) -> typing.Optional[typing.Iterable[typing.Any]]:
         if value is None:
             return None
         result = [self.inner._serialize(each, attr, obj, **kwargs) for each in value]
         return (
             result
             if self.serialization_type is list
-            else self.serialization_type(result)
+            else self.serialization_type(result)  # type: ignore
         )
 
-    def _deserialize(self, value, attr, data, **kwargs) -> typing.List[typing.Any]:
+    def _deserialize(self, value, attr, data, **kwargs) -> typing.Iterable[typing.Any]:
         if not utils.is_collection(value):
             raise self.make_error("invalid")
 
@@ -740,7 +740,7 @@ class Iterable(Field):
         return (
             result
             if self.deserialization_type is list
-            else self.deserialization_type(result)
+            else self.deserialization_type(result)  # type: ignore
         )
 
 
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -69,6 +69,12 @@ __all__ = [
 ]
 
 _T = typing.TypeVar("_T")
+IterableType = typing.Union[
+    typing.Type[typing.FrozenSet],
+    typing.Type[typing.List],
+    typing.Type[typing.Set],
+    typing.Type[typing.Tuple],
+]
 
 
 class Field(FieldABC):
@@ -666,8 +672,8 @@ class Iterable(Field):
     #: Default error messages.
     default_error_messages = {"invalid": "Not a valid iterable."}
 
-    serialization_type = list  # type: typing.Type[typing.Iterable[typing.Any]]
-    deserialization_type = list  # type: typing.Type[typing.Iterable[typing.Any]]
+    serialization_type = list  # type: IterableType
+    deserialization_type = list  # type: IterableType
 
     def __init__(self, cls_or_instance: typing.Union[Field, type], **kwargs):
         super().__init__(**kwargs)
@@ -698,7 +704,7 @@ class Iterable(Field):
         result = [self.inner._serialize(each, attr, obj, **kwargs) for each in value]
         if self.serialization_type is list:
             return result
-        return self.serialization_type(result)  # type: ignore
+        return self.serialization_type(result)
 
     def _deserialize(self, value, attr, data, **kwargs) -> typing.Iterable[typing.Any]:
         if not utils.is_collection(value):
@@ -717,7 +723,7 @@ class Iterable(Field):
             raise ValidationError(errors, valid_data=result)
         if self.deserialization_type is list:
             return result
-        return self.deserialization_type(result)  # type: ignore
+        return self.deserialization_type(result)
 
 
 class List(Iterable):
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -11,6 +11,7 @@ import math
 import typing
 import warnings
 from collections.abc import Mapping as _Mapping
+from typing_extensions import Protocol
 
 from marshmallow import validate, utils, class_registry, types
 from marshmallow.base import FieldABC, SchemaABC
@@ -69,12 +70,6 @@ __all__ = [
 ]
 
 _T = typing.TypeVar("_T")
-IterableType = typing.Union[
-    typing.Type[typing.FrozenSet],
-    typing.Type[typing.List],
-    typing.Type[typing.Set],
-    typing.Type[typing.Tuple],
-]
 
 
 class Field(FieldABC):
@@ -654,6 +649,16 @@ class Pluck(Nested):
         return self._load(value, data, partial=partial)
 
 
+class _SerializationType(Protocol):
+    def __call__(self, value: typing.List) -> typing.Iterable:
+        pass
+
+
+class _DeserializationType(Protocol):
+    def __call__(self, value: typing.List) -> typing.Iterable:
+        pass
+
+
 class Iterable(Field):
     """An abstract class for iterables.
 
@@ -672,8 +677,8 @@ class Iterable(Field):
     #: Default error messages.
     default_error_messages = {"invalid": "Not a valid iterable."}
 
-    serialization_type = list  # type: IterableType
-    deserialization_type = list  # type: IterableType
+    serialization_type = list  # type: typing.ClassVar[_SerializationType]
+    deserialization_type = list  # type: typing.ClassVar[_DeserializationType]
 
     def __init__(self, cls_or_instance: typing.Union[Field, type], **kwargs):
         super().__init__(**kwargs)
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -675,16 +675,18 @@ class Iterable(Field):
     Example: ::
 
         class FrozenSet(Iterable):
-            iterable_type = frozenset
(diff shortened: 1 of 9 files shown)


lovasoa:
@sloria do you have a feedback on that ? It would be useful for https://github.com/lovasoa/marshmallow_dataclass/issues/99

lafrech:
I just had a quick look.

Random thoughts:

- This adds a cast to each serialization/deserialization operation. Performance impact ?
- Should the type be used only on deserialization or also on serialization as in this PR ? I generally think of the field type as the type of the data in object world, not in serialized world.

KevinMFong:
Thanks for the feedback @lafrech !

> This adds a cast to each serialization/deserialization operation. Performance impact ?

I updated the code to only apply the cast on each (de)serialization operation if the iterable type is not `list`. For anyone curious about the performance impact of the cast, here's what I found by doing some back of the envelop checks in IPython:
```
n = 1_000_000
%timeit foo = [i*i for i in range(n)]  # 71.8 ms ± 799 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit bar = list([i*i for i in range(n)])  # 84.2 ms ± 1.45 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit baz = list(i*i for i in range(n))  # 87.9 ms ± 2.31 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
Based on these results, I decided it was better to keep the list comprehension and only cast non-list iterables instead of using a generator expression within the cast.

> Should the type be used only on deserialization or also on serialization as in this PR ? I generally think of the field type as the type of the data in object world, not in serialized world.

I went back and forth on this, especially since there is some discussion in https://github.com/marshmallow-code/marshmallow/issues/1263 around exposing the deserialization and serialization types separately in the `Iterable` field. Here are some of my reasons for applying the type on both:
- When serializing to a Python dictionary, the least surprising behavior for users would be to use the corresponding iterable type for a given field.
- Consistency with how the `Decimal` field works, which (de)serializes to the Python `decimal.Decimal` type. And since users already need to think about which Marshmallow fields are serializable by the default `json` module, it seemed reasonable to serialize to things like sets and tuples, even if they don't have corresponding JSON types.

Fongshway:
Thanks for the feedback @lafrech !

> This adds a cast to each serialization/deserialization operation. Performance impact ?

I updated the code to only apply the cast on each (de)serialization operation if the iterable type is not `list`. For anyone curious about the performance impact of the cast, here's what I found by doing some back of the envelop checks in IPython:
```
n = 1_000_000
%timeit foo = [i*i for i in range(n)]  # 71.8 ms ± 799 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit bar = list([i*i for i in range(n)])  # 84.2 ms ± 1.45 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit baz = list(i*i for i in range(n))  # 87.9 ms ± 2.31 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
Based on these results, I decided it was better to keep the list comprehension and only cast non-list iterables instead of using a generator expression within the cast.

> Should the type be used only on deserialization or also on serialization as in this PR ? I generally think of the field type as the type of the data in object world, not in serialized world.

I went back and forth on this, especially since there is some discussion in https://github.com/marshmallow-code/marshmallow/issues/1263 around exposing the deserialization and serialization types separately in the `Iterable` field. Here are some of my reasons for applying the type on both:
- When serializing to a Python dictionary, the least surprising behavior for users would be to use the corresponding iterable type for a given field.
- Consistency with how the `Decimal` field works, which (de)serializes to the Python `decimal.Decimal` type. And since users already need to think about which Marshmallow fields are serializable by the default `json` module, it seemed reasonable to serialize to things like sets and tuples, even if they don't have corresponding JSON types.

Fongshway:
@sloria @lafrech any interest in merging this PR? If not, I'll just close it.

sloria:
Thanks @Fongshway and sorry for the delay in reviewing. I think this would be a good feature to have, just haven't had time to give this the attention it deserves. Appreciate your patience 🙏 

@lafrech Will you have time to review this within the next week?

Fongshway:
@sloria no worries and totally understand. Thank you for all of your work on this project!

src/marshmallow/fields.py
```diff
@@ -701,7 +701,8 @@ def _serialize(
     ) -> typing.Optional[typing.List[typing.Any]]:
         if value is None:
             return None
-        return [self.inner._serialize(each, attr, obj, **kwargs) for each in value]
+        result = [self.inner._serialize(each, attr, obj, **kwargs) for each in value]
+        return result if self.iterable_type == list else self.iterable_type(result)
```
lafrech: Nitpick: I think the test should be 

```py
if self.iterable_type is list
```

lafrech (commented):
Could be worth adding a test showing how to subclass to use a set or a frozenset. We could even support those in the core, but this can be done in another PR.

I'm still concerned about the fact this applies to both deserialization and serialization.

As I wrote in [this comment](https://github.com/marshmallow-code/marshmallow/issues/1549#issuecomment-809123977), I generally see the field name as expressing the deserialized type, while parameters can be used to tweak serialization.

```py
class FooField(Field):  # Will deserialiaze as Foo instance.
    __init__(self,..., serialize_as="baz"):  # Will serialize as baz
        ...
```

This vision might be wrong. Other maintainers POV welcome.

Anyway, at this stage, it could be safer to use 2 different class attributes to define each type.

```py
class Iterable(Field):
    serialization_type = list
    deserialization_type = list
```

Fongshway:
@lafrech thanks! I'll give other maintainers a chance to opine on applying the field type for both deserialization and serialization and will plan on making updates this weekend.

src/marshmallow/fields.py
```diff
@@ -710,12 +710,15 @@ def _bind_to_schema(self, field_name, schema):
...
         self, value, attr, obj, **kwargs
-    ) -> typing.Optional[typing.List[typing.Any]]:
+    ) -> typing.Optional[typing.Iterable[typing.Any]]:
         if value is None:
             return None
-        return [self.inner._serialize(each, attr, obj, **kwargs) for each in value]
+        result = [self.inner._serialize(each, attr, obj, **kwargs) for each in value]
+        if self.serialization_type is list:
+            return result
+        return self.serialization_type(result)  # type: ignore
```
Fongshway: I _think_ this is a `mypy` issue. Here's the error I get without the ignore:
```
src/marshmallow/fields.py:719: error: Too many arguments for "Iterable"
```

Fongshway:
@lafrech I added tests showing how to subclass `fields.Iterable` using `frozenset` and updated the implementation to use separate `serialization_type` and `deserialization_type` class attributes. There are also a couple of other minor edits. Let me know if you have any additional feedback!

Fongshway:
@sloria @lafrech just wanted to follow up and see if you had any comments on my second round of edits from April.

tests/frozenset_field.py
```diff
@@ -0,0 +1,7 @@
+from marshmallow import fields
+
+
+class FrozenSet(fields.Iterable):
```
lafrech: I'd put this in tests/base.py

src/marshmallow/fields.py
```diff
@@ -730,7 +733,38 @@ def _deserialize(self, value, attr, data, **kwargs) -> typing.List[typing.Any]:
...
+
+    .. versionchanged:: 2.0.0
+        The ``allow_none`` parameter now applies to deserialization and
+        has the same semantics as the other fields.
+
+    .. versionchanged:: 3.0.0rc9
+        Does not serialize scalar values to single-item lists.
+
+    .. versionchanged:: 3.12.0
+        Inherits from Iterable instead of Field.
```
lafrech: I don't think this is worth mentioning.

`Mapping` docstring mentions the version in which it was added, but `Dict` does not mention when it inherited from it.

src/marshmallow/fields.py
lafrech: Strange. I'd be happy to get advice from someone more familiar with typing/mypy.

lafrech (changes requested):
Sorry about the delay.

Thanks for your work on this.

Two minor change requests.

I don't know about the mypy issue. Help welcome.

Fongshway:
> Sorry about the delay.
> 
> Thanks for your work on this.

No problem! I appreciate you taking the time to revisit this PR.

> Two minor change requests.

Done.

> I don't know about the mypy issue. Help welcome.

After sleeping on this, I think I found a better solution. Let me know what you think of the current implementation.

lafrech:
Thanks for the rework.

I'm not comfortable with redefining Iterable type.

```py
_T = typing.TypeVar("_T")
IterableType = typing.Union[
    typing.Type[typing.FrozenSet],
    typing.Type[typing.List],
    typing.Type[typing.Set],
    typing.Type[typing.Tuple],
]
```

Unfortunately, I can't investigate this right now.

Hint welcome, anyone.

Fongshway:
> I'm not comfortable with redefining Iterable type.

I took another stab using `Protocol`. Since `Protocol` was introduced in Python 3.8, I had to add the `typing-extensions` dependency for backwards computability. https://github.com/python/mypy/issues/5485 has some more context on why I chose to use `Protocol` instead of `Callable`.

Issue #686, opened 2017-10-09
Add format option for fields.time


BennyAlex:
ISO 8601 supports following formats:
> hh:mm:ss.sss or hhmmss.sss
> hh:mm:ss or hhmmss
> hh:mm or hhmm
> hh

It would be nice to have a format option for fields.time, like 
`begin = fields.Time(format='hh:mm')`

chkdmin:
I using custom fields for arrow type in marshmallow serialize/deserialize
``` python
class Arrow(Field):
    default_timezone = 'UTC'
    default_format = 'YYYY-MM-DD HH:mm:ss'
    default_error_messages = {
        'invalid': 'Invalid datetime value {value}'
    }

    def __init__(self, format_=None, timezone=None, to_timezone=None, **kwargs):
        super(Arrow, self).__init__(**kwargs)
        self.format = format_ or self.default_format
        self.timezone = timezone or self.default_timezone
        self.to_timezone = to_timezone

    def _serialize(self, value, attr, obj):
        if not value:
            return value
        return value.to(self.default_timezone).format(self.format)

    def _deserialize(self, value, attr, data):
        if value is None:
            return None
        try:
            v = arrow.get(value, self.format).replace(tzinfo=self.timezone)
            if self.to_timezone:
                return v.to(self.to_timezone)
            return v
        except (TypeError, ValueError, ParserError):
            return self.fail('invalid', value=value)
```
``` python
created_at = fields.Arrow(to_timezone='UTC', format_='hh:mm', required=False)
```
hope this helps

pranav93:
I wrote little dirty one for serialising,
```python
class TimeField(fields.Raw):
    def format(self, value):
        try:
            if value:
                return value.isoformat()
        except Exception as exc:
            raise MarshallingException(exc)

# usage
account_tz_format = {
    'id': fields.Integer,
    'target_country_timezone_id': fields.Integer,
    'start_time': TimeField,
    'end_time': TimeField,
    'is_active': fields.Boolean
}

```

wumadeline:
@sloria can I take a stab at this issue?

sloria:
@wumadeline Great! Thanks for taking it on.

infinityxxx:
@sloria could you take a look at this PR https://github.com/marshmallow-code/marshmallow/pull/1682?

PR #1645, opened 2020-08-10
Add format option for fields.Time

wumadeline:
Addresses https://github.com/marshmallow-code/marshmallow/issues/686

Adding a `format` param to the `Time` field, which supports the following formats:
- `hh`
- `hh:mm`
- `hh:mm:ss`
- `hh:mm:ss.sss`

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1295,6 +1295,9 @@ class AwareDateTime(DateTime):
 class Time(Field):
     """ISO8601-formatted time string.
 
+    :param format: ISO8601 string format of the value. Can either be
+        ``"hh"``, ``"hh:mm"``, ``"hh:mm:ss"``, or ``"hh:mm:ss.sss"``.
+        If not specified, defaults to built-in ``isoformat`` de/serialization.
     :param kwargs: The same keyword arguments that :class:`Field` receives.
     """
 
@@ -1304,10 +1307,14 @@ class Time(Field):
         "format": '"{input}" cannot be formatted as a time.',
     }
 
+    def __init__(self, format: str = "", **kwargs):
+        super().__init__(**kwargs)
+        self.format = format
+
     def _serialize(self, value, attr, obj, **kwargs):
         if value is None:
             return None
-        ret = value.isoformat()
+        ret = utils.to_iso_time(value, self.format)
         if value.microsecond:
             return ret[:15]
         return ret
@@ -1317,7 +1324,7 @@ class Time(Field):
         if not value:  # falsy values are invalid
             raise self.make_error("invalid")
         try:
-            return utils.from_iso_time(value)
+            return utils.from_iso_time(value, self.format)
         except (AttributeError, TypeError, ValueError) as error:
             raise self.make_error("invalid") from error
 
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1295,8 +1295,7 @@ class AwareDateTime(DateTime):
 class Time(Field):
     """ISO8601-formatted time string.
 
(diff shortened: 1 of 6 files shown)


src/marshmallow/utils.py
```diff
@@ -196,6 +231,25 @@ def to_iso_date(date: dt.date) -> str:
...
 
+def to_iso_time(time: dt.time, format: str = "") -> str:
+    """Return the ISO8601-formatted representation of a time object.
+
+    :param time: The time.
+    :param format: An ISO8601 time format string.
+    """
+    if format:
+        format_exp = {
+            "hh": "%H",
```
sloria: thought: Would it make sense for users to just pass the strftime format (`%H`, `%H:%M`, etc)?

Seems like that would keep consistency with the `format` argument of `fields.DateTime`.

sloria:
Thanks for the PR! I've only skimmed, but it looks good so far. Initial thought is about the format of the argument: I'm not sure if we should use `hh:MM` vs strftime formatting. What do you think?

src/marshmallow/utils.py
lafrech: Absolutely. User should input strftime format. Let's not invent our own format.

wumadeline:
Yes, agreed. Pushed a commit with those changes. I initially implemented the `hh:MM:..` formatting because those were listed in the issue description, but in retrospect `strftime` formatting makes way more sense 🙂

src/marshmallow/fields.py
```diff
@@ -1295,19 +1295,36 @@ def _deserialize(self, value, attr, data, **kwargs):
...
+        "%H:%M",
+        "%H:%M:%S",
+        "%H:%M:%S.%f",
+    ]
+
+    def __init__(self, format: typing.Optional[str] = None, **kwargs):
+        super().__init__(**kwargs)
+        self.format = format
+
+        if self.format and self.format not in Time.VALID_ISO_STRFTIME_FORMATS:
```
sloria, line 1321: thought: I'm not sure we should constrain the set of strftime formats; `fields.DateTime` does not do this, and we should keep this consistent.

src/marshmallow/utils.py
```diff
@@ -161,17 +161,28 @@ def from_iso_datetime(value):
...
+def from_iso_time(value, format: typing.Optional[str] = None):
     """Parse a string and return a datetime.time.
 
     This function doesn't support time zone offsets.
     """
+    invalid_format_value_error = ValueError("Not a valid ISO8601-formatted time string")
+
+    if format:
+        try:
+            return dt.datetime.strptime(value, format).time()
```
sloria, line 173: suggestion: Maybe this strptime logic should go in `fields.Time`?

That would keep things consistent with `fields.DateTime`.

src/marshmallow/utils.py
```diff
@@ -196,6 +207,18 @@ def to_iso_date(date: dt.date) -> str:
     return dt.date.isoformat(date)
 
 
+def to_iso_time(time: dt.time, format: typing.Optional[str] = None) -> str:
+    """Return the ISO8601-formatted representation of a time object.
+
+    :param time: The time.
+    :param format: An ISO8601 time format string.
+    """
+    if format:
```
sloria, line 216: suggestion: Related to above--move this to `fields.Time`?

sloria (commented):
Looking good! Just added a few drive-by thoughts that I have low confidence in. Curious to hear your thoughts.

By the way, you can use `tox -e lint` to format and lint your code. That should fix the build.

sloria:
Superseded by https://github.com/marshmallow-code/marshmallow/pull/1682 . Thanks for your work on this!

PR #1649, opened 2020-08-16
Optimise dumping to reduce unnecessary overhead

dsimidzija:
**NB:** Please take this MR more as a "starting a discussion" than "production-ready" code, because I hacked a lot of things.

When dumping many objects, marshmallow is calling the same field methods
over and over again, which return the same values. Parts of this process
can be called only once per dump, which reduces python method call
overhead significantly.

`Field.get_serializer` returns the optimized serializer for the current
dump operation, avoiding the expensive lookups for properties which will
not change during a single dump (such as `data_key`, `default`, etc)

Also, the default `Schema.get_attribute` is also not used because all it
does is calling `utils._get_value_for_key(s)`.

Benchmarks show around 30-35% improvement, which is quite significant even for this hacky patch:

```
T1: python benchmark.py
  Before: 395.60 usec/dump
  After:  261.04 usec/dump
T2: python benchmark.py --object-count 1000
  Before: 22508.80 usec/dump
  After:  14610.63 usec/dump
T3: python benchmark.py --iterations=5 --repeat=5 --object-count 20000
  Before: 442295.61 usec/dump
  After:  288202.98 usec/dump
T3: python benchmark.py --iterations=10 --repeat=10 --object-count 10000
  Before: 220163.94 usec/dump
  After:  142475.76 usec/dump
```

My motivation here is that marshmallow is excellent when it comes to schema validation, but according to the [benchmarks](https://voidfiles.github.io/python-serialization-benchmark/), there is a lot of overhead in there. The question is, is there a better way of improving serialization performance without sacrificing all the good things about marshmallow?

### DIFF
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -394,6 +394,9 @@ class Schema(base.SchemaABC, metaclass=SchemaMeta):
         self.fields = {}  # type: typing.Dict[str, ma_fields.Field]
         self.load_fields = {}  # type: typing.Dict[str, ma_fields.Field]
         self.dump_fields = {}  # type: typing.Dict[str, ma_fields.Field]
+        self.dump_serializers = (
+            self.dict_class()
+        )  # type: typing.Dict[str, typing.Callable]
         self._init_fields()
         messages = {}
         messages.update(self._default_error_messages)
@@ -466,7 +469,7 @@ class Schema(base.SchemaABC, metaclass=SchemaMeta):
         """
         pass
 
-    def get_attribute(self, obj: typing.Any, attr: str, default: typing.Any):
+    def default_get_attribute(self, obj: typing.Any, attr: str, default: typing.Any):
         """Defines how to pull values from an object to serialize.
 
         .. versionadded:: 2.0.0
@@ -475,6 +478,8 @@ class Schema(base.SchemaABC, metaclass=SchemaMeta):
             Changed position of ``obj`` and ``attr``.
         """
         return get_value(obj, attr, default)
+
+    get_attribute = default_get_attribute
 
     ##### Serialization/Deserialization API #####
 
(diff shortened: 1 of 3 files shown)


src/marshmallow/schema.py
```diff
@@ -511,19 +514,41 @@ def _serialize(
...
+                None
+            )  # type: typing.Optional[typing.Callable[[typing.Any, str, typing.Any], typing.Any]]
+            if self.get_attribute != self.default_get_attribute:
+                accessor = self.get_attribute
+
+            for field_name, field_obj in self.dump_fields.items():
+                key = (
+                    field_obj.data_key if field_obj.data_key is not None else field_name
+                )
+                self.dump_serializers[key] = field_obj.get_serializer(
```
deckar01, line 529: It looks like the test failure is being caused by `dump_serializers` being an unordered dict. Currently the ordering is dependent on `self.dump_fields`, which gets initialized to an ordered dict in [schema.py#L1008](https://github.com/marshmallow-code/marshmallow/blob/821c9316be173380f38780da94e26e5e92bf7a70/src/marshmallow/schema.py#L1008), but this PR iterates on `dump_serializers` instead.

Adding `self.dump_serializers = self.dict_class()` in this block should avoid the ordering issue for python3.5.

deckar01:
This lines up pretty well with the discussion in #805. I think the main difference is that we also proposed caching logic that depends on the object type. I think the scope of this PR would be a good incremental improvement.

src/marshmallow/schema.py
dsimidzija: Thanks, I figured as much! I don't have py35/6 on my system, otherwise I would have caught this before pushing :)

dsimidzija:
I didn't even know about #805, good to know! My ideas were something similar: figure out a way to cache values for certain types at least. For example, if I know that a specific field is a string property and it is 100% there, there should be no reason to have complicated lookups or enforcing types. I didn't have time to delve deeper into the various Field subclasses, but `line_profiler` implies a lot of repetition there.

As I'm typing this, I wonder if it would be possible to do something like [lima is doing](https://github.com/b6d/lima/blob/develop/lima/schema.py#L76), and compile a single function which dumps the entire object, where each Field subclass could in theory produce the "most optimized" serializer for itself.

dsimidzija:
I don't know if this is of any interest, but [with some minimal changes](https://github.com/dsimidzija/marshmallow/commit/fe8d898f7304c60b658ef609dd326e07cfcaaef7), marshmallow can be cythonized with [setuptools-cythonize](https://pypi.org/project/setuptools-cythonize/) and together with this MR, the performance is around ~45% better!

lafrech:
I'm not familiar with cythonize. Does cythonization itself - without this PR - bring a significant improvement?

Is there anything we could do to make it available more easily? Like, could/should we distribute binary packages? Would it be useful to many users or is it a niche?

dsimidzija:
I tested it without this MR a long time ago, so I don't have the numbers, but IIRC it did bring around 10% of speed on its own.  But I guess that should be examined in more detail.

I'm honestly not sure how niche it is, I started looking into it because I ran into these performance problems when dumping large(ish) datasets with marshmallow, but didn't want to give up dynamic schemas & validation that it provides. I haven't had the time yet, but I was hoping to look into speeding up individual fields as well, I feel like there is a lot of overhead there which should be easy to eliminate without breaking anything.

sloria:
closing this for now as it's stale. let's continue discussion on #805 . nice work on https://github.com/dsimidzija/python-althaia btw!

PR #1664, opened 2020-09-10
New Feature: `fields.nestedfield`

jbowen7:
A decorator to write inline nested schema

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -152,3 +152,4 @@ Contributors (chronological)
 - `@weeix <https://github.com/weeix>`_
 - Juan Norris `@juannorris <https://github.com/juannorris>`_
 - 장준영 `@jun0jang <https://github.com/jun0jang>`_
+- John Bowen `@jbowen7 <https://github.com/jbowen7>`_
--- a/docs/nesting.rst
+++ b/docs/nesting.rst
@@ -303,6 +303,102 @@ If the object to be marshalled has a relationship to an object of the same type,
     #     }
     # }
 
+.. _inline-nesting:
+
+Nesting Inline Schema Classes With a Decorator
+------------------------------
+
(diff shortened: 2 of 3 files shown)


jbowen7:
This PR adds a new decorator `nestedfield` to define nested schema with inline classes.

It's an optional way of building nested schema, that can help with organization.
Building nested schema inline with `Schema.from_dict` can get very unsightly, especially
when trying to add `post_dump, post_load, pre_dump, pre_load` decorators. 

```py
class APIRequest(Schema):
    url = fields.Url()
    method = fields.String(missing="POST", validate=validate.OneOf(("POST", "GET")))
  
    @fields.nestedfield()
     class headers(Schema):
         Authorization = fields.String()
         X - attr = fields.String()  

    @fields.nestedfield(required=True)
    class body(Schema):
        some_foo = fields.String()
        some_bar = fields.String()
  
         @fields.nestedfield()
         class pageSort(Schema):
             page = fields.Integer(missing=1)
             size = fields.Integer()
  
        @fields.nestedfield(required=True, many=True, list_required=True, list_validate=validate.Length(1, 100))
        class someResource(Schema):
            id = fields.UUID(required=True)
            name = fields.String()
```

sloria:
This is a neat idea--thanks for sharing it!

I don't think the added API surface (+ docs and tests to maintain) is worth adding to marshmallow core at this time. I'd be open to adding it to the Examples page of the docs, or perhaps you could release this as a separate package (looks like marshmallow-inlinenested is available on PyPI  😃 ).

jbowen7:
Yeah, that's a better idea. I didn't even poll the community to see if it's something that people would want.
I'll go ahead and package it under `marshmallow-inlinenested` as suggested.
We can see if people find it useful and you can determine if you want to bring it into core in the future.

Thanks Steven

PR #1666, opened 2020-09-15
Fix #1663 Typing - Use explicit optional

### LINKED ISSUE #1663: Remove usage of implicit Optional
> A past version of this PEP allowed type checkers to assume an optional type when the default value is None, as in this code:
> 
> def handle_employee(e: Employee = None): ...
> 
> This would have been treated as equivalent to:
> 
> def handle_employee(e: Optional[Employee] = None) -> None: ...
> 
> This is no longer the recommended behavior. Type checkers should move towards requiring the optional type to be made explicit.

https://www.python.org/dev/peps/pep-0484/

We currently use implicit `Optionals`. mypy still allows them but may change this in the future.

nadege:
To fix #1663
Also added the mypy option to check for implicit optional.

### DIFF
--- a/src/marshmallow/base.py
+++ b/src/marshmallow/base.py
@@ -7,6 +7,7 @@ These are necessary to avoid circular imports between core.py and fields.py.
     This module is treated as private API.
     Users should not need to use this module directly.
 """
+import typing
 
 
 class FieldABC:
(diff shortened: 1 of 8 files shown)


src/marshmallow/base.py
```diff
@@ -31,16 +32,29 @@ def _deserialize(self, value, attr, data, **kwargs):
 class SchemaABC:
     """Abstract base class from which all Schemas inherit."""
 
-    def dump(self, obj, *, many: bool = None):
+    def dump(self, obj, *, many: typing.Optional[typing.Optional[bool]] = None):
```
sloria:
```suggestion
    def dump(self, obj, *, many: typing.Optional[bool] = None):
```

Here and elsewhere

sloria (commented):
Thanks for the PR. Can you fix the "doubly-wrapped" `Optional` usages here?

src/marshmallow/base.py
nadege: oops, I'll do that :)

nadege: fixed

lafrech:
Thanks @nadege. Looks good. Do you mind rebasing (conflict in AUTHORS.rst file)?

nadege:
rebased :)

Issue #466, opened 2016-05-31
question: add `allow_none=True` in Meta


dbrumley:
Hi,
I'm trying to use `Meta` instead of re-declaring fields from a marshmallow-sqlalchemy project.  I'd like to declare some fields optional. I don't see a way to do that in Meta, so I was trying to add it to initialization a la:

```
class UserSchema(ma.Schema):
    class Meta:
        # fields from SQL model. Keeping it DRY with Meta
        fields = ('username', 'first_name')

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.fields['first_name'].allow_none = True

# inspecting user_schema.fields['first_name'].allow_none is True after this line
user_schema = UserSchema(strict=True)
...< a different file>

from user.models import user_schema

# if I check user_schema.fields['first_name'].allow_none, it's still True here 
# in this new file outside the class UserAPI

class UserAPI(Resource):
    @login_required
    def post(self):
           # For some reason user_schema.fields['first_name'].allow_none = False here. No idea why!
            user, errors=user_schema.load(json_data)

```

When I inspect `user_schema`, it seems that this sets `allow_none=True` as expected. 

But when I use it as part of `flask-restless`, for some reason I cannot explain the value is reset to `False` in my API extension above. I don't touch the object values anywhere, yet it mysteriously changes values!  No idea what I'm missing.
- Is there a better way to set `allow_none` than the above (I'd like to be DRY wrt to the model definition)
- Any idea why the value would be reset to false? What am I missing?

pkopac:
Hi! I have a similar problem, I'd like to allow_none globally, but it's not possible without some hacking.
It would be really cool, if Meta class had this param and Fields/Schema respected it. I didn't find a fast&easy way how to implement it though. Would you consider implementing it, please?

sloria:
We are holding off on adding `dump_only` as a class Meta options for the reasons given [here](https://github.com/marshmallow-code/marshmallow/issues/344#issuecomment-285226593) and [here](https://github.com/marshmallow-code/marshmallow/issues/453#issuecomment-280951726). 

> It's not clear that these class Meta options (required, dump_only...) belong in marshmallow core, since they are only relevant for libraries that auto-generate fields. This has been pointed out here. So for now we won't add required as a class Meta option and leave it up to consuming libraries to add it.

You can add `allow_none` as a class Meta option like so:

```python
from marshmallow import Schema, SchemaOpts, fields

[...]

Issue #344, opened 2015-11-26
Handle `required` (list/set) attribute of a Meta-class in Schema-inherited classes


frol:
The rationale behind the subject is that when I use webargs in combnation with marshmallow-sqlalchemy, I can only change `dump_only`, `load_only` options of the fields. I workaround this now in an overloaded `__init__` method, but I think it would be great if it is supported out-of-the-box.

``` python
class MyParameters(ModelSchema):
    class Meta:
        model = MyModel
        dump_only = ('id', )
        required = ('name', )
```

Note, while I see that my usecase is mostly valid for marshmallow-sqlalchemy `ModelSchema`, I found that `dump_only` logic is handled here, in marshmallow base project.

frol:
BTW, some switch for all-fields-required, or even make this the default behaviour (though this was discussed in #112 earlier) would be also helpful.

sloria:
I would have to give this more thought. 

There is a proposal/PR on marshmallow-sqlalchemy for customizing marshmallow fields by passing kwargs to SQLA fields: https://github.com/marshmallow-code/marshmallow-sqlalchemy/pull/22 .

lanzafame:
I am not sure that I entirely understand @frol's use case but my own sent me looking for a solution similar to what @frol describes.

My use case looks like the following:

``` python
class FooSchema(Schema):
    name = fields.Str(required=True)
    country = fields.Str(required=True)
    city = fields.Str(required=True)
    operator = fields.Str(required=True)
    campus = fields.Str(required=True)
    facility = fields.Str(required=True)
```

I would like to refactor the above to following:

``` python
class FooSchema(Schema):
    class Meta:
        required = ("name", "country", "city", "operator", "campus", "facility")
```

Just my 2 cents.

dbrumley:
+1 for Lanzafame's use case and idea; this is what I was thinking as well. Feels like `Str` is duplication.

sloria:
We don't plan on adding this in the near future, as it adds complexity to the API (> 1 way to do things), so closing this for now. Feel free to reopen an issue if you feel further discussion is necessary.

frol:
@sloria it is not the >1 way to do things, it might be the only way to do things when it comes to Marshmallow-SQLAlchemy, where the schema is inferred. Currently, we implement this support ourselves in our own base Schema class patching the auto-generated fields in `__init__`...

sloria:
[...]

PR #1670, opened 2020-09-25
Allow to define required fields in schema constructor and meta

mgetka:
Working with marshmallow I appreciated, that I can define `load_only` and `dump_only` fields via schema constructor and Meta. Also, I missed a feature that would allow to define required fields in similar way. This PR adds such functionality.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -550,6 +550,7 @@ class Nested(Field):
                     context=context,
                     load_only=self._nested_normalized_option("load_only"),
                     dump_only=self._nested_normalized_option("dump_only"),
+                    required=self._nested_normalized_option("required"),
                 )
         return self._schema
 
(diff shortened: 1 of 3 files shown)


mgetka:
Ok, so I can see that the topic is pretty controversial. My insights stem from discussion in the  following issues: #466, #344, #453. @lafrech  [commented](https://github.com/marshmallow-code/marshmallow/issues/453#issuecomment-280948445) in one of those issues that v3 could be a good moment to deprecate `load_only`, and `dump_only` meta options. At the moment, we are deep into v3 line and the depreciation did not happened.

Personally, I have found `load_only` and `dump_only` meta/constructor options extremely useful, and I am not working on any kind of automated schema creation tool. I'm using it to create specialized loader/dumper objects in which fields become `dump_only` or `load_only` based on authorization context. I would also like to mark some fields `required` in similar manner.

I would be grateful to receive some feedback on this topic. If `dump_only` and `load_only` meta/constructor options are still considered technical debt since 2017, maybe some actions on theirs depreciation could be made. For me it would be a clear signal no to utilize them in my projects. On the other hand, if those options are now considered a legitimate part of the API, including of `required`, as well as `allow_none`, could be considered.

Issue #1674, opened 2020-10-05
Passing an incorrect method name to Method


ghost:
Hi all, this question is in reference to making a typo for method fields such as:

```python
class MySchema(Schema):
    # Method constructor argument is misspelled but actual method is not
    foobar = fields.Method("get_footbar")

    def foobar(self, obj):
         return obj.some_other_data()
```

It will yield an error `ValueError: Object None is not callable.` from the Marshmallow utility `callable_or_raise()`:

```
  File ".../lib/python3.8/site-packages/marshmallow/schema.py", line 557, in dump
    result = self._serialize(processed_obj, many=many)
  File ".../lib/python3.8/site-packages/marshmallow/schema.py", line 521, in _serialize
    value = field_obj.serialize(attr_name, obj, accessor=self.get_attribute)
  File ".../lib/python3.8/site-packages/marshmallow/fields.py", line 312, in serialize
    return self._serialize(value, attr, obj, **kwargs)
  File ".../lib/python3.8/site-packages/marshmallow/fields.py", line 1663, in _serialize
    method = utils.callable_or_raise(
  File ".../lib/python3.8/site-packages/marshmallow/utils.py", line 285, in callable_or_raise
    raise ValueError("Object {!r} is not callable.".format(obj))
ValueError: Object None is not callable.
```

I dug a bit of digging and found that the method is being accessed from the schema class via `getattr(self.parent, self.serialize_method_name, None)`. 

Is there a reason for the `None`? Without it Python would be up an `AttributeError` with a message containing the class name and the attribute name, but with it Marshmallow simply passes up the `ValueError` which doesn't give much information on what went wrong.

Here is the code where it happens for reference, grabbed from the recent dev source:

https://github.com/marshmallow-code/marshmallow/blob/0689e5522638816ef8c24f1c3e9e6f9f6f15b126/src/marshmallow/fields.py#L1713-L1728

sloria:
I think this might be a bit of legacy. I agree that we should allow the `AttributeError` to be raised.

Since raising an `ValueError` is not part of the documented API, I think we can go ahead and change this. PRs welcome!

ghost:
@sloria great, thanks! I opened a PR https://github.com/marshmallow-code/marshmallow/pull/1675 for the issue

ghost:
Completed via merge of https://github.com/marshmallow-code/marshmallow/pull/1675

PR #1675, opened 2020-10-06
Raise AttributeError when missing Schema methods

ghost:
Currently mistyping a schema method name when using `fields.Method` will result in a `ValueError` rather than throwing an `AttributeError` that details what went wrong:

```python
class MySchema(Schema):
    # Method constructor argument is misspelled but actual method is not
    foobar = fields.Method("get_footbar")

    def foobar(self, obj):
         return obj.some_other_data()
```

```
  File ".../lib/python3.8/site-packages/marshmallow/schema.py", line 557, in dump
  File ".../lib/python3.8/site-packages/marshmallow/schema.py", line 521, in _serialize
  File ".../lib/python3.8/site-packages/marshmallow/fields.py", line 312, in serialize
  File ".../lib/python3.8/site-packages/marshmallow/fields.py", line 1663, in _serialize
    method = utils.callable_or_raise(
  File ".../lib/python3.8/site-packages/marshmallow/utils.py", line 285, in callable_or_raise
    raise ValueError("Object {!r} is not callable.".format(obj))
ValueError: Object None is not callable.
```

With this change the error would yield something like:

```python
AttributeError: 'MySchema' object has no attribute 'get_footbar'
```

https://github.com/marshmallow-code/marshmallow/issues/1674

### DIFF
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -287,7 +287,7 @@ def set_value(dct: typing.Dict[str, typing.Any], key: str, value: typing.Any):
 def callable_or_raise(obj):
     """Check that an object is callable, else raise a :exc:`ValueError`."""
     if not callable(obj):
-        raise ValueError("Object {!r} is not callable.".format(obj))
+        raise TypeError("Object {!r} is not callable.".format(obj))
     return obj
 
 
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -285,7 +285,7 @@ def set_value(dct: typing.Dict[str, typing.Any], key: str, value: typing.Any):
 
 
 def callable_or_raise(obj):
-    """Check that an object is callable, else raise a :exc:`ValueError`."""
+    """Check that an object is callable, else raise a :exc:`TypeError`."""
     if not callable(obj):
         raise TypeError("Object {!r} is not callable.".format(obj))
     return obj
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -470,7 +470,7 @@ class TestFieldSerialization:
             bad_field = fields.Method("invalid")
 
         u = User("Foo")
-        with pytest.raises(ValueError):
+        with pytest.raises(AttributeError):
             BadSerializer().dump(u)
 
     def test_method_field_passed_serialize_only_is_dump_only(self, user):
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -112,7 +112,7 @@ class TestFieldSerialization:
         assert "FOOBAR" == field.serialize("key", user)
 
     def test_function_field_passed_uncallable_object(self):
-        with pytest.raises(ValueError):
+        with pytest.raises(TypeError):
             fields.Function("uncallable")
 
     def test_integer_field(self, user):
@@ -416,7 +416,7 @@ class TestFieldSerialization:
         assert field.serialize("none", user) is None
 
     def test_function_with_uncallable_param(self):
-        with pytest.raises(ValueError):
+        with pytest.raises(TypeError):
             fields.Function("uncallable")
 
     def test_email_field_serialize_none(self, user):
@@ -489,7 +489,7 @@ class TestFieldSerialization:
             bad_field = fields.Method("foo")
 
         u = User("Foo")
-        with pytest.raises(ValueError):
+        with pytest.raises(TypeError):
             BadSerializer().dump(u)
 
     # https://github.com/marshmallow-code/marshmallow/issues/395
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -415,10 +415,6 @@ class TestFieldSerialization:
         assert field.serialize("falsy", user) is False
         assert field.serialize("none", user) is None
 
-    def test_function_with_uncallable_param(self):
-        with pytest.raises(TypeError):
-            fields.Function("uncallable")
-
     def test_email_field_serialize_none(self, user):
         user.email = None
         field = fields.Email()
@@ -469,9 +465,8 @@ class TestFieldSerialization:
         class BadSerializer(Schema):
             bad_field = fields.Method("invalid")
 
-        u = User("Foo")
         with pytest.raises(AttributeError):
-            BadSerializer().dump(u)
+            BadSerializer()
 
     def test_method_field_passed_serialize_only_is_dump_only(self, user):
         field = fields.Method(serialize="method")
@@ -488,9 +483,8 @@ class TestFieldSerialization:
             foo = "not callable"
             bad_field = fields.Method("foo")
 
-        u = User("Foo")
         with pytest.raises(TypeError):
-            BadSerializer().dump(u)
+            BadSerializer()
 
     # https://github.com/marshmallow-code/marshmallow/issues/395
     def test_method_field_does_not_swallow_attribute_error(self):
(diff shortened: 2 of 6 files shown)


lafrech:
Shouldn't the `callable_or_raise` check be performed at init time?

- Better for performance
- Better for debug, no runtime error

Or do we expect users to patch their schema and add the method after init?

ghost:
> Shouldn't the `callable_or_raise` check be performed at init time?
> 
> * Better for performance
> * Better for debug, no runtime error
> 
> Or do we expect users to patch their schema and add the method after init?

@lafrech I just pushed a commit 5969cf4 that adds a `Method._bind_to_schema()` override and performs the callable checks there rather than having them be performed at serialization/deserialization time. Let me know your thoughts!

Currently `callable_or_raise()` will raise `ValueError` if the attribute type is incorrect (not callable). Shouldn't this be a [`TypeError` based on the use case?](https://docs.python.org/3/library/exceptions.html#TypeError)

I don't think patching the schema afterwards should be expected behavior, but if they were to swap the attribute after init and make it non-callable they would receive `TypeError` in that situation rather than `ValueError` (with the changes of 5969cf4).

ghost:
Hi @lafrech 

I've just updated `callable_or_raise()` in c2764b7 to raise `TypeError` rather than `ValueError` as to be consistent with python when objects are not callable:

```python
>>> 1()
TypeError: 'int' object is not callable
```

src/marshmallow/utils.py
```diff
@@ -280,7 +280,7 @@ def set_value(dct: typing.Dict[str, typing.Any], key: str, value: typing.Any):
 def callable_or_raise(obj):
     """Check that an object is callable, else raise a :exc:`ValueError`."""
```
lafrech: s/ValueError/TypeError/

tests/test_serialization.py
```diff
@@ -356,7 +356,7 @@ def test_boolean_field_serialization(self, user):
         assert field.serialize("none", user) is None
 
     def test_function_with_uncallable_param(self):
-        with pytest.raises(ValueError):
+        with pytest.raises(TypeError):
             fields.Function("uncallable")
```
lafrech: Isn't this test a duplicate of `test_function_field_passed_uncallable_object`?

tests/test_serialization.py
```diff
@@ -412,7 +412,7 @@ class BadSerializer(Schema):
             bad_field = fields.Method("invalid")
 
         u = User("Foo")
-        with pytest.raises(ValueError):
+        with pytest.raises(AttributeError):
             BadSerializer().dump(u)
```
lafrech: Is the call to `dump()` necessary for the error to be raised?

tests/test_serialization.py
```diff
@@ -431,7 +431,7 @@ class BadSerializer(Schema):
             bad_field = fields.Method("foo")
 
         u = User("Foo")
-        with pytest.raises(ValueError):
+        with pytest.raises(TypeError):
             BadSerializer().dump(u)
```
lafrech: Is the call to `dump()` necessary for the error to be raised?

lafrech (commented):
Thanks.

Looks good to me, apart from a few comments.

I don't think it qualifies as a breaking change so it could be introduced in the next 3.x.

I don't mind a second reviewer on this.

tests/test_serialization.py
ghost: Yeah looks like it, I'll go ahead and remove this one

tests/test_serialization.py
ghost: Good catch, no `dump()` will not be necessary anymore

ghost:
@lafrech thank you! Updated in 124fad7 to address the comments

lafrech:
While we're at it, wouldn't it make sense to get the methods in `_bind_to_schema` and cache them in attributes?

```py
    def __init__(self, serialize: str = None, deserialize: str = None, **kwargs):
        # Set dump_only and load_only based on arguments
        kwargs["dump_only"] = bool(serialize) and not bool(deserialize)
        kwargs["load_only"] = bool(deserialize) and not bool(serialize)
        super().__init__(**kwargs)
        self.serialize_method_name = serialize
        self.deserialize_method_name = deserialize
        self._serialize_method = None
        self._deserialize_method = None

    def _bind_to_schema(self, field_name, schema):
        if self.serialize_method_name:
            self._serialize_method = utils.callable_or_raise(getattr(schema, self.serialize_method_name))

        if self.deserialize_method_name:
            self._deserialize_method = utils.callable_or_raise(getattr(schema, self.deserialize_method_name))

        super()._bind_to_schema(field_name, schema)

        if not self._serialize_method:
            return missing_
        return self._serialize_method(obj)

        if self._deserialize_method:
            return self.deserialize_method(value)
        return value
```

ghost:
> While we're at it, wouldn't it make sense to get the methods in `_bind_to_schema` and cache them in attributes?

Good call, yeah that's a much better approach and also addresses the attribute availability thing I mentioned before: users shouldn't monkey-patch schema methods. I updated this in b48e7da

There's one issue I encountered with that though is there is this test:

```python
def test_method_field_deserialize_only(self):
        class MethodDeserializeOnly(Schema):
            def lowercase_name(self, value):
                return value.lower()

        m = fields.Method(deserialize="lowercase_name")
        m.parent = MethodDeserializeOnly()

        assert m.deserialize("ALEC") == "alec"
```

From the way this test is written, `m` never has `_bind_to_schema()` called because it's a field completely outside the declaration of the schema itself... is this something that's supported by Marshmallow? 

I'm a bit confused on the use case, but I rewrote the test to utilize the Schema's `load()` instead:

```python
def test_method_field_deserialize_only(self):
        class MethodDeserializeOnly(Schema):
            name = fields.Method(deserialize="lowercase_name")

            def lowercase_name(self, value):
                return value.lower()

        assert MethodDeserializeOnly().load({"name": "ALEC"})["name"] == "alec"
```

Let me know your thoughts

lafrech:
This test already relied on

```py
    m.parent = MethodDeserializeOnly()
```

which is basically what `_bind_to_schema` does.

Apart from this PR, it would be a good idea to rewrite the test to use `_bind_to_schema` explicitly IMHO.

There's no way this field can be independent of the schema.

I'm fine with this change.

Note that before your last commit, the use case was broken already because the field didn't go through the callable or raise test, and this wasn't detected due to the test being badly written. Another reason to rewrite it.

Is the independent field use case supported? Well, I've been looking through the fields module, and the other fields actually using bind_to_schema are `DateTime` to get a format from the field class, and container/nested fields to pass `_bind_to_schema`. `DateTime` field doesn't need it to work.

So I suppose it works and could be supported in the general case but it makes no sense, by definition, for the `Method` field so I wouldn't mind making it an exception if the rule was to support the use case.

Regarding your implementation with `hasattr` vs. my proposal, unless there is a reason I'm missing, I tend to apply the general rule of declaring all attributes in `__init__`. pylint, for instance, complains about setting attributes in other methods as it is considered bad practice. Totally opinion-based. My proposal should be rewritten `if self._serialize_method is None`, though, to be more explicit and to skip a boolean evaluation.

ghost:
@lafrech awesome, thanks for the explanation!

I force pushed the last commit to 6d0edf8 to swap the `hasattr()` for the `None` init + `is not None` checks, I agree it's much better to explicitly create the attribute if it's expected to be used

lafrech (approved):
Great. Thanks for your work on this.

One last thing. Please also add yourself to AUTHORS file and this should be good to go.

Still, requesting review from @sloria.

ghost:
@sloria forgot that `Authors.rst` got updated recently, so just fixed the conflict. Should be good to review now!

sloria (approved):

PR #1676, opened 2020-10-09
Add serialization with timezone to AwareDateTime

theirix:
Allows AwareDateTime field to optionally serialize naive `datetime` with default timezone.
Behaviour is optional and enabled with `use_serialization=True` in ctor.

For example, if date-time is stored in a database without a timezone and we want to treat it as UTC datetime, new serializer helps a lot.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1284,6 +1284,8 @@ class AwareDateTime(DateTime):
     :param default_timezone: Used on deserialization. If `None`, naive
         datetimes are rejected. If not `None`, naive datetimes are set this
         timezone.
+    :param bool use_serialization: If `True`, naive datetimes are set to
+        `default_timezone` on serialization. If `False`, datetimes are not changed.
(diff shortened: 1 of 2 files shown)


odesenfans:
I was just looking for this change! I'm wondering: for some use cases, it makes sense IMO to localize aware datetimes as well (your fix just converts naive datetimes and leaves aware datetimes untouched). For example, using PostgreSQL + SQLAlchemy, datetimes are retrieved in the timezone of the DB connection (UTC by default). 

It would be cool to ignore this implementation detail by just specifying my preferred output timezone in the schema and be done with it. What do you think?

theirix:
Sorry, missed your comment. I think, your use case is better handled by creating a third `DateTime` descendant that stores a `timezone` passed in ctor and forces conversion an aware `datetime` (from DB) to this timezone for both serialization and deserialization. `NaiveDateTime` and `AwareDateTime` are a sort of mutually exclusive in handling awareness so it is not clear which class would handle proposed logic.

PR #1680, opened 2020-10-19
Test against Python 3.9

sloria:


lafrech:
Wondering if it is worth testing on all those versions or if we could just test on min and max (3.5 and 3.9). To shorten execution time and limit CPU waste.

sloria:
I have no strong opinion on it. FWIW, the execution time wouldn't be affected much since we get 10 parallel builds with AP.

lafrech:
> FWIW, the execution time wouldn't be affected much since we get 10 parallel builds with AP.

Fair enough.

CPU waste point still stands. I can't think of an issue what would appear in 3.7 then disappear in 3.8 for instance, so I don't think those tests are useful. But maybe the waste is peanuts and not worth considering.

PR #1685, opened 2020-11-02
Cast to mapping type in Mapping.serialize/deserialize

lafrech:
When no key/value fields are specified in a `Dict` field, the value is passed as is when loading or dumping.

This means that the dump is still related to the object and modifying the dump result modifies the object. My understanding is that marshmallow is meant to prevent that and guarantee that the dumped/loaded value is of proper type and not linked to the original value.

This triggered [this issue](https://github.com/Scille/umongo/issues/306).

This PR fixes this by always casting to mapping type even if no key/value fields are specified.

We could remove the line and rely on the generic logic below but I suppose it makes sense to keep this line as a shortcut for the "loose dict" case (no field specified), as a slight performance improvement. I don't mind either way. Dropping code is nice as well and the performance improvement might be limited.

Also, I had to remove two broken tests. Those tests check that `Dict` field pass invalid dicts transparently. I don't see the point of this. In practice, it only works in the "loose dict" case. Besides, `List` chokes on invalid iterables, for instance. I don't understand why we added those tests in the first place. I didn't dig in the history to find out.

------------------------------------------------------

~Finally, mypy complained about unused ignores, so I removed them.~

Edit: Looks like there's a mypy/pre-commit issue. When committing, mypy (launched by pre-commit) complained about unused ignores, so I removed the ignores, but then it would complain when run on the whole code. I put the ignores back and committed with `--no-verify`. Could it be mypy is executed by pre-commit on modified files, not all the files, and therefore it complains about unused stuff?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,16 @@
 Changelog
 ---------
 
+3.9.1 (unreleased)
+******************
+
+Bug fixes:
+
+- Cast to mapping type in ``Mapping.serialize`` and ``Mapping.deserialize``
+  (:pr:`1685`).
+- Fix bug letting ``Dict`` pass invalid dict on deserialization when no key or
+  value ``Field`` is specified (:pr:`1685`).
+
 3.9.0 (2020-10-31)
 ******************
 
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1505,7 +1505,7 @@ class Mapping(Field):
         if value is None:
             return None
         if not self.value_field and not self.key_field:
-            return value
+            return self.mapping_type(value)
 
         #  Serialize keys
         if self.key_field is None:
@@ -1532,7 +1532,7 @@ class Mapping(Field):
         if not isinstance(value, _Mapping):
             raise self.make_error("invalid")
         if not self.value_field and not self.key_field:
-            return value
+            return self.mapping_type(value)
 
         errors = collections.defaultdict(dict)
 
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -719,8 +719,13 @@ class TestFieldDeserialization:
         assert excinfo.value.args[0] == msg
 
     def test_dict_field_deserialization(self):
+        data = {"foo": "bar"}
         field = fields.Dict()
-        assert field.deserialize({"foo": "bar"}) == {"foo": "bar"}
+        load = field.deserialize(data)
+        assert load == {"foo": "bar"}
+        # Check load is a distinct object
+        load["foo"] = "baz"
+        assert data["foo"] == "bar"
         with pytest.raises(ValidationError) as excinfo:
             field.deserialize("baddict")
         assert excinfo.value.args[0] == "Not a valid mapping type."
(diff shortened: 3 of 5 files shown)


sloria:
> Could it be mypy is executed by pre-commit on modified files, not all the files, and therefore it complains about unused stuff?

Yes, I've also seen this strange behavior when mypy executes on single files vs all files. Not sure what the issue is ¯\_(ツ)_/¯

sloria (approved):
Makes sense to me.

Will you update the changelog then squash-n-merge?

lafrech:
I'd rather not squash since the two commits are independent. But I don't really mind if you do.

I pushed a changelog commit.

Concerning the removed test, I see this change as a bugfix because the tested behaviour is wrong, so I labeled it as such.

sloria:
I've no strong preference for squashing. Thanks for updating the changelog.

lafrech:
I like atomic meaningful commits to be preserved in the main branch, so I squash if the branch is a history of trial/error iterating to the result but I generally keep commits when they make sense individually.

drake-mer:
@sloria Sorry but this just breaks with python 3.9

drake-mer:
not sure if it's 3.9 related but it breaks

sloria:
@dkremer-ledger Would you mind opening an issue with more details, e.g. traceback and a repro case?

decaz:
I faced with the following problem: I need to dump the object which has attributes that cannot be modified easily (I need to modify some of them into dictionaries) so I can't modify them with `pre_dump`. Therefore I use `post_dump` for this and modify output data. And with marshmallow==3.9.0 it worked. But with marshmallow==3.9.1 there is type casting and now it doesn't work.

For instance:
```python
from marshmallow import Schema, fields, post_dump

class Foo:
    bar = 1

    def __setattr__(self, key, value):
        """Restricting setting of attribute to emulate SQLAlchemy behavior.

        For instance, SQLAlchemy can raise the following exception if list-like 
        attribute of ORM model object is being modified into dict:
            `TypeError: Incompatible collection type: dict is not list-like`
        """
        raise TypeError('Forbidden!')

class S(Schema):
    bar = fields.Dict()

    @post_dump
    def modify_bar(self, data, **kwargs):
        data['bar'] = {'foo': 'bar'}
        return data

obj = Foo()
schema = S()
print(schema.dumps(obj))
```
marshmallow==3.9.0:
```
$ python xxx.py
{"bar": {"foo": "bar"}}
```
marshmallow==3.9.1:
```
$ python xxx.py
    ...
    return self.mapping_type(value)
TypeError: 'int' object is not iterable
```
If I use `pre_dump` instead of `post_dump`:
```python
@pre_dump
def modify_bar(self, obj, **kwargs):
    obj.bar = {'foo': 'bar'}
    return obj
```
then I'll get `TypeError: Forbidden!` (look at `Foo.__setattr__`).

lafrech:
Your code worked because `Dict` was too permissive.

You can't modify the structure before dumping, so you use a post_dump, but why use a `Dict` field if the attribute is an `int`?

Might be a job for `Function` field.

decaz:
@lafrech `int` -> `dict` modification is just a theoretical example. But indeed, thanks for reminding about `Function` and `Method` fields, - they are more appropriate in my case. I haven't used them for a long time, so I forgot about them, thanks!

lafrech:
These are fields I'd rather avoid but there are still cases where I have no obvious alternative and I'm afraid this is one of those cases.

PR #1701, opened 2020-12-02
Add validators to hook order doc

### LINKED ISSUE #1684: Document order of `post_load` and `validates_schema`
I have more than once been unsure about this, and I think it should just be added to the docs.

`validates_schema` runs before `post_load`.

I'm happy to open a PR, but I'm not sure where would be best.
Maybe it should be included in [this bit about hook order](https://marshmallow.readthedocs.io/en/stable/extending.html#pre-post-processor-invocation-order)?

sirosen:
closes #1684


lafrech (approved):

sloria:
Thanks!

PR #1706, opened 2020-12-07
Fix pprint output in "ordering" quickstart docs

lafrech:
I think this was overlooked in #1586.

I also removed the `Function` field and `Meta.fields` to focus on `ordered` feature.


sloria (approved):
Good catch

PR #1720, opened 2021-01-15
decorator validates add order parameter

ficapy:
### This allows validate to be executed in order

For example, if field A needs to query the database when validating, and field B relies on the results of field A's query, 
then validation function A needs to query the database and validation function B needs to query the database if there is no guarantee that A will be executed first. 
Or merge validation functions A, B together, or some other very ugly solution

Adding the order parameter would look like this

```python
class User(Schema):
    name = fields.Str()
    age = fields.Int()
    sex = fields.Int()

    @validates("name", order=1)
    def validate_name(self, value):
        self.context["order"] += "1"
        return value

    @validates("age", order=2)
    def validate_age(self, value):
        self.context["order"] += "2"
        return value

    @validates("sex")
    def validate_sex(self, value):
        self.context["order"] += "0"
        return value

schema = User()
schema.context["order"] = ""
schema.load({"name": "name", "age": 24, "sex": 1})
assert schema.context["order"] == "210"
```

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -157,3 +157,4 @@ Contributors (chronological)
 - Nadège Michel `@nadege <https://github.com/nadege>`_
 - Tamara `@infinityxxx <https://github.com/infinityxxx>`_
 - Stephen Rosen `@sirosen <https://github.com/sirosen>`_
+- Ficapy `@ficapy <https://github.com/ficapy>`_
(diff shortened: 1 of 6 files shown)


sloria:
Thanks for the PR. This was also added in https://github.com/marshmallow-code/marshmallow/pull/600 but there's still discussion needed on whether this is something we want to add to core. Let's continue discussion on that PR.

PR #1725, opened 2021-01-21
Add `extra` argument to load/dump for extensions

### LINKED ISSUE #1726: Allow customized Schema _serialize and _deserialize to get custom data from load/dump calls
Although this is contained in #1725, I'm going to try to distill this down to exactly the feature I want, plus the context for why I want it.

-----

This issue originates in an interaction between `marshmallow-oneofschema` ("oneofschema") and `marshmallow-sqlalchemy`. Both of these are important to marshmallow, especially since oneofschema is the go-to answer for questions about polymorphic fields.

oneofschema overloads `load` and `dump` with special operations. This works for most simple use cases, and most features (e.g. `partial`, `many`) are supported.
However, it means that oneofschema doesn't support pre- and post-processing hooks built into marshmallow, and the only way to support them in the current implementation, in which oneofschema replaces `load` and `dump`, is to reimplement the marshmallow load and dump logic.

Looking into options for this, `_serialize` and `_deserialize` are in exactly the right place for oneofschema to overload them with its specialized behavior. As proven in https://github.com/marshmallow-code/marshmallow-oneofschema/pull/130 , the result is smaller code for oneofschema with added support for hooks.

It seems a perfect fit. However...

In https://github.com/marshmallow-code/marshmallow-oneofschema/pull/111 , oneofschema added support for passing through additional keyword arguments from `load` and `dump` to its child schemas. The reason is that `marshmallow-sqlalchemy` also customizes `load` and `dump`, but in a different way. For that library, additional arguments control added behaviors.

If oneofschema starts using `_serialize` and `_deserialize`, it will lose the ability to support the `marshmallow-sqlalchemy` use-case.

If marshmallow wants to support oneofschema in this, I believe it would be best to consider `_serialize` and `_deserialize` part of the public interface for the purpose of subclassing and customizing schemas. The parallel with these method names for Fields already suggests them as semi-public.

The most naive approach would be to allow arbitrary keyword arguments to `load` to go through `_do_load` and down to `_deserialize` (and likewise for `dump`). However, that allows `load` to accept and ignore arbitrary arguments like `parttial=True` -- it makes the experience for users worse.
In #1725 I proposed a specific argument, `extra`, a mapping which exists only to support extensions like oneofschema. I'm open to other approaches. A major version of oneofschema, but not of marshmallow, would be needed to implement the change.

sirosen:
resolves #1726 

I want to make some changes in `marshmallow-oneofschema` to support hooks (currently unsupported) and simplify the code.
Although it's [just a PR](https://github.com/marshmallow-code/marshmallow-oneofschema/pull/130), one of the issues I see with it is that it will not play nicely with `marshmallow-sqlalchemy` and there isn't a trivial way to get the information I need in `_deserialize` and `_serialize`.

I've been thinking about this on and off for a while and this approach is what I came up with.
It involves `marshmallow` being aware that someone might want to extend `Schema` by overloading `_serialize` and `_deserialize`, and a single new parameter to `load(s)` and `dump(s)` called `extra`. (Name inspired by stdlib `logging`'s use of `extra`.)

Although neither `marshmallow-oneofschema` nor `marshmallow-sqlalchemy` are part of the core project, they're both important parts of a healthy ecosystem of marshmallow extension packages. Supporting their interplay is valuable.

---

The purpose of `extra` is to allow libraries built on top of `marshmallow` to interact with one another cleanly by providing a space for additional data to be passed which `marshmallow` knows about.

In particular, at present the functionality of `marshmallow-sqlalchemy` relies upon adding keyword arguments to `load`, and a proposed refactor of `marshmallow-oneofschema` would not be able to pass through these additional arguments to child schemas.
At the crux of this is a desire in `marshmallow-oneofschema` to override the behavior of `_serialize` and `_deserialize` to take advantage of the hooks provided by marshmallow. This would simplify OneOfSchema and provide additional features, but cannot handle the extra arguments added by `marshmallow-sqlalchemy`.

One option is for these external libraries to coordinate in order to support one another via some other (indeterminate) mechanism. However, in order for `marshmallow` to support such 3rd-party libraries, the key issue here is that there is no way to take extensible data from `load()` calls and pass it down through to the inner (de)serialization methods.

`extra` could be exposed in additional contexts, for example in field loading, hook evaluation, error handling, and so forth. However, the current issue in these third party libraries only requires the addition of `extra` to be passed down to `_deserialize` and `_serialize`. Therefore, this feature addition is restricted to only this context for now.

For this particular problem, this will allow a new major version of `marshmallow-oneofschema` to support usage with marshmallow-sqlalchemy something like so:

    load(..., extra={"child_load_kwargs": {"transient": True}})

which is then unpacked and passed to a child sqlalchemy schema as `load(..., transient=True)`.

In a future version, if `marshmallow-sqlalchemy` started using `extra` instead of added keyword arguments, `marshmallow-oneofschema` could pass through the `extra` data it received or define other sophisticated behaviors.

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -2912,3 +2912,32 @@ def test_class_registry_returns_schema_type():
 
     SchemaClass = class_registry.get_class(DefinitelyUniqueSchema.__name__)
     assert SchemaClass is DefinitelyUniqueSchema
+
+
+def test_extra_args_passthrough():
+    class MySchema(Schema):
+        last_deserialize_side_effect = None
+        last_serialize_side_effect = None
+
+        def _deserialize(self, *args, extra, **kwargs):
+            if "side_effect" in extra:
+                MySchema.last_deserialize_side_effect = extra["side_effect"]
+            super()._deserialize(*args, extra=extra, **kwargs)
+
+        def _serialize(self, *args, extra, **kwargs):
+            if "side_effect" in extra:
+                MySchema.last_serialize_side_effect = extra["side_effect"]
+            super()._serialize(*args, extra=extra, **kwargs)
+
+        name = fields.Str()
+        email = fields.Email()
+
+    data = {"name": "Mick", "email": "mick@stones.com"}
+
+    loaded = MySchema().load(data, extra={"side_effect": "foo"})
+    assert MySchema.last_deserialize_side_effect == "foo"
+    assert MySchema.last_serialize_side_effect is None
+
+    MySchema().dump(loaded, extra={"side_effect": "bar"})
+    assert MySchema.last_deserialize_side_effect == "foo"
+    assert MySchema.last_serialize_side_effect == "bar"
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,14 @@
 Changelog
 ---------
 
+3.11.0 (Unreleased)
+*******************
+
+Features:
+
+- Add ``extra`` argument to ``Schema.load`` and ``Schema.dump`` as an extension
+  point for libraries built on marshmallow
+
 3.10.0 (2020-12-19)
 *******************
 
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -502,12 +502,17 @@ class Schema(base.SchemaABC, metaclass=SchemaMeta):
         return value
 
     def _serialize(
-        self, obj: typing.Union[_T, typing.Iterable[_T]], *, many: bool = False
+        self,
+        obj: typing.Union[_T, typing.Iterable[_T]],
+        *,
+        many: bool = False,
+        extra: typing.Optional[typing.Mapping[str, typing.Any]] = None
     ):
         """Serialize ``obj``.
 
         :param obj: The object(s) to serialize.
         :param bool many: `True` if ``data`` should be serialized as a collection.
+        :param extra: Extra data for serialization. Ignored by ``marshmallow``.
         :return: A dictionary of the serialized data
 
         .. versionchanged:: 1.0.0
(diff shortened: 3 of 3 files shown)


tests/test_schema.py
```diff
@@ -2912,3 +2912,32 @@ class DefinitelyUniqueSchema(Schema):
...
+
+        name = fields.Str()
+        email = fields.Email()
+
+    data = {"name": "Mick", "email": "mick@stones.com"}
+
+    loaded = MySchema().load(data, extra={"side_effect": "foo"})
+    assert MySchema.last_deserialize_side_effect == "foo"
+    assert MySchema.last_serialize_side_effect is None
+
```
lafrech, line 2940: I'd clear last_deserialize_side_effect here to make next test more selective.

lafrech (commented):
I didn't inspect thoroughly but the rationale makes sense.

deckar01:
> Although it's [just a PR](https://github.com/marshmallow-code/marshmallow-oneofschema/pull/130), ... there isn't a trivial way to get the information I need in `_deserialize` and `_serialize`.

This seems like a problem that would benefit from opening an issue in `marshmallow-oneofschema` to explore possible solutions.

sirosen:
I almost opened a marshmallow issue to discuss, but thought it would be better to come to the table with some concrete option for how to handle this. If you prefer to discuss in an issue to a PR, I can open one.

`marshmallow-oneofschema` isn't the right place to tackle this because that library isn't really capable of solving the problem. I want to define a type of schema with special meanings for `load` and `dump`, but still inheriting all of the goodness of the main marshmallow codebase -- namely, support for pre- and post-processing hooks.
That will only work if marshmallow itself supports it, and this is the best way I've thought of to do that.

sirosen:
I've opened #1726 to lay out the issue and discuss potential avenues for supporting extensions of `marshmallow` which may want to customize `Schema._deserialize` and `Schema._serialize`. This PR now can be said to fix #1726 .

sloria:
is the use care here served by `contextvars` (see proposal in https://github.com/marshmallow-code/marshmallow/pull/2707)?

sirosen:
I need to take more time to evaluate it, but I think so. I'd be interested in trying to get OneOfSchema to support pre/post load/dump hooks based on a contextvar based context API, if we land that.
I'm still invested in making OneOfSchema work more smoothly.

I'm going to close this, just to clean up the tracker. I haven't looked at it in ages and I bet it would need a rebase if we even wanted to accept it.

PR #1729, opened 2021-01-26
Remove :rtype: declarations

### LINKED ISSUE #1728: schema.py module type annotations
Is there a reason why type annotations are missing for some methods in `schema.py` (e.g. return types of `dump` and `load` methods)?

https://github.com/marshmallow-code/marshmallow/blob/4ee651014205155292249667fe70b1e646ec6d5d/src/marshmallow/schema.py#L530

However, there are some good old docstrings:

https://github.com/marshmallow-code/marshmallow/blob/4ee651014205155292249667fe70b1e646ec6d5d/src/marshmallow/schema.py#L538

Which can actually mislead IDEs and novices:

marshmallow_dump_type_annotation

Sphinx maintain the following [thing](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#python-signatures) if it's the reason:

> Multiple types in a type field will be linked automatically if separated by the word “or”

If it's just a bug, I would be happy to create the PR.

vemikhaylov:
Resolves: #1728

The `dump` method doesn't return only dictionaries right now, but `:rtype:` can be used by some type checkers in spite of Sphinx annotations being "rejected alternative".

In the issue decided with @deckar01 to remove two `:rtype:` declarations from the project. It resolves the problem with the Pycharm's type checker, which fallbacks to the `Any` type when an annotation is missing.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -157,3 +157,4 @@ Contributors (chronological)
 - Nadège Michel `@nadege <https://github.com/nadege>`_
 - Tamara `@infinityxxx <https://github.com/infinityxxx>`_
 - Stephen Rosen `@sirosen <https://github.com/sirosen>`_
+- Vladimir Mikhaylov `@vemikhaylov <https://github.com/vemikhaylov>`_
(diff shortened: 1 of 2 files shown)


deckar01:
This looks good to me. Thank you.

lafrech (approved):

PR #1731, opened 2021-01-27
Prohibiting null character in String field

lovmat:
Motivation: The default use case is that null character is not wanted - e.g. it causes issues with database since a string containing null character cannot be passed as a string literal

( django-rest-framework has similar solution: https://github.com/encode/django-rest-framework/blob/master/rest_framework/fields.py#L788 )


antonstjernquist:
lgtm

sloria:
we can implement this as part of or a follow-up to https://github.com/marshmallow-code/marshmallow/issues/2787

PR #1732, opened 2021-01-27
Issue #1721 Raise Exception if unknown arg is not INCLUDE, EXCLUDE, or RAISE

### LINKED ISSUE #1721: Is it intentional that setting parameter unknown=<any non-empty string> behaves the same as unknown=EXCLUDE?
I was playing around with some of the unknown options and noticed that if you set unknown to a random string when instantiating a schema, it behaves the same as setting `unknown=EXCLUDE`.

Is this intentional? 

```
class TestSchema(Schema):
    schema_field = fields.Str(required=True)

test = dict(schema_field="random", non_schema_field="also random")
schema = TestSchema(unknown=EXCLUDE)
print(schema.load(test))
# {'schema_field': 'random'}
```

```
schema = TestSchema(unknown=INCLUDE)
print(schema.load(test))
# {'schema_field': 'random', 'non_schema_field': 'also random'}
```

```
schema = TestSchema(unknown="1234")
print(schema.load(test))
# {'schema_field': 'random'}
```

chaynsaw:


lafrech:
Sorry for the late feedback.

I'd put this kind of test somewhere at init, not at runtime.

sirosen:
@sloria, I see you just merged #1989. That was meant to close this, but it didn't because I frequently forget that `closes #N, #M` only closes issue `N`, not issue `M`. Could you (or another maintainer) close?

PR #1735, opened 2021-01-29
Implements IPInterface fields

### LINKED ISSUE #1733: Feature Request:  Add IPv4Network and IPv6Network fields
Add the two network types added so that full network fields e.g. '192.168.1.0/24' can be validated along with a 'IPNetwork' type similar to the IP field that is a union of the two fields. 

These types are already in the ipaddress library <https://python.readthedocs.io/en/latest/library/ipaddress.html#ipaddress.IPv4Network>

madeinoz67:
Implements IPInterface fields that are a non-strict version of IPNetwork Types, where the IPInterfaces will always accept arbitrary host addresses and Mask, while IPNetwork will not allow this.

Implement in this PR are the following new fields:

- IPInterface
- IPv4Interface
- IPv6Interface

see <https://python.readthedocs.io/en/latest/library/ipaddress.html#interface-objects> for more information around these underlying types

closes #1733

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -55,6 +55,9 @@ __all__ = [
     "IP",
     "IPv4",
     "IPv6",
+    "IPInterface",
+    "IPv4Interface",
+    "IPv6Interface",
     "Method",
     "Function",
     "Str",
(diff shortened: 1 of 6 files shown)


src/marshmallow/fields.py
```diff
@@ -1711,6 +1714,66 @@ class IPv6(IP):
...
+
+    def _deserialize(
+        self, value, attr, data, **kwargs
+    ) -> typing.Optional[
+        typing.Union[ipaddress.IPv4Interface, ipaddress.IPv6Interface]
+    ]:
+        if value is None:
+            return None
+        try:
+            return (self.DESERIALIZATION_CLASS or ipaddress.ip_interface)(
```
sloria, line 1754: Why not set `DESERIALIZATION_CLASS` to `ipaddress.ip_interface` on the class?

sloria (changes requested):
Thanks for the PR! This is looking good. Can you please add tests?

src/marshmallow/fields.py
madeinoz67: Honestly this was a copy and paste from the already existing Marshmallow IPAddressField class, just the names were changed to protect the innocent :)

madeinoz67:
ok all tests have been added and passing locally, however I'm having a strange problem with the pre-commit where the mypy test in tests/base.py is failing on two lines that have the `# type: ignore` for the Filed.Integer() and field.Boolean() types, haven't even gone near these classes so am perplexed as-to why the mypy is failing when expressly told to ignore.

````
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

tests/base.py:270: error: unused 'type: ignore' comment
tests/base.py:278: error: unused 'type: ignore' comment
Found 2 errors in 1 file (checked 1 source file)
````

I removed the two troublesome "# type: ignore" pre-commit is fine but tox fails tests so they're there for a reason I guess 
 so I've re-added to how they were

if I run mypy against the file manually all tests fine
````
mypy tests/base.py 
Success: no issues found in 1 source file
````

**Comment by madeinoz67** [comment]
ok re the mypy issue above:

I'm thinking it might be my environment somehow.... 

```
pre-commit run --all-files
pyupgrade................................................................Passed
black....................................................................Passed
flake8...................................................................Passed
mypy.....................................................................Passed
blacken-docs.............................................................Passed
```

but the staged file was still failing as before when committing so I ended up committing with bypass of hooks and that pushed fine, all test are passing so not sure what was the cause the mypy error other than something in my environment causing the issue????

lafrech:
pre-commit doesn't use your environment but its own.

Mypy passes locally because your env is up-to-date. It passes in CI because the env is up-to-date because it is created on-the-fly. But somehow, pre-commit env gets outdated.

There must be a command to update pre-commit env.

The day it happened to me I did just like you: bypass hooks.

sloria (approved):

PR #1739, opened 2021-02-08
add a duplication validation

bonastreyair:
I've encountered that `marshmallow` did not have any duplicate validator. I found it might be useful so I'm sharing the  implementation I'm using so that could be merged as a new validator in a future marshmallow release.

This new validator is designed to work only with iterables.

In case the value is a list of objects which makes it difficult to check if one is repeated, you can use an `attribute` of the object to check if it is repeated or not.

I've also added a simple `pytest`. 

Hope it can be reviewed and at some point included in a future release.

### DIFF
--- a/src/marshmallow/validate.py
+++ b/src/marshmallow/validate.py
@@ -583,20 +583,20 @@ class ContainsNoneOf(NoneOf):
 
 class NoDuplicates(Validator):
     """
-    Validator which succeeds if the `value` is an `iterable` and has no duplicate
+    Validator which succeeds if the ``value`` is an ``iterable`` and has no duplicate
     elements. In case of a list of objects, it can easy check an internal
-    attribute by passing the `attribute` parameter.
+    attribute by passing the ``attribute`` parameter.
 
-    Validator which fails if `value` is not a member of `iterable`.
+    Validator which fails if ``value`` is not a member of ``iterable``.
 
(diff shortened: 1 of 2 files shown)


src/marshmallow/validate.py
```diff
@@ -579,3 +579,49 @@ def __call__(self, value) -> typing.Any:
...
+
+class NoDuplicates(Validator):
+    """
+    Validator which succeeds if the `value` is an `iterable` and has no duplicate
+    elements. In case of a list of objects, it can easy check an internal
+    attribute by passing the `attribute` parameter.
+
+    Validator which fails if `value` is not a member of `iterable`.
+
+    :param attribute: The name of the attribute of the object you want to check.
```
bonastreyair: Missing `str` before attribute

src/marshmallow/validate.py
```diff
@@ -579,3 +579,49 @@ def __call__(self, value) -> typing.Any:
...
         return value
+
+
+class NoDuplicates(Validator):
+    """
+    Validator which succeeds if the `value` is an `iterable` and has no duplicate
+    elements. In case of a list of objects, it can easy check an internal
+    attribute by passing the `attribute` parameter.
+
+    Validator which fails if `value` is not a member of `iterable`.
```
bonastreyair: `value` should have two `` instead of one `

PR #1742, opened 2021-02-09
Replace missing/default with {load,dump}_default

sirosen:
This is part of #778 , which has been open but inactive for a while.
I've made aggressive use of warnings here and chose the `load_default` and `dump_default` names (of the options proposed) because they are parallel with `load_key`, `dump_only`, etc in putting the de/serialization qualifier first.

I expect that if this is released, a lot of users of libraries like apispec would start seeing warnings.
I'm not sure what the best way to mitigate that is, but if removing or changing warnings on some of these behaviors would improve things, just let me know.

---

Add these new parameters to the Field initializer and to the attributes of all fields. Remove the existing `missing/default` attributes in lieu of the new values.

If a 'default' or 'missing' value is passed, respect it but emit a warning.

New `default` and `missing` field properties wrap access to load_default and dump_default, but also emit warnings on access.

Update all usage in tests to the newer form.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -169,20 +169,22 @@ class Field(FieldABC):
         **additional_metadata
     ) -> None:
         # handle deprecated `default` and `missing` parameters
-        if dump_default is missing_ and default is not missing_:
+        if default is not missing_:
             warnings.warn(
                 "The 'default' argument to fields is deprecated. "
                 "Use 'dump_default' instead.",
                 RemovedInMarshmallow4Warning,
             )
-            dump_default = default
-        if load_default is missing_ and missing is not missing_:
+            if dump_default is missing_:
+                dump_default = default
+        if missing is not missing_:
             warnings.warn(
                 "The 'missing' argument to fields is deprecated. "
                 "Use 'load_default' instead.",
                 RemovedInMarshmallow4Warning,
             )
-            load_default = missing
+            if load_default is missing_:
+                load_default = missing
         self.dump_default = dump_default
         self.load_default = load_default
 
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -76,10 +76,10 @@ class Field(FieldABC):
     data does not need to be formatted before being serialized or deserialized.
     On error, the name of the field will be returned.
 
-    :param default: If set, this value will be used during serialization if the input value
-        is missing. If not set, the field will be excluded from the serialized output if the
-        input value is missing. May be a value or a callable.
-    :param missing: Default deserialization value for the field if the field is not
+    :param dump_default: If set, this value will be used during serialization if the
+        input value is missing. If not set, the field will be excluded from the
+        serialized output if the input value is missing. May be a value or a callable.
+    :param load_default: Default deserialization value for the field if the field is not
         found in the input data. May be a value or a callable.
     :param data_key: The name of the dict key in the external representation, i.e.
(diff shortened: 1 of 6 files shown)


lafrech:
> I expect that if this is released, a lot of users of libraries like apispec would start seeing warnings.

We can ship a major using new param names. Or a minor with a condition to pick the appropriate parameter according to marshmallow version (extra work, but a major just to bump marshmallow version is not appealing).

lafrech:
I'm still willing to do this name change.

Question is should we postpone it to v4 or deprecate/warn (this PR) now and remove in v4.

I don't mind either way. Maybe we're overthinking it with the deprecation/warning cycle. OTOH, the job is done already.

We're about to ship webargs 8 so now would be a good time to drop the old names.

I had a quick look at apispec / webargs / flask-smorest and I think it should be easy to support without a breaking change by updating the tests / docstrings and add a few conditions to support the old names (not so many places).

sirosen:
If a deprecation warning is a big enough deal to potentially hold back this change, I am open to removing the warnings and leaving the deprecation in the docs only.

I'd really prefer to use `load_default` and `dump_default`. The `default`/`missing` names today are just too easy to mix up.

lafrech:
No, those warnings are good. Better than doc only. The alternative was postpone to v4. But I prefer to merge this now.

sirosen:
I just rebased this to cleanup the conflict.

After more thought, I think the easiest thing is for downstream projects to do
```
if MARSHMALLOW_VERSION < (3,12):
    ...
else:
    ...
```
as necessary if they want to avoid the warnings and be compatible with lower marshmallow 3.x versions.

src/marshmallow/fields.py
```diff
@@ -166,7 +168,24 @@ def __init__(
         metadata: typing.Optional[typing.Mapping[str, typing.Any]] = None,
         **additional_metadata
     ) -> None:
-        self.default = default
+        # handle deprecated `default` and `missing` parameters
+        if dump_default is missing_ and default is not missing_:
```
lafrech: Why not just this?

    if default is not missing_:

Do we want to accept setting both to let a lib write both as a means to manage both ma versions?

Wouldn't it be better to only accept the new argument and always warn if the deprecated one is passed?

lafrech (approved):
LGTM.

Just a minor comment. Not really blocking.

src/marshmallow/fields.py
sirosen: I think the reason is to ensure that `dump_default = default` (below the warning) is only done if `dump_default is missing_`. But this is somewhat lazy. I'll push a change to rewrite to the structure below, which is really what's wanted:
```python
if default is not missing_:
    warnings.warn(...)
    if dump_default is missing_:
        dump_default = default
```

sirosen:
There was a conflict because `dev` changed near some of this -- no actual changes needed -- so I've done a rebase.

lafrech:
@sloria, may we proceed with this?

sloria (approved):
I'm good with moving forward with this. @lafrech Do you have time to do the changelog and release?

lafrech:
OK, the first who gets the time does it.

But I'd rather settle on #1843 and ship a 3.12.2 bugfix release before.

lafrech:
I just released apispec 5.0.0b1 with marshmallow 3.13 support. 5.0.0 is almost ready, just minor stuff left, but it leaves a little time for people to test the beta version.

PR #1743, opened 2021-02-11
Add optional 'abs_tol' and `rel_tol` arguments to validate.Range

Pierre-Sassoulas:
Hello,

I think this could be merged in marshmallow as it's an optional argument that should not disturb existing code and make sense in Range().

There are more use cases in the tests but the basic principle is this:
```python
assert validate.Range(1, 8.450, abs_tol=0.4)(8.005) == 8.005
assert validate.Range(1, 8.450, abs_tol=0.4)(8.455) == 8.450

with pytest.raises(ValidationError) as exc:
    validate.Range(2, 3, abs_tol=1e-4)(1.999)
assert "must be greater than or equal to 2" in str(exc.value)
assert "(equality with an absolute tolerance of 1.0e-04)" in str(exc.value)

with pytest.raises(ValidationError) as exc:
    validate.Range(100, 200, rel_tol=1e-2)(95)
assert "95 must be greater than or equal to 100 (" in str(exc.value)
assert "(equality with a relative tolerance of 1.0e-02)" in str(exc.value)
 ```

### DIFF
--- a/src/marshmallow/validate.py
+++ b/src/marshmallow/validate.py
@@ -232,7 +232,7 @@ class Range(Validator):
             self._rel_tol_default = is_close.get("rel_tol").default  # type: ignore
             self._abs_tol_default = is_close.get("abs_tol").default  # type: ignore
         except ValueError:
-            self._rel_tol_default, self._abs_tol_default = None, None
+            self._rel_tol_default, self._abs_tol_default = 1e-09, 0.0
         self.rel_tol = self._rel_tol_default if rel_tol is None else rel_tol
         self.abs_tol = self._abs_tol_default if abs_tol is None else abs_tol
         self.error = error
--- a/src/marshmallow/validate.py
+++ b/src/marshmallow/validate.py
@@ -1,4 +1,6 @@
 """Validation classes for various types of data."""
+import inspect
+import math
 import re
 from itertools import zip_longest
 from operator import attrgetter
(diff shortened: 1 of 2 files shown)


lafrech:
Should we be using `math.isclose` to avoid floating point precision issues?

Pierre-Sassoulas:
Yeah, it would definitely make sense. Should we use `rel_tol`, `abs_tol` or both instead of `tolerance` though ? https://docs.python.org/3/library/math.html#math.isclose

Pierre-Sassoulas:
@lafrech this is ready for review, I implemented the abs_tol, rel_tol option in order to use `math.is_close()`.

Pierre-Sassoulas:
I did not realize that python 3.5 was supported by marshmallow, my bad ! Should be good now.

lafrech:
Thank you @Pierre-Sassoulas for your contribution.

My gut feeling is that this does not really belong in marshmallow core.

It is arguably a bit specific. And it is easy enough to achieve in user code. I mean pretty much all the code of the original `Range` validator is changed and this new validator could just as well be defined in user code. Nothing prevents it in marshmallow core. It would just be imported from user code rather than from `marshmallow.validate`.

I'm reluctant to add this API surface for a use case that is currently achievable in user code without any change in marshmallow. I don't feel like the use case is worth the added complexity.

I'll let the other maintainers express their opinion. Until then, I won't do a thorough review asking for changes if we don't intend to merge it in the end.

In any case, thanks for sharing this implementation.

Pierre-Sassoulas:
>  I mean pretty much all the code of the original Range validator is changed

I refactored a lot of it to separate message formatting and checking the conditions but I don't think there is a lot of change overall (It add two optional attributes, two conditions and a function to format the new part of the message). I could do it in two commit to show that better maybe.

sloria:
I agree with @lafrech 's gut feeling here. This use case is a bit too specific, it adds significant complexity to the code and API, and the problem can be solved simply in userland. I'm going to pass on this for now. Thanks anyway for your contribution.

PR #1767, opened 2021-03-19
Field take default on feature

### LINKED ISSUE #1766: Take default value when value is None
I would like to specify fields that will use the default value if their value is `None` (even if they are not missing), something similar to the following field declaration will match perfectly to my case:
```python
fields.Pluck(MySchema, 'x', default={'x': 0}, take_default_on=[None])
```

Is there any way doing this without `@post_dump` or overriding `get_attribute` method?

dormeiri:
Fixes #1766 

I noticed that most of the discussion is focused on deserialization, but I have use case where I need this feature for serialization.

This PR adds the ability for the user to extend the logic of when default value is taken during serialization.

TODO:
- [X] Tests
- [ ]  Docs


sloria:
Something similar is being worked on in https://github.com/marshmallow-code/marshmallow/pull/1381 but still needs discussion before landing. Let's continue discussion there

PR #1770, opened 2021-03-22
Remove useless hasattr / gettatr checks in `Field`

lafrech:
Minor code simplification.

Shouldn't break anything unless someone's been naughty deleting those attributes.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,10 @@ Features:
   ``IPv6Interface`` (:issue:`1733`). Thanks :user:`madeinoz67`
   for the suggestion and the PR.
 
+Other changes:
+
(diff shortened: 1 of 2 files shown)


sloria:
One possibly legitimate use case this could break is a wrapper field that doesn't call `super()` in its `__init__`.

```python
class OuterField(Field):
    def __init__(self, inner_field):
        self.inner_field = inner_field

    def _bind_to_schema(...):...
    def _serialize(...):...
```

Not sure if we should support this or not 🤔

lafrech:
> Not sure if we should support this or not thinking

I don't think we should. There is so much stuff happening in `__init__`.

In fact I don't think we do. `__repr__` doesn't make any check and assumes all attributes are defined, so we can argue it was never supported (or `__repr__` was broken).

I force-pushed to also remove `getattr` calls that were used for this purpose.

sloria (approved):

PR #1777, opened 2021-03-30
Add validate.And

### LINKED ISSUE #1768: And validator
Hello everyone!
I would like to add a new validator to combine multiple validators into one. Simple example:

```py
def calculate_age(birthday: date) -> int:
    today = date.today()
    offset = (today.month, today.day) < (birthday.month, birthday.day)
    return today.year - birthday.year - offset

def validate_not_future_date(value: date) -> None:
    if value > date.today():
        raise ValidationError("Not future date.")

def validate_14_years_old(value: date) -> None:
    if calculate_age(value) < 14:
        raise ValidationError("Must be 14 years old.")

class PassportSchema(Schema):
    issued_at = fields.Date(
        "%Y-%m-%d",
        validate=[
            validate_not_future_date,
            validate_14_years_old,
        ],
    )

    @post_load
    def make_passport(self, data: dict, **kwargs) -> Passport:
        return Passport(**data)

@attr.s(auto_attribs=True, slots=True, frozen=True)
class Passport:
    issued_at: date

    def __attrs_post_init__(self) -> None:
        validate_not_future_date(self.issued_at)
        validate_14_years_old(self.issued_at)
```

In this example, I am forced to pass validators as a list in the `issued_at` field (in `PassportSchema`) and call two functions to validate a single value in `Passport` entity.
I would like to combine the two functions into one and turn them into a single validator:

```py
T = TypeVar("T")

Validator = Callable[[T], T]

@attr.s(auto_attribs=True, slots=True, frozen=True)
class _AndValidator:
    validators: Sequence[Validator]

    def __call__(self, value: T) -> T:
        for validator in self.validators:
            validator(value)
        return value

def and_validator(*validators) -> _AndValidator:
    return _AndValidator(validators)

validate_passport_issued_at = and_validator(
    validate_14_years_old,
    validate_not_future_date,
)
```

Then the example above will look much more concise:

```py
class PassportSchema(Schema):
    issued_at = fields.Date("%Y-%m-%d", validate=validate_passport_issued_at)

    @post_load
    def make_passport(self, data: dict, **kwargs) -> Passport:
        return Passport(**data)

@attr.s(auto_attribs=True, slots=True, frozen=True)
class Passport:
    issued_at: date

    def __attrs_post_init__(self) -> None:
        validate_passport_issued_at(self.issued_at)
```

I would like to add such functionality to the `marshmallow.validate` package.

sloria:
close #1768

Follow's @deckar01 's suggested approach in https://github.com/marshmallow-code/marshmallow/issues/1768#issuecomment-804618935 for factoring out `Field._validate` into a separate validator.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -242,10 +242,10 @@ class Field(FieldABC):
         """Perform validation on ``value``. Raise a :exc:`ValidationError` if validation
         does not succeed.
         """
-        self._validate_callable(value)
+        self._validate_all(value)
 
     @property
-    def _validate_callable(self):
+    def _validate_all(self):
         return And(self.validators, error=self.error_messages["validator_failed"])
 
     def make_error(self, key: str, **kwargs) -> ValidationError:
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -246,7 +246,7 @@ class Field(FieldABC):
 
     @property
     def _validate_all(self):
-        return And(self.validators, error=self.error_messages["validator_failed"])
+        return And(*self.validators, error=self.error_messages["validator_failed"])
 
     def make_error(self, key: str, **kwargs) -> ValidationError:
         """Helper method to make a `ValidationError` with an error message
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -25,7 +25,7 @@ from marshmallow.exceptions import (
     StringNotCollectionError,
     FieldInstanceResolutionError,
 )
-from marshmallow.validate import Validator, Length
+from marshmallow.validate import And, Length
 from marshmallow.warnings import RemovedInMarshmallow4Warning
 
 __all__ = [
@@ -242,21 +242,11 @@ class Field(FieldABC):
         """Perform validation on ``value``. Raise a :exc:`ValidationError` if validation
         does not succeed.
         """
-        errors = []
-        kwargs = {}
-        for validator in self.validators:
-            try:
-                r = validator(value)
-                if not isinstance(validator, Validator) and r is False:
-                    raise self.make_error("validator_failed")
-            except ValidationError as err:
-                kwargs.update(err.kwargs)
-                if isinstance(err.messages, dict):
-                    errors.append(err.messages)
-                else:
-                    errors.extend(err.messages)
-        if errors:
-            raise ValidationError(errors, **kwargs)
+        self._validate_callable(value)
+
+    @property
+    def _validate_callable(self):
+        return And(self.validators, error=self.error_messages["validator_failed"])
 
     def make_error(self, key: str, **kwargs) -> ValidationError:
         """Helper method to make a `ValidationError` with an error message
--- a/src/marshmallow/validate.py
+++ b/src/marshmallow/validate.py
@@ -51,7 +51,7 @@ class And(Validator):
             if value % 2 != 0:
                 raise ValidationError("Not an even value.")
 
-        validator = validate.And([validate.Range(min=0), is_even])
+        validator = validate.And(validate.Range(min=0), is_even)
         validator(-1)
         # ValidationError: ['Must be greater than or equal to 0.', 'Not an even value.']
 
@@ -62,10 +62,7 @@ class And(Validator):
     default_error_message = "Invalid value."
 
     def __init__(
-        self,
-        validators: typing.Iterable[types.Validator],
-        *,
-        error: typing.Optional[str] = None
+        self, *validators: types.Validator, error: typing.Optional[str] = None
     ):
         self.validators = tuple(validators)
         self.error = error or self.default_error_message  # type: str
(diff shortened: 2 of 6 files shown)


src/marshmallow/validate.py
```diff
@@ -32,6 +32,60 @@ def _repr_args(self) -> str:
...
+
+    default_error_message = "Invalid value."
+
+    def __init__(
+        self,
+        validators: typing.Iterable[typing.Callable[[typing.Any], typing.Any]],
+        *,
+        error: typing.Optional[str] = None
+    ):
+        self.validators = validators
```
rugleb: If a `generator` of `Validator` is passed here, the validation will be called only once at the time of the first pass. I think there should be line like this:
```python
self.validators = tuple(validator)
```

src/marshmallow/validate.py
```diff
@@ -32,6 +32,60 @@ def _repr_args(self) -> str:
         """
         return ""
 
+    def __call__(self, value):
```
rugleb: What about adding a type annotation using generic? All (must part?) validators return the same type that passed for validation.
```python
T = TypeVar("T")

class Validator(ABC):

    @abstractmethod
    def __call__(self, value: T) -> T:
        pass
```
This will help us infer types and validate them through static analysis.

tests/test_validate.py
```diff
@@ -890,3 +890,16 @@ def test_containsnoneof_mixing_types():
...
     with pytest.raises(ValidationError):
         validate.ContainsNoneOf([1, 2, 3])((1,))
+
+
+def test_and():
+    def is_even(value):
+        if value % 2 != 0:
+            raise ValidationError("Not an even value.")
+
+    validator = validate.And([validate.Range(min=0), is_even])
```
rugleb: What about adding a factory to create such `validator.And` without combining sub-validators into a list?
Before:
```python
    validator = validate.And([validate.Range(min=0), is_even])
```
After:
```python
    validator = validate.and(validate.Range(min=0), is_even, is_magic)
```
?

src/marshmallow/validate.py
sloria: I'll probably do this in a separate PR--there's a few typing-related errors to address in this module.

tests/test_validate.py
sloria: Not really opposed to that (I saw attrs has something similar), but what's the benefit? Is it just to save keystrokes?

It would need to be named `and_` since `and` is a keyword. And at that point, we save only 1 keystroke 😅

src/marshmallow/validate.py
rugleb: Will be perfect.

tests/test_validate.py
rugleb: It's not only about the number of characters, but also about visual perception and the forced creation of an optimal sequence - tuple.

In addition, this will not allow to create "And" without validators.

sloria: OK. Seems fine since it's a fairly common pattern in Python. Added in my latest commit.

> this will not allow to create "And" without validators.

How? `validate.and_()` will pass without errors.

deckar01: Is there any reason not to make the *args syntax the default? It is easy to `(*spread)` an iterable into args when we use it internally, but as a user API, *args looks prettier. I doubt users are going to have their validators in a list for some other use very often.

src/marshmallow/validate.py
```diff
@@ -40,6 +40,66 @@ def __call__(self, value: typing.Any) -> typing.Any:
...
+        if errors:
+            raise ValidationError(errors, **kwargs)
+        return value
+
+
+def and_(*validators: types.Validator, error: typing.Optional[str] = None) -> And:
+    """Shortcut function for constructing a `marshmallow.validate.And` from
+    the validators passed positionally.
+    """
+    return And(validators, error=error)
```
deckar01, line 100: I initially imagined building this in the constructor and avoiding the errors list for basic validators, but I did not consider that the error object requires a list wrapper to keep things normalized. This won't provide the optimization I was hoping for, but I still think it is a worth while refactor.

src/marshmallow/fields.py
```diff
@@ -242,21 +242,11 @@ def _validate(self, value):
...
-                    errors.append(err.messages)
-                else:
-                    errors.extend(err.messages)
-        if errors:
-            raise ValidationError(errors, **kwargs)
+        self._validate_all(value)
+
+    @property
+    def _validate_all(self):
+        return And(self.validators, error=self.error_messages["validator_failed"])
```
deckar01: I initially imagined building this in the constructor and avoiding the errors list for basic validators, but I did not consider that the error object requires a list wrapper to keep things normalized. This won't provide the optimization I was expecting, but I still think it is a worth while refactor.

sloria: Yeah, I tried to build this in the constructor first, but it ended up being hairy to make it work with `Field`s that insert validators (e.g. Email, URL).

There still might be a clean way to do it, but I think we can leave that for the future.

tests/test_validate.py
sloria: @deckar01 By 'default', do you mean using `validate.and_` in `Field._validate_all`? Should we make `validate.And` private (`validate._And`)?

deckar01: I mean `And.__init__(*validators, error=None)` and no `_and()`. It would be nice to minimize the API surface area. When we use it internally we can just as easily call `And(*validators, error=error)`.

sloria: Good point; not much benefit in having 2 ways to do things here.

deckar01 (approved):

lafrech (approved):

Issue #1737, opened 2021-02-04
Implementation of "include" when "exclude" is declared on meta


ksdelacruz:
Good day.

I need help on implementing "include" when declaring "exclude" on meta. Before 3.x, I use the solution here: https://github.com/marshmallow-code/marshmallow/issues/244#issuecomment-140931116

I now transitioned to 3.x. I searched for some implementations and found this https://github.com/marshmallow-code/marshmallow/issues/614#issuecomment-450177491, but I can't make this work like the 2.x implementation.
I tried adding `_update_fields` as a `@pre_dump` and experimented that way, but it still not work. 

I also noticed that `self._declared_fields[field_name]` of the excluded field is None, thus I tried "re-initializing it on `__init__`.

Here's my code:

```
class SitesSchema(MARSHMALLOW.SQLAlchemyAutoSchema):
    def __init__(self, *args, **kwargs):
        self.include = kwargs.pop("include", [])
        for field_name in self.include:
            if field_name == "season_months":
                self._declared_fields[field_name] = MARSHMALLOW.Nested(
                    "SeasonsSchema", exclude=("sites",))
        super().__init__(*args, **kwargs)

    @pre_dump
    def _update_fields(self, data, **kwargs):
        if self.include:
            for field_name in self.include:
                self.fields[field_name] = self._declared_fields[field_name]

        return data

    season = MARSHMALLOW.Integer()
    season_months = MARSHMALLOW.Nested("SeasonsSchema", exclude=("sites",))

    class Meta:
        model = Sites
        ordered = False
        exclude = ["season_months"]
```

I'm stuck so any help is welcome. Thanks!

sloria:
Here's the "include" recipe adapted for marshmallow 3:

```python
from pprint import pprint

from marshmallow import Schema, fields


class Model(object):
    def __init__(self, **kwargs):
        self.__dict__.update(kwargs)


class BaseSchema(Schema):
    def __init__(self, *args, **kwargs):
        self.include = kwargs.pop("include", None)
        super().__init__(*args, **kwargs)

    def _init_fields(self, *args, **kwargs):
        super()._init_fields(*args, **kwargs)
        if self.include:
            for field_name in self.include:
                field_obj = self.declared_fields[field_name]
                self.fields[field_name] = field_obj
                self.dump_fields[field_name] = field_obj
                self.load_fields[field_name] = field_obj


class Member(BaseSchema):
    id = fields.Integer()
    email = fields.Email()
[...]

Issue #244, opened 2015-07-18
Append "include" param to schema contructor class ?


harobed:
I've this schema :

```
class Member(ma.Schema):
    id = ma.Integer()
    email = ma.Email()
    firstname = ma.String()
    lastname = ma.String()
    photo = ma.Raw()

    class Meta:
        exclude = ('photo',)
```

I would like `include` param like that : 

```
member_post_schema = schemas.Member(exclude=('id',), include=('photo',))
```

What do you think about append this include feature ?

sloria:
How would this behave differently from the `only` parameter?

jarmani:
As far as I know there is no way to define "optional" fields that are used in special context.

Maybe adding a way to define that a field is not used by default and add an include parameter to use them makes sense.

For now I'm using multiple Schema / and or exclude in all others contexts.

imhoffd:
+1

Marshmallow seems to use a "blacklist" serialization method instead of "whitelist." That is, you can `exclude` fields or you explicitly specify the entire subset you want, with `only`, but there is no way to add additional fields at any time between during/after schema instantiation and before serialization.

So, I've hacked it. In my schemas, I've had to override `__init__` to do some weird reverse logic with what I deem the "optional" fields, as @jarmani described. So it looks something like this:

``` python
MySchema(ModelSchema):
    optional_field_1 = fields.Str()
    optional_field_2 = fields.Str()

    class Meta:
        model = MyModel
        additional = ('field_3',)
        exclude = ('_field_3',)  # specifying them here won't work

    def __init__(self, *args, **kwargs):
        if 'exclude' not in kwargs:
            kwargs['exclude'] = {'optional_field_1', 'optional_field_2'}
        super(MySchema, self).__init__(*args, **kwargs)
```

And during instantiation, when I want to decide which fields to "add", I do this:

``` python
def get(id, fields=None):
    schema = MySchema()
    m = Model.query.get(id)

    if 'optional_field_1' in fields:
        schema.exclude.remove('optional_field_1')

    if 'optional_field_2' in fields:
        schema.exclude.remove('optional_field_2')

    return schema.dump(m).data
```

This works, but I don't recommend doing it this way to anyone, but there should definitely be an `include` option, which allows for additional fields during schema instantiation.

sloria:
I'm going to hold off on adding this feature.

You should, however, be able to implement this functionality in your own schemas like so:

``` python
from marshmallow import Schema, fields

class Model(object):
[...]

PR #1781, opened 2021-03-31
Add `embed` schema parameter and `embed_only` field option

bpneumann:
Implements the functionality described in https://github.com/marshmallow-code/marshmallow/issues/151.
Similar to the functionality of `only`, but  allows fields to be marked as `embed_only`. A common use case for this is for schemas which contain large collections which should only be serialized in specific cases. 

Add support for `embed_only` option on fields, and as a Meta option in schema classes.

If a field is marked as `embed_only`, the field will only be serialized if it is included in `embed` schema parameter.

Example:

```python
    class MySchema(Schema):
        foo = fields.Field(embed_only=True)

    sch = MySchema()
    assert "foo" not in sch.dump({"foo": "bar"})

    sch = MySchema(embed=['foo'])
    assert "foo" in sch.dump({"foo": "bar"})
```

This also supports nested fields using the same dotted identifiers used by the `only` option.

```python
    class ChildSchema(Schema):
        bar = fields.Field(embed_only=True)
        baz = fields.Field(embed_only=True)

    class ParentSchema(Schema):
        foo = fields.Nested(ChildSchema())

    sch = ParentSchema(embed=['foo.bar'])
    dump = sch.dump({"foo": {"bar": "val", "baz": "val"}})
    assert "foo" in dump
    assert "bar" in dump["foo"]
    assert "baz" not in dump["foo"] 
```

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -160,3 +160,4 @@ Contributors (chronological)
 - Vladimir Mikhaylov `@vemikhaylov <https://github.com/vemikhaylov>`_
 - Stephen Eaton `@madeinoz67 <https://github.com/madeinoz67>`_
 - Antonio Lassandro `@lassandroan <https://github.com/lassandroan>`_
+- Brady Neumann `@bpneumann <https://github.com/bpneumann>`_
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -162,6 +162,7 @@ class Field(FieldABC):
         allow_none: typing.Optional[bool] = None,
         load_only: bool = False,
         dump_only: bool = False,
+        embed_only: bool = False,
         error_messages: typing.Optional[typing.Dict[str, str]] = None,
         metadata: typing.Optional[typing.Mapping[str, typing.Any]] = None,
         **additional_metadata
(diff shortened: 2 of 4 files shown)


sloria:
I believe this is the same use case described in https://github.com/marshmallow-code/marshmallow/issues/244 . See the ["include" recipe](https://github.com/marshmallow-code/marshmallow/issues/244#issuecomment-140931116 ) for a way to address it.

Holding off on this PR for now.

bpneumann:
This recipe works for schemas without nested fields, but offers no way to include fields on nested fields. 

Extending the example from https://github.com/marshmallow-code/marshmallow/issues/1737#issuecomment-812969654
```python

class Organization(BaseSchema):
    members = fields.Nested(Member())
    name = fields.String()

organization = Model(name="Book Club", members=[member])

pprint(Organization(include=("members.photo",)).dump(organization))
# raises KeyError
```

Since `members.photo` is not a field on `Organization`, this raises a KeyError. The desired behavior would be to set `include` on the `Organization` schema to `members`, and propagate `photo` to the `include` parameter of the nested `Member` schema.

A full solution to this can currently only be accomplished by subclassing `Schema` as well as `Nested` to accommodate the include option, and overriding the initializers, `_init_fields`, and `_normalize_nested_options`. 

Alternatively, this could be made simpler with the changes proposed in https://github.com/marshmallow-code/marshmallow/pull/1725 allowing subclasses of schema to update fields before serialization/deserialization according to `extra` arguments.

PR #1789, opened 2021-04-07
Type hinted all the decorators in marshmallow/decorators.py

### LINKED ISSUE #1788: Type hinting decorators
By default `mypy` running in strict mode doesn't allow for untyped functions or decorators to be used. Since all the decorators in `/src/marshmallow/decorators.py` are untyped, the only way to use `mypy` with the `--strict` flag is to add a `# type: ignore` flag on every line referencing any of the decorators. Adding type hints would solve this problem.

Created a pull request here: https://github.com/marshmallow-code/marshmallow/pull/1789

Couldn't really think of any tests since these are just type hint changes. Used `Callable[..., Any]` to be as backwards compatible as possible towards other repos and not use overly convoluted type hints as `post_load` could return something very different from `pre_load`, etc.

The `MarshmallowHook` protocol was based on this `mypy` discussion: https://github.com/python/mypy/issues/2087. The `Protocol` class itself couldn't be used directly due to backwards compatibility issues. Alternatively I can update the PR to avoid the class and use `# type: ignore` on the two lines accessing the `__marshmallow_hook__` property, which normally isn't allowed since a `Callable` doesn't have that property.

michaeldimchuk:
Added a bunch of type hints to the `decorators` module making it easier to use `marshmallow` alongside `mypy` running in strict mode

close #1788

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -161,3 +161,4 @@ Contributors (chronological)
 - Stephen Eaton `@madeinoz67 <https://github.com/madeinoz67>`_
 - Antonio Lassandro `@lassandroan <https://github.com/lassandroan>`_
 - Javier Fernández `@jfernandz <https://github.com/jfernandz>`_
+- Michael Dimchuk  `@michaeldimchuk <https://github.com/michaeldimchuk>`_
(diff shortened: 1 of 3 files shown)


sloria (approved):
Thanks!

PR #1804, opened 2021-05-08
Fix markup of code literals in changelog

homeworkprod:
Makes it easier to recognize code/API references (IMHO).


sloria (approved):
Thanks!

Issue #1826, opened 2021-06-11
RFC: Use a context variable to pass Schema context


lafrech:
Python 3.7 brings context variables. I've been using this along with a context manager to pass contextual values in a thread-safe way.

https://github.com/Scille/umongo/blob/master/umongo/expose_missing.py
https://stackoverflow.com/a/67807219/4653485
https://stackoverflow.com/a/59911117/4653485

This could be suited to pass `Schema` context and it could provide a better API, allowing a `Schema` instance to have different contexts in different threads.

This could be done in different ways:

### Global context

Context unique to all schema classes.

```py
# Provided by marshmallow in schema.py
CONTEXT = ContextVar("session", default=None)

class Context(AbstractContextManager):
    def __init__(self, context):
        self.context = context

    def __enter__(self):
        self.token = CONTEXT.set(self.context)

    def __exit__(self, *args, **kwargs):
        CONTEXT.reset(self.token)

# User code
class Schema_1(Schema):
    """Schema 1"""

class Schema_2(Schema):
    """Schema 2"""


with Context({... insert context here ...}):
    Schema_1().dump(...)
    Schema_2().dump(...)
```

The good thing with this is that we don't even have to propagate the context through nested schemas.

The limitation is that the same context is used for both schemas. Is this an issue?

We could provide the context manager from the `Schema` class (as a static) but it would be ambiguous as it could let the user think the context is specific to that class while it is not the case.

### Schema-wise context

One context per schema class.

```py
# Provided by marshmallow in schema.py
class Schema():
    ctx = ContextVar("session", default=None)

    # TODO: Add method here to provide context manager
    # ...

# User code
class Schema_1:
    """Schema 1"""

class Schema_2:
    """Schema 2"""

schema_1 = Schema_1()
schema_2 = Schema_2()

with schema_1.context({... insert context here ...}):
    schema_1.dump(...)   # Uses context
    schema_2.dump(...)  # Does not use context
```

In this example, there is no reason to put `schema_2.dump` in the with block but maybe the with is outside the function. The point is it is not affected. Or there may be multiple contexts.

```py
with schema_1.context({... insert context here ...}), schema_2.context({... insert context here ...}):
    schema_1.dump(...)
    schema_2.dump(...)
```

This requires to propagate the context through nested schemas. Not that nice.

Or we could pass the context on call to load/dump, without a context manager:

```py
# Provided by marshmallow in schema.py
CONTEXT = ContextVar("session", default=None)

[...]

PR #1833, opened 2021-06-22
Fix nested schema context thread-safety

### LINKED ISSUE #1617: Context is not thread-safe (at least) for nested scheme
I'm developing a web application with Flask, having many marhsmallow schema classes with nested schemes. I also make use of marshmallows context.
Sometimes (e.g. first run of the application after a reboot) I noticed unexpected serialization results. It turned out that the context in one of the nested schemas differed from it's parent.

A simplified example, which shows the race condition for concurrent requests with a nested scheme:
```python
from datetime import datetime
from flask import Flask
from marshmallow import Schema, fields, post_dump
import time
app = Flask(__name__)


class TestNestSchema(Schema):
    name = fields.String()

    @post_dump()
    def context_dump(self, data, **kwargs):
        data['before'] = self.context.get('test')
        # Do some work
        time.sleep(5)
        data['after'] = self.context.get('test')
        return data


class TestSchema(Schema):
    name = fields.String()
    nested = fields.Nested(TestNestSchema())


class TestNestedClass:
    name = 'TestNestedClass'


class TestClass:
    name = 'TestClass'
    nested = TestNestedClass()


@app.route('/')
def test_endpoint():
    test_schema = TestSchema(context={'test': str(datetime.now())})
    dump = test_schema.dump(TestClass())
    return {'data': dump}


if __name__ == "__main__":
    app.run()

```

When troubleshooting, I also took a brief look at the marhsmallow code and got the feeling that it wasn't just the context that might cause such problems.
Is there a way to use marhsmallow thread-safe?

### LINKED ISSUE #1791: Context and instantiated Netsted schemas.
I like to use schema context to transmit information from one schema to another (child or parent) and the following behavior seems a little strange to me.

```python
import marshmallow as mrsh

class SchemaA(mrsh.Schema):
    a = mrsh.fields.Integer()

    @mrsh.pre_load
    def _pre_load(self, data, **_):
        self.context["from_a"] = "a"

        return data

class SchemaB1(mrsh.Schema):
    ac = mrsh.fields.Nested(SchemaA)

    @mrsh.pre_load
    def _pre_load(self, data, **_):
        self.context["from_b1"] = "b1"

        return data

class SchemaB2(mrsh.Schema):
    ac = mrsh.fields.Nested(SchemaA())

    @mrsh.pre_load
    def _pre_load(self, data, **_):
        self.context["from_b2"] = "b2"

        return data
```
The only difference between SchemaB1 and SchemaB2 is that I provide the SchemaA class to the first one and an instance of SchemaA to the other.
This is the kind of things I usually do to provide a specific context to SchemaA when generated from SchemaB.
Something like that : `ac = mrsh.fields.Nested(SchemaA(context={"x": 1))`

```python
sb1 = SchemaB1()
sb2 = SchemaB2()

sb1.load({"ac": {"a": 5}})
sb2.load({"ac": {"a": 5}})

print(sb1.context)
>> {'from_b1': 'b1', 'from_a': 'a'}

print(sb2.context)
>> {'from_b2': 'b2'}
```
Using an instance of SchemaA result in the context not being fully shared between the Nested schema and its parent.

```python
sb1.fields["ac"].schema.context is sb1.context
>> True
```
Same object if using a class.

```python
sb2.fields["ac"].schema.context is sb2.context
>> False
```
Different objects if using an instance.

```python
sb2.fields["ac"].schema.context
>> {'from_b2': 'b2', 'from_a': 'a'}
```
SchemaA correctly inherited from its parent context.

I'm not sure if that's a bug or not but that behavior somehow feels inconsistent to me (or at least hard to predict).

Is this the expected behavior ?

Thanks for your great library!

```
# Name                    Version                   Build  Channel
marshmallow               3.11.1             pyhd3eb1b0_0    defaults
python                    3.8.5                h7579374_1    defaults
```

lafrech:
Fixes #1617.
Closes #1791.

This fixes the thread-safety issue but might be a breaking change for people relying on the update: parent schema updates nested schema context. See #1791.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -512,7 +512,7 @@ class Nested(Field):
 
             if isinstance(nested, SchemaABC):
                 self._schema = copy.copy(nested)
-                self._schema.context.update(context)
+                self._schema.context = context
                 # Respect only and exclude passed from parent and re-initialize fields
                 set_class = self._schema.set_class
                 if self.only is not None:
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -2580,6 +2580,25 @@ class TestContext:
         obj = {"inner": {"foo": 42}}
         assert outer.dump(obj)
 
+    def test_context_thread_safe(self):
+        class InnerSchema(Schema):
+            i = fields.Integer()
+
+        class OuterSchema(Schema):
+            inner = fields.Nested(InnerSchema)
+
+        outer_1 = OuterSchema()
+        outer_1.context = {1: 1}
+        outer_1.dump({"inner": {"i": 1}})
+        outer_2 = OuterSchema()
+        outer_2.context = {2: 2}
+        outer_2.dump({"inner": {"i": 1}})
+
+        assert outer_1.context == {1: 1}
+        assert outer_2.context == {2: 2}
+        assert outer_1.fields["inner"].context == {1: 1}
+        assert outer_2.fields["inner"].context == {2: 2}
+
 
 def test_serializer_can_specify_nested_object_as_attribute(blog):
     class BlogUsernameSchema(Schema):


lafrech:
The alternative, to avoid breaking the update case, would be to document it as not thread-safe and invite the users to use a class  rather than an instance, as this is thread-safe (but does not allow context update from parent, only replacement, and doesn't allow to use modifiers in the nested schema init).

Then in 4.0, drop the update case if this is acceptable. And perhaps implement #1826.

The update case allows users to specify a context at both parent and nested level. In practice, I'm not sure this is worth supporting but I could be proven wrong. Besides, the design smells a bit, as the update flattens parent and nested contexts without namespacing nested context. I hope I'm making sense.

lafrech:
See discussion in #1791.

It might be better not to merge this (breaking change) and instead perhaps raise a warning about the thread-safety issue.

sloria:
closing as addressed by https://github.com/marshmallow-code/marshmallow/pull/2707

PR #1834, opened 2021-06-22
Update documentation around deserializing objects

insacuri:
There wasn't anything in the documentation about how deserializing to an object
changed in the upgrade from 2.x to 3.x.  This PR mentions it and tries to give
a clean example of what the before and after looks like

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -163,3 +163,4 @@ Contributors (chronological)
 - Javier Fernández `@jfernandz <https://github.com/jfernandz>`_
 - Michael Dimchuk  `@michaeldimchuk <https://github.com/michaeldimchuk>`_
 - Jochen Kupperschmidt  `@homeworkprod <https://github.com/homeworkprod>`_
+- Adam Mann  `@insacuri <https://github.com/insacuri>`_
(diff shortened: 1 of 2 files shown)


deckar01:
I think this is covered in the ["Schemas are always strict"](https://github.com/marshmallow-code/marshmallow/blob/31b784f/docs/upgrading.rst#schemas-are-always-strict) section of the 3.0 upgrade docs.

> Schema().load <marshmallow.Schema.load> and Schema().dump <marshmallow.Schema.dump> don't return a (data, errors) tuple any more. Only data is returned.

insacuri:
Does it? That example covers the amount of variables returned but doesn't say that the type of object being returned is different.

I had to figure this out myself in a recent upgrade as nowhere in the documentation (that I could find) said that deserializing to an object would now just give you the object rather then an `UnmarshalResult` that you'd have to dig your object out of.

insacuri:
Do you want me to close this PR?  I still think that the doco you linked isn't quite the same as what my PR is describing but if nobody replies to my comment then trying to discuss it is pointless

lafrech:
Then change has nothing to do with `post_load` instantiating an object. The result of the load is always the data rather than a (data, errors) tuple, as written in the doc quote above, whether `post_load` is involved or not.

I think this PR can be closed.

Thanks for contributing, though.

PR #1845, opened 2021-07-07
Added support for single character TLDs

### LINKED ISSUE #1844: email validation should support single character TLDs
Right now the email schema dont support single character TLDs:

```python
class EmailSchema(Schema):
    email = fields.Email()

print(EmailSchema().validate({"email": "test@a.a"}))
#> {'email': ['Not a valid email address.']}
```
The above behavior is wrong since `a.a` is a valid TLD according to `RFC-1123`, Please see the stackoverflow thread: https://stackoverflow.com/questions/7411255/is-it-possible-to-have-one-single-character-top-level-domain-name/21872376#21872376

This validation should pass.

gautamkrishnar:
Added support for single character TLDs for the email and domain validation. It will now support TLDs like `a.a`. Refer #1844  for more detaills.

fixes: #1844

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -93,3 +93,4 @@ Contributors (chronological)
 - Bryce Drennan `@brycedrennan <https://github.com/brycedrennan>`_
 - Cristi Scoarta `@cristi23 <https://github.com/cristi23>`_
 - Nathan `@nbanmp <https://github.com/nbanmp>`_
+- Gautam Krishna R `@gautamkrishnar <https://github.com/gautamkrishnar>`_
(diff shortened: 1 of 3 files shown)


lafrech:
Thanks for the contribution and sorry for about the delay.

I have no objection to this, even if [the SO answer](https://stackoverflow.com/a/21872376) underlines the fact that in practice there are currently no single character TLDs in the root.

Two things I'd like to check before merging:

- Does this also affect URL?
- What does Django do about this?

Also, please add yourself to AUTHORS.

gautamkrishnar:
@lafrech Added myself to authors list. Eventhough there are no single character TLDs currently available right now, this will make the library future proof.

> Does this also affect URL?

yes, this regex will affect the domain validations as well since email and domain validation shares the same regex. 

> What does Django do about this?

I don't use django personally, I am not sure about this.

lafrech:
> yes, this regex will affect the domain validations as well since email and domain validation shares the same regex.

I had a quick look and I thought it uses a different regex (see _regex_generator). I shall double-check.

> I don't use django personally, I am not sure about this.

I don't either, but I remember some regexes being copied from there, and since I'm not proficient in regex, I find it nice to get them from there, or at least discuss the subject with them.

PR #1849, opened 2021-07-19
docs: document nested field behavior with Meta.unknown

ikonst:


ikonst:
Not true since https://github.com/marshmallow-code/marshmallow/pull/963

PR #1852, opened 2021-07-22
Validate strict numbers during serialization, fixes #1851

### LINKED ISSUE #1851: Strict Integer validation not called during serialization (dump)
When `strict=True` is set in  `fields.Integer`, the validation is only called during deserialization. During serialization, float values are converted to int.

```python
class MySchema(Schema):
    some_int = fields.Integer(strict=True)

s = MySchema()
result = s.dump({'some_int': 10.01})
print(result)
```

output:
```
{'some_int': 10}
```

Expected output:
Marshmallow should raise a validation error:
```
```

vivekkr12:


sloria:
As mentioned in the linked issue: marshmallow assumes that data to serialize is valid and therefore does not validate upon serialization. This is an explicit design choice that improves serialization performance significantly.

PR #1862, opened 2021-08-12
Make the build reproducible

lafrech:
See https://github.com/marshmallow-code/apispec/pull/669.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -28,7 +28,7 @@ issues_github_path = "marshmallow-code/marshmallow"
 
 templates_path = ["_templates"]
 
-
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -27,13 +28,18 @@
 
 templates_path = ["_templates"]
 
+
```
sloria: might be worth a bit of docco. i wasn't aware of this envvar before now

```suggestion
# Use SOURCE_DATE_EPOCH for reproducible build output https://reproducible-builds.org/docs/source-date-epoch/
```

sloria (approved):

lafrech:
Good point about the comment.

I'll merge equivalent PRs in webargs and apispec, with the added comment.

PR #1867, opened 2021-08-20
Fix: TimeDelta Precision Errors

### LINKED ISSUE #1865: TimeDelta serialization precision
Hi there!

I just found quite strange behaviour of `TimeDelta` field serialization

```python3
from marshmallow.fields import TimeDelta
from datetime import timedelta

td_field = TimeDelta(precision="milliseconds")

obj = dict()
obj["td_field"] = timedelta(milliseconds=345)

print(td_field.serialize("td_field", obj))
```

Output of this snippet is `344`, but it seems that `345` is correct.

Looks like a rounding issue here: https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/fields.py#L1474

deckar01:
Use microsecond integer arithmetic to fix high precision timedelta errors.

Fixes #1865

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1471,7 +1471,9 @@ class TimeDelta(Field):
         if value is None:
             return None
         base_unit = dt.timedelta(**{self.precision: 1})
-        return int(value.total_seconds() / base_unit.total_seconds())
+        delta = utils.timedelta_to_microseconds(value)
+        unit = utils.timedelta_to_microseconds(base_unit)
+        return delta // unit
 
     def _deserialize(self, value, attr, data, **kwargs):
         try:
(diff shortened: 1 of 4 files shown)


src/marshmallow/fields.py
```diff
@@ -1471,7 +1471,9 @@ def _serialize(self, value, attr, obj, **kwargs):
         if value is None:
             return None
         base_unit = dt.timedelta(**{self.precision: 1})
-        return int(value.total_seconds() / base_unit.total_seconds())
+        delta = round(decimal.Decimal(value.total_seconds()), 9)
+        unit = round(decimal.Decimal(base_unit.total_seconds()), 9)
```
lafrech: Should we add a comment to explain where the `9` comes from?

Also I'm wondering if it could be worth computing `unit` at init for each precision. Or even hardcode all of them. But this might not be worth the trouble.

deckar01: Reverting to integer arithmetic ended up being simpler than I anticipated. See https://github.com/marshmallow-code/marshmallow/issues/1865#issuecomment-904787313.

PR #1882, opened 2021-09-21
Include affected keyword argument names in metadata deprecation warning

traherom:
I decided to tackle the fixing the Marhmallow 4 deprecation warning, but found it very challenging to determine which fields were actually triggering the warning. This PR simply includes the `additional_metadata` in the warning message, making searching for the offending fields easier.


sloria (approved):
This is a good idea. Thanks!

PR #1888, opened 2021-10-14
Support Python 3.10

lafrech:


lafrech:
I suggest we run CI on min/max Python versions only. The only benefit would be saving resources. Tests are fast and free, but it just seems like CPU waste. WDYT?

sloria:
Looks like it didn't actually use Python 3.10

Firefox 2021-10-15 at 08 30 24


update: oh, scratch that. python 3.10 does get installed via apt

Firefox 2021-10-15 at 08 29 34

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = lint,py36,py37,py38,py39,docs
+envlist = lint,py36,py37,py38,py39,p310,docs
```
sloria:
```suggestion
envlist = lint,py36,py37,py38,py39,py310,docs
```

sloria (approved):
i'm fine with testing only 3.6 and 3.10. 3.7-3.9 don't have any significant changes that would affect marshmallow, and it'd be nice to reduce our footprint.

tox.ini
lafrech: 🤦

lafrech:
Good. Thanks for catching the typo.

Next step, [drop Python 3.6 for Christmas](https://endoflife.date/python).

Xmas

Issue #1744, opened 2021-02-12
[RFC] set_class = OrderedSet by default?


lafrech:
I assume the performance impact would be minimal. From a quick look, it would only affect schema instantiation.

The benefit is that when using Python 3.7+, all schemas would be ordered. They still wouldn't return `OrderedDict` instances unless `ordered` Meta is passed, but they would keep fields declaration order in their output and in their schema list, so that order would be respected in API responses and apispec users would get their fields documented in declaration order.

Currently, users can achieve this in a base schema with

```py
class OrderedBaseSchema(Schema):
    set_class = OrderedSet
```

My gut feeling is that this change would have no functional downside and a negligible perf impact at field init (less that we care about when adding a new feature), while using `ordered=True` does have an impact on serialization performance (I didn't measure it) and functionality (equality test depends on order), hence my proposal. There would still be an interest in allowing the use of `OrderedDict`.

If we go this route, we could use `OrderedSet` by default in marshmallow 3.x, and in marshmallow 4 perhaps rework the `ordered` feature because schemas would already be ordered so it would be specifically for users really needing an `OrderedDict` instance. Perhaps drop the `ordered` Meta option and let the user specify a dict class of their choice another way. But that's another story.

See https://github.com/marshmallow-code/flask-smorest/issues/228#issuecomment-777836226.

sloria:
I'm not opposed to this, so long as there is no perf impact (I think your intuition on this is right).

lafrech:
Or since this is easily achieved in user code thanks to a base schema as shown above, we postpone this to marshmallow 4 as described in my last paragraph.

Adding v4 milestone for now.

deckar01:
> using `ordered=True` does have an impact on serialization performance

I believe that is caused by `OrderedDict`. None of the `OrderedSet` instance attributes are ever accessed during `dump()`, only the `OrderedDict` `dump_fields`. Even though `dict` is ordered now, it is still faster than `OrderedDict`. Waiting until 3.7+ would avoid using `OrderedDict` also.

<details><summary>OrderedDict Benchmark</summary>

```py
#!/usr/bin/env python3.9

from datetime import datetime
from collections import OrderedDict


def run(D):
    A = [('a', None), ('b', None), ('r', None), ('c', None), ('d', None)]
    B = [('1', None), ('2', None), ('3', None)]
    t0 = datetime.now()
    for i in range(5000000):
        s = D(A)
        s['e'] = None
        del s['e']
[...]

Issue #546, opened 2016-11-03
Include Ordered Set as a package instead of writing it yourself?


sabinem:
I studied marshmellows code, since I was looking for a project, that I understand well enough, in order to start contributing to it. I really like marshmellow and I feel that its code and documentation are very well written. One thing, that I noticed though are the custom written collectibles. I understand that marshmellow takes its custom OrderedDict only for Python 2.6, where there is no other choice, but for the custom OrderedSet there would be other choices: I found two packages: 
1. the first is very similar to the function used now: https://pypi.python.org/pypi/ordered-set. It is written in pure Python, but has some improvements that are all stated on the projects github page. It seems more clean and performs better.
2. the second is written in Cython: https://pypi.python.org/pypi/orderedset. It comes with better documentation then the first package and differs more form the current solution.
Both are based on Raymond Hettinger's OrderedSet recipe, just as the one used now is. I would suggest to use one of these two options instead of the custom OrderedSet. 
If the community agrees I would be happy to work on this. What do you think about this suggestion?

sloria:
Thanks for taking the time to study marshmallow, @sabinem . Marshmallow currently has no hard external dependencies, which keeps the codebase self-contained. This is not to say marshmallow will never have external dependencies--sufficiently stable dependencies that add value will certainly be considered.

Re: OrderedSet--marshmallow already exposes a way to change the set class:

``` python
from myorderedset import OrderedSet

class MySchema(Schema):
    @property
    def set_class(self):
        return OrderedSet if self.ordered else set
```

As long we have this hook, I don't think adding a dependency is necessary. It's the same as the `Meta.json_module` option: though there are alternative JSON encoding/decoding libaries (e.g. simplejson, ujson), it suffices to expose an option that lets the user choose.

sabinem:
@sloria, fair enough, I understand that. Then I would suggest to consider coping the more mature recipe from LuminosoInsight: [https://github.com/LuminosoInsight/ordered-set/blob/master/ordered_set.py](https://github.com/LuminosoInsight/ordered-set/blob/master/ordered_set.py). This recipe is very similar to the one used now, but the code looks a lot nicer and the reasons why he changed the original recipe seem convincing to me:

> Based on a recipe originally posted to ActiveState Recipes by Raymond Hettiger,
> and released under the MIT license.
> Rob Speer's changes are as follows:
>     - changed the content from a doubly-linked list to a regular Python list.
[...]

PR #1891, opened 2021-10-19
unbundle ordered-set

xsuchy:
This module was added in 2014, but it is available in PyPI for long time.


xsuchy:
This patch is actually in package in Fedora Linux for ages. No issue has been reported during that time.

setup.py
```diff
@@ -2,7 +2,7 @@
 from setuptools import setup, find_packages
 
 EXTRAS_REQUIRE = {
-    "tests": ["pytest", "pytz", "simplejson"],
+    "tests": ["ordered-set", "pytest", "pytz", "simplejson"],
```
lafrech, line 5: Shouldn't be needed if ordered-set is an install_requires already.

However this might be the right place to pin a test version so that new ordered-set releases trigger dependabot.

lafrech (commented):
I like this. I've always been bothered by this code being duplicated here. It breaks the "marshmallow is dependency-free" thing, but oh well. We used to depend on dateutil.

Besides, I'd like to be as distro-friendly as possible, so whatever we can do to minimize packaging work, I'm happy to merge.

We may investigate in marshmallow 4 whether using ordered-sets is still justified.

Meanwhile, I don't think this is a breaking change, but other maintainers feedback welcome.

setup.py
xsuchy: Should I remove it from here then?

Not sure what you meant by pinning.

lafrech: I meant something like `"ordered-set==1.0.0"` (didn't check the actual value). So that dependabot launches a build whenever a new ordered-set version is released.

And perhaps in `install_requires`, set a minimal version and prevent next major: `"ordered-set>=1.2.0,<2.0.0"`.

lafrech:
See #546. Looks like there are (were ?) several competing implementations. Let's make sure we pick the right one if we go on with this.

lafrech:
Looks like none is packaged on Debian.

Since we don't use all features of an ordered set, and since performance doesn't matter that much on schema creation, I'm wondering if we even need an ordered set in the first place. But I've been giving it a quick try to see if we could replace this with a list or a dict and I have the feeling that it would take more code than the recipe we currently use.

xsuchy:
The other option can be:

```
try:
  from ordered_set import OrderedSet
except ImportError:
  from marshmallow.orderedset import OrderedSet
```

lafrech:
> ```
>     try:
>       from ordered_set import OrderedSet
>     except ImportError:
>       from marshmallow.orderedset import OrderedSet
> ```

I'd rather settle one way or another than maintain both ways using the conditional import above.

While I like to avoid code duplication, this part of the code is small and stable and there doesn't even seem to be a consensus on a single implementation (#546). It is mostly a question of principle. The "no-duplication" principle meets the "don't go making a lib for any piece of code that is more than 10 lines long, this is not Javascript" principle.

Thinking this again, I totally respect the choice of Fedora devs, but I don't feel strongly in favour of this. I'm fine with the status quo.

Leaving this open for others to comment.

#1744 is partly related.

Issue #1356, opened 2019-08-20
RFC: Remove implicit field creation


sloria:
Proposal: Remove Implicit/inferred field creation (`fields` and `additional` class Meta options).

At the very least, it's an unnecessary feature that adds complexity to the codebase. At worst, it interacts in weird ways with consuming libraries that autogenerate fields (see #923). The feature was added pre-1.0, when there were no ORM integration libs, as a poor man's [ModelSerializer](https://www.django-rest-framework.org/api-guide/serializers/#modelserializer). At this point, though, I think we can remove it.

MikeBadescu:
To clarify, do you still plan to keep the `Schema` class meta `unknown`?
```
    class Meta:
        unknown = INCLUDE
```

sloria:
Yes, the unknown field API will remain the unchanged.

MikeBadescu:
Thank you!

sloria:
Let's consider this accepted for 4.0.

lafrech:
Achieved in #2705.

PR #1897, opened 2021-10-27
Let data_key and attribute default to field name

### LINKED ISSUE #1864: Always set data_key
Hello all,

in my application we use and love marshmallow. We have a lot of dynamic coding, therefore we access fields e.g. over the `declared_fields` attribute. Also we need most of the time the `Field.data_key` attribute which is defined as `str | None`.

I would like to suggest an option, which always sets the data_key so that it will never be None. If the user provides it, fine. Else, just set it to the field name. I guess this could also simplify some coding, since you can expect that this attribute is always set.

What do you think?

lafrech:
Closes #1864.

Trying to address #1864, I get typing issues, as foreseen in https://github.com/marshmallow-code/marshmallow/issues/1864#issuecomment-909225360. I wouldn't mind a bit of help, here.

This change spares users the hassle of checking if `data_key` is `None`, as can be seen in the changes in schema.py. This would be used in apispec, for instance.

Notes:

- In Pluck, we don't check for `None` but only for truthyness, which may be wrong for empty strings. I didn't look into it any further but this PR would fix it.
- I also intend to add `data_key` to the repr. The printed name will be the actual value, be it the parameter or the field name (default). I don't think it is an issue.
- I also intend to do the same with attribute, symmetrically.

### DIFF
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -38,6 +38,23 @@ from marshmallow.warnings import RemovedInMarshmallow4Warning
 _T = typing.TypeVar("_T")
 
 
+def _set_field_name(field_obj, field_name):
+    try:
+        field_obj._set_name(field_name)
+    except TypeError as error:
+        # Field declared as a class, not an instance. Ignore type checking because
+        # we handle unsupported arg types, i.e. this is dead code from
+        # the type checker's perspective.
+        if isinstance(field_obj, type) and issubclass(field_obj, base.FieldABC):
+            msg = (
+                'Field for "{}" must be declared as a '
+                "Field instance, not a class. "
+                'Did you mean "fields.{}()"?'.format(field_name, field_obj.__name__)
+            )
+            raise TypeError(msg) from error
+        raise error
+
+
 def _get_fields(attrs, ordered=False):
     """Get fields from a class. If ordered=True, fields will sorted by creation index.
 
@@ -51,6 +68,9 @@ def _get_fields(attrs, ordered=False):
     ]
     if ordered:
         fields.sort(key=lambda pair: pair[1]._creation_index)
+    # Set field name on each field
+    for field_name, field_value in fields:
+        _set_field_name(field_value, field_name)
     return fields
 
 
@@ -111,6 +131,8 @@ class SchemaMeta(type):
         # get_declared_fields
         klass.opts = klass.OPTIONS_CLASS(meta, ordered=ordered)
         # Add fields specified in the `include` class Meta option
+        for field_name, field_obj in klass.opts.include.items():
+            _set_field_name(field_obj, field_name)
         cls_fields += list(klass.opts.include.items())
 
         dict_cls = OrderedDict if ordered else dict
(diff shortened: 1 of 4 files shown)


src/marshmallow/schema.py
```diff
@@ -634,9 +633,7 @@ def _deserialize(
         else:
             partial_is_collection = is_collection(partial)
             for attr_name, field_obj in self.load_fields.items():
-                field_name = (
-                    field_obj.data_key if field_obj.data_key is not None else attr_name
-                )
+                field_name = field_obj.data_key
                 raw_value = data.get(field_name, missing)
```
deckar01: > error: Argument 1 to "get" of "Mapping" has incompatible type "Optional[str]"; expected "str"

The problem is that `Fields.__init__` assigned a `Optional[str]` type to `data_key`. `_bind_to_schema` guarantees that `field_obj.data_key` is a `str`, but the type checker doesn't know that `_deserialize` is guaranteed to run after `_bind_to_schema`. I guess the `get` type annotation enforces this argument, but AFAICT, `None` doesn't actually error. 🤷

It wants you to ensure `data_key` is never `None`. The only thing I can think of is to make `data_key` a `@property` that raises an exception if the field is unbound.

lafrech: Yes, we agree on the reason behind the error.

It would be a shame to "complicate" the code, even slightly, to please the linter.

I suppose we could add comments to skip test on each failing line, but I'd rather find a way to centralize that by somehow forcing the type of `self.data_key` once and for all.

We could use a [cast](https://mypy.readthedocs.io/en/stable/casts.html):

```py
self.data_key = typing.cast(str, data_key)
```

but doing this in `__init__` before field binding would be wrong.

Or we do it just before the type check error:

```py
field_name = typing.cast(str, field_obj.data_key)
```

Still not ideal because:

- It is not done everywhere, only where the error was triggered. Other errors may occur later here or in other libs.
- It is executed on each deserialization, although IIUC it is neglectable.

I just pushed a commit doing that.

lafrech:
Notes:

- I renamed a few local variables in `Schema` methods to be more explicit: `data_key` instead of `field_name`. We might want to rename `field_name` attribute of `_call_and_store` because it not necessarily the field name. That can be done later.

- In `Field.get_value`, I had to keep the `if self.attribute is None` check because the field can be used while not bound to a schema.

lafrech:
I performed a quick test on apispec codebase to see how things could be simplified there and the result is a bit disappointing. apispec operates on schema instances or classes, and when operating on a class, fields are not bound to the schema so we can't rely on `data_key` or `attribute` being set.

This feature would be more useful if it occurred on schema class creation, not instantiation. Technically, that could mean adding a `_set_name` method to `Field` that would be called from `Schema._get_fields`. This is what I do in the last commit. Although setting name at schema creation seems more logical to me, the downside of this is that custom fields, especially container fields, may now have to override both `_bind_to_schema` and `_set_name`. (Part of the trouble there is related to inferred or `Meta.include` fields, which may be dropped in marshmallow 4 (at least inferred ones: #1356).)

With this last commit, the changes in apispec work.

Notes:

- The change in apispec is minimal (2 lines changed), so the benefit of this is rather limited.)
- If I understand @kasium correctly in #1864, they are using `_declared_fields` so they are working on `Schema` classes, not instances, therefore, they fall in the same case as apispec and the change would only be beneficial if done at `Schema` creation.

kasium:
@lafrech yes, that's right. I use the `declared_fields` attribute. But I guess using a schema instance would not add too much trouble if somehow cached (not your worry). However, having it on class level would still be nice

PR #1903, opened 2021-11-03
fix: distutils deprecation warning in Python 3.10

### LINKED ISSUE #1902: Deprecation warning
```
DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. 
Use setuptools or check PEP 632 for potential alternatives
```
https://github.com/marshmallow-code/marshmallow/blob/35225039900be03536cfde7abf121d11066e3183/src/marshmallow/__init__.py#L14

Environment:
- CPython 3.10.0
- marshmallow 3.14.0

kkirsche:
Fixes: #1902

This fixes the distutils deprecation warning by using `packaging.version`

```
In [4]: from packaging.version import Version
In [5]: Version("3.14.0").release
Out[5]: (3, 14, 0)
In [6]: from distutils.version import LooseVersion
In [7]: tuple(LooseVersion("3.14.0").version)
Out[7]: (3, 14, 0)
````

### DIFF
--- a/src/marshmallow/__init__.py
+++ b/src/marshmallow/__init__.py
@@ -11,10 +11,10 @@ from marshmallow.decorators import (
 )
 from marshmallow.utils import EXCLUDE, INCLUDE, RAISE, pprint, missing
 from marshmallow.exceptions import ValidationError
-from distutils.version import LooseVersion
+from packaging.version import Version
 
 __version__ = "3.14.0"
-__version_info__ = tuple(LooseVersion(__version__).version)
+__version_info__ = Version(__version__).release
 __all__ = [
     "EXCLUDE",
     "INCLUDE",
--- a/src/marshmallow/__init__.py
+++ b/src/marshmallow/__init__.py
@@ -14,7 +14,10 @@ from marshmallow.exceptions import ValidationError
 from packaging.version import Version
 
 __version__ = "3.14.0"
-__version_info__ = Version(__version__).release
+__parsed_version__ = Version(__version__)
+__version_info__ = __parsed_version__.release
+if __parsed_version__.pre:
+    __version_info__ += __parsed_version__.pre
 __all__ = [
     "EXCLUDE",
     "INCLUDE",
--- a/src/marshmallow/__init__.py
+++ b/src/marshmallow/__init__.py
@@ -13,7 +13,7 @@ from marshmallow.utils import EXCLUDE, INCLUDE, RAISE, pprint, missing
 from marshmallow.exceptions import ValidationError
 from packaging.version import Version
 
-__version__ = "3.14.0"
+__version__ = "3.14.1"
 __parsed_version__ = Version(__version__)
 __version_info__ = __parsed_version__.release
 if __parsed_version__.pre:
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -165,3 +165,5 @@ Contributors (chronological)
 - Jochen Kupperschmidt  `@homeworkprod <https://github.com/homeworkprod>`_
 - `@yourun-proger <https://github.com/yourun-proger>`_
 - Ryan Morehart '@traherom <https://github.com/traherom>`_
+- Ben Windsor  '@bwindsor <https://github.com/bwindsor>`_
+- Kevin Kirsche  '@kkirsche <https://github.com/kkirsche>`_
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,20 @@
 Changelog
 ---------
 
+3.15.0 (unreleased)
+*******************
+
+* Address distutils deprecation warning in Python 3.10 (:pr:`1903`).
+  Thanks :user:`kkirsche` for the PR.
+
+3.14.1 (2021-11-13)
+*******************
+
+Bug fixes:
+
+- Fix publishing type hints per `PEP-561 <https://www.python.org/dev/peps/pep-0561/>`_.
+  Thanks :user:`bwindsor` for the catch and patch.
+
 3.14.0 (2021-10-17)
 *******************
 
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,6 @@
 include LICENSE
 include *.rst NOTICE
+include src/marshmallow/py.typed
 recursive-include tests *
 recursive-include examples *
 recursive-include docs *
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

File: src/marshmallow/__init__.py [hunk 1]
```diff
@@ -11,10 +11,10 @@
 )
 from marshmallow.utils import EXCLUDE, INCLUDE, RAISE, pprint, missing
 from marshmallow.exceptions import ValidationError
-from distutils.version import LooseVersion
+from packaging.version import Version
```
deckar01, line 14: I believe the `packaging` module would need to be listed as a dependency.

src/marshmallow/__init__.py
```diff
@@ -11,10 +11,10 @@
...
 
 __version__ = "3.14.0"
-__version_info__ = tuple(LooseVersion(__version__).version)
+__version_info__ = Version(__version__).release
```
deckar01: This is missing the prerelease information that the previous version info contained.

```py
>>> tuple(LooseVersion('3.0.0b16').version)
(3, 0, 0, 'b', 16)
>>> Version('3.0.0b16').release
(3, 0, 0)
>>> v = Version('3.0.0b16')
>>> v.release + v.pre
(3, 0, 0, 'b', 16)
```

We could avoid the dependency by parsing the version string ourself.

```py
parts = filter(None, re.split(r'\.|(\D+)', __version__))
__version_info__ = tuple(int(p) if p.isdigit() else p for p in parts)
```

deckar01 (changes requested):

src/marshmallow/__init__.py
kkirsche: I wouldn't be willing to go that route personally and would ask someone else to take this on in that case. I think it's bad practice to purposefully avoid dependencies to avoid the recommended migration path. 

Your call which way you want to go. I'm fine closing this as a wontfix in this state.

src/marshmallow/__init__.py
kkirsche: Ah correct, sorry about that. I had it from poetry it seems. Will add based on feedback from other comment.

lafrech:
I'm fine with the new dependency. After all, it is recommended in https://www.python.org/dev/peps/pep-0632/#toc-entry-6.

Perhaps this could have been left to client libs. I suppose it was included here in #1074 to spare them the job as it was a no-cost for marshmallow. Maybe it wouldn't have been done in the first place if it required a dependency back then, but never mind.

We have plenty of time to act upon this as Python 3.12 is far away, but I see no reason to leave this open for months while that fix is acceptable (with the changes required by @deckar01).

BTW, we have the same issue in webargs, and we also use LooseVersion in apispec, although for another purpose.

kkirsche:
> I'm fine with the new dependency. After all, it is recommended in https://www.python.org/dev/peps/pep-0632/#toc-entry-6.
> 
> Perhaps this could have been left to client libs. I suppose it was included here in #1074 to spare them the job as it was a no-cost for marshmallow. Maybe it wouldn't have been done in the first place if it required a dependency back then, but never mind.
> 
> We have plenty of time to act upon this as Python 3.12 is far away, but I see no reason to leave this open for months while that fix is acceptable (with the changes required by @deckar01).
> 
> BTW, we have the same issue in webargs, and we also use LooseVersion in apispec, although for another purpose.

Happy to make a pull request for those if the update looks good 👍

src/marshmallow/__init__.py
kkirsche: Added as install requirement

lafrech (approved):

lafrech:
Thanks.

The apispec case is a bit different as we use it for another purpose and it would be a breaking change. We'll have to do it in the next major.

kkirsche:
Validated with the list of versions in distutils from StrictVersion

```python
from distutils.version import LooseVersion
from packaging.version import Version

versions = [
    "0.4.1",
    "0.5a1",
    "0.5b3",
    "0.5",
    "0.9.6",
    "1.0",
    "1.0.4a3",
    "1.0.4b1",
    "1.0.4",
]

if __name__ == "__main__":
    for version in versions:
        lv = LooseVersion(version)
        v = Version(version)
        parsed = v.release
        if v.pre:
            parsed += v.pre

        assert parsed == tuple(lv.version)
```

src/marshmallow/__init__.py
kkirsche: Let me know if you want all the segments represented from the version regex in it's order sequence:

```
VERSION_PATTERN = r"""
    v?
    (?:
        (?:(?P<epoch>[0-9]+)!)?                           # epoch
        (?P<release>[0-9]+(?:\.[0-9]+)*)                  # release segment
        (?P<pre>                                          # pre-release
            [-_\.]?
            (?P<pre_l>(a|b|c|rc|alpha|beta|pre|preview))
            [-_\.]?
            (?P<pre_n>[0-9]+)?
        )?
        (?P<post>                                         # post release
            (?:-(?P<post_n1>[0-9]+))
            |
            (?:
                [-_\.]?
                (?P<post_l>post|rev|r)
                [-_\.]?
                (?P<post_n2>[0-9]+)?
            )
        )?
        (?P<dev>                                          # dev release
            [-_\.]?
            (?P<dev_l>dev)
            [-_\.]?
            (?P<dev_n>[0-9]+)?
        )?
    )
    (?:\+(?P<local>[a-z0-9]+(?:[-_\.][a-z0-9]+)*))?       # local version
"""
```

deckar01 (approved):

kkirsche:
WebArgs PR submitted.

I looked at the APISpec change and it doesn't look like it'll be too hard. Version uses the term `micro` instead of `patch` for that property, otherwise the major and minor properties are already built into Version. A few other minor changes are needed, but doesn't appear to be anything too bad and should make the code more readable since it won't require indexes to access pieces of the version as is being done today, though it still is an option depending on the location.

Let me know if you'd like me to submit that one. As it'd probably require a major version update, I didn't want to submit it if you are a long way away from another major version as it'll just slowly get out of sync with other changes

lafrech:
Thanks.

Regarding apispec, I took a quick stab at it and indeed it is easy and it will make the code nicer, but since it is an easy breaking change, I'll do it just before next major.

kkirsche:
Sounds good, thank you Jérôme. Let me know if you need any support from me on it

sloria (approved):
Thanks!

jon-betts:
Just FYI, I believe this affects more than just Python 3.10. We're seeing it in 3.8.9 for example.

PR #1905, opened 2021-11-03
Add py.typed to MANIFEST.in to include in source distributions

### LINKED ISSUE #1904: PyPi source distribution does not work with mypy
When importing `marshmallow` into another project which uses mypy, I get the following error:
```
error: Skipping analyzing "marshmallow": found module but no type hints or library stubs
```

This is strange, because `marshmallow` should be compliant with [PEP-561](https://www.python.org/dev/peps/pep-0561), as the `src/marshmallow/py.typed` file is present, is has `package_data={"marshmallow": ["py.typed"]}` included in `setup.py`, and has `zip_safe=False` as explained in mypy's [guide to creating PEP-561 compatible packages](https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages).

I investigated this further and found that the issue is that `setup.py` also has `include_package_data=True`. As explained in the `setuptools` [documentation](https://setuptools.pypa.io/en/latest/userguide/datafiles.html):

> If using the setuptools-specific include_package_data argument, files specified by package_data will not be automatically added to the manifest unless they are listed in the MANIFEST.in file

So the issue here is that `py.typed` is not explicity included in marshmallow's `MANIFEST.in` file. This issue would be solved by adding `include src/marshmallow.py.typed` to `MANIFEST.in`.

I have tested this locally and it works. I'll open a PR in a moment.

bwindsor:
Including `py.typed` via `MANIFEST.in` will ensure that it is included in source distributions, and therefore remain compatible with Mypy for anybody who installs marshmallow via source distribution. A full explanation of the issue can be found in #1904.

Fixes #1904

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -165,3 +165,4 @@ Contributors (chronological)
 - Jochen Kupperschmidt  `@homeworkprod <https://github.com/homeworkprod>`_
 - `@yourun-proger <https://github.com/yourun-proger>`_
 - Ryan Morehart '@traherom <https://github.com/traherom>`_
+- Ben Windsor  '@bwindsor <https://github.com/bwindsor>`_
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,14 @@
 Changelog
 ---------
 
+3.14.1 (unreleased)
+*******************
+
+Bug fixes:
+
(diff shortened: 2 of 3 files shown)


deckar01 (approved):

lafrech (approved):
Thanks.

Note we have the same issue with webargs.

sloria (approved):

PR #1925, opened 2021-12-16
Fix type information for `__version_info__` to handle prereleases

sirosen:
~Running `__version_info__ += __parsed_version__.pre` in `__init__.py` is incorrect (adding tuples). This bug will only be tripped on a prerelease version, so it hasn't had any impact to-date.~
For background, see also: https://github.com/marshmallow-code/webargs/issues/667
EDIT: The issue is not a bug at runtime, but a type annotation is needed for `__version_info__`. Right now, it combines strings and ints but has a type of `tuple[int, ...]`

`__version_info__` was being computed by adding together tuples. `mypy` has started flagging this (I think due to improved `packaging` type information).

To protect against flaws in the correction, and to present accurate type information, the `__version_info__` tuple is built by a utility function.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,7 @@ repos:
   rev: v0.910-1
   hooks:
   - id: mypy
-    additional_dependencies: [types-simplejson, types-pytz]
+    additional_dependencies: [types-simplejson, types-pytz, packaging]
 - repo: https://github.com/asottile/blacken-docs
   rev: v1.12.0
   hooks:
--- a/src/marshmallow/__init__.py
+++ b/src/marshmallow/__init__.py
@@ -1,5 +1,4 @@
 from __future__ import annotations
-import typing
 
 from marshmallow.schema import Schema, SchemaOpts
 
@@ -18,11 +17,11 @@ from packaging.version import Version
 
 __version__ = "3.14.1"
 __parsed_version__ = Version(__version__)
-__version_info__: tuple[int, int, int] | tuple[int, int, int, str, int] = typing.cast(
-    tuple[int, int, int], __parsed_version__.release
-)
+__version_info__: tuple[int, int, int] | tuple[
+    int, int, int, str, int
+] = __parsed_version__.release  # type: ignore[assignment]
 if __parsed_version__.pre:
-    __version_info__ += __parsed_version__.pre  # type: ignore
+    __version_info__ += __parsed_version__.pre  # type: ignore[assignment]
 __all__ = [
     "EXCLUDE",
     "INCLUDE",
--- a/src/marshmallow/__init__.py
+++ b/src/marshmallow/__init__.py
@@ -1,3 +1,6 @@
+from __future__ import annotations
+import typing
+
 from marshmallow.schema import Schema, SchemaOpts
 
 from . import fields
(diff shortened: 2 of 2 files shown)


deckar01:
> Unsupported operand types for + ("Tuple[int, ...]" and "Tuple[str, int]")

The current logic appears to be correct, the mypy error is just complaining that a string is being mixed in with integers. These tuples are intended to be compared regardless of length/type differences, so the type should probably be loosened to reflect that.

```py
__version_info__: tuple[int | str, ...] = __parsed_version__.release
```

sirosen:
> The current logic appears to be correct, the mypy error is just complaining that a string is being mixed in with integers.

You're 100% right. I almost never use `+` with tuples, so I got mixed up!
I thought this was citing lack of support for plus, but I was wrong.

> These tuples are intended to be compared regardless of length/type differences, so the type should probably be loosened to reflect that.

I know that in other repos we've used something akin to `MARSHMALLOW_MAJOR = marshmallow.__version_info__[0]` followed by checks like `if MARSHMALLOW_MAJOR < 3: ...`.
That makes me want to annotate this such that the first three elements are known to be ints.

Is what you suggest, `tuple[int | str, ...]` really preferable? Or is this better?
```python
__version_info__: tuple[int, int, int] | tuple[int, int, int, str, int]
```

(I think a helper is needed in this case, since the alternative is some pretty awkward use of `cast` in `__init__.py`)

lafrech:
> That makes me want to annotate this such that the first three elements are known to be ints.

Good point.

Would it be possible to just
- set type to `tuple[int, int, int] | tuple[int, int, int, str, int]`
- add some directive to skip the mypy check for the line since we (think we) know what we are doing, without going through the hassle of defining a helper
?

I can't think of a way to both set `type: ignore` and `type: tuple[int, int, int] | tuple[int, int, int, str, int]`.

If there's no simpler way, I'm fine with the helper. I'm just thinking this is a case where CI gets in the way.

sirosen:
mypy is satisfied if we write
```
__version_info__: tuple[int, int, int] | tuple[int, int, int, str, int] = typing.cast(
    tuple[int, int, int], __parsed_version__.release
)
if __parsed_version__.pre:
    __version_info__ += __parsed_version__.pre  # type: ignore
```

The first line sets the type and the `+=` line doesn't change it.

I went ahead and amended to this version, since the helper is another thing to maintain and I'm not deeply committed to it.

`mypy` failed in CI as a result claiming that the `type: ignore` was unused. It was passing for me locally, so I'm going to try adding `packaging` to the requirements of the `mypy` pre-commit hook. I suspect it needs that type information and it's getting it for me from user site-packages or something similar.

sirosen:
I've had to tinker further, but this is now passing CI again. The sample I wrote above is acceptable to `mypy` and works on newer pythons, but fails at runtime under 3.7. It's surprising at first, but `cast(tuple[int, int, int], ...)` won't work under 3.7, even with the future import. My last commit message explains why in detail.

To workaround, I've switched to
```
__version_info__: tuple[int, int, int] | tuple[int, int, int, str, int] = (
    __parsed_version__.release  # type: ignore[assignment]
)
```
and a quick check of `reveal_type(__version_info__)` shows that this works as intended.

It also does appear to be necessary to include `packaging` for `mypy` to run consistently. As I suspected, that was the source of (spurious) warnings about unused ignore comments.

lafrech (approved):
Thanks for working on this.

This is exactly what I meant. Don't bother with a helper to construct the version, set the type to what it is and ignore mypy complaints.

lafrech:
See https://github.com/marshmallow-code/webargs/pull/680.

I think the mypy issue was an error in mypy that was fixed in mypy 0.931, although I couldn't find the changelog entry explaining what was fixed.

Still, this change makes the typing more specific without being too heavy (compared to the earlier version involving a helper method).

@sloria, @deckar01 any objection to this?

sloria (approved):
Looks good!

PR #1932, opened 2022-01-03
Use type hints made available in Python 3.9 and 3.10.

Isira-Seneviratne:


lafrech:
Thanks again.

Just to be sure I understand.

- What did you change between your former PR and this one to let Python 3.7 tests pass?
- According to [the docs](https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations) this feature was introduced in 3.7 and becomes default (no `__future__`) in 3.10, so I don't understand your commit comment and PR title.

lafrech:
OK, I spotted a difference between both branches, related to `dict` being used as a generic type. Detailed explanation [here in SO](https://stackoverflow.com/a/37087556). This only works for 3.9+. This answers the first question, and probably the second. Should the commit comment be rephrased to focus on the change (postponed evaluation of annotations) introduced in 3.7 and not mention 3.9, then?

Edit: In fact, I still don't get it. With 3.7+, `dict` should be usable as a generic type with the `from __future__ import annotations` directive.

Isira-Seneviratne:
> OK, I spotted a difference between both branches, related to `dict` being used as a generic type. Detailed explanation [here in SO](https://stackoverflow.com/a/37087556). This only works for 3.9+. This answers the first question, and probably the second. Should the commit comment be rephrased to focus on the change (postponed evaluation of annotations) introduced in 3.7 and not mention 3.9, then?
> 
> Edit: In fact, I still don't get it. With 3.7+, `dict` should be usable as a generic type with the `from __future__ import annotations` directive.

Yeah, adding the future import allows `dict` to be used as a generic type.

lafrech:
Then I don't understand why the first PR failed since you added the import to all modified files. And it seems the build logs have already been deleted from CI servers.

In this PR, there are places where you use the `dict` form 

```
- _registry = {}  # type: typing.Dict[str, typing.List["SchemaType"]]
+ _registry = {}  # type: dict[str, list[SchemaType]]
```

and others where you don't

```
hooks = defaultdict(list)  # type: typing.Dict[types.Tag, typing.List[str]]
```

Is there a good reason or did you just missed a few lines by accident?

Also, your commit message mentions Python 3.9 but I don't see what changes with 3.9. AFAIU, the changes are in 3.7 (feature introduced as future) and 3.10 (feature becomes default).

sirosen:
> Then I don't understand why the first PR failed since you added the import to all modified files.

I think I know the cause, as it's something I've bumped up against recently. In python < 3.10 you can't use the new types in `cast` or assignments. e.g. This line:
https://github.com/Isira-Seneviratne/marshmallow/blob/2b355bf8c9e03caf3d5fd82040743795fdf37692/src/marshmallow/schema.py#L794

`from __future__ import annotations` avoids evaluating the annotations at runtime. However, if you write `list[int]` in a context other than an annotation, it needs to evaluate -- the interpreter can't avoid it anymore.
So `x: list[int]` is okay, but `TypeX = list[int]; x: TypeX` is not.

(This is probably my least favorite thing I've run into in all of the python typing work.)

You can play games with this to make it work by stringifying annotations. e.g. `typing.cast("list[int]", [])` works.

> Is there a good reason [not to use the new annotation style everywhere] or did you just miss a few lines by accident?

I think these are by accident. I don't see any cases where changes are made but `typing.Dict`, `typing.List`, `typing.Union`, or `typing.Optional` are used.

I'd let this slide rather than insisting that we catch everything.
It's an initial step in this direction, and we can get the rest over time. Plus, I see existing type comments which should maybe become `:`-delimited annotations, so I expect we'll do another pass soon.

> commit message mentions Python 3.9 but I don't see what changes with 3.9.

3.9 introduced `dict`, `tuple`, and `list` subscripting, and 3.10 introduced `|` for unions.
3.10 is also where deferred evaluation of annotations becomes the default, which makes me think of this feature as generally "python3.10-style annotations". I'm guessing the mention of 3.9 refers to the use of subscripts on builtin collection types.

lafrech:
> I'd let this slide rather than insisting that we catch everything.
> It's an initial step in this direction, and we can get the rest over time. Plus, I see existing type comments which should maybe become :-delimited annotations, so I expect we'll do another pass soon.

Makes sense. Let's merge, then. Anyone willing to go on with this is welcome to join the party.

PR #1935, opened 2022-01-05
Allow `fields.Nested` to consume a dict

sirosen:
To make it easier to leverage `Schema.from_dict`, allow `Nested` to consume a dict mapping names to fields. This is internally passed to `Schema.from_dict`.
That is, what previously would be written

    Nested(Schema.from_dict({"foo": String()}))

can now be written

    Nested({"foo": String()})

This relates to the original use-case for `Schema.from_dict`: easier lightweight schema generation.

---

Currently `webargs` redefines `Nested` to add this behavior, but it leads to a messy situation which I'm trying to sort out.
(References: https://github.com/marshmallow-code/webargs/issues/677 , https://github.com/marshmallow-code/webargs/pull/681 , https://github.com/marshmallow-code/webargs/issues/671 )

Presumably if this feature is useful in `webargs` -- and it looks like support for `webargs`-like use-cases was [the original motivation for `from_dict`](https://github.com/marshmallow-code/marshmallow/issues/1312) -- it would be useful to the broader `marshmallow` userbase.

The only thing which feels messy here is that the `fields` and `schema` modules are carefully avoiding direct dependencies via `base`. That won't work for `from_dict`, so I've wrapped that in a helper which does the import of `Schema` when it is called.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -508,7 +508,8 @@ class Nested(Field):
         # No
         author = fields.Nested(UserSchema(), only=('id', 'name'))
 
-    :param nested: `Schema` instance, class, class name (string), or callable that returns a `Schema` instance.
+    :param nested: `Schema` instance, class, class name (string), dictionary, or callable that
+        returns a `Schema` or dictionary. Dictionaries are converted with `Schema.from_dict`.
     :param exclude: A list or tuple of fields to exclude.
     :param only: A list or tuple of fields to marshal. If `None`, all fields are marshalled.
         This parameter takes precedence over ``exclude``.
@@ -523,7 +524,11 @@ class Nested(Field):
 
     def __init__(
         self,
-        nested: SchemaABC | type | str | typing.Callable[[], SchemaABC],
+        nested: SchemaABC
+        | type
+        | str
+        | dict[str, Field | type]
+        | typing.Callable[[], SchemaABC | dict[str, Field | type]],
         *,
         dump_default: typing.Any = missing_,
         default: typing.Any = missing_,
@@ -568,6 +573,8 @@ class Nested(Field):
                 nested = self.nested()
             else:
                 nested = self.nested
+            if isinstance(nested, dict):
+                nested = self._nested_from_dict(nested)
 
             if isinstance(nested, SchemaABC):
                 self._schema = copy.copy(nested)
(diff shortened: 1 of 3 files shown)


src/marshmallow/fields.py
```diff
@@ -566,6 +571,8 @@ def schema(self):
             context = getattr(self.parent, "context", {})
             if callable(self.nested) and not isinstance(self.nested, type):
                 nested = self.nested()
+            elif isinstance(self.nested, dict):
+                nested = self._nested_from_dict(self.nested)
```
lafrech: What about a callable returning a dict?

sirosen: It would be very easy to support. (Just push this down a couple of lines and update the type annotation.)

If I do that, the `:param nested:` doc needs an update. Would the following work?

> `Schema` instance, class, class name (string), dictionary, or callable that returns a `Schema` or dictionary. Dictionaries are converted with `Schema.from_dict`.

lafrech: Yes. Not that the use case is that important but when I looked at the code I figured it wouldn't be more complicated and I saw no reason not to do it.

sirosen:
I found a minor error at the same time that I was modifying to support a callable returning a dict: I used `Field | type[Field]` in the dict type for `from_dict`.

Field classes don't appear to be supported -- only field instances -- so I'm not sure why `from_dict` accepts `Field | type`. However, I've followed what `from_dict` supports in order to avoid potential issues.

src/marshmallow/fields.py
```diff
@@ -614,6 +621,13 @@ def _nested_normalized_option(self, option_name: str) -> list[str]:
             if field.startswith(nested_field)
         ]
 
+    # this helper defers the import of `marshmallow.schema` to avoid circular imports
+    # however, it otherwise aliases Schema.from_dict
+    def _nested_from_dict(self, fields: dict[str, Field | type]) -> type:
```
lafrech: Wouldn't it work the same by just adding the `Schema` import in` __init__` and calling `Schema.from_dict` from there?

Functionally equivalent, but less awkward, perhaps. In any case, just a matter of preference, no big deal.


```py
            if isinstance(nested, dict):
                from marshmallow.schema import Schema
                nested = Schema.from_dict(nested)
```

sirosen: Yeah, that would work equivalently-well. Before I added the annotations, I thought it would be nice to isolate this into a dedicated method, but it does look a bit messy now.

I'll switch to just doing the import in the spot you suggested. I think it will read more nicely.

lafrech (approved):
LGTM!

@sloria, any objection?

sloria (approved):
Makes sense to me. Thanks!

PR #1945, opened 2022-02-08
Fix Memory Leak

### LINKED ISSUE #1943: Memory leak?
I'm using `marshmallow` with Falcon and have observed a quite strange memory leak. This minimal application does nothing but provide a GET endpoint serializing a dictionary and profiling the memory usage with `memory_profiler`:

```python
from falcon import App, Request, Response
from memory_profiler import profile


class ArtistSchema(Schema):
    name = fields.Str()


class ArtistResource:
    @profile
    def on_get(self, request: Request, response: Response):
        response.text = ArtistSchema().dumps({"name": "David Bowie" * 10000})


application = App()
application.add_route('/artist', ArtistResource())
```

Sending now 100,000 requests like:

```python
import requests


for _ in range(100000):
    requests.get("http://127.0.0.1:8000/artist")
```

will slowly increase the memory footprint from 53.0 MiB to 54.5 MiB (and would apparently continue to increase like this).

I'm constructing a new `ArtistSchema` object for every request. But if I do this only once an reuse it for every request:

```python
ARTIST_SCHEMA = ArtistSchema()

class ArtistResource:
    @profile
    def on_get(self, request: Request, response: Response):
        response.text = ARTIST_SCHEMA.dumps({"name": "David Bowie" * 10000})
```

the memory usage does not increase. How is this possible? Or is this even a `marshmallow` issue?

My setup:
- Python 3.9.6
- `marshmallow` 3.14.1
- `falcon` 3.0.1
- `gunicorn` 20.1.0
- `memory-profiler` 0.60.0

deckar01:
Fixes #1943 

The garbage collector is not cleaning up circular references between schemas and fields, so the `__del__()` methods must explicitly clear these references.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -247,6 +247,10 @@ class Field(FieldABC):
     def __deepcopy__(self, memo):
         return copy.copy(self)
 
+    def __del__(self):
+        self.parent = None
+        self.root = None
+
     def get_value(self, obj, attr, accessor=None, default=missing_):
         """Return the value for a given key from an object.
 
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -404,6 +404,11 @@ class Schema(base.SchemaABC, metaclass=SchemaMeta):
         messages.update(self.error_messages or {})
         self.error_messages = messages
 
+    def __del__(self):
+        self.fields = None
+        self.load_fields = None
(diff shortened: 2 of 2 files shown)


deckar01:
Once I ran the extended test on dev, I saw the same behavior of the garbage collector being delayed. This change does not have any effect.

Issue #1960, opened 2022-03-15
Support multiple fields in the @validates decorator


dpriskorn:
Currently the decorator only supports one field. pydantic supports multiple see https://pydantic-docs.helpmanual.io/usage/validators/

e.g.
![bild](https://user-images.githubusercontent.com/68460690/158322713-27fb6caf-396b-4bef-b072-50745fb14281.png)

This makes it easier to specify multiple fields.

SJLC:
Agree that would be nice to have, but do you know the existing workaround of using `@validates_schema` decorator to validate multiple fields?

dpriskorn:
> Agree that would be nice to have, but do you know the existing workaround of using `@validates_schema` decorator to validate multiple fields?

Nope, is an example available somewhere?

caspii:
Example: I have a schema that has multiple fields for colors:
- `background_color`
- `text_color`
- `secondary_color`

Then I want a validator method that works like this

```python
    @validates("secondary_color")
    @validates("background_color")
    @validates("text_color")
    def validate_color(self, value):
        if not value or len(value) == 0:
            return
        if not value.startswith("#"):
            raise ValidationError("Color must start with #")
        if not all(c.upper() in "0123456789ABCDEF" for c in value[1:]):
            raise ValidationError("Invalid characters in color")
        if len(value) != 9:
            raise ValidationError("Color must be 9 characters long")
```

I am surprised that this issue is not more active.

lafrech:
Achieved in #1965.

PR #1965, opened 2022-03-24
@validates accepts multiple field names

dharani7998:
Fix for #1960


src/marshmallow/decorators.py
```diff
@@ -77,12 +77,12 @@ class MarshmallowHook:
...
 
 
-def validates(field_name: str) -> Callable[..., Any]:
+def validates(*field_names: str) -> Callable[..., Any]:
     """Register a field validator.
 
     :param str field_name: Name of the field that the method validates.
     """
-    return set_hook(None, VALIDATES, field_name=field_name)
+    return set_hook(None, VALIDATES, field_names=field_names)
```
deckar01, line 94: Should we try to preserve this hook interface for a minor release? The docstring for `set_hook` says "You should not need to use this method directly". Any objection to making `validates` return a wrapper that loops the calls to `set_hook` instead?

dharani7998: I don't see how the wrapper will work if we loop and call set_hook, `__marshmallow_hook__[VALIDATES]` value gets overwritten each time and ends with the last argument passed in the validates decorator.

deckar01: https://github.com/marshmallow-code/marshmallow/commit/c7bda74799ffd82640505f64b8b0de33213a84e8

dharani7998: It won't work, try this. It will only register the last field given in the decorator

```
from marshmallow import (
    Schema,
    fields,
    pre_dump,
    post_dump,
    pre_load,
    post_load,
    # pre_validate,
    validates,
    validates_schema,
    ValidationError,
    EXCLUDE,
    INCLUDE,
    RAISE,
)

class TestValidates(Schema):
    value = fields.Str()
    another_value = fields.Str()
    
    @validates('value', 'another_value')
    def validates_function(self, data, **kwargs):
        print(f'Inside validates: {{ {data}:{data} }}')


schema = TestValidates()
print(schema._hooks['validates'])
validator = getattr(schema, schema._hooks['validates'][0])
print(validator.__marshmallow_hook__)
print()



schema.load({
    'value': 'value',
    'another_value': 'another_value'
})
```

sloria:
apologies for the long delay on this. i hesitated initially because i had the same concern as @deckar01 about this being a breaking change. now that we're closing to releasing the next major, i think this is one that we can get merged. i'll update this when i have some time

sloria:
should validator methods receive `field_name`?

```python
from marshmallow import Schema, fields, validates, ValidationError

class UserSchema(Schema):
    name = fields.Str(required=True)
    nickname = fields.Str(required=True)

    @validates("name", "nickname")
    def validate_names(self, value: str, field_name: str) -> str:
        if len(value) < 3:
            raise ValidationError(f"{field_name} too short")
        return value
```

sloria:
looks like SQLAlchemy does pass a `key` argument: https://docs.sqlalchemy.org/en/20/orm/mapped_attributes.html#simple-validators

```python
from sqlalchemy.orm import validates


class EmailAddress(Base):
    __tablename__ = "address"

    id = mapped_column(Integer, primary_key=True)
    email = mapped_column(String)

    @validates("email")
    def validate_email(self, key, address):
        if "@" not in address:
            raise ValueError("failed simple email validation")
        return address
```

sloria:
Updated to pass `data_key` as a keyword argument to decorated methods. Would like one review on this since it's a breaking change

docs/quickstart.rst
```diff
@@ -301,12 +301,30 @@ It is sometimes convenient to write validators as methods. Use the `validates <m
...
+
+
+        class UserSchema(Schema):
+            name = fields.Str(required=True)
+            nickname = fields.Str(required=True)
+
+            @validates("name", "nickname")
+            def validate_names(self, value: str, data_key: str) -> None:
+                if len(value) < 3:
+                    raise ValidationError("Too short")
```
lafrech, line 326: This would be the right place to show the use of `data_key` like in the upgrading guide.

lafrech (commented):
No strong opinion about this. It adds a bit of API surface for a case that may not be so frequent, but addressing it with `validates_schema` is less convenient so this makes sense.

I'm not a fan of the "string or list of strings" pattern. In fact I didn't see where the magic operates but I suppose it does, otherwise all tests would be broken. And expecting a list of strings for a single field (99% case) would be a pity.

sloria:
> I'm not a fan of the "string or list of strings" pattern. In fact I didn't see where the magic operates but I suppose it does, otherwise all tests would be broken. And expecting a list of strings for a single field (99% case) would be a pity.

there's actually no special-casing here because 

```python
def validates(*field_names: str)
```

allows _one or more_ field names to be passed positionally.

lafrech:
But then in `validates`, the code shall not iterate on the string if a single field name is passed. If we blindly iterate on `field_names` when a single string is passed, it won't work. I expected to see a type check such as `iterable_not_string`.

(I'm not talking about mypy / type checking but about runtime.)

sloria:
`field_names` will always be a tuple, even when a single argument is passed.

```python
>>> def validates(*field_names: str):
...     print(type(field_names))
...     
>>> validates("foo")
<class 'tuple'>
```

lafrech:
Oh sorry, I missed the obvious.

The user never passes a tuple or list, just N positional args.

lafrech:
I just rebased this but couldn't figure out how to (force) push into the original PR branch.

Here's the rebase: https://github.com/marshmallow-code/marshmallow/compare/pr/1965

Is there anything else that needs to be done to merge this?

sloria:
no blockers to merging!

lafrech:
Looks like we're done with [milestone 4.0](https://github.com/marshmallow-code/marshmallow/milestone/15).

Should we ship? Review open issues/PR to see if anything else should be crammed in before releasing?

#2799 and #2810 are not breaking so they could be added later.

sloria:
nice! let's let `dev` bake for a few weeks. i'm traveling a bit for the next couple weeks. once i'm back home i'll review issues/PRs that could be breaking-but-low-impact that could go into v4

lafrech:
Great. No rush on my side.

Enjoy your trip!

PR #1970, opened 2022-04-01
Run pre-commit autoupdate to update `black`

sirosen:
@sloria, @lafrech, can we fast-track this, assuming CI shows no issues? At `black==22.1.0`, a fresh install (as in CI) gets `click==8.1.0` and fails. There are issues on `black` and on `click` which explain how this happened, but the fix is just to bump `black`.

---

Update to latest tool versions, in particular `black==22.3.0` to fix a bug which caused it to be incompatible with the latest `click` version and crashfail.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,11 @@
 repos:
 - repo: https://github.com/asottile/pyupgrade
-  rev: v2.31.0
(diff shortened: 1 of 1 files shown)


deckar01:
https://github.com/psf/black/issues/2964

deckar01 (approved):

PR #1972, opened 2022-04-04
print OPTIONS_CLASS on TypeError (#597)

milahu:
part of #597

example output

> ```TypeError: __init__() got an unexpected keyword argument 'ordered' in OPTIONS_CLASS = <class 'portia_api.jsonapi.serializers.JsonApiSerializerOpts'>```

so the user should fix

```diff
 class JsonApiSerializerOpts(SchemaOpts):
-    def __init__(self, meta):
+    def __init__(self, meta, ordered):
```

with

```py
class JsonApiSerializer(with_metaclass(JsonApiSerializerMeta, BaseSchema)):
    OPTIONS_CLASS = JsonApiSerializerOpts
```

### DIFF
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -109,9 +109,15 @@ class SchemaMeta(type):
         inherited_fields = _get_fields_by_mro(klass, ordered=ordered)
 
         meta = klass.Meta
+
         # Set klass.opts in __new__ rather than __init__ so that it is accessible in
         # get_declared_fields
-        klass.opts = klass.OPTIONS_CLASS(meta, ordered=ordered)
+        try:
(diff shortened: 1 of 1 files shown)


sloria (commented):
thanks for the PR. the code looks good, but i don't think the benefit here is worth the extra LOC (i realize this is totally debatable!). if a user is already aware of `OPTIONS_CLASS` (which they must be to hit the relevant error), then they should be able to use the existing traceback to debug ctor issues.

milahu:
> if a user is already aware of `OPTIONS_CLASS` (which they must be to hit the relevant error)

no. when debugging 3rd party code, i dont even know what marshmallow is

> [the user] should be able to use the existing traceback to debug ctor issues.

my goal is to make this easier, faster
the current code is hiding valuable information from the user

PR #1976, opened 2022-04-18
Add messages_dict property to ValidationError

### LINKED ISSUE #1442: Typing issue with ValidationError.messages?
Not sure if this is a documentation or typing issue, but `ValidationError.messages` is given as `Union[List[Any], Dict[Any, Any]]` while the docs say in a few places it's going to be a dictionary, at least when coming from `load()`.

For example, in the [validation](https://marshmallow.readthedocs.io/en/stable/quickstart.html#validation) documentation it says, "Schema.load() (and its JSON-decoding counterpart, Schema.loads()) raises a ValidationError error when invalid data are passed in. You can access the dictionary of validation errors from the ValidationError.messages attribute."

A slightly modified example from the documentation that shows it:

```python
from pprint import pprint

from marshmallow import Schema, fields, ValidationError


class BandMemberSchema(Schema):
    name = fields.String(required=True)
    email = fields.Email()


user_data = {"email": "mick@stones.com", "name": "Mick"}

try:
    BandMemberSchema().load(user_data)
    errors: dict = {}
    errors = err.messages
    pprint(err.messages)
```

Running it through MyPy:

```
> $ mypy m3example.py
m3example.py:17: error: Incompatible types in assignment (expression has type "Union[List[Any], Dict[Any, Any]]", variable has type "Dict[Any, Any]")
Found 1 error in 1 file (checked 1 source file)
```

- Python 3.7.4
- Marshmallow 3.2.1
- MyPy 0.740

Side note, the example in the docs gives MyPy errors as-is as well: `error: Argument 1 to "load" of "Schema" has incompatible type "List[Dict[str, str]]"; expected "Mapping[Any, Any]"`.

sirosen:
'messages_dict' is an alias for 'messages' which checks if the value is a dict and raises a TypeError if it is not.

This "type asserting property" leads to easier usage with type checkers. In many cases, the 'messages' attribute is known to be a dict, but is annotated as a union type, leading to false negatives from type checkers. Using 'messages_dict', one can write

```python
x: dict = err.messages_dict
```

rather than

```python
x: dict = typing.cast(dict, err.messages)
```

To validate this behavior at type checking time and runtime, two tests are added:

1. A pytest test which checks that the desired TypeError is raised if the value is not a dict

2. A new file to act as a "mypy test case", using `mypy --warn-unused-ignores` to check that things are annotated correctly

(1) is like existing tests, but (2) requires a little bit of new infrastrucutre. A new dir for the tests, which is listed in pytest norecursedirs, and a new tox env (listed in CI as well) which runs `mypy` with the desirable flags.

resolves #1442

---

Some minor notes:
- This behavior was one of a couple of possible solutions for #1442, where it got a 👍 from the OP. It has some faults, but I think it's the simplest solution.
- The testing strategy used for annotations is recommended [in the standalone `typing` docs](https://typing.readthedocs.io/en/latest/source/quality.html#testing-using-mypy-warn-unused-ignores). As far as I know, it's the simplest way of adding testing for type annotations, especially as it doesn't require learning any new tools.

### DIFF
--- a/tests/mypy_test_cases/test_validation_error.py
+++ b/tests/mypy_test_cases/test_validation_error.py
@@ -1,0 +1,20 @@
+from __future__ import annotations
+
+import marshmallow as ma
+
+# OK types for 'message'
+ma.ValidationError("foo")
+ma.ValidationError(["foo"])
+ma.ValidationError({"foo": "bar"})
+
+# non-OK types for 'message'
+ma.ValidationError(0)  # type: ignore[arg-type]
+
+# 'messages' is a dict|list
+err = ma.ValidationError("foo")
+a: dict | list = err.messages
+# union type can't assign to non-union type
+b: str = err.messages  # type: ignore[assignment]
+c: dict = err.messages  # type: ignore[assignment]
+# 'messages_dict' is a dict, so that it can assign to a dict
+d: dict = err.messages_dict
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,6 +19,8 @@ repos:
   hooks:
   - id: mypy
     additional_dependencies: [types-simplejson, types-pytz, packaging]
+    # these files are checked under `tox -e mypy-test`
+    exclude: ^tests/mypy_test_cases/.*$
 - repo: https://github.com/asottile/blacken-docs
   rev: v1.12.1
   hooks:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -15,7 +15,7 @@ resources:
 jobs:
 - template: job--python-tox.yml@sloria
   parameters:
-    toxenvs: [lint, py37, py310]
+    toxenvs: [lint, mypy-test, py37, py310]
     os: linux
 - template: job--pypi-release.yml@sloria
   parameters:
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,7 +8,7 @@ max-complexity = 18
 select = B,C,E,F,W,T4,B9
 
 [tool:pytest]
-norecursedirs = .git .ropeproject .tox docs env venv
+norecursedirs = .git .ropeproject .tox docs env venv tests/mypy_test_cases
 addopts = -v --tb=short
 
 [mypy]
--- a/src/marshmallow/exceptions.py
+++ b/src/marshmallow/exceptions.py
@@ -47,6 +47,15 @@ class ValidationError(MarshmallowError):
             return self.messages
         return {self.field_name: self.messages}
 
+    @property
+    def messages_dict(self) -> dict[str, typing.Any]:
+        if not isinstance(self.messages, dict):
+            raise TypeError(
+                "cannot access 'messages_dict' when 'messages' is of type "
+                + type(self.messages).__name__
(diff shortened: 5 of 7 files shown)


lafrech:
I'm not very familiar with typing as I don't use that in the apps we develop. There are still no type hints in flask-smorest (next step now that webargs and apispec are covered). All the typing issues here give me the feeling that although it may be the right thing to do, the cost is high.

This said, since the solution you used is the recommended one, and since you did the job already, I have no objection to merging this.

Thanks.

scottbelden:
Is the plan to merge this? I hit this same bug within 30 minutes of using marshmallow.

tests/mypy_test_cases/test_validation_error.py
```diff
@@ -0,0 +1,20 @@
...
+
+# non-OK types for 'message'
+ma.ValidationError(0)  # type: ignore[arg-type]
+
+# 'messages' is a dict|list
+err = ma.ValidationError("foo")
+a: dict | list = err.messages
+b: str = err.messages  # type: ignore[assignment]
+# union type can't assign to non-union type
+c: dict = err.messages  # type: ignore[assignment]
```
lafrech, line 18: Looks like the comment applies to both b and c. Shouldn't it be above b?

lafrech (approved):
I didn't investigate the alternative proposed in https://github.com/marshmallow-code/marshmallow/issues/1442#issuecomment-1085987284 involving a subclass guaranteeing a `dict`.

While it kinda sucks to add a runtime check to please a linter, this is not on the critical path and users are free not to use it (if they don't use a linter or if they prefer to add a comment on each line where mypy complains to tell it to ignore).

I can merge and release. @sloria, any objection?

tests/mypy_test_cases/test_validation_error.py
sirosen: Yeah, looks like an oversight. Let me amend before you merge. 👍

sirosen:
This isn't a bugfix. I think users should be crystal clear about this. `ValidationError.messages` is correctly annotated today.

This is a new feature, intended to improve usage within type-checked codebases.

I think this is one of the two best ways to address #1442, the other being to close it as not-a-bug.

> I didn't investigate the alternative proposed in [#1442 (comment)](https://github.com/marshmallow-code/marshmallow/issues/1442#issuecomment-1085987284) involving a subclass guaranteeing a `dict`.

I think it would work, but it requires changes to ensure that the new subclass is used where appropriate. That seems kind of difficult and fiddly, relative to the benefit delivered.

Plus, it means users need to catch and check for the new subclass on their errors, which is at least as much work as checking `isinstance(err.messages, dict)` themselves.

> While it kinda sucks to add a runtime check to please a linter, this is not on the critical path and users are free not to use it (if they don't use a linter or if they prefer to add a comment on each line where mypy complains to tell it to ignore).

I waffle on this topic a lot. Is `x = err.messages_dict` more correct than `x = cast(dict, err.messages)`?

`messages_dict` would raise an informative `TypeError` at runtime if the data is a list. For some use-cases, that might be useful. The developer gets faster feedback that the data doesn't meet their expectations.

Issue #1982, opened 2022-05-03
Non-obvious error when using fields.Nested with schema instance and many


mmulqueen:
I discovered the hard way that fields.Nested silently ignores the many parameter when it's passed an instance. It's my fault, I can see in the docstring that this behaviour is described, but that didn't stop me wasting a load of time debugging the error. I see that using many=True is no longer recommended anyway, but it's not deprecated yet either - #779 may change that. The issue here isn't how Marshmallow behaves, but that it isn't obvious.

Actual behaviour: 

Preferred, works: `fields.List(fields.Nested(Child()))`
Old style with nested class, works: `fields.Nested(Child, many=True)`
Old style with nested instance, broken: `fields.Nested(Child(), many=True)`

When trying the nested instance way, the error you get is unhelpful because it's the result of Marshmallow treating the nested schema as singular:

     {'_schema': ['Invalid input type.']}

Confusing things even further, many and a nested instance seem to work when dumping (even though it shouldn't), but not when loading (as described). Since the broken way half works, I'm sure there will be code out there that relies on it. 

Pull request to follow.

PR #1983, opened 2022-05-03
Warnings and errors for incompatible use of schema instance and many in fields.Nested

mmulqueen:
Addresses the issue I've described in #1982. Behaviour should be unchanged, but more obvious.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -25,9 +25,10 @@ from marshmallow.exceptions import (
     ValidationError,
     StringNotCollectionError,
     FieldInstanceResolutionError,
+    FieldConfigurationError,
 )
 from marshmallow.validate import And, Length
-from marshmallow.warnings import RemovedInMarshmallow4Warning
+from marshmallow.warnings import RemovedInMarshmallow4Warning, FieldConfigurationWarning
 
 __all__ = [
     "Field",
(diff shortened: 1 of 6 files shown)


src/marshmallow/fields.py
```diff
@@ -634,9 +640,14 @@ def _serialize(self, nested_obj, attr, obj, **kwargs):
         return schema.dump(nested_obj, many=many)
 
     def _test_collection(self, value):
+        if self.many and not self.schema.many:
+            raise FieldConfigurationError(
+                '"many" may not be used in conjunction with an instance in a nested field.'
+            )
```
lafrech, line 646: I'd remove this test. This should be coverered by the init warning you added above, right?

src/marshmallow/fields.py
```diff
@@ -634,9 +640,14 @@ def _serialize(self, nested_obj, attr, obj, **kwargs):
...
         many = self.schema.many or self.many
         if many and not utils.is_collection(value):
             raise self.make_error("type", input=value, type=value.__class__.__name__)
+        # Check whether many on the nested field and the schema are in conflict.
```
lafrech, line 650: I guess this comment was meant for the change above?

src/marshmallow/fields.py
```diff
@@ -551,6 +552,11 @@ def __init__(
                 "Use `Nested(lambda: MySchema(...))` instead.",
                 RemovedInMarshmallow4Warning,
             )
+        if isinstance(nested, SchemaABC) and many:
+            warnings.warn(
+                'Passing "many" is ignored if using an instance of a schema. Consider fields.List(fields.Nested(...)) instead.',
+                FieldConfigurationWarning,
+            )
```
lafrech, line 559: I'd remove the suggestion.

lafrech (commented):
Thanks for the contribution and sorry about the delay.

I understand the issue. Yet I'm reluctant to add code just to warn about this specific case.

I left a few comments anyway. I might change my mind and accept the test at init. No strong opinion against it.

deckar01:
#779 lists other unexpected edge cases like this. I would rather bite the bullet and deprecate `many=True` than go down the path of maintaining edge case warnings.

sloria:
closing as this has gone stale, and i agree with @deckar01 that we should just get rid of the `many=True` API altogether (eventually)

PR #1984, opened 2022-05-05
Fix PyCharm warnings on fields

rice0208:
When using marshmallow with PyCharm, there's a warning like this:
image

The problem is caused by the type hint of argument `validate` of `Field` in [fields.py](https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/fields.py), and this problem can be fixed by the PR.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -155,11 +155,9 @@ class Field(FieldABC):
         default: typing.Any = missing_,
         data_key: str | None = None,
         attribute: str | None = None,
-        validate: typing.Optional(
-            typing.Union(
-                typing.Callable[[typing.Any], typing.Any],
-                typing.Iterable[typing.Callable[[typing.Any], typing.Any]]
-            )
+        validate: None | (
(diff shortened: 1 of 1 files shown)


rice0208:
Linter failed, but if don't do this, pycharm warning occurs.

greyli:
Maybe do something like this (keep `|`  with `None`):

```python
validate: None |
(
    typing.Callable[[typing.Any], typing.Any]
    | typing.Iterable[typing.Callable[[typing.Any], typing.Any]]
) = None,
```

rice0208:
> Maybe do something like this (keep `|` with `None`):
> 
> ```python
> validate: None |
> (
>     typing.Callable[[typing.Any], typing.Any]
>     | typing.Iterable[typing.Callable[[typing.Any], typing.Any]]
> ) = None,
> ```

linter will fail if do this, I'll try typing to pass the linter :-)
but if linter passes, pycharm will display warnings, that's why I did this.

lafrech:
I'd like to close this. Seems like a PyCharm issue.

See #2066.

PR #1986, opened 2022-05-10
Add stacklevel to print appropriate location in warnings.

tirkarthi:
Adding stacklevel will print the location of the warning in source code than Marshmallow which is more helpful.

https://docs.python.org/3/library/warnings.html#warnings.warn

```python
# /tmp/foo.py
import json
from marshmallow.utils import pprint
from marshmallow import Schema, fields, SchemaOpts

class CustomOpts(SchemaOpts):
    pass


class CustomField(fields.Field):
    pass


class FooSchema(Schema):
    custom_field = CustomField(default=True, missing=True, foo=1)
    custom_nested_field = fields.Nested('self')

    class Meta:
        json_module = json

custom_field = CustomField()
custom_field.default
custom_field.default = True
custom_field.missing
custom_field.missing = True
pprint("hello")
custom_field.fail("custom_field")

```

Current warning

```
/home/karthikeyan/stuff/python/marshmallow/src/marshmallow/fields.py:173: RemovedInMarshmallow4Warning: The 'default' argument to fields is deprecated. Use 'dump_default' instead.
  warnings.warn(
```

With PR : 

```
/tmp/foo.py:14: RemovedInMarshmallow4Warning: The 'default' argument to fields is deprecated. Use 'dump_default' instead.
  custom_field = CustomField(default=True, missing=True, foo=1)
```

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -168,3 +168,4 @@ Contributors (chronological)
 - Ben Windsor `@bwindsor <https://github.com/bwindsor>`_
 - Kevin Kirsche `@kkirsche <https://github.com/kkirsche>`_
 - Isira Seneviratne `@Isira-Seneviratne <https://github.com/Isira-Seneviratne>`_
+- Karthikeyan Singaravelan `@tirkarthi  <https://github.com/tirkarthi>`_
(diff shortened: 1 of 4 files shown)


lafrech:
This makes sense.

Should we wrap this in a `utils.deprecate` function?

Also please add yourself to AUTHORS file.

tirkarthi:
Thanks, added myself to authors. `stacklevel` depends on how deeply the deprecation warning is raised and 2 is more common since it's direct call. In case of nested calls 2 might print stacklevel that is not useful. I feel it's okay to add stacklevel to individual instances instead of a wrapper function for this.

sloria (approved):

PR #1989, opened 2022-05-23
Validate values passed for 'unknown'

### LINKED ISSUE #1721: Is it intentional that setting parameter unknown=<any non-empty string> behaves the same as unknown=EXCLUDE?

sirosen:
There are three places that a user can pass a value for 'unknown':
- `Meta` in a Schema class definition
- `Schema.__init__`
- `Schema.load`

For each of these, call an internal check to confirm that the value is valid. In cases where `unknown or self.unknown` is used to support `None`, do the validation after the `or`.

A new test case covers these three locations and confirms that passing an unknown (hah!) value causes a ValueError.

closes #1721, #1732

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,11 @@ Changelog
 (unreleased)
 ************
 
+Features:
+
+- Raise `ValueError` if an invalid value is passed to the ``unknown`` argument (:issue:`1721`, :issue:`1732`).
+  Thanks :user:`sirosen` for the PR.
+
 Other changes:
 
 - Set lower bound for `packaging` requirement (:issue:`1957`).
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -390,7 +390,11 @@ class Schema(base.SchemaABC, metaclass=SchemaMeta):
         self.load_only = set(load_only) or set(self.opts.load_only)
         self.dump_only = set(dump_only) or set(self.opts.dump_only)
         self.partial = partial
-        self.unknown = validate_unknown_parameter_value(unknown or self.opts.unknown)
+        self.unknown = (
+            self.opts.unknown
+            if unknown is None
+            else validate_unknown_parameter_value(unknown)
+        )
         self.context = context or {}
         self._normalize_nested_options()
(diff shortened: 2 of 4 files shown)


lafrech:
Thanks for this.

> In cases where unknown or self.unknown is used to support None, do the validation after the or.

I suppose it would be slightly more efficient to only validate `unknown` in this case as `self.unknown` is supposed to be valid since it was validated earlier (thanks to this PR).

The code would be a little longer, though. Not a one-liner anymore.

Not sure this is relevant. The validation itself is just one statement after all:

```py
    if obj not in (INCLUDE, EXCLUDE, RAISE):
```

(Would a (precomputed at import) `set` be marginally faster?)

sirosen:
> I suppose it would be slightly more efficient to only validate `unknown` in this case as `self.unknown` is supposed to be valid since it was validated earlier (thanks to this PR).
> 
> The code would be a little longer, though. Not a one-liner anymore.

I was thinking the same thing and then I realized that there's a nice way to write this:
```python
unknown = self.unknown if unknown is None else validate_unknown_parameter_value(unknown)
```

I'll make that change since it should be slightly faster when `load` is called without the `unknown` arg (which is where I think it really matters to try to make things fast).

I don't think it really matters -- once these changes are tweaked as above -- to use a module-scoped set, but I'll make that change too. We're talking about it and I see no reason _not_ to do it that way.

lafrech (approved):

sloria (approved):

PR #1992, opened 2022-05-25
Remove Meta.order_by from peewee example

### LINKED ISSUE #1980: Confusing Meta option in examples
In  [one of the examples](https://github.com/marshmallow-code/marshmallow/blob/5f7e6172acdadc6221b21faf921fcf3fe8f1aa23/examples/peewee_example.py#L42)  there is `Meta.order_by` defined, however I don't see trace of support for something like this in codebase after all.

sirosen:
order_by was deprecated and removed from peewee

closes #1980


sloria (approved):
Thanks!

PR #1998, opened 2022-06-03
Support serialization as float in TimeDelta field

marcosatti:
Add functionality to (de)serialize to float values as a representation of the total seconds. 

In our use case, we are trying to specify a period which is a combination of seconds, milliseconds and microseconds, which is easier to represent as a single floating point value rather than specifying the smallest unit (microseconds). We have currently been serializing to a `fields.Float()` using `.total_seconds()` and the equivalent to deserialize.

The feature does not promise absolute precision, which is stated in the docs (same restrictions as floating point values).

This does not change the current way `TimeDelta` is used.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1481,7 +1481,10 @@ class TimeDelta(Field):
     }
 
     def __init__(
-        self, precision: str = SECONDS, serde_type: type[int | float] = int, **kwargs
+        self,
+        precision: str = SECONDS,
+        serialization_type: type[int | float] = int,
+        **kwargs,
     ):
         precision = precision.lower()
         units = (
@@ -1500,17 +1503,11 @@ class TimeDelta(Field):
             )
             raise ValueError(msg)
 
-        serde_types = (
-            int,
-            float,
-        )
-
-        if serde_type not in serde_types:
-            msg = f"The (de)serialization must be one of {serde_types}"
-            raise ValueError(msg)
+        if serialization_type not in (int, float):
+            raise ValueError("The serialization type must be one of int or float")
 
         self.precision = precision
-        self.serde_type = serde_type
+        self.serialization_type = serialization_type
         super().__init__(**kwargs)
 
     def _serialize(self, value, attr, obj, **kwargs):
@@ -1519,17 +1516,17 @@ class TimeDelta(Field):
 
         base_unit = dt.timedelta(**{self.precision: 1})
 
-        if self.serde_type is int:
+        if self.serialization_type is int:
             delta = utils.timedelta_to_microseconds(value)
             unit = utils.timedelta_to_microseconds(base_unit)
             return delta // unit
         else:
-            assert self.serde_type is float
+            assert self.serialization_type is float
             return value.total_seconds() / base_unit.total_seconds()
 
     def _deserialize(self, value, attr, data, **kwargs):
         try:
-            value = self.serde_type(value)
+            value = self.serialization_type(value)
         except (TypeError, ValueError) as error:
             raise self.make_error("invalid") from error
 
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1440,7 +1440,7 @@ class TimeDelta(Field):
     :param precision: Influences how the integer or float is interpreted during
         (de)serialization. Must be 'days', 'seconds', 'microseconds',
         'milliseconds', 'minutes', 'hours' or 'weeks'.
-    :param serde_type: Whether to (de)serialize to a `int` or `float`.
+    :param serialization_type: Whether to (de)serialize to a `int` or `float`.
     :param kwargs: The same keyword arguments that :class:`Field` receives.
 
     Integer Caveats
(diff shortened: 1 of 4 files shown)


tests/test_serialization.py
```diff
@@ -721,6 +733,19 @@ def test_timedelta_field(self, user):
...
+            days=6,
+            hours=2,
+            minutes=5,
+            seconds=51,
+            milliseconds=10,
+            microseconds=742,
+        )
+        field = fields.TimeDelta(fields.TimeDelta.TOTAL_SECONDS)
+        # Test for reasonable approximate equality. No guarantees are made about accuracy.
+        assert abs(field.serialize("d10", user) - 1130751.010742) < 0.1
```
lafrech: I'd use math.isclose for this.

lafrech (commented):
Thanks for the contribution.

I understand your use case.

I'm reluctant to integrate this as is, though. At this point, we might as well offer int and float for any precision. This would have the benefit of symmetry, the possibilities would be the cartesian product of {all precisions} x {int, float}. And one could serialize to a floating point number of days.

You'd just have to add a new parameter expressing int or float (int by default).

Just my thoughts, I could be convinced otherwise and maybe other maintainers will chime in with a different view.

marcosatti:
Having a `serde_type` parameter does sound like a more complete approach to the problem - I'd be ok to add that in. I'll wait for more discussion first.

marcosatti:
Went ahead and updated anyway to support all precisions through use of the `serde_type` parameter.

src/marshmallow/fields.py
```diff
@@ -1479,20 +1500,36 @@ def __init__(self, precision: str = SECONDS, **kwargs):
             )
             raise ValueError(msg)
 
+        serde_types = (
+            int,
+            float,
+        )
+
+        if serde_type not in serde_types:
```
lafrech: Nitpick: Either I'd write `if serde_type not in (int, float)` or I'd set `SERDE_TYPES` as class attribute.

I use a class attribute when I expect subclassing. In this case, I don't see much use for subclassing, so both suit me.

src/marshmallow/fields.py
```diff
@@ -1461,7 +1480,9 @@ class TimeDelta(Field):
         "format": "{input!r} cannot be formatted as a timedelta.",
     }
 
-    def __init__(self, precision: str = SECONDS, **kwargs):
+    def __init__(
+        self, precision: str = SECONDS, serde_type: type[int | float] = int, **kwargs
```
lafrech: I don't really like the name. I suppose you mean serialization/deserialization type. Although it is involved in both serialization and deserialization, to me the name implies the destination type, so it is the serialization type while the deserialization type is `dt.timedelta`.

Therefore, I'd call it `serialization_type` or `ser_type`. Those doesn't sound so nice, so I'm open to suggestions.

src/marshmallow/fields.py
```diff
@@ -1479,20 +1500,36 @@ def __init__(self, precision: str = SECONDS, **kwargs):
...
             raise ValueError(msg)
 
+        serde_types = (
+            int,
+            float,
+        )
+
+        if serde_type not in serde_types:
+            msg = f"The (de)serialization must be one of {serde_types}"
+            raise ValueError(msg)
```
lafrech: I don't see a test for this, but I'm not sure it is required.

lafrech (commented):
Nice, thanks!

Just minor comments.

src/marshmallow/fields.py
marcosatti: I put one into `test_serialization.py` although maybe its not the correct place for it...

src/marshmallow/fields.py
marcosatti: Went with `serialization_type`... Maybe `primitive_type` could be used but I feel the former describes it better.

src/marshmallow/fields.py
marcosatti: Cleaned this up to be more inline.

src/marshmallow/fields.py
lafrech: Oh, right. Didn't see it.

src/marshmallow/fields.py
```diff
@@ -1434,17 +1434,36 @@ def _make_object_from_format(value, data_format):
...
+    Smaller units than microseconds cannot be properly deserialized and will result
+    in the truncation of any fractional data. This is only possible when using `float`.
+    For example, 1.12345 interpreted as microseconds will result in `timedelta(microseconds=1)`.
+
     .. versionchanged:: 2.0.0
         Always serializes to an integer value to avoid rounding errors.
         Add `precision` parameter.
+    .. versionchanged:: 3.17.0
+        Allow (de)serialization to `float` through use of a new `serde_type` parameter.
+        `int` is the default to retain previous behaviour.
```
lafrech, line 1465: > Allow (de)serialization to `float` through use of a new `serde_type` parameter.

Wrong param name.

src/marshmallow/fields.py
```diff
@@ -1434,17 +1434,36 @@ def _make_object_from_format(value, data_format):
...
+
+    Float Caveats
+    -------------
+    Use of `float` when (de)serializing may result in data precision loss due
+    to the way machines handle floating point values.
+
+    Microseconds is the smallest unit able to be represented wholly.
+    Smaller units than microseconds cannot be properly deserialized and will result
+    in the truncation of any fractional data. This is only possible when using `float`.
+    For example, 1.12345 interpreted as microseconds will result in `timedelta(microseconds=1)`.
```
lafrech: I'm not sure about the meaning of "This is only possible when using `float`." I suppose you mean only float is affected because there is no fractional part in int. But I found the sentence confusing. Just removing it might make things clearer. The example clarifies things anyway.

Or perhaps rephrase and state that whatever the precision chosen, the fractional part when using float will always be truncated to microseconds.

lafrech (approved):
LGTM. Just a few comments.

I'll wait a few days in case other maintainers want to review as well.

marcosatti:
Fixed up remaining documentation issues.

PR #2002, opened 2022-06-15
Show errors on serialization of broken property

symstu:


src/marshmallow/utils.py
```diff
@@ -258,7 +259,8 @@ def _get_value_for_key(obj, key, default):
 
     try:
         return obj[key]
-    except (KeyError, IndexError, TypeError, AttributeError):
+    except (KeyError, IndexError, TypeError, AttributeError) as e:
+        logging.error(e, exc_info=True)
```
deckar01, line 263: If we did want to log this it would be a warning, not a logging error. Falling back to property access is a documented behavior, so an unconditional warning would be irrelevant noise for the vast majority of calls. It is impossible to differentiate between an intentional and an unexpected `AttributeError` coming from a property, so there is no way for us to actually determine if the user should be warned.

lafrech:
Thanks for contributing. Unfortunately, we can't merge this.

Perhaps the situation could be improved if there was a way to let the getter know whether it needs to getitem or getattr. I think there has been discussions about this for performance issue.

PR #2011, opened 2022-07-04
Improved search of data_keys_duplicates and attributes_duplicates

izveigor:
Counter sorts elements by the number of occurrence of elements and has the same asymptotic notation like a set. This improvement speeds up the code.

### DIFF
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -995,26 +995,28 @@ class Schema(base.SchemaABC, metaclass=SchemaMeta):
             field_obj.data_key if field_obj.data_key is not None else name
             for name, field_obj in dump_fields.items()
         ]
-        if len(dump_data_keys) != len(set(dump_data_keys)):
-            data_keys_duplicates = {
-                x for x in dump_data_keys if dump_data_keys.count(x) > 1
-            }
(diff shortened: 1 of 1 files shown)


src/marshmallow/schema.py
```diff
@@ -995,26 +995,38 @@ def _init_fields(self) -> None:
...
-            }
+
+        data_keys_counter = Counter(dump_data_keys)
+        if len(dump_data_keys) != len(data_keys_counter):
+            data_keys_duplicates = []
+            for data_key, count in data_keys_counter.items():
+                if count > 1:
+                    data_keys_duplicates.append(data_key)
+                else:
+                    break
```
lafrech: Why the break?

lafrech (commented):
AFAIU, `Counter` might speed-up the calculation of the count for each key, at the cost of an extra cost at instantiation.

Since instantiation is on the critical path and count only occurs on errors, I'd say using `set` is better.

Also, why remove the `dict` comprehension?

izveigor:
Hello, @lafrech!
Thanks to the quick response.

I agree that Counter instantiation is a expensive operation. So, I suggest another idea to optimize this part of code.
If assignment of a variable is not expensive operation in Python, this improvement can speed up the code.

PR #2014, opened 2022-07-19
docs: fix simple typo, stict -> strict

timgates42:
There is a small typo in src/marshmallow/fields.py.

Should read `strict` rather than `stict`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md


lafrech:
Good catch, thanks!

PR #1885, opened 2021-10-05
Add EnumString

orenc17:
EnumString is an attempt to parse Enums from String

Issue #267, opened 2015-08-28
Support Enum for Select-field


floqqi:
I have got some use cases where I use a `Enum` in my models.

Example:

``` python
# definitions.py
from enum import Enum

class Gender(Enum):
    male = 'm'
    female = 'f'
```

``` python
# models.py
from definitions import Gender

class Person:
    def __init__(self, gender: Gender):
        self.gender = gender
```

Now it would be great if I could use the Enum in `marshmallow.fields.Select`:

``` python
# schemas.py
from marshmallow import fields, Schema
from definitions import Gender
from models import Person

class PersonSchema(Schema):
    gender = fields.Select(Gender)

    @staticmethod
    def make_object(data) -> Person:
        return Person(**data)
```

For backwards-compatibility [enum34](https://pypi.python.org/pypi/enum34) could be used.

floqqi:
If you think this might be a good idea I could provide a pull request if I find some time over the weekend.

floqqi:
Ah, I see it's deprecated. Shouldn't then `marshmallow.validate.OneOf` support `Enum`?

sloria:
@floqqi I would accept a PR adding `Enum` support to `OneOf`. 

I'd rather not add `enum34` as a dependency. We can just support Enum if it is available in the user's Python environment:

``` python
HAS_ENUM = False  # Check this in OneOf
try:
    import enum
except ImportError:
    pass
else:
    HAS_ENUM = True
```

sloria:
Another option is to add a separate `Enum` validator. That way, we can avoid type-checking in `OneOf`.

ctolsen:
I've been using an `Enum` field like this:

``` python
class Enum(fields.Field):
    """Validates against a given set of enumerated values."""
    def __init__(self, enum, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.enum = enum
        self.validators.insert(0, OneOf([v.value for v in self.enum]))

    def _serialize(self, value, attr, obj):
        return self.enum(value).value

    def _deserialize(self, value):
        return self.enum(value)

    def _validate(self, value):
        if type(value) is self.enum:
            super()._validate(value.value)
        else:
            super()._validate(value)
```

Feel free to grab some of that if it helps.

floqqi:
I will take a look at this tomorrow. But I think `Enum` should be a separate field kinda like @ctolsen mentioned. Or would you prefer a validator like `OneOfEnum`, @sloria?

[...]

Issue #2012, opened 2022-07-18
Custom Field Logic Bypass


deckar01:
If you want to create a field that transforms `None` to a preset value, a custom field is required.

Currently the documentation for "Creating A Field Class" states:

> To create a custom field class, create a subclass of [marshmallow.fields.Field](https://marshmallow.readthedocs.io/en/latest/marshmallow.fields.html#marshmallow.fields.Field) and implement its [_serialize](https://marshmallow.readthedocs.io/en/latest/marshmallow.fields.html#marshmallow.fields.Field._serialize) and/or [_deserialize](https://marshmallow.readthedocs.io/en/latest/marshmallow.fields.html#marshmallow.fields.Field._deserialize) methods.

https://marshmallow.readthedocs.io/en/latest/custom_fields.html#creating-a-field-class

```py
class Integer(fields.Integer):
    def _deserialize(self, value, *args, **kwargs):
        value = value or 0
        return super()._deserialize(value, *args, **kwargs)

class Test(Schema):
    foo = Integer(allow_none=True)

Test().load({'foo': None}) # but it produces {"foo": None} instead of {"foo": 0}
```

When `None` (or `missing`) is encountered `_deserialize()` is never executed, because `Field.deserialize()` returns immediately. The current work around appears to require an understanding of undocumented parts of the `Field` implementation and extending `deserialize()` instead. The documentation should probably explain the `_/de/serialize()` and None/missing interaction.

```py
class Amount(fields.Integer):
    def deserialize(self, value, *args, **kwargs):
        value = value or 0
        return super(Amount, self).deserialize(value, *args, **kwargs)
```

A similar use case is having a `Nested` field supply the default values for the nested schema's fields when the parent field is `None`/missing. We may want to explore improving the default behavior in a future version, but for now providing a pattern for this in the docs would be helpful.

```py
class Options(fields.Nested):
    def deserialize(self, value, *args, **kwargs):
        value = value or {}
        return super(Options, self).deserialize(value, *args, **kwargs)
```

Related to #388

ThiefMaster:
Being able to use the `load_default` logic when receiving a `None` value would probably already cover many usecases...

lafrech:
> Being able to use the `load_default` logic when receiving a `None` value would probably already cover many usecases...

I believe this is the point of https://github.com/marshmallow-code/marshmallow/pull/1381 (from https://github.com/marshmallow-code/marshmallow/issues/713#issuecomment-529154890).

[...]

PR #2017, opened 2022-07-20
Add Enum field

lafrech:
Replaces @orenc17's #1885.

This is a long standing feature request. See for instance discussions in #267.

The status quo is that people can use @justanr's [marshmallow_enum](https://github.com/justanr/marshmallow_enum).

Currently, this lib doesn't seem actively maintained. I don't blame the author. A few things are outdated, deprecated stuff. The changes since MA3 have been user contributed and some are still missing.

I'm thinking if we recommend this lib as the reference choice, why not integrate it in code? This is what I've been trying to do.

I started by importing the code and updating it for PY3/MA3.

I also removed the dump/load asymmetry (dump by value, load by name) that I found awkward and revamped the error message generation.

I was still not happy with the `by_value`/`by_name` mechanism. I serialize by name, so I could live with only this, but I understand the need for serializing by value (to get an int or a nicer string). However, I like the type to be well-defined so that we can inherit deserialization from basic fields, and to help documenting the type (e.g. in apispec).

My proposal is therefore an `Enum` field serializing by name, and typed enum fields `StringEnum` and `IntegerEnum`, to serialize by value with a given type. Users would be free to create new types but I believe those cover most cases.

Transition from `marshmallow_enum` should be relatively smooth for users who don't use the dump/load asymmetry and who don't rely on exact error messages.

This still lacks a bit of doc but I'd rather get feedback before polishing.

Note we can't use `OneOf` validator because validation occurs after deserialization and wrong values are caught at deserialization already.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -59,6 +59,7 @@ __all__ = [
     "IPInterface",
     "IPv4Interface",
     "IPv6Interface",
+    "Enum",
     "Method",
     "Function",
     "Str",
@@ -1853,6 +1854,51 @@ class IPv6Interface(IPInterface):
     default_error_messages = {"invalid_ip_interface": "Not a valid IPv6 interface."}
 
     DESERIALIZATION_CLASS = ipaddress.IPv6Interface
+
+
+class Enum(Field):
+
+    default_error_messages = {
+        "invalid": "Not a valid string.",
+        "unknown": "Must be one of: {choices}",
+    }
+
+    def __init__(
+        self,
+        enum,
+        by_value=False,
+        *args,
+        **kwargs,
+    ):
+        self.enum = enum
+        self.by_value = by_value
+        self.choices = ", ".join([str(e.value if by_value else e.name) for e in enum])
+        super().__init__(*args, **kwargs)
+
+    def _serialize(self, value, attr, obj, **kwargs):
+        if value is None:
+            return None
+        return value.value if self.by_value else value.name
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        if self.by_value:
+            return self._deserialize_by_value(value, attr, data)
+        else:
+            return self._deserialize_by_name(value, attr, data)
+
+    def _deserialize_by_value(self, value, attr, data):
+        try:
+            return self.enum(value)
+        except ValueError as exc:
+            raise self.make_error("unknown", choices=self.choices) from exc
+
+    def _deserialize_by_name(self, value, attr, data):
+        if not isinstance(value, (str, bytes)):
+            raise self.make_error("invalid")
+        try:
+            return getattr(self.enum, value)
+        except AttributeError as exc:
+            raise self.make_error("unknown", choices=self.choices) from exc
 
 
 class Method(Field):
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1905,8 +1905,6 @@ class EnumValue(Field):
 
     def __init__(self, cls_or_instance: Field | type, enum: type[Enum], **kwargs):
         super().__init__(**kwargs)
-        self.enum = enum
-        self.choices = ", ".join([str(m.value) for m in enum])
         try:
             self.field = resolve_field_instance(cls_or_instance)
         except FieldInstanceResolutionError as error:
@@ -1914,6 +1912,10 @@ class EnumValue(Field):
                 "The enum field must be a subclass or instance of "
                 "marshmallow.base.FieldABC."
             ) from error
+        self.enum = enum
+        self.choices = ", ".join(
+            [str(self.field._serialize(m.value, None, None)) for m in enum]
+        )
 
     def _serialize(self, value, attr, obj, **kwargs):
         if value is None:
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1864,15 +1864,10 @@ class EnumSymbol(String):
         "unknown": "Must be one of: {choices}.",
     }
 
-    def __init__(
-        self,
-        enum,
-        *args,
-        **kwargs,
-    ):
+    def __init__(self, enum: type[Enum], **kwargs):
         self.enum = enum
         self.choices = ", ".join(enum.__members__)
-        super().__init__(*args, **kwargs)
+        super().__init__(**kwargs)
 
     def _serialize(self, value, attr, obj, **kwargs):
         if value is None:
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -59,9 +59,9 @@ __all__ = [
     "IPInterface",
     "IPv4Interface",
     "IPv6Interface",
-    "Enum",
-    "StringEnum",
-    "IntegerEnum",
+    "EnumSymbol",
+    "StringEnumValue",
+    "IntegerEnumValue",
     "Method",
     "Function",
     "Str",
@@ -1858,7 +1858,7 @@ class IPv6Interface(IPInterface):
     DESERIALIZATION_CLASS = ipaddress.IPv6Interface
 
 
-class Enum(String):
+class EnumSymbol(String):
 
     default_error_messages = {
         "unknown": "Must be one of: {choices}.",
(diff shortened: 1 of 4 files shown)


src/marshmallow/fields.py
```diff
@@ -1855,6 +1856,73 @@ class IPv6Interface(IPInterface):
...
+        *args,
+        **kwargs,
+    ):
+        self.enum = enum
+        self.choices = ", ".join([str(m.value) for m in enum])
+        super().__init__(*args, **kwargs)
+
+    def _serialize(self, value, attr, obj, **kwargs):
+        if value is None:
+            return None
```
deckar01, line 1922: Tangent: This pattern is repeated in nearly every field. Is there any reason this isn't in `Field.serialize`? The complimentary `None` handling is implemented in `Field.deserialize`. Maybe something to tack onto #2012.

src/marshmallow/fields.py
```diff
@@ -1855,6 +1856,73 @@ class IPv6Interface(IPInterface):
...
+        return value.value
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        value = super()._deserialize(value, attr, data, **kwargs)
+        try:
+            return self.enum(value)
```
deckar01, line 1928: The main difference between `Enum` and `TypedEnum` seems to be that the former expects enum symbols, while the latter expects enum values. The field names don't currently communicate that.

lafrech: Well `TypedEnum` is not meant to be used, I just didn't make it private to allow for other user-defined typed enums. So it is more about `Enum` vs. `StringEnum` and `IntegerEnum`.

But I agree the names don't really convey the meaning. I'm open to suggestions. Docstrings will help clarifying.

deckar01: I would probably avoid preferring one use case over another with the shorthand `Enum`. `EnumSymbol` and `EnumValue` would be clear regardless of what the user's intuition is.

lafrech: Why not?

So should it be

- `EnumSymbol`
- `EnumValue`
- `EnumString`
- `EnumInteger`

?

Or

- `EnumSymbol`
- `EnumValue`
- `StringEnumValue`
- `IntegerEnumValue`

?

I don't like the former. Kinda like the latter.

justanr: Latter is clearer to me.

tests/base.py
```diff
@@ -12,6 +13,19 @@
 central = pytz.timezone("US/Central")
 
 
+class GenderEnum(IntEnum):
+    male = 1
+    female = 2
+    non_binary = 3
```
justanr, line 20: It's me (:

src/marshmallow/fields.py
```diff
@@ -1855,6 +1858,73 @@ class IPv6Interface(IPInterface):
...
+        except ValueError as exc:
+            raise self.make_error("unknown", choices=self.choices) from exc
+
+
+class StringEnum(TypedEnum, String):
```
justanr: If I was doing everything all over again, I'd leave it at just the string representation of the enum and call it a day. The only value with using an underlying numeric is that when sending to some other languages they can successfully decode an unknown value into their enum type. But that just seems like asking for subtle bugs instead of being hit with "Cannot deserialize MyEnum.FooBar15" or whatever they'd spit out. I've been away from python long enough that I don't recall what Python does off hand but if I had to guess Python barfs when attempting that.

justanr:
Funnily enough, I showed up here because I got notified I'm a maintainer of a critical package, which turns out is the one I  question.

src/marshmallow/fields.py
lafrech: Rather than `str(value)`, which would expose the enum name, I'd rather use the member name (`SymbolEnum`). Offering to serialize by value allows users to provide human-readable names as enum values, for instance, or at least decouple business logic and API layer a bit.

But thanks for the feedback. Your comment confirms my intuition that the dump/load asymmetry is too much API surface for the need.

tests/base.py
lafrech: Yeah, I figured while touching this part of the code I might as well make it more inclusive.

I hope no one will complain about the hair colour choice being too restrictive.

src/marshmallow/fields.py
justanr: Yeah the symbol name is what I meant, just worded it kind of poorly. The asymmetry was a pain in the ass for tests and it tripped me up a few times. Pretending that they're just a union of string literals is the way to go imo

lafrech:
Renamed  Enum -> EnumSymbol, TypedEnum -> EnumValue. I'll squash later on.

Docstrings still missing.

src/marshmallow/fields.py
deckar01: What about `EnumValue(String)` and `EnumValue(Integer)`? That is consistent with `Nested`/`List` API, improves readability, and minimizes the API surface area.

lafrech: Good idea. I shall investigate this.

tests/base.py
deckar01: Hair color should be a `Flag` so that you can dye multiple colors. 😛

lafrech:
> What about `EnumValue(String)` and `EnumValue(Integer)`? That is consistent with `Nested`/`List` API, improves readability, and minimizes the API surface area.

Just tried this and it worked a treat!

I'm happy with this implementation.

Still requires docstrings and perhaps testing with a DateTime field or some other field for which ser/deser is less trivial to assert ser/deser actually happens.

lafrech:
I think I'm done.

@sloria @deckar01 @justanr I'll wait for a few days in case someone wants to review this and then I'll merge.

deckar01 (approved):

justanr (approved):

src/marshmallow/fields.py
```diff
@@ -1855,6 +1858,76 @@ class IPv6Interface(IPInterface):
...
+    """
+
+    default_error_messages = {
+        "unknown": "Must be one of: {choices}.",
+    }
+
+    def __init__(self, cls_or_instance: Field | type, enum: type[Enum], **kwargs):
+        super().__init__(**kwargs)
+        self.enum = enum
+        self.choices = ", ".join([str(m.value) for m in enum])
```
lafrech: This is wrong, we need to use the field to dump the value. `str` only work for trivial cases.

lafrech (commented):
As PR author, I can't request changes.

Please don't merge until I fix.

src/marshmallow/fields.py
lafrech: I just pushed a change to let the field serialize the values.

Our API kinda sucks because serialize expects `attr` and `obj` parameters and we don't have those and actually I don't think they are ever used. There might be a nice opportunity for a refactor here. I opened #2039.

lafrech:
PR fixed. I'm not 100% happy with `self.field._serialize(m.value, None, None)` but that's our API.

src/marshmallow/fields.py
lafrech:
```py
self.field._serialize(m.value, None, None)
```

This won't work for fields that need `attr` or `obj` (but is there any ? #2039) or fields that are not symmetric (don't dump what they loaded). I'm not talking about fields that don't round-trip, I mean if several equivalent representation are accepted at load time and only one is used at dump time, this one is the one that will be displayed here and that's fine, I believe. It would be more of a problem for custom field that would purposely have asymmetric load/dump methods, like using different time formats, for instance.

Since those must be quite rare, and using them in this enum field would be even more rare, I'm fine with this limitation.

We could remove the list of accepted values from the error message but I think it is useful in the "normal" case (99% of the time) and it will be needed for documentation purpose anyway (apispec).

lafrech:
```py
self.field._serialize(m.value, None, None)
```

 is already used in `Mapping`.

lafrech:
Alright, moving on with this.

lafrech:
Just after merging this, as I was updating the changelog to release, I realized the naming didn't follow our habits of naming fields by object type.

I pushed a new PR to propose merge of both fields into an `Enum` field: #2044.

ThiefMaster:
Would you consider adding a short migration guide for `marshmallow_enum` users? :)

PR #2018, opened 2022-07-25
Add return type to `fields.Email.__init__` so that it's not untyped

kkirsche:
This merge request adds a type hint to `fields.Email`'s `__init__ method to avoid no untyped call errors in mypy for library consumers.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1731,7 +1731,7 @@ class Email(String):
     #: Default error messages.
     default_error_messages = {"invalid": "Not a valid email address."}
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         # Insert validation into self.validators so that multiple errors can be stored.
         validator = validate.Email(error=self.error_messages["invalid"])


lafrech:
Is this specific to `Email` field?

kkirsche:
This was the only `field` from `marshmallow` in my codebase that was triggering this mypy behavior. I'm not sure if any others do though

lafrech:
From those discussions

- https://github.com/python/mypy/issues/3358
- https://github.com/python/mypy/issues/604
- https://github.com/python/mypy/pull/5677

Explicit `-> None` is needed if init func has no argument. I guess `*args` and `**kwargs` don't count.

Form a quick check, `Email` is the only concerned field. Although `IPInterface.__init__` lacks typing (fixed in #2036).

lafrech (approved):

kkirsche:
Thank you :) I can look at expanding types if you are interested in a future PR, I just wasn't sure what the responsiveness of the team was, so I was rather hesitant to expand the scope too far.

lafrech:
Well, don't expect us to be quick to review and merge (see current backlog), but quality PRs are absolutely welcome.

(And if you're good at type hinting, I wouldn't mind a helping hand for #1896.)

Thanks!

kkirsche:
Course, I'll take a look when I have a few moments. I've been trying to work a lot more lately with the typeshed team to get feedback from the source about how to use type hints correctly and effectively, so I'm happy to see if I may be able to assist with that. I'll need to pull it down and run mypy against it as currently it looks like the build log was deleted, so can't view the historic error from mypy

PR #2047, opened 2022-09-11
Add SECURITY.md

sloria:
Per https://github.com/ossf/scorecard/blob/main/docs/checks.md#security-policy, places security contact information in a SECURITY.md file.


lafrech (approved):
Good.

The only part that is not addressed is 

> The file should contain information on what constitutes a vulnerability

We can merge this anyway.

Up till now, that hasn't been much debate about it (also because there hasn't been many security issues).

PR #2062, opened 2022-10-16
RFC alternative key in serialized data

vito-c:
The problem:
`data_key` provides a way to map keys in a symmetric manner I often have to use two schemas to accomplish asymmetric mapping of keys so that the output matches my data classes.

Example:
```json
{ "myFoo": "bar" }
```
```python
class MySchema(Schema):
    foo = String(data_key="myFoo")

class SchemaTwo(Schema):
    foo = String()

symmetric = MySchema().dump(MySchema.load({"myFoo":"bar"}))
# >> {"myFoo":"bar"}
asymmetric = SchemaTwo().dump(MySchema.load({"myFoo":"bar"}))
# >> {"foo":"bar"}
```

Proposal:
We should add a new attribute in the fields class that supports an alternative key to the schema key name. This would allow schemas to load data either from the schema the field_objects name or from an alternate key in the data.

Example:
```python
class SingleSchema(Schema):
    foo = String(alternative="myFoo")

asymmetric = SingleSchema().dump(SingleSchema.load({"myFoo":"bar"}))
# >> {"foo":"bar"}
```

Reference From other frameworks:
- https://www.javadoc.io/doc/com.google.code.gson/gson/2.8.0/com/google/gson/annotations/SerializedName.html (see alternatives)

Benefits:
- One schema is less than two :)
- Less questions about 

Status:
I have done some work here to implement this feature. I have a bunch more tests to write and I wanted to submit this RFC before spending more time writing tests.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -138,6 +138,7 @@ class Field(FieldABC):
     #  for those fields
     _CHECK_ATTRIBUTE = True
     _creation_index = 0  # Used for sorting
+    _deserialized_key: str | None = None
 
     #: Default error messages for various kinds of errors. The keys in this dictionary
     #: are passed to `Field.make_error`. The values are error messages passed to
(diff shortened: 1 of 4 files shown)


lafrech:
Hi. Sorry about the response delay.

For the rationale about `data_key`, see #717. I'm not sure we want to bring back asymmetry here.

Issue #1962, opened 2022-03-21
Support union type syntax as defined in PEP 604 and implemented in Python 3.10


zyv:
A new syntax for union types has been introduced in [PEP 604](https://peps.python.org/pep-0604/) and implemented in Python 3.10:

```python
dict[str, Union[str, bool]] == dict[str, str | bool]
```

Unfortunately, it is not currently supported in Marshmallow:

> UserWarning: Unknown type str | bool at XXX: list[dict[str, str | bool]] It's advised to pass the correct marshmallow type to `YYY`.

Because of that, we can't ban `Union` in our linter.

sirosen:
> Unfortunately, it is not currently supported in Marshmallow
> ...
> Because of that, we can't ban `Union` in our linter.

Could you clarify what you mean by this? I don't think anything that marshmallow does should impact annotations you have in your own code.

---

Looking to see if we can remove marshmallow's own explicit `Union`s, I see them being used in type aliases in `marshmallow.types`. That means that without runtime support for unions -- meaning the removal of support for python3.9 and below -- we can't just substitute the nicer `|` syntax.

Today we have:
```python
# in types.py
StrSequenceOrSet = typing.Union[typing.Sequence[str], typing.Set[str]]
Tag = typing.Union[str, typing.Tuple[str, bool]]
```

We can pull in `typing_extensions` and do this
```python
from typing_extensions import TypeAlias

StrSequenceOrSet: TypeAlias = "typing.Sequence[str] | set[str]"
Tag: TypeAlias = "str | tuple[str, bool]"
```

but I don't feel great about requiring `typing_extensions;python_version<3.10` just for this purpose.
And I'm not sure how a stringified type alias is treated with and without the explicit TypeAlias annotation.

zyv:
> Could you clarify what you mean by this? I don't think anything that marshmallow does should impact annotations you have in your own code.

Unfortunately, yes, it does: `dataclasses_json` uses `marshmallow` under the hood - type annotations of data classes are passed to `marshmallow` to perform validation. However, if `marshmalllow` cannot work with those type annotations it issues the warning like posted above.

This means that in our code for dataclasses, we are forced to use type annotations, which `marshmallow` can understand. So we can't simply ban `from typing import Union` in our linter, because we must still use `Union` on dataclasses, which will go through `marshmallow`.

Hope that clears it up!

sirosen:
[...]

PR #2071, opened 2022-11-15
Use PEP 604 where possible

lafrech:
I couldn't remove `Union` from types.py (#1962) so this is the only place where I could remove it.


lafrech:
I also reworked a few union types. It seems more logical to write `(None | A | B)` than `None | (A | B)`.

PR #2078, opened 2022-12-08
Use fromisoformat from standard library (or backport)

lafrech:
This PR shows what it would look like if we used Python 3.11's improved `fromisoformat` (backported for older releases).

Less code on our side and it should be more efficient (C implementation).

(We could simplify even more by passing those functions directly to the class and remove everything from utils.py.)

There are 2 failing tests.

- "2018-01-01" is accepted as a datetime while it is currently rejected in marshmallow. I guess this would be a breaking change. We may keep the old behaviour with a temporary trick until marshmallow 4 (keep the regex or even just check input string length).

- Time deserialization respects timezones. It is not clear to me what a time with timezone but without date means. It could be ambiguous for timezones with DST. In any case, this can be considered either bugfix or breaking change, and it could be "fixed" to keep old behaviour.

-------------------------------------

Consider this as an RFC.

Meanwhile, anyone willing to do it in their app may set those functions as class attributes and use the backport if needed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,14 @@ Changelog
 4.0.0 (unreleased)
 ******************
 
-- `marshmallow.fields.TimeDelta` no longer truncates float values when
+- *Backwards-incompatible*: Use `datetime.date.fromisoformat`, `datetime.time.fromisoformat`, and `datetime.datetime.fromisoformat` from the standard library to deserialize dates, times and datetimes (:pr:`2078`). 
+
+As a consequence of this change:
+  - Time with time offsets are now supported.
+  - YYYY-MM-DD is now accepted as a datetime and deserialized as naive 00:00 AM.
+  - `from_iso_date`, `from_iso_time` and `from_iso_datetime` are removed from `marshmallow.utils`
+
+- *Backwards-incompatible*: `marshmallow.fields.TimeDelta` no longer truncates float values when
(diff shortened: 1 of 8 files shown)


lafrech:
We may want to keep things as is in marshmallow 3 but I think we should be using standard lib `fromisoformat` in marshmallow 4 and this is at least a preview of the minor changes that involves.

lafrech:
I'm leaning towards using this and accepting the slightly breaking changes.

The docs mention limitations. We could check wether our current implementation also has those limitations.

https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat

ddelange:
nice cleanup! 4.0?

sloria:
Lgtm for 4.0. I think we can also get rid of the `utils.py` functions, so long as there's a clear migration path documented in the changelog

lafrech:
Changelog suggestion:


- *Backwards-incompatible*: Use ``fromisoformat`` from standard library to
  deserialize dates, times and datetimes.
  Time with time offset is now supported.
  `YYYY-MM-DD` is now accepted as a datetime and deserialized as naive 00:00 AM.
  Remove ``from_iso_date``, ``from_iso_time`` and ``from_iso_datetime`` from utils.py.
  (:pr:`2078`)

lafrech:
I just moved a few issues from 4.0 milestone to 5.0.

We may postpone more stuff if there's still too much left.

sloria (approved):

ddelange (approved):
nice

PR #2098, opened 2023-02-08
Move from Azure to GHA

lafrech:
This is something I've been willing to do for a while.

Now would be a good time. We have issues in Azure due to ubuntu-18.04 brownout. Fixing this would require an update to @sloria's Azure template (see https://github.com/sloria/azure-pipeline-templates/issues/5).

Switching to GHA seems like an easier move.


lafrech:
@sloria, if we go this route, we need to

- remove Azure hook (not sure I can do that)
- provide PyPI API_KEY

Could you please do that?

Thanks.

sloria (approved):
thanks! i've added the `PYPI_API_KEY` secret and removed the AP branch protection. I'll go ahead and merge this and add the GHA branch protection.

lafrech:
Thanks. Not sure what you mean by GHA branch protection.

Did you also change Azure config to remove the hook? I'd need the same in marshmallow-sqlalchemy. I don't think I can do it, can I?

sloria:
There was a branch protection rule that prevented merging until the AP builds were passing. I changed that to the GHA builds

sloria:
I think you have the privs to do marshmallow-sqlachemy now. You might need to accept the email invite for pypi

lafrech:
I do, thanks.

I thought there might be something to do in Azure itself to disable the CI hook.

PR #2102, opened 2023-02-15
fix: unhandled OSError and OverflowErrors from utils.from_timestamp

TheBigRoomXXL:
When running [schemathesis](https://github.com/schemathesis/schemathesis) against my smorest app I found some unexpected 500 errors.  They where caused by  `utils.from_timestamp` which doesn't  handle the [OverflowError or OSError that can be raised by  datetime.fromtimestamp](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp)

What's has been done:

- OverflowError and OSError are caught and re-throwed as ValueError so that they can be handle properly.
- Tests for `utiles.from_timestamp` were added to avoid regression

### DIFF
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -197,7 +197,12 @@ def from_timestamp(value: typing.Any) -> dt.datetime:
 
     # Load a timestamp with utc as timezone to prevent using system timezone.
     # Then set timezone to None, to let the Field handle adding timezone info.
-    return dt.datetime.fromtimestamp(value, tz=dt.timezone.utc).replace(tzinfo=None)
+    try:
+        return dt.datetime.fromtimestamp(value, tz=dt.timezone.utc).replace(tzinfo=None)
+    except OverflowError as exc:
+        raise ValueError("Timestamp is too large") from exc
+    except OSError as exc:
+        raise ValueError("Error converting value to datetime") from exc
 
 
 def from_timestamp_ms(value: typing.Any) -> dt.datetime:
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -4,6 +4,8 @@ import ipaddress
 import decimal
 import math
 
+from unittest.mock import patch
+
 import pytest
 
 from marshmallow import EXCLUDE, INCLUDE, RAISE, fields, Schema, validate
@@ -19,6 +21,20 @@ from tests.base import (
     HairColorEnum,
     DateEnum,
 )
+
+
+class MockDateTimeOverflowError(dt.datetime):
+    """Used to simulate the possible OverflowError of datetime.fromtimestamp"""
+
+    def fromtimestamp(self, *args, **kwargs):
+        raise OverflowError()
+
+
+class MockDateTimeOSError(dt.datetime):
+    """Used to simulate the possible OSError of datetime.fromtimestamp"""
+
+    def fromtimestamp(self, *args, **kwargs):
+        raise OSError()
 
 
 class TestDeserializingNone:
@@ -565,9 +581,20 @@ class TestFieldDeserialization:
         ["", "!@#", 0, -1, dt.datetime(2013, 11, 10, 1, 23, 45)],
     )
     def test_invalid_timestamp_field_deserialization(self, fmt, in_value):
-        field = fields.DateTime(format="timestamp")
+        field = fields.DateTime(format=fmt)
         with pytest.raises(ValidationError, match="Not a valid datetime."):
             field.deserialize(in_value)
+
+    #  Regression test for https://github.com/marshmallow-code/marshmallow/pull/2102
+    @pytest.mark.parametrize("fmt", ["timestamp", "timestamp_ms"])
+    @pytest.mark.parametrize(
+        "mock_fromtimestamp", [MockDateTimeOSError, MockDateTimeOverflowError]
+    )
+    def test_oversized_timestamp_field_deserialization(self, fmt, mock_fromtimestamp):
+        with patch("datetime.datetime", mock_fromtimestamp):
+            field = fields.DateTime(format=fmt)
+            with pytest.raises(ValidationError, match="Not a valid datetime."):
+                field.deserialize(99999999999999999)
 
     @pytest.mark.parametrize(
         ("fmt", "timezone", "value", "expected"),
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -172,3 +172,4 @@ Contributors (chronological)
 - Marco Satti `@marcosatti  <https://github.com/marcosatti>`_
 - Ivo Reumkens `@vanHoi <https://github.com/vanHoi>`_
 - Aditya Tewary `@aditkumar72 <https://github.com/aditkumar72>`_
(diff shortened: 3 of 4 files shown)


TheBigRoomXXL:
One thing to note is that a large timestamp will sometime raise ValueError 'year XXXXX is out of range' instead of OSError, that depends on the machine or OS used.

lafrech:
Thanks.

It would be nice to test both exceptions, if this is feasible (not sure how to reproduce "localtime() or gmtime() failure").

Also, while it is nice to keep the unit test close to the util function, what we really want to assert is that the field generates a `ValidationError` rather than a crash. Maybe this ought to be tested too.

I had never heard about schemathesis. I shall try this out.

TheBigRoomXXL:
> It would be nice to test both exceptions, if this is feasible (not sure how to reproduce "localtime() or gmtime() failure").

I could not find information on how to reproduce. I don't think it's possible without running the CI on multiple differents systems.

> Also, while it is nice to keep the unit test close to the util function, what we really want to assert is that the field generates a ValidationError rather than a crash. Maybe this ought to be tested too.

I added a test at the field level in `test_deserialization.py`. Also, I realized while doing it that the test above was parametrized with a parameter `ftm` which was not used because `format="timestamp"` was used instead of `format=fmt` so I fixed that.

lafrech:
Good!

> I could not find information on how to reproduce. I don't think it's possible without running the CI on multiple differents systems.

What you could do is mock `fromtimestamp` to force `OSError`, if only to have a non-reg test in case we'd be tempted someday to remove the `OSError` catch.

TheBigRoomXXL:
I don't have much experience with mocking in Python. I couldn't mock` datetime.datetime.fromtimestamp` directly because python built-in types are immutable so i mocked the entire `datetime` class instead.

src/marshmallow/utils.py
```diff
@@ -197,7 +197,10 @@ def from_timestamp(value: typing.Any) -> dt.datetime:
 
     # Load a timestamp with utc as timezone to prevent using system timezone.
     # Then set timezone to None, to let the Field handle adding timezone info.
-    return dt.datetime.fromtimestamp(value, tz=dt.timezone.utc).replace(tzinfo=None)
+    try:
+        return dt.datetime.fromtimestamp(value, tz=dt.timezone.utc).replace(tzinfo=None)
+    except (OSError, OverflowError) as exc:
+        raise ValueError("Timestamp is too large") from exc
```
lafrech, line 203: The message is fine for overflow but probably wrong for OSError, right?

> [fromtimestamp()](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp) may raise [OverflowError](https://docs.python.org/3/library/exceptions.html#OverflowError), if the timestamp is out of the range of values supported by the platform C localtime() or gmtime() functions, and [OSError](https://docs.python.org/3/library/exceptions.html#OSError) on localtime() or gmtime() failure.

We should use a more generic message to include `OSError`. Or use two different messages, but the message doesn't matter that much since it is not used to build the `ValidationError`.

tests/test_deserialization.py
```diff
@@ -565,10 +581,27 @@ def test_timestamp_field_deserialization(self, fmt, value, expected):
...
     def test_invalid_timestamp_field_deserialization(self, fmt, in_value):
-        field = fields.DateTime(format="timestamp")
+        field = fields.DateTime(format=fmt)
         with pytest.raises(ValidationError, match="Not a valid datetime."):
             field.deserialize(in_value)
 
+    @pytest.mark.parametrize("fmt", ["timestamp", "timestamp_ms"])
+    def test_oversized_timestamp_field_deserialization(self, fmt):
+        field = fields.DateTime(format=fmt)
+        with pytest.raises(ValidationError, match="Not a valid datetime."):
```
lafrech: I just realized this error message is not that great for timestamps. But that's another story.

lafrech (changes requested):
Great. We're close. Just minor comments.

Do you think you could rebase your branch for a nicer history? Like one commit for AUTHORS file and one for the code? Otherwise, no pb.

src/marshmallow/utils.py
TheBigRoomXXL: I changed it to two differents messages.

TheBigRoomXXL:
OK! It should be the last change to the merge request. Thank you for your time and all the feedbacks!

tests/test_deserialization.py
TheBigRoomXXL: Do you mean it should return `"Not a valid timestamp."` when failing to deserialize a timestamp ?

lafrech (approved):
The informative error messages from utils.py don't make it to the final validation error message but this would require a wider change and the current state of this PR is better than nothing.

Sorry about the delay. Do you think you could rebase?

Thanks.

TheBigRoomXXL:
Rebase done. Don't worry for the delay. 

Thank you.

tests/test_deserialization.py
```diff
@@ -565,10 +581,27 @@ def test_timestamp_field_deserialization(self, fmt, value, expected):
...
+            field.deserialize(999999999999999999999999)
+
+    #  Regression test for https://github.com/marshmallow-code/marshmallow/pull/2102
+    @pytest.mark.parametrize("fmt", ["timestamp", "timestamp_ms"])
+    @pytest.mark.parametrize(
+        "mock_fromtimestamp", [MockDateTimeOSError, MockDateTimeOverflowError]
```
lafrech, line 591: Do we need `MockDateTimeOverflowError`? Isn't this case covered by `test_oversized_timestamp_field_deserialization` above?

src/marshmallow/utils.py
```diff
@@ -197,7 +197,12 @@ def from_timestamp(value: typing.Any) -> dt.datetime:
 
     # Load a timestamp with utc as timezone to prevent using system timezone.
     # Then set timezone to None, to let the Field handle adding timezone info.
-    return dt.datetime.fromtimestamp(value, tz=dt.timezone.utc).replace(tzinfo=None)
+    try:
+        return dt.datetime.fromtimestamp(value, tz=dt.timezone.utc).replace(tzinfo=None)
+    except OverflowError as exc:
+        raise ValueError("Timestamp is too large") from exc
+    except OSError as exc:
+        raise ValueError("Failed to convert this value to datetime") from exc
```
lafrech: The following sentence might be slightly better and more in line with other errors:

> Error converting value to datetime

Just my feeling. I'm not a native English speaker.

lafrech (commented):
Thanks. Just a comment about a test that seems redundant.

tests/test_deserialization.py
TheBigRoomXXL: `test_oversized_timestamp_field_deserialization` will most probably raise OverflowError internally, but they is not guarantee, it could also be OSError depending on the system.  

But indeed it is redundant as it was written before the mock. I decided to delete the first test and keep the two mocks as it make each error case explicit.

src/marshmallow/utils.py
TheBigRoomXXL: Ok, it's updated.

PR #2111, opened 2023-03-17
Minor DateTime field logic refactor

lafrech:
I also removed a few unneeded else after return. Pylint complains about this. I agree this is arguable and some may think the else makes it more readable.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -920,8 +920,7 @@ class UUID(String):
         try:
             if isinstance(value, bytes) and len(value) == 16:
                 return uuid.UUID(bytes=value)
-            else:
-                return uuid.UUID(value)
+            return uuid.UUID(value)
         except (ValueError, AttributeError, TypeError) as error:
             raise self.make_error("invalid_uuid") from error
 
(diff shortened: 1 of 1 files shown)


src/marshmallow/fields.py
```diff
@@ -1281,28 +1281,23 @@ def _serialize(self, value, attr, obj, **kwargs) -> str | float | None:
...
-                ) from error
-        else:
-            try:
-                return self._make_object_from_format(value, data_format)
-            except (TypeError, AttributeError, ValueError) as error:
-                raise self.make_error(
-                    "invalid", input=value, obj_type=self.OBJ_TYPE
-                ) from error
+        if not func:
+            func = functools.partial(
```
sloria: (non-blocking) is there any performance hit from creating this partial?

sloria (approved):

src/marshmallow/fields.py
lafrech: Good point. Didn't check.

We can just revert and move the try/catch out of the `if func` block. This is a totally acceptable scope. It might even be more readable.

PR #2132, opened 2023-05-26
Add OpenSSF scorecard

lafrech:
Trying to see what it would do if we added an [OpenSFF Scorecard](https://securityscorecards.dev/).

This GHA file was automatically generated. I just added the "pull request" condition. Not sure we should keep it but we needed it at least to test in this PR.

We should get stuff in the security -> code-scanning menu.

I also tried this in [flask-smorest](https://github.com/marshmallow-code/flask-smorest/pull/511). Got a few errors, mostly false-positives, IMHO, so I'm not convinced it is worth it but I'm open to discussion about it.

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,73 @@
+# This workflow uses actions that are not certified by GitHub. They are provided
+# by a third-party and are governed by separate terms of service, privacy
+# policy, and support documentation.
+
+name: Scorecard supply-chain security
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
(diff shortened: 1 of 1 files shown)


lafrech:
We get the same kind of warnings I got in flask-smorest: https://github.com/marshmallow-code/marshmallow/security/code-scanning?query=is%3Aopen+pr%3A2132.

This is not really a surprise since the repo configs are similar.

Those don't seem so useful to me.

Also when I checked the other day I didn't find a nice dashboard to display. The docs point to a place where to get a badge, but the average mark is not that useful without the details. Apparently, it can't point to the list of issues above, as it is private.

lafrech:
Closing as not useful.

PR #2135, opened 2023-06-09
Drop Python 3.7

lafrech:
EOL on June 27th: https://endoflife.date/python

Also adds py311 in target_version in pyproject.toml, forgotten in earlier PR.


sdd:
Hi @lafrech! Thanks for all your work on Marshmallow over the years, I've found it very useful! 🥇 

Do you have an estimate for when 3.20 will be released? I'm looking to upgrade a project that uses marshmallow to run on 3.11.

Thanks again 😎

lafrech:
Hi. Thanks for the kind words.

No idea, really, and no rush. Note that the only thing introduced in 3.20 is the fact that it is tested against 3.11, but the code in 3.19 works on 3.11 since it is the one tested in main branch, so you're safe using it on 3.11 already.

There won't be changes in 3.19 anymore, like 3.19.x, that could break anything and wouldn't be covered by tests.

sdd:
Aah great, that's good to know - we'll see how we get on with v3.19 then. Thanks for the quick reply, much appreciated! 😁

lafrech:
I had a quick look, there are in fact a few changes, mostly typing related. You may want to diff or just run tests on 3.11 locally to be sure.

PR #2141, opened 2023-06-27
Use Abstract Base Classes to define FieldABC and SchemaABC

### LINKED ISSUE #1449: Use Abstract Base Classes to define FieldABC and SchemaABC
https://docs.python.org/3/library/abc.html

https://stackoverflow.com/questions/44315961/when-to-use-raise-notimplementederror/44316506#44316506

aditkumar72:
This PR fixes #1449 to refactor base.py to use Abstract Base Classes. It also fixes meta class conflict by using ABCMeta as a super class for SchemaMeta.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -171,3 +171,4 @@ Contributors (chronological)
 - Karthikeyan Singaravelan `@tirkarthi  <https://github.com/tirkarthi>`_
 - Marco Satti `@marcosatti  <https://github.com/marcosatti>`_
 - Ivo Reumkens `@vanHoi <https://github.com/vanHoi>`_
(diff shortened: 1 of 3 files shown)


lafrech (approved):
Looks good.

Thanks!

aditkumar72:
@lafrech since all the check has passed, Could we merge this PR ?

lafrech:
I'd be happy to get the feedback of another maintainer.

lafrech:
I gave this another look. The abstract classes are private API, not meant to be used directly, as stated in file docstring, so on one hand, enforcing abstractness shouldn't be needed, on the other hand, it shouldn't break anyone's workflow, so let's merge this.

PR #2153, opened 2023-07-20
Fix call to get_declared_fields: pass dict_cls

### LINKED ISSUE #2152: 3.20.0 seems to break marshmallow_sqlalchemy
As you can see in the title, my CI just broke:
```
/env/lib/python3.11/site-packages/marshmallow_sqlalchemy/schema.py:143: in <module>
    class SQLAlchemySchema(
/env/lib/python3.11/site-packages/marshmallow/schema.py:116: in __new__
    klass._declared_fields = mcs.get_declared_fields(
E   TypeError: SQLAlchemySchemaMeta.get_declared_fields() missing 1 required positional argument: 'dict_cls'
```
Possibly an issue on their end more than yours, but someone had to report it somewhere. :-)

lafrech:
Fixes #2152.

Now that `dict`s are ordered, there is no point using an `OrderedDict` to store fields, so while working on #1896, I simplified the logic to always use a `dict`.

This screws up classes overriding `get_declared_fields` that still expect a dict class to be passed (#2152).

Step back and pass dict class again. (Keep default value in `get_declared_fields` implementation.)

### DIFF
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -117,6 +117,7 @@ class SchemaMeta(ABCMeta):
             klass=klass,
             cls_fields=cls_fields,
             inherited_fields=inherited_fields,
+            dict_cls=dict,
         )
         return klass
 


Cheaterman:
Sorry for Hyrum's law :grin: thanks for fixing this!

PR #2164, opened 2023-08-11
Fix NestedField Type hint for Lambda Schema types

somethingnew2-0:
Providing a lambda with a returning callable type is handled by Marshmallow (such as below), but not allowed by the type-hinting. Fix that

```
class PrimarySchema(Schema):
    other = fields.Nested(lambda: OtherSchema)
```

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -173,3 +173,4 @@ Contributors (chronological)
 - Ivo Reumkens `@vanHoi <https://github.com/vanHoi>`_
 - Aditya Tewary `@aditkumar72 <https://github.com/aditkumar72>`_
 - Sebastien Lovergne `@TheBigRoomXXL <https://github.com/TheBigRoomXXL>`_
+- Peter C `@somethingnew2-0 <https://github.com/somethingnew2-0>`_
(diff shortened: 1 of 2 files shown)


lafrech:
Thanks.

Would you like to add yourself to AUTHORS file?

I'm wondering if we could be more specific than `type`. Could we enforce that it is just not any `type` but only schema types? Like, by specifying `SchemaMeta` rather than `type`? This question applies to the whole codebase, so we may merge anyway, but I'd be interested to gather thoughts about this.

somethingnew2-0:
@lafrech added myself to AUTHORS

`SchemaMeta` does appear to work in my codebase to be more specific than `type`.  But I agree that it should probably be updated everywhere in marshmallow where `type` was used such as on line 536 of the same method https://github.com/marshmallow-code/marshmallow/pull/2164/files#diff-0cddc4aac0e4e16946830a5807c264d65b7f14a2a1b1cb23b03028a36df3681aR536

lafrech:
Yes, exactly. Thanks for the feedback. This can be done in another (welcome) PR.

Issue #2145, opened 2023-07-12
Clarify behavior of `attribute` for load-only fields


Defman21:
In the docs, it says that:

>    :param attribute: The name of the attribute to get the value from when **serializing**.

However, it also affects the **deserialization** part:

https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/schema.py#L681

e.g.

```python
from marshmallow import Schema, fields

class SomeSchema(Schema):
    field_dump = fields.Str(dump_only=True, attribute="some_name", data_key="field")
    field_load = fields.Str(load_only=True, attribute="field", data_key="some_other_name")

schema = SomeSchema()
schema.load({"some_other_name": "some field"}) # -> {"field": "some field"}
schema.dump({"some_name": "some value"}) # -> {"field": "some value"}
```

I've used `data_key` with different names to show distinction between `attribute` and `data_key` usage. My use case is:
* I want to have `{"some_field: "http://test"}` as the result of `Schema.dump(SomeThing(some_field_url="http://test"))` 
* I want to have `{"some_field": <BytesIO>}` as the result of `Schema.load({"some_field": "base64 representation of something"})`.

lafrech:
You're right about the documentation issue. I think this was discussed a while back and I'm surprised it wasn't fixed already.

I'm afraid you're out of luck here because you'd need to set `data_key` to the same value for both fields, which is not allowed (to protect users from their own mistakes).

Your use case is a corner case that might not be worth removing that protection. It can be achieved by using two schemas, obviously. I don't see any other mean.

Defman21:
>I'm afraid you're out of luck here because you'd need to set data_key to the same value for both fields, which is not allowed (to protect users from their own mistakes).

Perhaps I'm missing something?

<img width="688" alt="image" src="https://github.com/marshmallow-code/marshmallow/assets/7100645/4072cbc0-c3b1-4693-8dff-0f5bb8990f69">

```
{'field': 'some field', 'id': 1}
{'field': 'some field', 'id': 2}
```

lafrech:
Oh, right. That's because we only check data_keys for fields that are not load-only. The rationale is that one may want to load the same data with multiple fields, while it would be wrong to dump multiple attributes to the same key.

Well, good for you. I'm labeling this a doc issue, then.

lafrech:
I just pushed a commit in `dev` with a phrasing that should fix this.

```py
[...]

PR #2166, opened 2023-08-13
Update field docstring to reflect "attribute" argument affecting deserialization

MikeMazalahi:
I believe this simple change should resolve the documentation ambiguity in #2145. However, I'm a first-time contributor so please let me know if an explanation here would be better.

Also, should I perhaps open a PR following this to reflect the data_key condition?

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -87,7 +87,7 @@ class Field(FieldABC):
     :param data_key: The name of the dict key in the external representation, i.e.
         the input of `load` and the output of `dump`.
         If `None`, the key will match the name of the field.
-    :param attribute: The name of the attribute to get the value from when serializing.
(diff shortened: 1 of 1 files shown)


lafrech:
Thanks for contributing.

I just pushed a commit in `dev` with a phrasing that I find more explicit.

```py
    :param data_key: The name of the dict key in the external representation, i.e.
        the input of `load` and the output of `dump`.
        If `None`, the key will match the name of the field.
    :param attribute: The name of the key/attribute in the internal representation, i.e.
        the output of `load` and the input of `dump`.
        If `None`, the key/attribute will match the name of the field.
        Note: This should only be used for very specific use cases such as
        outputting multiple fields for a single attribute, or using keys/attributes
        that are invalid variable names, unsuitable for field names. In most cases,
        you should use ``data_key`` instead.
```

PR #2181, opened 2023-09-10
Typing: replace "type" with specific metaclass

lafrech:
See #2164.

Feedback welcome.

### DIFF
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -502,9 +502,7 @@ class Schema(base.SchemaABC, metaclass=SchemaMeta):
             return error.valid_data or missing
         return value
 
-    def _serialize(
-        self, obj: typing.Any | typing.Iterable[dict], *, many: bool = False
-    ):
+    def _serialize(self, obj: typing.Any, *, many: bool = False):
         """Serialize ``obj``.
 
         :param obj: The object(s) to serialize.
@@ -588,7 +586,7 @@ class Schema(base.SchemaABC, metaclass=SchemaMeta):
         partial=None,
         unknown=RAISE,
         index=None,
-    ) -> dict | list[dict]:
+    ) -> typing.Any | list[typing.Any]:
         """Deserialize ``data``.
 
         :param dict data: The data to deserialize.
@@ -602,26 +600,24 @@ class Schema(base.SchemaABC, metaclass=SchemaMeta):
             fields in the data. Use `EXCLUDE`, `INCLUDE` or `RAISE`.
         :param int index: Index of the item being serialized (for storing errors) if
             serializing a collection, otherwise `None`.
-        :return: A dictionary of the deserialized data.
+        :return: The deserialized data as `dict_class` instance or list of `dict_class`
+        instances if `many` is `True`.
         """
         index_errors = self.opts.index_errors
         index = index if index_errors else None
         if many:
             if not is_collection(data):
                 error_store.store_error([self.error_messages["type"]], index=index)
-                ret_l = []  # type: typing.List[dict]
+                ret_l = []
             else:
                 ret_l = [
-                    typing.cast(
-                        dict,
-                        self._deserialize(
-                            typing.cast(dict, d),
-                            error_store=error_store,
-                            many=False,
-                            partial=partial,
-                            unknown=unknown,
-                            index=idx,
-                        ),
+                    self._deserialize(
+                        typing.cast(dict, d),
+                        error_store=error_store,
+                        many=False,
+                        partial=partial,
+                        unknown=unknown,
+                        index=idx,
                     )
                     for idx, d in enumerate(data)
                 ]
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -40,8 +40,6 @@ from marshmallow.utils import (
 )
 from marshmallow.warnings import RemovedInMarshmallow4Warning
 
-_T = typing.TypeVar("_T")
-
 
 def _get_fields(attrs):
     """Get fields from a class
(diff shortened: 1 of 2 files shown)


deckar01:
It looks like there are two last anonymous `type` annotations that could be fixed. https://github.com/marshmallow-code/marshmallow/commit/f8f4c8f26f4dc8b986ca58a08a985cdfa4391d0a

Changing the return type of `dict_class()` to `type[Mapping]` reveals that `Mapping` doesn't provide some of the `dict` interface we are using.

The `_T` `TypeVar` is hiding some of that incompatibility from the type checker also. `serialize` doesn't actually provide a generic pass through of the input type. It takes any `dict` type as input, and regardless of that type it returns an instance of the configured `dict_class()`.

deckar01 (approved):

lafrech:
I applied @deckar01's commit and rebased.

Guys, feel free to merge.

src/marshmallow/schema.py
```diff
@@ -497,7 +498,7 @@ def _call_and_store(getter_func, data, *, field_name, error_store, index=None):
             return error.valid_data or missing
         return value
 
-    def _serialize(self, obj: _T | typing.Iterable[_T], *, many: bool = False):
+    def _serialize(self, obj: dict | typing.Iterable[dict], *, many: bool = False):
```
sloria: this isn't quite right, is it? `obj` can be any arbitrary object type or an iterable objects.

src/marshmallow/schema.py
```diff
@@ -510,7 +511,7 @@ def _serialize(self, obj: _T | typing.Iterable[_T], *, many: bool = False):
         if many and obj is not None:
             return [
                 self._serialize(d, many=False)
-                for d in typing.cast(typing.Iterable[_T], obj)
+                for d in typing.cast(typing.Iterable[dict], obj)
```
sloria: related to comment above: i think this was already correct

src/marshmallow/schema.py
```diff
@@ -584,7 +585,7 @@ def _deserialize(
         partial=None,
         unknown=RAISE,
         index=None,
-    ) -> _T | list[_T]:
+    ) -> dict | list[dict]:
```
sloria: related to above: `_deserialize` is expected to return an arbitrary object or list of objects, not dictionaries.

sloria (changes requested):
the changes in fields.py look right to me, but i'm less sure about the changes to schema.py. can you double check them?

src/marshmallow/schema.py
deckar01: `ret_d = self.dict_class() ... return ret_d` implies that the return type of `_deserialize()` is the return type of `dict_class()` (`Mapping`). That change reveals that our usage of the return value actually depends on the `dict` interface. The fact that the type checker doesn't error on key access to a `TypeVar` which is effectively `Any`, seems like a mypy bug, but maybe it just needs explicit test coverage with a concrete type.

Enveloping occurs in `_do_load()`, so the return type of `_deserialize()` should not be constrained by the `TypeVar`. Now that I understand the purpose of the `TypeVar` it looks like it should be moved to `load()` rather than removed.

src/marshmallow/schema.py
deckar01: ~~Per https://github.com/marshmallow-code/marshmallow/pull/2181/files#r1401344892, the issue was that `_T` was being used incorrectly in `_deserialize`. Moving it to `load()` would avoid the need to change this `TypeVar`.~~

src/marshmallow/schema.py
deckar01: ~~Per https://github.com/marshmallow-code/marshmallow/pull/2181/files#r1401346431, yes, this can be reverted.~~

src/marshmallow/schema.py
deckar01: When I tested my proposed change mypy errored with "A function returning TypeVar should receive at least one argument containing the same TypeVar". I thought `_T` was trying to enforce a generic type between two methods, but it can't actually do that. The type vars were only used to cast list items in recursive calls.

Since none of these methods actually pass through an input type, `Any` should be used and those recursive cast operations are unnecessary.

src/marshmallow/schema.py
deckar01: Per https://github.com/marshmallow-code/marshmallow/pull/2181#discussion_r1406499380 this should be `Any` and the cast is not necessary.

src/marshmallow/schema.py
deckar01: Per https://github.com/marshmallow-code/marshmallow/pull/2181#discussion_r1406499380 `obj` should be `Any`.

lafrech:
I tried to apply the changes discussed in comment. I'm not 100% sure I got everything right.

mypy seems to be happy with those changes.

lafrech:
I just fixed `_deserialize` as per https://github.com/marshmallow-code/marshmallow/pull/2181#discussion_r1401276788.

I think I applied all suggested changes.

deckar01 (approved):

sloria (approved):

src/marshmallow/schema.py
```diff
@@ -501,7 +502,9 @@ def _call_and_store(getter_func, data, *, field_name, error_store, index=None):
             return error.valid_data or missing
         return value
 
-    def _serialize(self, obj: _T | typing.Iterable[_T], *, many: bool = False):
+    def _serialize(
+        self, obj: typing.Any | typing.Iterable[dict], *, many: bool = False
```
lafrech: `obj: typing.Any | typing.Iterable[dict]`

should have been

`obj: typing.Any | typing.Iterable[typing.Any]`

which anyway can be simplified as

`obj: typing.Any`

from mypy perspective at least, for a human `| typing.Iterable[typing.Any]` indicates a clearer intent about the `many` case.

lafrech: I just force-pushed a modified commit with just `obj: typing.Any`.

PR #2188, opened 2023-09-16
Add support for Python 3.12

hugovk:
The [second Python 3.12 release candidate](https://discuss.python.org/t/python-3-12-0rc2-final-release-candidate-released/33105?u=hugovk) is out! :rocket:

> ## Call to action
> 
> We strongly encourage maintainers of third-party Python projects to prepare their projects for 3.12 compatibilities during this phase, and where necessary publish Python 3.12 wheels on PyPI to be ready for the final release of 3.12.0.

Python 3.12.0 final is due out in two weeks: [2023-10-02](https://peps.python.org/pep-0693/).

See also https://dev.to/hugovk/help-test-python-312-beta-1508/

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -36,7 +36,6 @@ jobs:
       matrix:
         include:
           - {name: '3.8', python: '3.8', tox: py38}
-          - {name: '3.11', python: '3.11', tox: py311}
           - {name: '3.12', python: '3.12', tox: py312}
     steps:
       - uses: actions/checkout@v4.0.0
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -10,8 +10,8 @@ jobs:
     name: mypy
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3.1.0
-      - uses: actions/setup-python@v4.3.0
+      - uses: actions/checkout@v4.0.0
+      - uses: actions/setup-python@v4.7.0
         with:
           python-version: '3.11'
       - run: python -m pip install --upgrade pip
@@ -21,8 +21,8 @@ jobs:
     name: docs
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3.1.0
-      - uses: actions/setup-python@v4.3.0
+      - uses: actions/checkout@v4.0.0
+      - uses: actions/setup-python@v4.7.0
         with:
           python-version: '3.11'
       - run: python -m pip install --upgrade pip wheel
@@ -37,11 +37,13 @@ jobs:
         include:
           - {name: '3.8', python: '3.8', tox: py38}
           - {name: '3.11', python: '3.11', tox: py311}
+          - {name: '3.12', python: '3.12', tox: py312}
     steps:
-      - uses: actions/checkout@v3.1.0
-      - uses: actions/setup-python@v4.3.0
+      - uses: actions/checkout@v4.0.0
+      - uses: actions/setup-python@v4.7.0
         with:
           python-version: ${{ matrix.python }}
+          allow-prereleases: true
       - run: python -m pip install --upgrade pip
       - run: python -m pip install tox
       - run: python -m tox -e ${{ matrix.tox }}
@@ -52,8 +54,8 @@ jobs:
     if: startsWith(github.ref, 'refs/tags')
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3.1.0
-      - uses: actions/setup-python@v4.3.0
+      - uses: actions/checkout@v4.0.0
+      - uses: actions/setup-python@v4.7.0
         with:
           python-version: '3.11'
       - run: python -m pip install --upgrade pip
@@ -65,8 +67,8 @@ jobs:
     if: startsWith(github.ref, 'refs/tags')
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3.1.0
-      - uses: actions/setup-python@v4.3.0
+      - uses: actions/checkout@v4.0.0
+      - uses: actions/setup-python@v4.7.0
       - name: install requirements
         run: python -m pip install build twine
       - name: build dists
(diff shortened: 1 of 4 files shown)


.github/workflows/build-release.yml
```diff
@@ -37,11 +37,13 @@ jobs:
         include:
           - {name: '3.8', python: '3.8', tox: py38}
           - {name: '3.11', python: '3.11', tox: py311}
+          - {name: '3.12', python: '3.12', tox: py312}
```
deckar01: @lafrech Should we drop 3.11 out of the test matrix to keep it min/max?

.github/workflows/build-release.yml
```diff
@@ -10,8 +10,8 @@ jobs:
     name: mypy
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3.1.0
-      - uses: actions/setup-python@v4.3.0
+      - uses: actions/checkout@v4.0.0
+      - uses: actions/setup-python@v4.7.0
         with:
           python-version: '3.11'
```
deckar01, line 16: We might want to consider bumping the tooling to 3.12 `allow-prereleases` while we are at it.

deckar01:
Here are the changes I proposed: fe6b3a5d

A little off topic, but is there a reason we are limiting the push trigger to tagged `dev` branches? The release job is already restricted to tags. It would be a nice assurance for contributors to know that their branches are stable before opening a PR (and maybe less noise for maintainers). bc08f5bd

.github/workflows/build-release.yml
lafrech: I don't mind either way. As long as we keep it not too old so that it runs on the CI VM.

.github/workflows/build-release.yml
lafrech: I always do. This time it is a bit different, it is a pre-release but this has no consequence, not like if the pre-release success status was optional, for instance.

Yes, let's drop Python 3.11 here and we'll have no change to make when 3.12 is finally out.

setup.py
```diff
@@ -83,6 +83,7 @@ def read(fname):
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: 3.11",
+        "Programming Language :: Python :: 3.12",
```
lafrech, line 86: Should we advertise 3.12 here? What if 3.12 stable actually breaks something we can't fix right away?

I don't mind cause I don't think it'll happen and this tag is essentially informative. Just a theoretical issue.

lafrech (changes requested):

.github/workflows/build-release.yml
hugovk: For tooling, how about `'3.x'`? 

It will use the "latest stable" version, currently 3.11, and will move to 3.12 in the near future. And we won't need to bump this particular line once a year.

setup.py
hugovk: I think it's fine, especially during release candidate phase when the ABI is frozen:

> There will be **no ABI changes** from this point forward in the 3.12 series. The intent is for the final release of 3.12.0, scheduled for Monday, 2023-10-02, to be identical to this release candidate. **This is the last chance to find critical problems in Python 3.12.**

https://discuss.python.org/t/python-3-12-0rc2-final-release-candidate-released/33105?u=hugovk

lafrech:
> A little off topic, but is there a reason we are limiting the push trigger to tagged dev branches? The release job is already restricted to tags. It would be a nice assurance for contributors to know that their branches are stable before opening a PR (and maybe less noise for maintainers). https://github.com/marshmallow-code/marshmallow/commit/bc08f5bd5ac3312f8373e60a32b5655d5f04c510

I don't mind. I often allow CI on any branch push in my repos to fix things before a PR. I suppose the idea was to avoid 2 CI runs for a push in a PR but I'm realizing that once we enable it on push, we don't need it on PR, do we?

I always thought it didn't matter because most contributors push on their repo so only the few people here with push permissions would be affected. I'm realizing that this config may also affect contributors in their own repos. Right? In this case it's nice if CI runs in their repo before the PR is created. And if not, then we need to keep the PR trigger for external contributions.

hugovk:
TODO for a repo admin:
* [ ] Remove 3.11 from required checks and add 3.12:

image

.github/workflows/build-release.yml
lafrech: Sounds good, but then we lose traceability. But I see no real practical issue (it's not like using latest version of all dependencies), so I think it would be fine.

hugovk:
> A little off topic, but is there a reason we are limiting the push trigger to tagged `dev` branches? The release job is already restricted to tags. It would be a nice assurance for contributors to know that their branches are stable before opening a PR (and maybe less noise for maintainers). [bc08f5b](https://github.com/marshmallow-code/marshmallow/commit/bc08f5bd5ac3312f8373e60a32b5655d5f04c510)

Yes, this is a very good idea!

I pushed to `dev` to make sure I could test my changes before opening a PR. If I'd have wanted to create another PR, I wouldn't be able to test it on CI. I strongly believe contributors should use feature branches and should be able to test them first :)

deckar01:
> once we enable it on push, we don't need it on PR, do we?

idk. Does that rebase on the target branch or trigger the checks UI?

lafrech:
All checks appear in CI checks UI: https://github.com/marshmallow-code/flask-smorest/pull/551.

hugovk:
Anything more I need to do here? Thanks!

lafrech:
I guess not. I merged. I updated the required checks.

The `allow-prereleases` line is not needed anymore but it doesn't hurt and it's there for next time.

@deckar01 

> A little off topic, but is there a reason we are limiting the push trigger to tagged dev branches? The release job is already restricted to tags. It would be a nice assurance for contributors to know that their branches are stable before opening a PR (and maybe less noise for maintainers). https://github.com/marshmallow-code/marshmallow/commit/bc08f5bd5ac3312f8373e60a32b5655d5f04c510

No problem with this. And it is probably better to keep both push and PR (for the rebase on the target branch).

Thanks @hugovk.

PR #2226, opened 2024-01-17
Dev chores 

sloria:
- Migrate to pyproject.toml
  - Reference: https://packaging.python.org/en/latest/guides/modernize-setup-py-project/
  - Mostly borrows from https://github.com/pallets/flask/blob/main/pyproject.toml
- Migrate to ruff
  - Because it's faster and consolidates tooling we already use 
- Update release workflow based on https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

I've already done these in a few of the other marshmallow-code projects and my personal repos without issue so far.

see

https://github.com/marshmallow-code/flask-marshmallow/pull/285
https://github.com/marshmallow-code/flask-marshmallow/pull/287


sloria:
@lafrech tagged you for review since there's some tooling changes here. no need to review if it's not interesting to you; just wanted you to be aware before i do the same for the other repos (marshmallow-sqlalchemy, webargs, apispec, etc). i'll plan to merge this at COB tomorrow

src/marshmallow/__init__.py
```diff
@@ -16,7 +18,8 @@
 
 from . import fields
 
-__version__ = "3.20.2"
+# TODO: Deprecate __version__?
```
sloria, line 21: I'm planning to deprecate `__version__`, `__parsed_version__`, and `__version_info__` in a follow-up PR

lafrech:
Thanks for updating all the repos. That's a great deal of work I meant to tackle some day but didn't find the time to. Also, best practices and publication methods change and improve all the time, so I was waiting for things to settle a bit until I could copy from some over project.

I shall do the same to flask-smorest and probably my own repos. Time permitting.

Many thumbs up.

PR #2229, opened 2024-01-19
Deprecate __version__ and related attributes

### LINKED ISSUE #2227: Deprecate __version__ and related attributes
See https://github.com/pallets/flask/issues/5230 for rationale and the linked PR for implementation.

Additionally, this will allow us to remove `packaging` as a dependency

sloria:
close #2227 

Follow-ups:

- Remove usages of `__version__` et al. in other marshmallow-code repos

### DIFF
--- a/src/marshmallow/__init__.py
+++ b/src/marshmallow/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import importlib.metadata
+import typing
 
 from packaging.version import Version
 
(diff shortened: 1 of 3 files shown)


src/marshmallow/__init__.py
```diff
@@ -18,14 +19,49 @@
...
-__version__ = importlib.metadata.version("marshmallow")
-__parsed_version__ = Version(__version__)
-__version_info__: tuple[int, int, int] | tuple[
-    int, int, int, str, int
-] = __parsed_version__.release  # type: ignore[assignment]
-if __parsed_version__.pre:
-    __version_info__ += __parsed_version__.pre  # type: ignore[assignment]
+
+def __getattr__(name: str) -> typing.Any:
+    import warnings
```
lafrech, line 24: I suppose this is imported here to avoid a useless import in normal use.
I don't mind either way.

lafrech (approved):
Sorry, I missed this review requests.

Feel free to go ahead and apply this to all repos.

src/marshmallow/__init__.py
sloria: This turns out to be necessary to avoid an error. If the import is moved to the top, you get

```python
AttributeError: module 'marshmallow.warnings' has no attribute 'warn'
```

flask does the same https://github.com/pallets/flask/blob/b90a4f1f4a370e92054b9cc9db0efcb864f87ebe/src/flask/__init__.py#L49

lafrech: Oh, OK. Strange at first sight, but alright.

Issue #2232, opened 2024-01-25
URL validation fails if URL is missing user field


ddennerline3:
The test code below fails to parse the URL when the user field is missing, but the password is present.

```
from marshmallow import Schema, fields, ValidationError

def test_marshmallow_url():
    """Test marshmallow validating a URL without a username"""

    # When tracing through the code, this is the regex that was built to validate the URL 
    validate_url_regex = r"^(?:[a-z0-9\.\-\+]*)://(?:[^:@]+?(:[^:@]*?)?@|)(?:(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+(?:[A-Z]{2,6}\.?|[A-Z0-9-]{2,}\.?)|localhost|\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}|\[[A-F0-9]*:[A-F0-9:]+\]|(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.?))(?::\d+)?(?:/?|[/?]\S+)\Z"
    url_happy = "rediss://bill:pass1234@www.amazonaws.com:6379"

    class UrlTest(Schema):
        my_url = fields.Url(schemes=["redis", "rediss"], require_tld=False)

    url_test = UrlTest().load({"my_url": url_happy})
    print(url_test)

    url_unhappy = "rediss://:pass1234@www.amazonaws.com:6379"
    try:
        url_test = UrlTest().load({"my_url": url_unhappy})
        print(url_test)
    except ValidationError as err:
        print(url_unhappy, err)
```

RFC1738 specifies that either field is optional.

https://datatracker.ietf.org/doc/html/rfc1738#section-3.1

sloria:
Thanks for reporting! I don't have time to dig into this right now, but would def appreciate an investigation and/or a PR from anyone who'd be so kind

deckar01:
> RFC1738 specifies that either field is optional.

A password is only optional after a user name is included.

> Some or all of the parts "\<user\>:\<password\>@", ":\<password\>", ":\<port\>", and "/\<url-path\>" may be excluded.

> password - An optional password. If present, it follows the user name separated from it by a colon.

> there is no way to specify a password without specifying a user name

sloria:
Oh good catch. So the current validation appears to be correct then. I think we can close this

deckar01:
[...]

PR #2244, opened 2024-02-22
URL User Info Validation

deckar01:
Conform to NWG RFC 3986 per #2232.


sloria:
Great! I tested this against the code sample in https://github.com/marshmallow-code/marshmallow/issues/2232#issue-2100789429 and confirmed this fixes it

PR #2250, opened 2024-03-15
Let IP fields inherit String

lafrech:
Any reason not to inherit `String` here?

Due to this, in apispec, the type wouldn't be set to string. This is being addressed there by setting type to string and the proper format, but I suppose it makes sense to "fix" this here as well.

- https://github.com/marshmallow-code/apispec/pull/892
- https://github.com/marshmallow-code/flask-smorest/issues/623

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1741,7 +1741,7 @@ class Email(String):
         self.validators.insert(0, validator)
 
 
-class IP(Field):
+class IP(String):
     """A IP address field.
 
     :param bool exploded: If `True`, serialize ipv6 address in long form, ie. with groups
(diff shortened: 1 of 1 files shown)


sloria:
i'm unsure about this. we purposely decided to inherit from `Field` instead of `String`: https://github.com/marshmallow-code/marshmallow/pull/1485/commits/62fdfa88acacccbfb2e5730c6ed02ca7b8f6b33c . see this comment: https://github.com/marshmallow-code/marshmallow/pull/1485#issuecomment-579922293

> As I can see, fields inheriting from String, are in general deserialized from strings to strings with additional input value format validation. We can see it in the Url and Email fields cases. Exception from this rule is the UUID field, which deserializes from string into UUID object representation. Therefore, it is not clear for me why UUID inherits from String and not Field, like in the case of DateTime or TimeDelta which also deserializes to object representation.

lafrech:
I guess I get the rationale. The type expressed in the field name (and its parents in the class hierarchy) represents the type in object world, not in serialized world.

Alright, I'm fine with this. This came from apispec where the string type must then be specified in the field class -> JSON type mapping, but that's not a blocker.

Thanks for digging into this.

PR #2264, opened 2024-04-30
DateTime field: allow timestamp 0

### LINKED ISSUE #2133: timestamp 0 is not valid for DateTime('timestamp')
0 is a valid timestamp, however, an error occurs during parsing

code example:
```python


class A(Schema):
    x = fields.DateTime('timestamp')


a = A().load({"x": 0})
print(a)
```
expected:
```python
{'x': datetime.datetime(1970, 1, 1, 0, 0)}
```
actual
```
marshmallow.exceptions.ValidationError: {'x': ['Not a valid datetime.']}
```
[bug line](https://github.com/marshmallow-code/marshmallow/blob/656ef145270db8667e7d945453ba1e85ad0c587e/src/marshmallow/fields.py#L1286)

lafrech:
Fixes #2133.

### DIFF
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -422,6 +422,8 @@ class TestFieldDeserialization:
         [
             "not-a-datetime",
             42,
+            True,
+            False,
             0,
             "",
             [],
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -422,6 +422,7 @@ class TestFieldDeserialization:
         [
             "not-a-datetime",
             42,
+            0,
             "",
             [],
             "2018",
@@ -576,7 +577,7 @@ class TestFieldDeserialization:
     @pytest.mark.parametrize("fmt", ["timestamp", "timestamp_ms"])
     @pytest.mark.parametrize(
         "in_value",
-        ["", "!@#", 0, -1, dt.datetime(2013, 11, 10, 1, 23, 45)],
+        ["", "!@#", -1, dt.datetime(2013, 11, 10, 1, 23, 45)],
     )
     def test_invalid_timestamp_field_deserialization(self, fmt, in_value):
         field = fields.DateTime(format=fmt)
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1285,6 +1285,8 @@ class DateTime(Field):
         return value.strftime(data_format)
 
     def _deserialize(self, value, attr, data, **kwargs) -> dt.datetime:
+        if value is True or value is False:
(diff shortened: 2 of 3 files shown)


tests/test_deserialization.py
```diff
@@ -576,7 +577,7 @@ def test_timestamp_field_deserialization(self, fmt, value, expected):
     @pytest.mark.parametrize("fmt", ["timestamp", "timestamp_ms"])
     @pytest.mark.parametrize(
         "in_value",
-        ["", "!@#", 0, -1, dt.datetime(2013, 11, 10, 1, 23, 45)],
+        ["", "!@#", -1, dt.datetime(2013, 11, 10, 1, 23, 45)],
```
sloria, line 582: Nit: might be worth testing None here

tests/test_deserialization.py
```diff
@@ -422,6 +422,7 @@ def test_field_toggle_show_invalid_value_in_error_message(self):
         [
             "not-a-datetime",
             42,
+            0,
```
sloria, line 427: thought(non-blocking): I believe this will also make False a valid value which isn't totally intuitive. But I'm not sure there's a way around that

sloria (approved):
Just a tiny suggestion. Good to merge and release

tests/test_deserialization.py
lafrech: `None` fails validation before reaching `_deserialize`.

    ValidationError: Field may not be null.

tests/test_deserialization.py
lafrech: I added a condition to reject boolean inputs, copied from numeric fields above.

deckar01: I would be tempted to ~~inherit `Float` to match `datetime.fromtimestamp()`'s type signature and~~ avoid duplicating that logic. ~~That would also allow removing `value = float(value)` from `utils.from_timestamp()` which is equally guilty for the unexpected bool to int promotion.~~

deckar01: On second thought inheritance wouldn't really work with the other date formats.

I would push this logic down to `utils.from_timestamp()` since all of the other date time `utils.from_*` properly error on bools already.

deckar01: https://github.com/marshmallow-code/marshmallow/commit/55cfa193068421906681894479342e2415c751df

lafrech: Right. I added your commit to the PR.

deckar01 (approved):

sloria (approved):

PR #2279, opened 2024-06-06
Simplify Hooks

deckar01:
- Replace slow tuple-keyed dicts with string-keyed dicts
- Load the hook config into `cls._hook` at setup rather than searching `fn.__marshmallow_hook__` at run-time
- Skip mismatched `many` hooks while iterating rather than indexing and filtering them

* pypy (3.10): ~30% speedup
* cpython (3.10): <1% difference
* cpython (3.11, 3.12): ~3.5% speedup

### DIFF
--- a/performance/benchmark.py
+++ b/performance/benchmark.py
@@ -100,7 +100,7 @@ def run_timeit(quotes, iterations, repeat, profile=False):
         profile.disable()
         profile.dump_stats("marshmallow.pprof")
 
-    usec = best * 1e6 / iterations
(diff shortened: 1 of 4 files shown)


deckar01:
Follow up to https://github.com/marshmallow-code/marshmallow/pull/2277#issuecomment-2150978462.

sloria (approved):

lafrech:
Great.

If the code is simpler and (even slightly) more performant, go for it.

-- 
Jérôme

PR #2300, opened 2024-08-20
Reduce pre-commit.ci updates to monthly

sloria:
To reduce PR noise


lafrech (approved):
Lifesaver.

Let's span this across all repos.

PR #2323, opened 2024-10-17
Remove min Python 3 minor version in docs and README

lafrech:
I forgot to update those before releasing and it might not be the first time.

What about we remove the mention of the minor version from the doc pages?

I'd be tempted to also remove it from the README. Especially since we support all maintained Python versions anyway.

### DIFF
--- a/docs/upgrading.rst
+++ b/docs/upgrading.rst
@@ -62,7 +62,7 @@ Upgrading to 3.0
 Python compatibility
 ********************
 
-The marshmallow 3.x series supports Python >= 3.8.
+The marshmallow 3.x series requires Python 3.
 
 
 Schemas are always strict
--- a/README.rst
+++ b/README.rst
@@ -77,16 +77,10 @@ Get It Now
 
     $ pip install -U marshmallow
 
-
 Documentation
 =============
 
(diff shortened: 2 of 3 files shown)


docs/upgrading.rst
```diff
@@ -62,7 +62,7 @@ Upgrading to 3.0
 Python compatibility
 ********************
 
-The marshmallow 3.x series supports Python >= 3.8.
+The marshmallow 3.x series require Python 3.
```
sloria:
```suggestion
The marshmallow 3.x series requires Python 3.
```

sloria (approved):
just a grammar nit, but otherwise good 👍

docs/upgrading.rst
lafrech: Oh right. Series is singular here...

lafrech: Shouldn't I remove the whole Python 3 mention, in 2024?

lafrech: Oh, no, that's the upgrading to 3.0.0 section. Leaving it for history.

However, I'm tempted to also remove the Requirements section of the README.

sloria: >  remove the Requirements section of the README.

👍

sloria (approved):

PR #2324, opened 2024-10-18
:sparkles: Improve error messages for URL validation

### LINKED ISSUE #2243: fields.URL validation error should specify "not a valid FQDN" if trying to pass hostname without domain
The current validation error message: `'Not a valid URL.'` is not descriptive enough to help the user, especially when they are passing a value that would appear to be a valid URL to most people, such as a docker hostname: `http://test-url:8001`

Sample code:
```py

class TestSchema(Schema):
  url = fields.URL(required=True)

TestSchema().load({"url": "http://192.168"}) // this works
TestSchema().load({"url": "http://test-url"}) // this doesn't work
TestSchema().load({"url": "derp"}) // and this has the same error message as http://test-url
```

This is because fields.URL has a default: `require_tld=True`, for `require_tld: Whether to reject non-FQDN hostnames.`
FQDN = fully-qualified domain name. i.e. `http://server1.example.com` is a FQDN, `http://server1` is not.

Suggestion: error message should be more descriptive when require_tld is True, and validation fails FQDN requirement.

ff137:
Resolves #2243

### DIFF
--- a/src/marshmallow/validate.py
+++ b/src/marshmallow/validate.py
@@ -7,6 +7,7 @@ import typing
 from abc import ABC, abstractmethod
 from itertools import zip_longest
 from operator import attrgetter
+from urllib.parse import urlparse
 
 from marshmallow import types
 from marshmallow.exceptions import ValidationError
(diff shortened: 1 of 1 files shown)


lafrech:
I'd do it differently. In init function, build the error string from the parameters to specify the kind of URL that is expected (relative, absolute, FQDN,...).

I gave it a quick shot but didn't find a satisfying phrasing. It might look weird for a client to get a response mentioning FQDN for the (default) URL case.

PR #2654, opened 2024-12-10
Remove TimeDelta `serialization_type` parameter; deserialize from float; improve serialization perf and accuracy

ddelange:
This PR allows propagating floats/strings with decimals to TimeDelta.deserialize without silently downcasting the input to int. Currently, `serialization_type` is not only being used for serialization, but also being used for deserialization, which adds no added value: a float cast can always safely be used for deserialization and is backwards compatible (but fixes this bug).

originally reported here: https://github.com/sloria/environs/issues/372

MRE:
```pycon
>>> field = fields.TimeDelta()
>>> field.deserialize(12.9)
datetime.timedelta(seconds=12)
```
After this PR:
```pycon
>>> field = fields.TimeDelta()
>>> field.deserialize(12.9)
datetime.timedelta(seconds=12, microseconds=900000)
```

I added one test for string input, and changed the existing test for float input.

There are already `math.isclose` tests when passing floats to microseconds (which will be rounded by the datetime.timedelta constructor).

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1435,26 +1435,18 @@ class Date(DateTime):
 
 
 class TimeDelta(Field):
-    """A field that (de)serializes a :class:`datetime.timedelta` object to an
-    integer or float. The integer or float can represent any time unit that the
-    :class:`datetime.timedelta` constructor supports.
+    """A field that (de)serializes a :class:`datetime.timedelta` object to a `float`.
+    The `float` can represent any time unit that the :class:`datetime.timedelta` constructor
+    supports.
 
     :param precision: The time unit used for (de)serialization. Must be one of 'weeks',
         'days', 'hours', 'minutes', 'seconds', 'milliseconds' or 'microseconds'.
-    :param serialization_type: Whether to serialize to an `int` or `float`.
-        Ignored during deserialization: both `int` and `float` inputs are supported.
     :param kwargs: The same keyword arguments that :class:`Field` receives.
-
-    Integer Caveats
-    ---------------
-    When serializing using ``serialization_type=int`` and depending on the ``precision``
-    used, any fractional parts might be truncated (downcast to integer).
 
     Float Caveats
     -------------
     Precision loss may occur when serializing a highly precise :class:`datetime.timedelta`
-    object using ``serialization_type=float`` and a big ``precision`` unit due to floating
-    point arithmetics.
+    object using a big ``precision`` unit due to floating point arithmetics.
 
     When necessary, the :class:`datetime.timedelta` constructor rounds `float` inputs
     to whole microseconds during initialization of the object. As a result, deserializing
@@ -1466,7 +1458,9 @@ class TimeDelta(Field):
         Add `precision` parameter.
     .. versionchanged:: 3.17.0
         Allow serialization to `float` through use of a new `serialization_type` parameter.
-        Defaults to `int` for backwards compatibility.
+        Defaults to `int` for backwards compatibility. Also affects deserialization.
+    .. versionchanged:: 4.0.0
+        Deprecate `serialization_type` parameter, always serialize to float.
     """
 
     WEEKS = "weeks"
@@ -1477,6 +1471,17 @@ class TimeDelta(Field):
     MILLISECONDS = "milliseconds"
     MICROSECONDS = "microseconds"
 
+    # cache this mapping on class level for performance
+    _unit_to_microseconds_mapping = {
+        WEEKS: 1000000 * 60 * 60 * 24 * 7,
+        DAYS: 1000000 * 60 * 60 * 24,
+        HOURS: 1000000 * 60 * 60,
+        MINUTES: 1000000 * 60,
+        SECONDS: 1000000,
+        MILLISECONDS: 1000,
+        MICROSECONDS: 1,
+    }
+
     #: Default error messages.
     default_error_messages = {
         "invalid": "Not a valid period of time.",
@@ -1486,47 +1491,36 @@ class TimeDelta(Field):
     def __init__(
         self,
         precision: str = SECONDS,
-        serialization_type: type[int | float] = int,
+        serialization_type: typing.Any = missing_,
         **kwargs,
-    ):
+    ) -> None:
         precision = precision.lower()
-        units = (
-            self.DAYS,
-            self.SECONDS,
-            self.MICROSECONDS,
-            self.MILLISECONDS,
-            self.MINUTES,
-            self.HOURS,
-            self.WEEKS,
-        )
 
-        if precision not in units:
-            msg = 'The precision must be {} or "{}".'.format(
-                ", ".join([f'"{each}"' for each in units[:-1]]), units[-1]
-            )
+        if precision not in self._unit_to_microseconds_mapping:
+            units = ", ".join(self._unit_to_microseconds_mapping)
+            msg = f"The precision must be one of: {units}."
             raise ValueError(msg)
 
-        if serialization_type not in (int, float):
-            raise ValueError("The serialization type must be one of int or float")
+        if serialization_type is not missing_:
+            warnings.warn(
+                "The 'serialization_type' argument to TimeDelta is deprecated.",
+                RemovedInMarshmallow4Warning,
+                stacklevel=2,
+            )
 
         self.precision = precision
-        self.serialization_type = serialization_type
         super().__init__(**kwargs)
 
-    def _serialize(self, value, attr, obj, **kwargs):
+    def _serialize(self, value, attr, obj, **kwargs) -> float | None:
         if value is None:
             return None
 
-        base_unit = dt.timedelta(**{self.precision: 1})
+        # limit float arithmetics to a single division to minimize precision loss
+        microseconds: int = utils.timedelta_to_microseconds(value)
+        microseconds_per_unit: int = self._unit_to_microseconds_mapping[self.precision]
+        return microseconds / microseconds_per_unit
 
-        if self.serialization_type is int:
-            delta = utils.timedelta_to_microseconds(value)
-            unit = utils.timedelta_to_microseconds(base_unit)
-            return delta // unit
-        assert self.serialization_type is float
-        return value.total_seconds() / base_unit.total_seconds()
-
-    def _deserialize(self, value, attr, data, **kwargs):
+    def _deserialize(self, value, attr, data, **kwargs) -> dt.timedelta:
         try:
             value = float(value)
         except (TypeError, ValueError) as error:
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1436,44 +1436,46 @@ class Date(DateTime):
 
 class TimeDelta(Field):
     """A field that (de)serializes a :class:`datetime.timedelta` object to an
-    integer or float and vice versa. The integer or float can represent the
-    number of days, seconds or microseconds.
+    integer or float. The integer or float can represent any time unit that the
+    :class:`datetime.timedelta` constructor supports.
 
-    :param precision: Influences how the integer or float is interpreted during
-        (de)serialization. Must be 'days', 'seconds', 'microseconds',
-        'milliseconds', 'minutes', 'hours' or 'weeks'.
-    :param serialization_type: Whether to (de)serialize to a `int` or `float`.
+    :param precision: The time unit used for (de)serialization. Must be one of 'weeks',
+        'days', 'hours', 'minutes', 'seconds', 'milliseconds' or 'microseconds'.
+    :param serialization_type: Whether to serialize to an `int` or `float`.
+        Ignored during deserialization: both `int` and `float` inputs are supported.
     :param kwargs: The same keyword arguments that :class:`Field` receives.
 
     Integer Caveats
     ---------------
-    Any fractional parts (which depends on the precision used) will be truncated
-    when serializing using `int`.
+    When serializing using ``serialization_type=int`` and depending on the ``precision``
+    used, any fractional parts might be truncated (downcast to integer).
 
     Float Caveats
     -------------
-    Use of `float` when (de)serializing may result in data precision loss due
-    to the way machines handle floating point values.
+    Precision loss may occur when serializing a highly precise :class:`datetime.timedelta`
+    object using ``serialization_type=float`` and a big ``precision`` unit due to floating
+    point arithmetics.
 
-    Regardless of the precision chosen, the fractional part when using `float`
-    will always be truncated to microseconds.
-    For example, `1.12345` interpreted as microseconds will result in `timedelta(microseconds=1)`.
+    When necessary, the :class:`datetime.timedelta` constructor rounds `float` inputs
+    to whole microseconds during initialization of the object. As a result, deserializing
+    a `float` might be subject to rounding, regardless of `precision`. For example,
+    ``TimeDelta().deserialize("1.1234567") == timedelta(seconds=1, microseconds=123457)``.
 
     .. versionchanged:: 2.0.0
         Always serializes to an integer value to avoid rounding errors.
         Add `precision` parameter.
     .. versionchanged:: 3.17.0
-        Allow (de)serialization to `float` through use of a new `serialization_type` parameter.
-        `int` is the default to retain previous behaviour.
+        Allow serialization to `float` through use of a new `serialization_type` parameter.
+        Defaults to `int` for backwards compatibility.
     """
 
-    DAYS = "days"
-    SECONDS = "seconds"
-    MICROSECONDS = "microseconds"
-    MILLISECONDS = "milliseconds"
-    MINUTES = "minutes"
-    HOURS = "hours"
     WEEKS = "weeks"
+    DAYS = "days"
+    HOURS = "hours"
+    MINUTES = "minutes"
+    SECONDS = "seconds"
+    MILLISECONDS = "milliseconds"
+    MICROSECONDS = "microseconds"
 
     #: Default error messages.
     default_error_messages = {
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1526,7 +1526,7 @@ class TimeDelta(Field):
 
     def _deserialize(self, value, attr, data, **kwargs):
         try:
-            value = self.serialization_type(value)
+            value = float(value)
         except (TypeError, ValueError) as error:
             raise self.make_error("invalid") from error
 
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1460,7 +1460,8 @@ class TimeDelta(Field):
         Allow serialization to `float` through use of a new `serialization_type` parameter.
         Defaults to `int` for backwards compatibility. Also affects deserialization.
     .. versionchanged:: 4.0.0
-        Deprecate `serialization_type` parameter, always serialize to float.
+        Remove `serialization_type` parameter and always serialize to float.
+        Value is cast to a `float` upon deserialization.
     """
 
     WEEKS = "weeks"
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1491,7 +1491,6 @@ class TimeDelta(Field):
     def __init__(
         self,
         precision: str = SECONDS,
-        serialization_type: typing.Any = missing_,
         **kwargs,
     ) -> None:
         precision = precision.lower()
@@ -1500,13 +1499,6 @@ class TimeDelta(Field):
             units = ", ".join(self._unit_to_microseconds_mapping)
             msg = f"The precision must be one of: {units}."
             raise ValueError(msg)
-
-        if serialization_type is not missing_:
-            warnings.warn(
-                "The 'serialization_type' argument to TimeDelta is deprecated.",
-                RemovedInMarshmallow4Warning,
-                stacklevel=2,
-            )
 
         self.precision = precision
         super().__init__(**kwargs)
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -698,7 +698,7 @@ class TestFieldDeserialization:
         assert field.deserialize(value) == expected
 
     def test_invalid_timedelta_precision(self):
-        with pytest.raises(ValueError, match='The precision must be "days",'):
+        with pytest.raises(ValueError, match="The precision must be one of: weeks,"):
             fields.TimeDelta("invalid")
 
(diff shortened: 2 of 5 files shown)


sloria (approved):
Thank you for the PR ❤️ This seems like the right behavior.

Lgtm but I'll give @lafrech some time to take a look in case I'm not thinking of some case where end users might rely on the current behavior.

tests/test_deserialization.py
```diff
@@ -741,7 +748,7 @@ def test_timedelta_field_deserialization(self):
         assert isinstance(result, dt.timedelta)
         assert result.days == 0
         assert result.seconds == 12
-        assert result.microseconds == 0
+        assert result.microseconds == 900000
```
sloria, line 751: 🤔 This was explicitly asserting that microseconds were lost. I'm not sure why, though

lafrech:
[I enabled issues again. The spam streak is over on flask-smorest. Let's hope it is here too.]

I don't have a full understanding of the issue but in case it went unnoticed, here are related issues/PRs:

- I think current behaviour was introduced a while back in #105.
- Possibly related: #1998.

Also, the docstring should be updated if we merge.

ddelange:
Hi @lafrech 👋 

You're right, the docstring needed some work. Please see the last commit. I hope that also clarifies the bug that this PR fixes: needlessly using `serialization_type` during deserialization.

sloria:
Docs update looks good to me. Thanks @ddelange .

Thanks for digging up the original issue and PRs @lafrech. Looking back at those, I think allowing truncation to an integer was an oversight on my part rather than an explicit decision. Even the [ISO8601](https://en.wikipedia.org/wiki/ISO_8601#Durations) spec mentioned in https://github.com/marshmallow-code/marshmallow/issues/105#issuecomment-66910264 allows for fractional components:

> The smallest value used may also have a decimal fraction,[[39]](https://en.wikipedia.org/wiki/ISO_8601#cite_note-39) as in "P0.5Y" to indicate half a year. This decimal fraction may be specified with either a [comma](https://en.wikipedia.org/wiki/Comma) or a [full stop](https://en.wikipedia.org/wiki/Full_stop), as in "P0,5Y" or "P0.5Y".

@lafrech @deckar01  Do we consider this a breaking change? Even though this is a much smaller change than previous major version bumps, I'm not opposed to a major version bump with some explanatory text in the changelog that says that the change is very limited in scope and only affects users of the `TimeDelta` field.

ddelange:
fwiw, this change only affects (helps) users that were serializing using `serialization_type=float`, but then not taking care when deserializing.

I would say this does not warrant a major version bump, but rather a bugfix release.

sloria:
It would also affect users who didn't explicitly pass `serialization_type` or passed `serialization_type=int`, right (as your MRE shows)? Users who were relying on floats getting truncated would experience a breakage (not saying this is likely or desirable, but it's possible).

ddelange:
right. the good ol' _rely on a bug_ motto. [programming sucks](https://www.stilldrinking.org/programming-sucks)...

sloria:
lol, yeah... i guess i'm not 100% sure it's a bug, given that there was an explicit test for the current behavior. i'm like 80% sure it was oversight, but it was long ago so want to give a bit of time for others to chime in before shipping the change

src/marshmallow/fields.py
```diff
@@ -1526,7 +1528,7 @@ def _serialize(self, value, attr, obj, **kwargs):
 
     def _deserialize(self, value, attr, data, **kwargs):
         try:
-            value = self.serialization_type(value)
+            value = float(value)
```
lafrech, line 1523: Is it possible that when deserializing an `int` this introduces an error due to floating point precision?

Like `float(42)` -> `42.0000000000000000001`

Answering to myself, no because this will be rounded in timedelta:

```
>>> field.deserialize(12.9000000001)
datetime.timedelta(seconds=12, microseconds=900000)
```

src/marshmallow/fields.py
```diff
@@ -1436,44 +1436,46 @@ def _make_object_from_format(value, data_format):
...
+    ``TimeDelta().deserialize("1.1234567") == timedelta(seconds=1, microseconds=123457)``.
 
     .. versionchanged:: 2.0.0
         Always serializes to an integer value to avoid rounding errors.
         Add `precision` parameter.
     .. versionchanged:: 3.17.0
-        Allow (de)serialization to `float` through use of a new `serialization_type` parameter.
-        `int` is the default to retain previous behaviour.
+        Allow serialization to `float` through use of a new `serialization_type` parameter.
+        Defaults to `int` for backwards compatibility.
```
lafrech: Should we add a new versionchanged line here?

lafrech (commented):
I'm not sure.

One of our concerns when designing a field is how values round-trip.

Current behaviour lets the user decide whether the serialized version should be `int` or `float`, and considers it is being given serialized values accordingly. If values are serialized as `int`, then the type that is expected as input when deserializing is `int`.

The original name of the parameter was `serde_type`, not very nice but showing it works both ways.

Before this PR:

```
    # TimeDelta(serialization_type=int)
    12 -> timedelta(seconds=12) -> 12
    12.42 -> timedelta(seconds=12) -> 12

    # TimeDelta(serialization_type=float)
    12 -> timedelta(seconds=12) -> 12.0
    12.42 -> timedelta(seconds=12, microseconds=420000) -> 12.42
```

After this PR:

```
    # TimeDelta(serialization_type=int)
    12 -> timedelta(seconds=12) -> 12
    12.42 -> timedelta(seconds=12, microseconds=420000) -> 12

    # TimeDelta(serialization_type=float)
    12 -> timedelta(seconds=12) -> 12.0
    12.42 -> timedelta(seconds=12, microseconds=420000) -> 12.42
```

From this perspective, I'm not sure the latter is less surprising behaviour.

The PR makes the field a bit more permissive on inputs, accepting floats when expecting integers. This is not in line with `Integer` field which truncates floats.

If floats are to be expected, then `serialization_type=float` should be used.

If the issue is that `TimeDelta` defaults to `int` then maybe the default could be changed in next major version. Until then, nothing prevents the user to specify `float`, right?

I might be totally mistaken, in which case please ignore.

src/marshmallow/fields.py
ddelange:
```
Never truncate a `float` to `int` during deserialization.
```

too concise?

ddelange:
Thanks for your elaborate review!

> From this perspective, I'm not sure the latter is less surprising behaviour.

To me the latter definitely looks 'right'. From another perspective: that float came from marshmallow (starting off with a timedelta). It got serialized and sent elsewhere for deserialization. When the receiving party deserializes, in the latter example he doesn't need prior knowledge about the sender's settings. Deserialization has now become robust.

src/marshmallow/fields.py
ddelange: or maybe
```
Never truncate a `float` to an `int` before deserializing it into a :class:`datetime.timedelta`.
```

lafrech:
Is there anything wrong with ` TimeDelta(serialization_type=float)`, apart from the fact that it is not the default?

I mean does the change in this PR allow a use case that can't be achieved with current parameters? Or is the only benefit to make it the default choice?

ddelange:
re: anything wrong - my argument here is that serialization_type should be used only for serialization and not for deserialization. but that's about it :)

lafrech:
So would it be fine if the parameter was named `serialization_deserialization_type`, `de_serialization_type` or `serialized_type` of whatever we can come up with that makes it explicit it works both ways?

If it is just that, we could use a better naming and deprecate the old one.

ddelange:
from my side there's no need for a breaking change here, this change just seemed like a nice robustness improvement to me (not considering this breaking).

it only came up because a user was surprised that his float was being silently truncated when feeding into @sloria's environs package. since that package only ever deserializes (values from unknown sources), this seemed like the right way forward. but feel free to close / take over if you do consider this a 'bad' change, no hard feelings :)

will look uglier than [now](https://github.com/sloria/environs/blob/11.2.1/src/environs/__init__.py#L377), but environs _could_ fix this without any PR om marshmallow by subclassing `TimeDelta.__init__`

sloria:
I generally agree with @ddelange's comment in https://github.com/marshmallow-code/marshmallow/pull/2654#issuecomment-2537447810, and the proposed change certainly makes sense for the use case in environs. On one hand, we might as well be robust on input. `Boolean` will accept various strings and integers while outputting to a `bool`. On the other hand, marshmallow doesn't always attempt to be lenient on inputs. For example, `Date` and `DateTime` assume that the input and output formats are the same.

@lafrech @deckar01 Any strong opinions?

sloria:
Another possible path would be to

1. Add a `deserialization_type` to this PR and make it default to `int` (backwards-compatible)
2. Add a warning if a deserializing a float

something like

```diff
diff --git a/src/marshmallow/fields.py b/src/marshmallow/fields.py
index b564aee..19d82f6 100644
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1487,6 +1487,8 @@ def __init__(
         self,
         precision: str = SECONDS,
         serialization_type: type[int | float] = int,
+        *,
+        deserialization_type: type[int | float] = int,
         **kwargs,
     ):
         precision = precision.lower()
@@ -1508,9 +1510,12 @@ def __init__(
 
         if serialization_type not in (int, float):
             raise ValueError("The serialization type must be one of int or float")
+        if deserialization_type not in (int, float):
+            raise ValueError("The deserialization type must be one of int or float")
 
         self.precision = precision
         self.serialization_type = serialization_type
+        self.deserialization_type = deserialization_type
         super().__init__(**kwargs)
 
     def _serialize(self, value, attr, obj, **kwargs):
@@ -1527,8 +1532,15 @@ def _serialize(self, value, attr, obj, **kwargs):
         return value.total_seconds() / base_unit.total_seconds()
 
     def _deserialize(self, value, attr, data, **kwargs):
+        if isinstance(value, float) and self.deserialization_type is not float:
+            warnings.warn(
+                f"float value for attribute {attr!r} will truncated to an integer value in {self.precision}. "
+                + "It will be deserialized as a float in marshmallow 3.25.0.",
+                UserWarning,
+                stacklevel=2,
+            )
         try:
-            value = float(value)
+            value = self.deserialization_type(value)
         except (TypeError, ValueError) as error:
             raise self.make_error("invalid") from error
```

3. Release this in 3.24.0.
4. Change the default `deserialization_type` to `float` and remove the warning.
5. Release 3.25.0

this warns potentially-affected users of the change. but maybe it's not worth the added API surface? ¯\\_(ツ)_/¯

deckar01:
`total_seconds()` returns a float, so that should be the default. I don’t think it needs an extra arg (deserialize from == serialize to). Truncation was documented and tested, so I think it needs a major release, even though the use case is questionable.

We might consider deprecating the type arg while we are at it. The truncation can be enveloped or a custom field and does not seem to warrant a convenience API.

ddelange:
do you suggest to deprecate `serialization_type` altogether (major release) and make float the default? sounds good to me

lafrech:
It seems to me that the current code allows everything using the right
parameters and it's just the default behaviour that is being discussed,
so there's no need to rush fixing/improving this in 3.x by adding
another parameter or a deprecation notice.

Simplifying things and removing parameters can be done in a 4.x change.
We could pick the low-hanging fruits from the 4.0 milestone and ship
soonish, postponing the other issues to 5.0.

-- 
Jérôme

sloria:
@lafrech if i'm understanding your suggestion correctly:

1. change this PR to remove the `serialization_type` parameter and mark this for 4.0
2. optionally merge other small-but-breaking changes

in environs, `serialization_type` can be defaulted to `float` for marshmallow 3 compatibility.

if this is the proposal, i'm good with it

lafrech:
Yes, scratch this and keep things as is in 3.x, then drop
`serialization_type` and only support the `float` case in 4.0 as
@deckar01 suggests.

And rather than postpone 4.0 endlessly, select PRs from the 4.0
milestone that can be finalized easily.

src/marshmallow/fields.py
```diff
@@ -1484,49 +1480,60 @@ class TimeDelta(Field):
...
+        if self.precision == self.HOURS:
+            return value.days * 24 + seconds / 3600
+        if self.precision == self.MINUTES:
+            return value.days * 24 * 60 + seconds / 60
+        if self.precision == self.SECONDS:
+            return value.days * 24 * 60 * 60 + seconds
+        if self.precision == self.MILLISECONDS:
+            return value.days * 24 * 60 * 60 * 1000 + seconds * 1000
+        if self.precision == self.MICROSECONDS:
+            return value.days * 24 * 60 * 60 * 1000000 + seconds * 1000000
```
ddelange, line 1532: I had to write this out in order to avoid precision loss that occurs with the (current) `total_seconds` approach:
```py
base_unit = dt.timedelta(**{self.precision: 1})
return value.total_seconds() / base_unit.total_seconds()
```

it's shorter, but it's a bad implementation when you see what it does to the d6 test:
```diff
        field = fields.TimeDelta(fields.TimeDelta.WEEKS)
-       assert field.serialize("d6", user) == 1.148910385252976
+       assert field.serialize("d6", user) == 1.1489103852529763
        field = fields.TimeDelta(fields.TimeDelta.DAYS)
        assert field.serialize("d6", user) == 8.042372696770833
        field = fields.TimeDelta(fields.TimeDelta.HOURS)
-       assert field.serialize("d6", user) == 193.0169447225
+       assert field.serialize("d6", user) == 193.01694472250003
        field = fields.TimeDelta(fields.TimeDelta.MINUTES)
        assert field.serialize("d6", user) == 11581.01668335
        field = fields.TimeDelta(fields.TimeDelta.SECONDS)
        assert field.serialize("d6", user) == 694861.001001
        field = fields.TimeDelta(fields.TimeDelta.MILLISECONDS)
        assert field.serialize("d6", user) == 694861001.001
        field = fields.TimeDelta(fields.TimeDelta.MICROSECONDS)
-       assert field.serialize("d6", user) == 694861001001
+       assert field.serialize("d6", user) == 694861001001.0001
```

src/marshmallow/fields.py
```diff
@@ -1484,49 +1480,63 @@ class TimeDelta(Field):
...
```
ddelange: I had to write this out in order to avoid precision loss that occurs with the (current) `total_seconds` approach:
```py
base_unit = dt.timedelta(**{self.precision: 1})
return value.total_seconds() / base_unit.total_seconds()
```

it's shorter, but it's a bad implementation when you see what it does to [the d6 test](https://github.com/ddelange/marshmallow/blob/fd171e80bcee994c8e663e2ef55af825137b2bfa/tests/test_serialization.py#L743):
```diff
        field = fields.TimeDelta(fields.TimeDelta.WEEKS)
-       assert field.serialize("d6", user) == 1.148910385252976
+       assert field.serialize("d6", user) == 1.1489103852529763
        field = fields.TimeDelta(fields.TimeDelta.DAYS)
        assert field.serialize("d6", user) == 8.042372696770833
        field = fields.TimeDelta(fields.TimeDelta.HOURS)
-       assert field.serialize("d6", user) == 193.0169447225
+       assert field.serialize("d6", user) == 193.01694472250003
        field = fields.TimeDelta(fields.TimeDelta.MINUTES)
        assert field.serialize("d6", user) == 11581.01668335
        field = fields.TimeDelta(fields.TimeDelta.SECONDS)
        assert field.serialize("d6", user) == 694861.001001
        field = fields.TimeDelta(fields.TimeDelta.MILLISECONDS)
        assert field.serialize("d6", user) == 694861001.001
        field = fields.TimeDelta(fields.TimeDelta.MICROSECONDS)
-       assert field.serialize("d6", user) == 694861001001
+       assert field.serialize("d6", user) == 694861001001.0001
```

looking at the [`total_seconds`](https://github.com/python/cpython/blob/v3.13.1/Lib/_pydatetime.py#L728) implementation, it looks like these artifacts are due to scaling up to microseconds before addition, and then scaling back down to seconds after addition (resulting in a float). then finally one more float operation, namely dividing the two `total_seconds` floats

ddelange (commented):
OK! now targeting 4.0: I've demolished `serialization_type` (with deprecation warning), updated tests and added to docstring:
```
.. versionchanged:: 4.0.0
    Deprecate `serialization_type` parameter, always serialize to float.
```

src/marshmallow/fields.py
ddelange: switched to using `utils.timedelta_to_microseconds(value)` so we only need a single float operation (dividing two ints)

src/marshmallow/fields.py
```diff
@@ -1484,49 +1491,38 @@ class TimeDelta(Field):
...
-            self.WEEKS,
-        )
 
-        if precision not in units:
-            msg = 'The precision must be {} or "{}".'.format(
-                ", ".join([f'"{each}"' for each in units[:-1]]), units[-1]
-            )
+        if precision not in self._unit_to_microseconds_mapping:
+            units = ", ".join(self._unit_to_microseconds_mapping)
+            msg = f"The precision must be one of: {units}."
```
ddelange, line 1506: unified this msg formatting with the Enum field [ref](https://github.com/marshmallow-code/marshmallow/blob/3470680cdfea01ddfdd5a1257b5553cf5153b1eb/tests/test_deserialization.py#L1185).

ddelange:
`TimeDelta._serialize` is now 5x faster:
```python
In [1]: from marshmallow import utils
   ...: from datetime import timedelta
   ...:
   ...: WEEKS = "weeks"
   ...: DAYS = "days"
   ...: HOURS = "hours"
   ...: MINUTES = "minutes"
   ...: SECONDS = "seconds"
   ...: MILLISECONDS = "milliseconds"
   ...: MICROSECONDS = "microseconds"
   ...:
   ...: _unit_to_microseconds_mapping = {
   ...:     WEEKS: 1000000 * 60 * 60 * 24 * 7,
   ...:     DAYS: 1000000 * 60 * 60 * 24,
   ...:     HOURS: 1000000 * 60 * 60,
   ...:     MINUTES: 1000000 * 60,
   ...:     SECONDS: 1000000,
   ...:     MILLISECONDS: 1000,
   ...:     MICROSECONDS: 1,
   ...: }
   ...:
   ...: def serialize_new(value):
   ...:     precision = 'weeks'
   ...:     microseconds: int = utils.timedelta_to_microseconds(value)
   ...:     microseconds_per_unit: int = _unit_to_microseconds_mapping[precision]
   ...:     return microseconds / microseconds_per_unit
   ...:
   ...: def serialize_old(value):
   ...:     precision = 'weeks'
   ...:     base_unit = timedelta(**{precision: 1})
   ...:     return value.total_seconds() / base_unit.total_seconds()
   ...:
   ...: value = timedelta(weeks=1, microseconds=1)

In [2]: %timeit serialize_old(value)
558 ns ± 3.16 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [3]: %timeit serialize_new(value)
113 ns ± 0.583 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

src/marshmallow/fields.py
```diff
@@ -1484,49 +1491,38 @@ class TimeDelta(Field):
...
-    def _serialize(self, value, attr, obj, **kwargs):
+    def _serialize(self, value, attr, obj, **kwargs) -> float | None:
         if value is None:
             return None
 
-        base_unit = dt.timedelta(**{self.precision: 1})
+        # limit float arithmetics to a single division to minimize precision loss
+        microseconds: int = utils.timedelta_to_microseconds(value)
+        microseconds_per_unit: int = self._unit_to_microseconds_mapping[self.precision]
+        return microseconds / microseconds_per_unit
```
ddelange, line 1519: apart from [float precision issues](https://github.com/marshmallow-code/marshmallow/pull/2654#discussion_r1894992958) with the old `total_seconds` approach, the new approach is 5x faster:
```py
In [1]: from marshmallow import utils
   ...: from datetime import timedelta
   ...:
   ...: WEEKS = "weeks"
   ...: DAYS = "days"
   ...: HOURS = "hours"
   ...: MINUTES = "minutes"
   ...: SECONDS = "seconds"
   ...: MILLISECONDS = "milliseconds"
   ...: MICROSECONDS = "microseconds"
   ...:
   ...: _unit_to_microseconds_mapping = {
   ...:     WEEKS: 1000000 * 60 * 60 * 24 * 7,
   ...:     DAYS: 1000000 * 60 * 60 * 24,
   ...:     HOURS: 1000000 * 60 * 60,
   ...:     MINUTES: 1000000 * 60,
   ...:     SECONDS: 1000000,
   ...:     MILLISECONDS: 1000,
   ...:     MICROSECONDS: 1,
   ...: }
   ...: 
   ...: precision = WEEKS
   ...:
   ...: def serialize_old(value):
   ...:     base_unit = timedelta(**{precision: 1})
   ...:     return value.total_seconds() / base_unit.total_seconds()
   ...:
   ...: def serialize_new(value):
   ...:     microseconds: int = utils.timedelta_to_microseconds(value)
   ...:     microseconds_per_unit: int = _unit_to_microseconds_mapping[precision]
   ...:     return microseconds / microseconds_per_unit
   ...:
   ...: value = timedelta(weeks=1, microseconds=1)

In [2]: %timeit serialize_old(value)
558 ns ± 3.16 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [3]: %timeit serialize_new(value)
113 ns ± 0.583 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

src/marshmallow/fields.py
```diff
@@ -1484,49 +1491,38 @@ class TimeDelta(Field):
...
-                ", ".join([f'"{each}"' for each in units[:-1]]), units[-1]
-            )
+        if precision not in self._unit_to_microseconds_mapping:
+            units = ", ".join(self._unit_to_microseconds_mapping)
+            msg = f"The precision must be one of: {units}."
             raise ValueError(msg)
 
-        if serialization_type not in (int, float):
-            raise ValueError("The serialization type must be one of int or float")
+        if serialization_type is not missing_:
```
sloria: Since we're actually removing the serialization_type rather than deprecating, this code block becomes unnecessary

ddelange: ah right. so I'll remove? and people on v4 still passing the arg will get silently propagated into `super().__init__(**kwargs)` (going into `additional_metadata`)

ddelange: 7519edc

sloria: additional_metadata will be removed in 4.0, so it passing `serialization_type` will fail loudly (which we want)

sloria (approved):

PR #2657, opened 2024-12-12
Improve type hint formatting to resolve PyCharm warnings

### LINKED ISSUE #2268: fields.Nested Type Warning Since Version 3.15.0
Hi,

I'm encountering a warning from PyCharm related to `fields.Nested` after upgrading to `marshmallow` version 3.15.0. The warning states that `fields.Nested` expects a type `SchemaABC`, but it gets a type `[UserSerializer]` instead. This issue did not occur in earlier versions.

Here is a simple example to demonstrate the issue:

```python

class A(Schema):
    id = fields.Integer()

class B(Schema):
    a = fields.Nested(A)
```

**Warning:**
```
Expected type 'SchemaABC', got 'Type[A]' instead 
```

This warning appears when using `fields.Nested` with a schema class. It seems that the type checking introduced in version 3.15.0 does not recognize schema classes correctly.

Is there a new way to define nested schemas, or is this a bug that needs to be addressed?

Thank you for your help!

Fares-Abubaker:
This PR refactors the type hints for `Field`, `Nested`, and `Function` fields to ensure compatibility with PyCharm’s type checker. Previously, certain multiline unions and formatting choices caused PyCharm to show spurious warnings. By adjusting the formatting—using parentheses for multiline unions and, where needed, Optional and Union—these warnings are resolved without altering the underlying logic. This change will improve the developer experience for anyone using PyCharm with this codebase.

Fixes #2268

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -175,3 +175,4 @@ Contributors (chronological)
 - Sebastien Lovergne `@TheBigRoomXXL <https://github.com/TheBigRoomXXL>`_
 - Peter C `@somethingnew2-0 <https://github.com/somethingnew2-0>`_
 - Marcel Jackwerth `@mrcljx` <https://github.com/mrcljx>`_
+- Fares Abubaker `@Fares-Abubaker <https://github.com/Fares-Abubaker>`_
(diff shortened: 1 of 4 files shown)


sloria:
Very odd 🤔 , wouldn't think the formatting would matter to type-checkers. But no objection to the change. Thanks for the PR!

PR #2698, opened 2024-12-29
Use format specifiers instead of percent format

lafrech:
Should fix #2331.

### DIFF
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -138,7 +138,7 @@ def get_fixed_timezone(offset: int | float | dt.timedelta) -> dt.timezone:
     if isinstance(offset, dt.timedelta):
         offset = offset.total_seconds() // 60
     sign = "-" if offset < 0 else "+"
-    hhmm = "%02d%02d" % divmod(abs(offset), 60)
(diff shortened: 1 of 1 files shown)


src/marshmallow/utils.py
```diff
@@ -138,7 +138,7 @@ def get_fixed_timezone(offset: int | float | dt.timedelta) -> dt.timezone:
     if isinstance(offset, dt.timedelta):
         offset = offset.total_seconds() // 60
     sign = "-" if offset < 0 else "+"
-    hhmm = "%02d%02d" % divmod(abs(offset), 60)
+    hhmm = "%02d%02d".format(*divmod(abs(offset), 60))
```
sloria, line 141: Wouldn't this need to be

```python
hhmm = "{:02d}{:02d}".format(*divmod(abs(offset), 60))

```

sloria: https://github.com/marshmallow-code/marshmallow/pull/2699

lafrech: Yeah, probably. String formatting is a pain and I suck at it.
I was happy enough that CI passed, I figured this was fine.
Thanks for double-checking.

PR #2704, opened 2024-12-30
Remove ordered class Meta option

### LINKED ISSUE #2146: Remove `ordered` option
`set_class` defaults to `OrderedSet` which is convenient enough, and setting `dict_class` to `OredredDict` is a corner case, so we can get rid of the `ordered` attribute and pass those as class attributes.

This is a follow-up to #1744 and #1896, the backwards incompatible part that should happen in marshmallow 4.

sloria:
close #2146

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -37,8 +37,8 @@ As a consequence of this change:
 
 Thanks :user:`ddelange` for the PR.
 
-- The `ordered` class Meta option is removed. Set `Schema.dict_class` to `OrderedDict` to maintain the previous behavior.
-  Note that field order already preserved by default (:issue:`2146`).
+- The `ordered` class Meta option is removed  (:issue:`2146`). Field order is already preserved by default.
+  Set `Schema.dict_class` to `OrderedDict` to maintain the previous behavior.
 
 Previously-deprecated API have been removed, including:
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.rst
```diff
@@ -35,6 +35,9 @@ As a consequence of this change:
 
 Thanks :user:`ddelange` for the PR.
 
+- The `ordered` class Meta option is removed. Set `Schema.dict_class` to `OrderedDict` to maintain the previous behavior.
+  Note that field order already preserved by default (:issue:`2146`).
```
lafrech: s/field order already preserved/field order is already preserved

lafrech (commented):
Nice cleanup!

Should we add a test where `OrderedDict` is used as `dict_class`?

We may want to rename schema classes in test_options as they don't make much sense when everything schema is ordered. This is something I should have done in #1896 but I lazily postponed. It can wait a little more...

sloria:
> Should we add a test where OrderedDict is used as dict_class?

looks like we already have one here https://github.com/marshmallow-code/marshmallow/blob/747c198bc13317e45dae00f160122b79858b3811/tests/test_schema.py#L2663-L2664

> We may want to rename schema classes in test_options as they don't make much sense when everything schema is ordered. 

good idea. done

lafrech (approved):

PR #2705, opened 2024-12-30
Remove implicit field creation

### LINKED ISSUE #1356: RFC: Remove implicit field creation

sloria:
close #1356

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@ Changelog
 4.0.0 (unreleased)
 ******************
 
+- *Backwards-incompatible*: Remove implicit field creation, i.e. using the ``fields`` or ``additional`` class Meta options with undeclared fields (:issue:`1356`).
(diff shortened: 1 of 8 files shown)


lafrech (approved):
Nice cleanup here, too!

sloria:
i've left `Schema.TYPE_MAPPING`--even though it's no longer used internally--because it's useful for schema autogen libs like marshmallow-sqlalchemy: https://github.com/marshmallow-code/marshmallow-sqlalchemy/blob/5318e4458b59d38ecd894ca1d2660d008c029cfd/src/marshmallow_sqlalchemy/convert.py#L108-L112

Issue #2173, opened 2023-08-18
Feature request: immutable schemas and context passing


matejsp:
When using context passing into schema you have to always create new instance making schema not immutable.
```
sch = SomeSchema(context=http_request)
data = {"ac1": {"a": 1}, "ac2": {"a": 1}}
sch.load(data)
```

I would prefer a bit different syntax to pass context inside load.
```
sch = SomeSchema()
data = {"ac1": {"a": 1}, "ac2": {"a": 1}}
sch.load(data, context=http_request1)
sch.load(data, context=http_request2)
```

This would make Schema definition immutable, thread safe and faster since you don't have to initialise fields for each and every request.

lafrech:
Related to #1826.

matejsp:
This looks like possible solution. I will follow the discussion in related issue.

Issue #1825, opened 2021-06-11
Race Conditions in Marshmallow - Context Not thread safe?


alnikolaj:
Hello everybody,

I am having a serious problem. We are serializing a list of objects up to 100 objects at a time/per request and we have some custom field methods.

In these field methods, we pass the user_id in the context and perform some queries specific to that user_id. Now... I have noticed that sometimes this extra object field is wrong and after much investigation I found out that in some cases (within the list of objects in the same request!) the user_id was wrong and suddenly changed. **The user_id passed through the context changed mid serialization of these 100 objects...** I checked this "wrong" user and saw that his/her last activity was the same as mine what leads me to the conclusion that for some reason Marshmallow is not thread-safe and there is a race condition? Here is an example of my code:

```python
# Schema
class MySchema(SQLAlchemyAutoSchema):
    class Meta:
        model = Object

    extra = fields.Method('custom_field', dump_only=True)

    def custom_field(self, obj):
        user_id = self.context.get('user_id')

        # User Favorites
        favorited = False
        in_favorites = Favorite.query.filter_by(user_id = user_id).first()
        if in_favorites:
            favorited = True
        
        return {
            'user_id' : user_id, 
            'favorited' : favorited
        }

# API
objects = Object.query.all()
user = User.query.get(id=id)

many_schema = MySchema(many=True)
many_schema.context['user_id'] = user.id
return many_schema.dump(objects)
```

I am unsure how to proceed now. I would need to fix this as soon as possible. 

Thanks in advance.

lafrech:
I suspect the same schema instance is shared among all requests. This shouldn't happen if the view is passed a schema class and instantiates the schema on each request.

Please post a sample code to show the way you use the schema in the view function.

alnikolaj:
Example:

```
#api.py
from flask_classful import FlaskView
from app.models import User, Object
from serializer.py import MySchema

schema = MySchema()

class ObjectListAPI(FlaskView):
    def get(self):
        # Query User
        user_id = request.args.get('user_id')
        user_obj = User.query.get(user_id)
        objects = Object.query.all()

        schema.context['user_id'] = user.id
        return schema.dump(objects)

```

I had the MySchema initialisation outside of the Get Function / FlaskView, which might have caused the problem?

lafrech:
Exactly.

This is not exactly a marshmallow thread-safety issue. If a object instance is shared by multiple threads, setting an attribute in one thread modifies the object for the other threads as well.

[...]

PR #2707, opened 2024-12-30
Use a context variable to pass Schema context

lafrech:
Implements #1826.

Hopefully solves/fixes/replaces/obsoletes #1791, #1833, #1700, #2173, #1617, #1825.

I think most current use cases are covered.

Context can be passed in dump/load call or set in a context manager. The former overrides the latter.

Things can be renamed if `CONTEXT` and `Context` are not specific enough.

`context` used to be a `dict` but can now be anything the user wants. I didn't see any reason to enforce a `dict`. We could let `CONTEXT` default to an empty `dict` (immutable, to avoid issues) but since the type may be anything I didn't do it.

To access the context from a function, one must use `CONTEXT.get()`. This may look a little less nice than before (`self.context`) but people familiar with context variables shouldn't frown upon that. We could make it available from a schema or field property, but it would hide the technical implementation a little too much IMHO and could lead to misunderstandings.

I should add tests showing it works from any hook.

Docs need to be updated but I'd rather wait for feedback before I do it.


tests/test_schema.py
```diff
@@ -360,15 +363,13 @@ class NestedSchema(Schema):
             bar = fields.Str()
 
             def on_bind_field(self, field_name, field_obj):
-                field_obj.metadata["fname"] = self.context["fname"]
```
lafrech, line 353: I modified this test because I couldn't reproduce this with the new context.

I assume this was a very contrived example, not really something we want to do in real life.

This test is now so close to the test above, I'm wondering if we ought to keep it.

tests/test_schema.py
```diff
@@ -2348,51 +2349,67 @@ class ValidatingSchema(Schema):
 
 class UserContextSchema(Schema):
     is_owner = fields.Method("get_is_owner")
-    is_collab = fields.Function(lambda user, ctx: user in ctx["blog"])
+    is_collab = fields.Function(
+        lambda user: user in typing.cast(dict, CONTEXT.get())["blog"]
+    )
```
lafrech: For some reason I couldn't figure, mypy complains if I don't cast, but it does not complain about a later equivalent line where I didn't cast.

```py
likes_bikes = fields.Function(lambda obj: "bikes" in CONTEXT.get()["info"])
```

tests/test_schema.py
```diff
@@ -2348,51 +2349,67 @@ class ValidatingSchema(Schema):
...
-        context = {"blog": blog}
-        serializer = UserContextSchema()
-        serializer.context = context
-        data = serializer.dump(collab)
-        assert data["is_collab"] is True
-        noncollab = User("Foo")
-        data = serializer.dump(noncollab)
-        assert data["is_collab"] is False
-
-    def test_function_field_raises_error_when_context_not_available(self):
```
lafrech, line 2194: This does not make much sense anymore. User code should manage the case where the context is unset.

lafrech:
Thinking of it, this PR does either too much or not enough.

If we kept only the explicit context manager form, this could be achieved in user code. The added value in providing this from marshmallow code is to allow passing the context in the load/dump call rather than having to call the context manager. And of course to avoid users recreating the wheel on each project.

We could remove the whole feature.

Or we could go a bit further and expose the context with schema or field properties to hide the context variable. I just pushed a new commit doing this. `Function` field is a special case as it cannot access the field or schema, so I reintroduced the 3.0 logic.

Overall, I prefer that latter option to the full removal, at least for now. It makes for a smoother transition.

The context feature was introduced when context variables where not a thing and IIUC with uses cases in mind that were related to `Function` and `Method` fields, although users have found other ways to use it since then. BTW, I think these fields introduce too much complexity in the code and I'd like to remove/replace them ([#2039](https://github.com/marshmallow-code/marshmallow/issues/2039#issuecomment-1281932694)).

I didn't check but I don't think the performance overhead is significant. If it was, we could publish this in a `ContextSchema` subclass.

I'd like to go on with this.

- [ ] Decide whether or not the context should be a dict and if so provide an empty (ideally immutable) dict as default value
- [ ] Add tests showing how it works with hooks
- [ ] Update docs

tests/test_schema.py
lafrech: We could do this:

```py
def test_nested_on_bind_field_hook():
    class MySchema(Schema):
        class NestedSchema(Schema):
            bar = fields.Str()

                field_obj.metadata["fname"] = self.context["fname"]

        foo = fields.Nested(NestedSchema)

    with Context({"fname": "foobar"}):
        schema1 = MySchema()
        assert schema1.fields["foo"].schema.fields["bar"].metadata["fname"] == "foobar"

    with Context({"fname": "quxquux"}):
        schema2 = MySchema()
        assert schema2.fields["foo"].schema.fields["bar"].metadata["fname"] == "quxquux"
```

But it is not the kind of usage I'd like to promote.

lafrech:
We can provide an empty mapping as default:

```py
DEFAULT: MappingProxyType = MappingProxyType({})

CONTEXT = contextvars.ContextVar("context", default=DEFAULT)
```

Then, there is always a context, be it empty, so the test in `Function` field doesn't work (and the code will crash later if context is needed).

The point is do we want to distinguish no context and empty context?

It might be wiser not to trade semantics for ease of use and keep `None` as default (or don't even set `None` as default and use a try/catch instead but the perf impact might be negligible). This allows users to really set any value as context.

Since my latest commit, the context variable is only accessed directly in our code, users access it via schema/field properties, so we can wrap the access with whatever code we want, it doesn't have to be a one-liner.

src/marshmallow/context.py
```diff
@@ -0,0 +1,17 @@
...
+
+class Context(contextlib.AbstractContextManager):
+    def __init__(self, context):
+        self.context = context
+
+    def __enter__(self):
+        self.token = CONTEXT.set(self.context)
+
+    def __exit__(self, *args, **kwargs):
+        CONTEXT.reset(self.token)
```
sloria: suggestion: maybe can unify these into a single class?

```python
class Context(contextlib.AbstractContextManager):
    current_context: contextvars.ContextVar = contextvars.ContextVar("context")

    def __init__(self, context):
        self.context = context

    def __enter__(self):
        self.token = self.current_context.set(self.context)

    def __exit__(self, *args, **kwargs):
        self.current_context.reset(self.token)

    @classmethod
    def get(cls):
        return cls.current_context.get()

File: src/marshmallow/schema.py [hunk 5]
```diff
@@ -408,6 +406,10 @@ def dict_class(self) -> type[dict]:
         else:
             return dict
 
+    @property
+    def context(self) -> typing.Any:
```
sloria: question: is this property still necessary? when would it be used?

src/marshmallow/fields.py
```diff
@@ -230,6 +231,10 @@ def __repr__(self) -> str:
     def __deepcopy__(self, memo):
         return copy.copy(self)
 
+    @property
+    def context(self) -> typing.Any:
```
sloria: same question here: when would this be useful?

sloria (commented):
i'm on board with replacing the current marshmallow implementation of "context" with a `contextvars`-based API, so i'm good with moving forward with this feature 👍 so it's just little API decisions to make from this point.

> Decide whether or not the context should be a dict and if so provide an empty (ideally immutable) dict as default value

imo we should let the user decide which data structure to use for their context

> It might be wiser not to trade semantics for ease of use and keep None as default (or don't even set None as default and use a try/catch instead but the perf impact might be negligible). This allows users to really set any value as context.

might make sense to just not set a default. `Function` already fails loudly if the passed function expects `context` but it isn't set, so there's an opportunity to simplify the code if we let `ContextVar.get()` to raise the error.

src/marshmallow/context.py
sloria: thought: it might even make sense to make `current_context` private.

```python
class Context(contextlib.AbstractContextManager):
    __current_context: contextvars.ContextVar = contextvars.ContextVar("context")

File: src/marshmallow/schema.py [hunk 7]
```diff
@@ -512,13 +514,16 @@ def _serialize(self, obj: typing.Any, *, many: bool = False):
             ret[key] = value
         return ret
 
-    def dump(self, obj: typing.Any, *, many: bool | None = None):
+    def dump(
+        self, obj: typing.Any, *, many: bool | None = None, context: typing.Any = None
```
sloria: question: what's the rationale behind adding `context` to `dump` and `load`? i'm not sure the added API surface is worth the convenience.

lafrech: I think my initial intent was to offer a mechanism that was close to the existing one, using context vars and context manager under the hood.

As I wrote above, we could remove this and only expose the context manager.

And we could also remove `schema.context` and `field.context`. I kept those to be close to the existing code and also to offer a nicer API than `CONTEXT.get()`. Because in most cases (apart from the function provided to the `Function` field), the schema or field should be passed to the function where the context is used. But I admit that it does not make much sense to make this a schema or field method.

We can make `get` a method of `Context` as you suggest. It will still require an import in the file that defines the function using the context. No big deal, though.

Note that if we go this way, the only added value of this is to provide the context variable and manager, but technically, this could be left to the user, since it is not called anywhere in the core. I still think it is nice to provide, if only to offer a clear migration path.

I'll give it a go. Thanks for the feedback.

lafrech:
I applied said changes.

I removed the default. I don't mind `Function` field not catching the error about no context. In fact, this shouldn't deserve a `ValidationError`. The error lies in the code, not in the input.

However, I find it less convenient to use:

```py
                # With None as default
                if (context := Context.get()) is not None:
                    val *= context.get("factor", 1)

                # Without default
                try:
                    context = Context.get()
                except LookupError:
                    pass
                else:
                    val *= context.get("factor", 1)
```

The first form makes it nicer when using the context in a function that is meant to work either with or without.

I have never been using contexts. Maybe functions are meant to work only one way and it is up to the user to provide a context on each call, even a default one. Makes sense. If it is the way you see it, I'll leave it as is and just remove the part of the test that does what I copied above. I agree it means even less API surface.

I still need to add tests checking the use in hooks.

sloria:
> However, I find it less convenient to use...

i suppose we could allow optionally passing a default to `Context.get`

```python
from marshmallow import Context 


Context.get() # => LookupError
Context.get(None) # => None
```

src/marshmallow/context.py
```diff
@@ -0,0 +1,21 @@
+"""Objects related to serializtion/deserialization context"""
+
+import contextlib
+import contextvars
+
+
+class Context(contextlib.AbstractContextManager):
+    _current_context: contextvars.ContextVar = contextvars.ContextVar("context")
```
sloria: just came across this in the [python docs](https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar):

> Important: Context Variables should be created at the top module level and never in closures. [Context](https://docs.python.org/3/library/contextvars.html#contextvars.Context) objects hold strong references to context variables which prevents context variables from being properly garbage collected.

i'm not sure if this is an issue in this case, though. the `Context` class object should only get instantiated once in a program's lifetime 🤔

src/marshmallow/context.py
```diff
@@ -0,0 +1,21 @@
...
+        self.context = context
+
+    def __enter__(self):
+        self.token = self._current_context.set(self.context)
+
+    def __exit__(self, *args, **kwargs):
+        self._current_context.reset(self.token)
+
+    @classmethod
+    def get(cls):
```
sloria: thought: should we also expose API for setting the context? 

```python
from marshmallow import Context

@app.before_request
def bind_marshmallow_context():
    Context.set({"user_id": current_user.id})
```

lafrech:
Good idea. I tried that.

I had to create a singleton to distinguish `None` from `no value`. I clumsily copied `missing`.

Pandas does it differently using an `Enum`.

src/marshmallow/context.py
```diff
@@ -0,0 +1,25 @@
...
+        self.context = context
+
+    def __enter__(self):
+        self.token = self._current_context.set(self.context)
+
+    def __exit__(self, *args, **kwargs):
+        self._current_context.reset(self.token)
+
+    @classmethod
+    def get(cls, default=no_default):
```
sloria: suggestion: could use `Ellipsis`

```suggestion
    def get(cls, default=Ellipsis):
```

i encountered similar usage in structlog: https://github.com/hynek/structlog/blob/e51f34705d0e6c6551ca0915eded697dc99f0e57/src/structlog/contextvars.py#L133

sloria (commented):
> Note that if we go this way, the only added value of this is to provide the context variable and manager, but technically, this could be left to the user, since it is not called anywhere in the core. I still think it is nice to provide, if only to offer a clear migration path.

i see what you mean. i'm torn on this. on one hand, the current `context` implementation has rough edges that `contextvars` obviates. on the other hand, i don't want code ourselves in a corner with an underpowered API.

one incremental approach would be to deprecate the current `context` implementation for 4.0 with a warning that recommends using `contextvars` and leave it up to users to handle context on their own (we could even provide some guidance in the Upgrading Guide docs).

we'd postpone the addition of a convenience API for a future release

src/marshmallow/context.py
```diff
@@ -0,0 +1,25 @@
+"""Objects related to serializtion/deserialization context"""
+
+import contextlib
+import contextvars
+
+from marshmallow.utils import no_default
+
+
+class Context(contextlib.AbstractContextManager):
```
sloria: thought: should we also expose API for setting the context? 

```python
from marshmallow import Context

@app.before_request
def bind_marshmallow_context():
    Context.set({"user_id": current_user.id})
```

src/marshmallow/context.py
lafrech: Nice! I used the `...` form, which looks equivalent.

src/marshmallow/context.py
lafrech: Hmmm. I think this is meant to be used as a context manager, so that the context is reset on exit.

We'd need to reset in `@app.after_request` or something. And manage the request crash case. Things context managers do.

I'm not sure how I'd deal with this use case, though.

src/marshmallow/context.py
lafrech: I think I created it the way I did because that's how it was shown in the docs and it fitted the use cases I've met so far.

The implications are unclear to me but In doubt, we can use a module private attribute.

lafrech:
We did have a little adherence in `Function` field where `Context` is called. I just removed it, which is a nice cleanup. It is now up to the user to define a function that gets the context from `Context`. `Context` code is now independent from the core.

This feature is something I had in mind for a while so I'd be happy if it made it through, but that's not a good reason to merge.

A better reason is the cleanup of the existing feature. But in fact, I wouldn't mind removing the current feature (rather than deprecating) even if we only provide `Context` as a doc hint. The point is that context vars make the original feature not so useful.

I have no idea how widely the feature is used.

sloria:
ok, i'm good with removing the current context implementation for 4.0 and documenting how to use contextvars in the docs.

we could even add the `Context` implementation as provisional/experimental API (e.g. put it in `marshmallow.experimental.context`).

src/marshmallow/fields.py
```diff
@@ -1941,18 +1938,13 @@ def __init__(
...
+        return self.serialize_func(obj)
 
     def _deserialize(self, value, attr, data, **kwargs):
         if self.deserialize_func:
-            return self._call_or_raise(self.deserialize_func, value, attr)
+            return self.deserialize_func(value)
         return value
 
-    def _call_or_raise(self, func, value, attr):
-        if len(utils.get_func_args(func)) > 1:
```
sloria, line 2047: suggestion: now that this is removed, i think `marshmallow.utils.get_func_args` can also be removed.

src/marshmallow/context.py
sloria: i'm not opposed to it being at the module level, especially if that's the best practice

lafrech:
> we could even add the `Context` implementation as provisional/experimental API (e.g. put it in `marshmallow.experimental.context`).

I can do that. Is it any different than a docstring in the top of the file stating it is experimental? What would be the exact meaning? Expressing the fact that next major may ship a modified version or remove the feature altogether?

sloria:
putting it in `experimental` makes it clear to users that they're using unstable API even if they don't read the docs, e.g. if they discover it through autocomplete/introspection. it's a pattern i've seen in other libraries like [scikit-learn](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/experimental/__init__.py) and [strawberry](https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/experimental/__init__.py).

> What would be the exact meaning? Expressing the fact that next major may ship a modified version or remove the feature altogether?

i would treat it like pre-1.0 software, i.e. breaking changes could be made with minor releases. here's the warning in scikit-learn's docs

```rst
.. warning::

    The features and estimators that are experimental aren't subject to
    deprecation cycles. Use them at your own risks!
```

lafrech:
OK, I put it in `experimental`.

- [ ] Add tests demonstrating the use with hooks
- [ ] Changelog
- [ ] Documentation
- [ ] Dostrings (versionchanged for Schema, Function and Method)

src/marshmallow/experimental/context.py
```diff
@@ -0,0 +1,25 @@
...
+import typing
+
+_T = typing.TypeVar("_T")
+_CURRENT_CONTEXT: contextvars.ContextVar = contextvars.ContextVar("context")
+
+
+class Context(contextlib.AbstractContextManager, typing.Generic[_T]):
+    def __init__(self, context: _T) -> None:
+        self.context = context
+
```
sloria, line 45: nitpick(non-blocking): maybe initialize self.token in the ctor? 

```suggestion
self.token: contextvars.Token | None = None
```

just to follow the best practice of well-formed instances, and perhaps to allow stuff like

```python
user_context = Context({"user": user))

# ...
user_context_is_active = user_context.token is not None
```

CHANGELOG.rst
```diff
@@ -48,6 +48,29 @@ As a consequence of this change:
 
 Thanks :user:`ddelange` for the PR.
 
+- *Backwards-incompatible*: Remove schema ``context`` property. Passing a context
+  should be done using a context variable.
+  marshmallow 4.0 provides an experimental `Context <marshmallow.experimental.context.Context>`
+  manager class that can be used both to set and retrieve the context.
```
sloria: nit(non-blocking): add relevant issues here

docs/upgrading.rst
```diff
@@ -58,6 +58,49 @@ If you want to use anonymous functions, you can use this helper function.
         password = fields.String(validate=predicate(lambda x: x == "password"))
 
 
+Schema Context is Removed
```
sloria: nit:

```suggestion
Schema context is removed
```

docs/upgrading.rst
```diff
@@ -58,6 +58,49 @@ If you want to use anonymous functions, you can use this helper function.
...
+*************************
+
+The feature allowing to pass a context to the schema has been removed. Users should
+use context variables for that.
```
sloria: nit:

```suggestion
use `contextvars` for that.
```

so that it links to the relevant Python docs

src/marshmallow/experimental/context.py
```diff
@@ -0,0 +1,26 @@
...
+class Context(contextlib.AbstractContextManager, typing.Generic[_T]):
+    def __init__(self, context: _T) -> None:
+        self.context = context
+        self.token: contextvars.Token | None = None
+
+    def __enter__(self) -> None:
+        self.token = _CURRENT_CONTEXT.set(self.context)
+
+    def __exit__(self, *args, **kwargs) -> None:
+        _CURRENT_CONTEXT.reset(typing.cast(contextvars.Token, self.token))
```
sloria, line 50: you should be able to remove the cast if you check against None

```suggestion
        if self.token is not None:
            _CURRENT_CONTEXT.reset(self.token)
```

sloria (approved):
looks great. just a few small things. go ahead and merge after those

docs/upgrading.rst
sloria: erm, actually it looks like the casing is inconsistent in this file 🤦 other files seem to use title case, so let's just stick with that (so no change needed)

sloria: fwiw, the Python docs say

> In the Python documentation, the use of sentence case in section titles is preferable, but consistency within a unit is more important than following this rule. If you add a section to a chapter where most sections are in title case, you can either convert all titles to sentence case or use the dominant style in the new section title.

https://devguide.python.org/documentation/style-guide/#capitalization

we should prob just edit everything to be sentence case, but that can happen later

lafrech: OK. I thought we always used title case but didn't check.

I prefer sentence case. +1 for fixing this anytime.

src/marshmallow/experimental/context.py
lafrech: I figured the cast was better because it is optimized out at compile time (right ?).

The perf impact is probably peanuts but as a principle, it kinda sucks to add a useless instruction to silent a linter.

I admit that the cast is less eye-candy, so I'm open to the change.

sloria (approved):

src/marshmallow/experimental/context.py
sloria: that's fair. most of the time it's better to avoid casting for robustness. in this case, it can easily be reasoned that this line will never be reached unless `self.token` is set.

sloria:
looks great! 🚢 🇮🇹 go ahead and merge unless you have any further changes to make

docs/custom_fields.rst
```diff
@@ -95,40 +95,6 @@ Both :class:`Function <marshmallow.fields.Function>` and :class:`Method <marshma
     result = schema.load({"balance": "100.00"})
     result["balance"]  # => 100.0
 
-.. _adding-context:
-
-Adding Context to `Method` and `Function` Fields
```
sloria: thought(non-blocking): should we replace this with the equivalent using the new `Context` API?

docs/upgrading.rst
sloria: i've updated the rest of the docs to use sentence case, so good to change it here, too

sloria: went ahead and did this because i felt bad for causing so many merge conflicts 😅

docs/custom_fields.rst
sloria: #2727 addresses this

sloria (approved):
@lafrech no more changes on my end (i promise!). go ahead and merge if you have no other edits to make

lafrech:
OK, let's go.

It's never too late to further improve this anyway.

Glad we finally did this. Thanks.

PR #2708, opened 2025-01-02
Don't check for False from validators

### LINKED ISSUE #1775: Don't check for False return values from validators
Currently, validators are allowed to return `False` which results in `ValidationError` with a generic message.

https://github.com/marshmallow-code/marshmallow/blob/9ac3241b78b444033f901d51b431ac862d1e21cc/src/marshmallow/fields.py#L249-L251

I'm not convinced this behavior is that useful, and we could improve validation/deserialization performance by removing it. It also is a simpler API, avoiding confusion like in https://github.com/marshmallow-code/marshmallow/issues/1458

Slating this for 4.0 since it's a breaking change.

sloria:
close #1775

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,8 +12,9 @@ As a consequence of this change:
   - YYYY-MM-DD is now accepted as a datetime and deserialized as naive 00:00 AM.
   - `from_iso_date`, `from_iso_time` and `from_iso_datetime` are removed from `marshmallow.utils`
 
+- *Backwards-incompatible*: Custom validators must raise a `ValidationError <marshmallow.exceptions.ValidationError>` for invalid values.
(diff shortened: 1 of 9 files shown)


lafrech (approved):
Good.

It turned up to be a lot of work for a limited API surface decrease, and the fact you had to come up with that `predicate` function could make you wonder if the use case was not legit after all, but I stand with you on this. The use of so many lambdas in the tests is due to tests being contrived examples. I believe in practice, people use real functions (I have a bias here as I tend to think that a function is always easier to read than a lambda) and it is better API to require a `ValidationError`.

sloria:
yeah, in practice it's much better to use a function so that you can have helpful error messages. as you mentined, the lambdas in the tests are only there out of convenience.

PR #2711, opened 2025-01-03
Add types to Field subclass kwargs

### LINKED ISSUE #2285: Static typing of Field subclass kwargs
Today, the subclasses of Field have their constructor kwargs untyped, which means that static typecheckers like mypy/pyright treat them as Unknown. This can lead to bugs or typos like passing `fields.Integer(validates=...)` instead of `fields.Integer(validate=...)`. In addition, editors like VSCode or neovim can't autocomplete or show you the types of these kwargs:
Screenshot 2024-07-12 at 4 06 00 PM

In https://github.com/microsoft/pyright/issues/6344#issuecomment-2226519311, it was recommended to use the TypedDict and Unpack features to statically type kwargs. For example:

```py
class BaseFieldKwargs(TypedDict, total=False):
    load_default: typing.Any
    missing: typing.Any
    dump_default: typing.Any
    default: typing.Any
    data_key: str | None
    attribute: str | None
    validate: (
        None
        | typing.Callable[[typing.Any], typing.Any]
        | typing.Iterable[typing.Callable[[typing.Any], typing.Any]]
    )
    ...

class Integer(Number):
    ...
    def __init__(self, *, strict: bool = False, **kwargs: Unpack[BaseFieldKwargs]):
        ...
```

(The presence of a catch-all `additional_metadata` makes it a little tricky. We might need to union it with a generic `Dict[str, typing.Any]` until that parameter is deprecated.)

Thoughts on adding this to improve static typechecking?

sloria:
close #2285

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,13 @@ Changelog
 
 See :ref:`upgrading_4_0` for a guide on updating your code.
 
-- *Backwards-incompatible*: Remove implicit field creation, i.e. using the ``fields`` or ``additional`` class Meta options with undeclared fields (:issue:`1356`).
+Features:
+
+- Typing: Add types to all `Field <marshmallow.fields.Field>` subclass kwargs (:issue:`2285`).
+  Thanks :user:`navignaw` for the suggestion.
+
+Other changes:
+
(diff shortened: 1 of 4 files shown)


lafrech (approved):
Good for me.

IIUC, the FIXMEs are not blockers, just hints.

sloria:
> IIUC, the FIXMEs are not blockers, just hints.

that's correct

PR #2716, opened 2025-01-03
Update upgrading guide

sloria:


sloria:
@lafrech i think this PR and #2707 are the last ones to merge before 4.0 is release-ready 🙌 

we should also ensure that the other marshmallow libs are compatible with this 4.0 before we release

sloria:
we already test the other libs against the marshmallow `dev` branch: https://github.com/marshmallow-code/apispec/blob/6583194325b5cd2033537d34d58cbf6ce2b99d2b/.github/workflows/build-release.yml#L30-L35

so i think the correct order of operations is

1. merge this PR and #2707 
2. create `3.x-line` branch off of `dev`. enable docs builds in readthedocs
3. merge `4.0` branch into `dev` and remove `4.0`
4. ensure libraries pass against marshmallowdev. make any necessary releases.
5. release 4.0

sloria:
@lafrech @deckar01 how long should we support 3.x for? should we continue merging features into 3.x, and if so, for how long? how long should we continue with bugfixes, i.e. when should be its EOL?

lafrech:
Compared to marshmallow 3.0, the migration should be very smooth for the user. But libs using marshmallow may take a little time to migrate.

I propose to feature freeze 3.x right now and merge bugfixes for a few months.

lafrech (approved):
I don't see anything in the changelog that should appear in the upgrading guide but does not.

Issue #2039, opened 2022-08-25
[RFC] Don't pass attr, obj, data to fields (de)serialize methods


lafrech:
From a quick look, those are not used by existing fields. I suspect this might be or have been useful in specific fields such as `Method`.

I can't investigate this thoroughly right now and anyway that would be a breaking change, but opening this here for discussion.

For separation of concerns, (de)ser methods shouldn't need the parent obj/data and field name.

Also, this means that one needs an object and a field name to serialize some data with a field, which may not always be the case. They may pass `None` but then they don't respect the API (since it won't work with a field (de)ser method that actually relies on those). I faced this specific problem with `EnumValue` field, when building `self.choices`, see #2017.

lafrech:
I gave it a quick look.

`attr` and `obj` are needed in `serialize` to get the value from the object.

In `_serialize`, `deserialize` and `_deserialize`, they seem to be only used by `Function` / `Method` fields, so removing those would allow us to simplify the implementation.

I guess some users rely on those fields. OTOH, if they didn't exist, I doubt we'd modify (de)ser methods to pass `attr` and `obj` just to add them.

I don't have a simple idea to replace those fields, though. pre_load/dump methods don't really answer because we don't want to modify the object. Unless we add some sort of proxy around the object but this sounds complicated.

lafrech:
An option could be to remove `Function` and `Method` fields and add an argument or two to `Field` to allow the user to pass a custom function/method to get the value from the object. This function/method could be called from `get_value`. In fact, it would be a `get_value` override.

So it wouldn't be a field by itself but rather a customization of an existing field. There would be a benefit in terms of auto-documentation, since the type of the value would be known (the user may choose to use `Field` for a type-free value).

We could add this feature and deprecate `Function` and `Method` (as well as the use of `attr` and `obj`) in marshmallow 3.x.

-----------------------------

I was thinking of opening another issue about moving `get_value` logic to the schema. The schema would just need to access the `attribute` attribute of the field. If we do what I describe above, the schema would also need to call the custom accessor (the function/method). It's not obvious to me where the logic should be, but I still find it strange to pass an obj, an accessor function, etc.

lafrech:
We could add `accessor` parameter to `Field`.

```py
class Field(FieldABC):
   def __init__(..., accessor,...):
        ...
        self.get_value = accessor
        ...
```

Allowing the user to do that:

```py

def ser_function(...):
    ...

class MySchema(Schema):

[...]

PR #2720, opened 2025-01-04
Remove `Field._CHECK_ATTRIBUTE`

lafrech:
This is a first step towards #1046/#2039 (those issues could probably be merged).

It introduces an asymmetry because there is no accessor concept when deserializing.

Ultimately, I'd like to remove those fields. As I suggest in #2039, the accessor method could be passed to any field, so as to combine the capability to customize the accessor (like in `Function`/`Method`) and the field feature related to its type (`String`, `Integer`, etc.).

I don't mind postponing this to 5.0 if we're not confident this is the right direction to take and we want to have the full rework complete.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -153,11 +153,6 @@ class Field(FieldABC):
         Replace ``missing`` and ``default`` parameters with ``load_default`` and ``dump_default``.
     """
 
-    # Some fields, such as Method fields and Function fields, are not expected
-    #  to exist as attributes on the objects to serialize. Set this to False
-    #  for those fields
-    _CHECK_ATTRIBUTE = True
-
     #: Default error messages for various kinds of errors. The keys in this dictionary
     #: are passed to `Field.make_error`. The values are error messages passed to
     #: :exc:`marshmallow.exceptions.ValidationError`.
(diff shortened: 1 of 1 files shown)


sloria (approved):
seems like a good incremental step to me. i wouldn't even consider this backwards-incompatible since `_CHECK_ATTRIBUTE` was private

sloria:
@lafrech feel free to merge if there's no other changes you plan to make

lafrech:
I think https://github.com/marshmallow-code/marshmallow/issues/2039#issuecomment-2571769052 might be the way to go.

Not sure I'll carve the time to work on it in the next days, although if my intuition is correct, it might be a nice and complete solution and it may not require so much work.

Depends on how eager we are to release. No need to rush IMHO.

sloria:
https://github.com/marshmallow-code/marshmallow/issues/2039#issuecomment-2571769052 sounds like a good plan. no rush on releasing, but i would like to prevent 4.0 from having too many large breaking changes (this is not to discourage implementation of #2039--it's mostly my fault for the 4.0 scope creep 🙃 ).

lafrech:
Let's hold this this until #2039. I doesn't bring much by itself and might catch people in the wild with fancy setups, like `Method` field in `List` field or whatever.

If we go on with #2039, `serialize` signature changes so all custom fields are broken and most libs must be updated and all. No problem if we postpone it to make the transition more incremental.

sloria:
sounds good. i think postponing #2039 is a good idea

sloria:
@lafrech should we close this PR for now?

lafrech:
Or make it a draft but I don't think we can do that.

I just moved it to 5.0. Keeping it to keep that commit somewhere. We may close it, no big deal.

sloria:
woops, this auto-closed when i merged 4.0 into dev and deleted 4.0. We can reopen this when the time comes

lafrech:
Yeah, no pb, really.

Le 6 janvier 2025 18:33:12 GMT+01:00, Steven Loria ***@***.***> a écrit :
>woops, this auto-closed when i merged 4.0 into dev and deleted 4.0. We can reopen this when the time comes
>

-- 
Jérôme

PR #2722, opened 2025-01-04
Remove FieldABC and the base module

sloria:
refactors the code so that `FieldABC` and the entire `base` module are no longer needed.

note: this is technically backwards incompatible because `resolve_field_instance` and `FieldInstanceResolutionError` are privatized but I wouldn't expect anyone to actually be using those


sloria:
gonna go ahead and merge this one because i'm reasonably confident in it and it will unblock some further refactoring i want to do

PR #2723, opened 2025-01-04
Various typing improvements to fields module

sloria:


emorikawa:
@sloria heads up that this is likely a major breaking change. This took us down today because it broke Airflow's APIs: https://github.com/apache/airflow/pull/45442

Since this does schema validation, this effectively caused all of my Airflow API requests to suddenly start 400'ing on my next deploy (deploys re pip-install things when the container is rebuilt).

I have business critical functions that depend on programatically launching DAGs. All of those broke suddenly.

lafrech:
The real change is in 4.0.

3.24 adds deprecation warnings.

(The change did break my build, but AFAIU, this was due to an issue in marshmallow-sqlalchemy that is now fixed.)

PR #2725, opened 2025-01-04
Make Field a generic type; refactor inheritance hierarchies

sloria:
the main changes here are:

- Make `Field` a generic type with a type argument for the internal value type. This vastly improves the typing for all the fields. as a consequence, `Field` is no longer usable within Schemas
- Other "base" fields incl. `Number` and `Mapping` are also no longer usable in schemas.
- Change the inheritance hierarchy of the temporal fields to make more sense

```
Before:
                     DateTime
                  /    |    |    \
                /      |     |     \
            Time    Date  NaiveDateTime  AwareDateTime


After:
              _TemporalField
                /    |    \
              /      |      \
           Time    Date    DateTime
                              /  \
                            /      \
                  NaiveDateTime  AwareDateTime

```

- `Boolean` no longer serializes non-booleans. this makes it more consistent with the other fields (and more performant)

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -682,7 +682,7 @@ class Pluck(Nested):
         return self._load(value, partial=partial)
 
 
-class List(Field[list[_InternalType | None]]):
+class List(Field[list[typing.Optional[_InternalType]]]):
     """A list field, composed with another `Field` class or
     instance.
 
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import abc
 import collections
 import copy
 import datetime as dt
@@ -84,6 +85,8 @@ __all__ = [
     "Pluck",
 ]
 
+_InternalType = typing.TypeVar("_InternalType")
+
 
 class _BaseFieldKwargs(typing.TypedDict, total=False):
     load_default: typing.Any
@@ -118,7 +121,7 @@ def _resolve_field_instance(cls_or_instance: Field | type[Field]) -> Field:
         return cls_or_instance
 
 
-class Field:
+class Field(typing.Generic[_InternalType]):
     """Basic field from which other fields should extend. It applies no
     formatting by default, and should only be used in cases where
     data does not need to be formatted before being serialized or deserialized.
@@ -258,7 +261,7 @@ class Field:
             typing.Callable[[typing.Any, str, typing.Any], typing.Any] | None
         ) = None,
         default: typing.Any = missing_,
-    ):
+    ) -> _InternalType:
         """Return the value for a given key from an object.
 
         :param object obj: The object to get the value from.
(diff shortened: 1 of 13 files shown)


src/marshmallow/fields.py
```diff
@@ -855,16 +868,14 @@ def _deserialize(self, value, attr, data, **kwargs) -> typing.Any:
             raise self.make_error("invalid_utf8") from error
 
 
-class UUID(String):
+class UUID(Field[uuid.UUID]):
```
sloria, line 890: UUID inheriting from String was always a suboptimal choice, as was mentioned in https://github.com/marshmallow-code/marshmallow/pull/1485#issuecomment-579922293

> As I can see, fields inheriting from String, are in general deserialized from strings to strings with additional input value format validation. We can see it in the Url and Email fields cases. Exception from this rule is the UUID field, which deserializes from string into UUID object representation. Therefore, it is not clear for me why UUID inherits from String and not Field, like in the case of DateTime or TimeDelta which also deserializes to object representation.

src/marshmallow/fields.py
```diff
@@ -334,13 +336,33 @@ def serialize(
...
-    ):
+    ) -> _InternalType: ...
+
+    def deserialize(
+        self,
+        value: typing.Any,
+        attr: str | None = None,
+        data: typing.Mapping[str, typing.Any] | None = None,
+        **kwargs,
+    ) -> _InternalType | None:
```
lafrech, line 365: I'm not familiar with overload but from the example in [the docs](https://docs.python.org/3/library/typing.html#typing.overload), I'm wondering if the typing in the actual implementation is redundant.

lafrech (approved):
It is nice to improve typing.

My only concern would be that it might make it more difficult to work on the code (because of more typing info to maintain, especially if using advanced typing features), but that's always the case when working with typing, I just need to learn how to use type hints. It sometimes looks so cryptic...

src/marshmallow/fields.py
sloria: i'm going to leave the explicit typing because when I remove it the the inferred types aren't quite right:

CleanShot 2025-01-05 at 14 07 38@2x

notice `attr` and `data` are `Any`

sloria:
yeah, i agree it does add some complexity but i think the DX improvement on the user's end is worth it in this case.

lafrech:
While adapting flask-smorest to 4.0, I realized that I don't get any error when using `Field` instead of `Raw`. In fact, I replaced `Raw` with `Field` in all marshmallow tests and everything passes. Is it just a matter or typing?

sloria:
@lafrech yes, type checkers will error when using Field.

```python


class MySchema(Schema):
    # mypy error: Need type annotation for "foo"
    foo = fields.Field()
```

you _could_ address the error by passing a type argument

```python
class MySchema(Schema):
    foo = fields.Field[str]()
```

but at that point you might as well use `fields.String`.

lafrech:
OK this explains why tests wouldn't fail.
Thanks.

PR #2730, opened 2025-01-05
Allow fields to accept their internal value types

### LINKED ISSUE #1415: Feature request: Validate should accept deserialized values as valid for all fields

sloria:
resolves #1415

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1243,6 +1243,9 @@ class _TemporalField(Field[_D], metaclass=abc.ABCMeta):
         return value.strftime(data_format)
 
     def _deserialize(self, value, attr, data, **kwargs) -> _D:
+        internal_type: type[_D] = getattr(dt, self.OBJ_TYPE)
+        if isinstance(value, internal_type):
+            return value
         data_format = self.format or self.DEFAULT_FORMAT
         func = self.DESERIALIZATION_FUNCS.get(data_format)
         try:
@@ -1516,6 +1519,8 @@ class TimeDelta(Field[dt.timedelta]):
         return microseconds / microseconds_per_unit
 
     def _deserialize(self, value, attr, data, **kwargs) -> dt.timedelta:
+        if isinstance(value, dt.timedelta):
+            return value
         try:
             value = float(value)
         except (TypeError, ValueError) as error:
@@ -1863,7 +1868,7 @@ class Enum(Field[_EnumType]):
         or Field class or instance to use to (de)serialize by value. Defaults to False.
 
     If `by_value` is `False` (default), enum members are (de)serialized by symbol (name).
-    If it is `True`, they are (de)serialized by value using :class:`Field`.
+    If it is `True`, they are (de)serialized by value using `marshmallow.fields.Raw`.
     If it is a field instance or class, they are (de)serialized by value using this field.
 
     .. versionadded:: 3.18.0
@@ -1893,7 +1898,7 @@ class Enum(Field[_EnumType]):
         # Serialization by value
         else:
             if by_value is True:
-                self.field = Field()
+                self.field = Raw()
             else:
                 try:
                     self.field = _resolve_field_instance(by_value)
@@ -1918,6 +1923,8 @@ class Enum(Field[_EnumType]):
         return self.field._serialize(val, attr, obj, **kwargs)
 
     def _deserialize(self, value, attr, data, **kwargs) -> _EnumType:
+        if isinstance(value, self.enum):
+            return value
         val = self.field._deserialize(value, attr, data, **kwargs)
         if self.by_value:
             try:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,8 @@ Features:
 
 - Typing: Add types to all `Field <marshmallow.fields.Field>` subclass kwargs (:issue:`2285`).
   Thanks :user:`navignaw` for the suggestion.
+- All fields accept their internal value type as valid input (:issue:`1415`).
+  Thanks :user:`bitdancer` for the suggestion.
 
 Other changes:
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,7 +10,9 @@ Features:
 
 - Typing: Add types to all `Field <marshmallow.fields.Field>` subclass kwargs (:issue:`2285`).
   Thanks :user:`navignaw` for the suggestion.
-- All fields accept their internal value type as valid input (:issue:`1415`).
+- `DateTime <marshmallow.fields.DateTime>`, `Date <marshmallow.fields.Date>`, `Time <marshmallow.fields.Time>`,
+  `TimeDelta <marshmallow.fields.TimeDelta>`, and `Enum <marshmallow.fields.Enum>`
+  accept their internal value types as valid input (:issue:`1415`).
   Thanks :user:`bitdancer` for the suggestion.
 
 Other changes:
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -579,7 +579,7 @@ class TestFieldDeserialization:
     @pytest.mark.parametrize("fmt", ["timestamp", "timestamp_ms"])
     @pytest.mark.parametrize(
         "in_value",
-        ["", "!@#", -1, dt.datetime(2013, 11, 10, 1, 23, 45)],
+        ["", "!@#", -1],
     )
     def test_invalid_timestamp_field_deserialization(self, fmt, in_value):
         field = fields.DateTime(format=fmt)
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -1462,6 +1462,43 @@ class TestFieldDeserialization:
             with pytest.raises(ValidationError, match="Invalid value."):
                 field.deserialize("invalid")
 
+    @pytest.mark.parametrize(
+        ("field", "value"),
+        (
+            pytest.param(fields.List(fields.String()), ["foo", "bar"], id="List"),
+            pytest.param(
+                fields.Tuple((fields.String(), fields.Integer())),
+                ("foo", 42),
+                id="Tuple",
+            ),
+            pytest.param(fields.String(), "valid", id="String"),
+            pytest.param(fields.UUID(), uuid.uuid4(), id="UUID"),
+            pytest.param(fields.Integer(), 42, id="Integer"),
+            pytest.param(fields.Float(), 42.3, id="Float"),
+            pytest.param(fields.Decimal(), decimal.Decimal("42.3"), id="Decimal"),
+            pytest.param(fields.Boolean(), True, id="Boolean"),
+            pytest.param(fields.DateTime(), dt.datetime(2014, 8, 21), id="DateTime"),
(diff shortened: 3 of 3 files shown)


sloria:
looks like the date and time related fields and `Enum` are the only ones that need to be updated

```
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[List] PASSED                                               [  5%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[Tuple] PASSED                                              [ 10%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[String] PASSED                                             [ 15%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[UUID] PASSED                                               [ 21%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[Integer] PASSED                                            [ 26%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[Float] PASSED                                              [ 31%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[Decimal] PASSED                                            [ 36%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[Boolean] PASSED                                            [ 42%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[DateTime] FAILED                                           [ 47%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[Time] FAILED                                               [ 52%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[Date] FAILED                                               [ 57%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[TimeDelta] FAILED                                          [ 63%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[Dict] PASSED                                               [ 68%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[Url] PASSED                                                [ 73%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[Email] PASSED                                              [ 78%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[IP] PASSED                                                 [ 84%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[IPv4] PASSED                                               [ 89%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[IPv6] PASSED                                               [ 94%]
tests/test_deserialization.py::TestFieldDeserialization::test_fields_accept_internal_types[Enum] FAILED                                               [100%]
```

src/marshmallow/fields.py
```diff
@@ -1243,6 +1243,9 @@ def _serialize(self, value: _D | None, attr, obj, **kwargs) -> str | float | Non
         return value.strftime(data_format)
 
     def _deserialize(self, value, attr, data, **kwargs) -> _D:
+        internal_type: type[_D] = getattr(dt, self.OBJ_TYPE)
```
sloria, line 1246: could also add a separate class attribute, like `INTERNAL_TYPE`. i don't have a strong feeling either way

src/marshmallow/fields.py
```diff
@@ -1243,6 +1243,9 @@ def _serialize(self, value: _D | None, attr, obj, **kwargs) -> str | float | Non
...
+        if isinstance(value, internal_type):
+            return value
```
lafrech, line 1248: This will let `datetime`s be accepted as `date`s because `datetime` is a `date` subclass (https://github.com/marshmallow-code/marshmallow/issues/1415#issuecomment-684948224).

This may not be an issue, since as a subclass it behaves like a `date`, but it may be surprising if the user then serialize the object and doesn't expect to see hours and minutes. At least this what I remember from when I encountered this in umongo.

src/marshmallow/fields.py
lafrech: The perf impact must be minimal. We can keep this.

lafrech (commented):
Good. It works with minimal changes.

I'm wondering if the fact that it works for stock fields means it generalizes well. I'm pretty sure one may come up with custom fields for which it won't work.

Imagine a field that transforms a string into another string, e.g. a CeasarCipher field. You can't tell in which form the input is.

I'm interested in @deckar01's philosophical point of view on this.

src/marshmallow/fields.py
sloria: i think that is OK and well within an acceptable range of permissibility (along the lines of how `String` accepts `bytes`)

sloria:
>I'm wondering if the fact that it works for stock fields means it generalizes well. I'm pretty sure one may come up with custom fields for which it won't work.
> Imagine a field that transforms a string into another string, e.g. a CeasarCipher field. You can't tell in which form the input is. 

@lafrech i'm not fully following. is this just an observation, or are you concerned about a possible drawback of the `isinstance(value, InternalType)` approach here?

the `CaesarCipher` case is similar to `URL` and `Email`, right? extra validation logic is required upon deserialization

lafrech:
My point is that, in stock fields, either deserialization is idempotent and there's no problem going through deserialization again as it is just a validation (string, URL,...), either it changes the type (these are the cases this PR fixes) and you skip deserialization based on the type. But we could imagine fields that are not deserialization-idempotent, like the CeasarCypher field (if we rotate an already rotated string, it won't work) for which we couldn't tell from the value type whether or not it was already deserialized (because in this case it is always a string).

It's late. Maybe I'm just confused.

sloria:
transforming a string to another format is a valid use case and indeed there are cases where the field can't know which form is it receiving. i'm missing the connection with this PR, though. is there a blocker i'm not seeing?

lafrech:
What I mean is that this PR "only" modifies existing fields so that they accept their internal type but this is not a global change, just a per-field change, and in fact there is no guarantee that this may be achievable with any new or custom field.

In other words, this is not "Ensure all fields can always accept their internal types" but rather "Allow all currently known field to accept their internal type".

My point is that this may leave users with the feeling that deserialization is always sort of idempotent (not really, because of the attribute/data_key parameters, processors, etc.), and frameworks could be tempted to build upon this assumption, then realize that it may be broken in some cases (custom fields).

I believe this initial issue and [your comment](https://github.com/marshmallow-code/marshmallow/issues/1415#issuecomment-533185603) aimed at a generic solution, but the solution based on the type as drafted in your comment and implemented where needed in this PR does not generalize.

Or maybe there is something I don't get.

sloria:
oh, i understand now, thanks for clarifying. you're correct: this isn't a global solution and users are still left to handle this case for their custom fields.

i hesitate to implement a generalized solution to this because

1. this problem only affects fields that deserialize from a basic type (str, int, etc.) to a complex type (`UUID`, `IP`, date and time fields, etc.). fields that deserialize from a basic type to another basic type would end up with an unnecessary type check (-performance)
2. there may be use cases where the user don't want to accept the internal type as input (i'm less confident in this reason)
3. there isn't one clear "best" way to handle this, i.e. `isinstance` doesn't necessarily work in all use cases. as you pointed out, `isinstance` allows `Date` to accept `datetimes`, which is OK here, but there may be other use cases where you wouldn't want to allow subclasses
4. the user burden is minimal

lafrech:
> oh, i understand now, thanks for clarifying. you're correct: this isn't a global solution and users are still left to handle this case for their custom fields.
> the user burden is minimal

Well, I'm afraid it is worse than this. There are cases where it is just not doable at all. E.g. `CeasarCypher`. How would the user implement that if there is no way for the deserialization function to know if input data is in serialized or deserialized form?

My point is not that the user has a bit of work to do to make custom fields compliant. In fact, I stand with you on the compromise. It is that sometimes (hypothetical corner cases) it may not be feasible. It just happens to work on all fields we currently ship.

sloria:
i understand. indeed we can't guarantee that custom fields will be always be able to accept their internal types. that said, i don't think that was the intent of the request in #1415, which was specifically about accepting datetimes.

lafrech:
It was expressed in generic terms, with a focus on the fields that were problematic (those you "fix" in this PR). The original issue as I understand it is not really solved. More like won't fix / can't fix but works in most cases anyway (after this PR is merged).

I agree it would be a shame to not do this if it works in 98% cases and we can't to anything for the 2% hypothetical edge cases anyway. This is an improvement. Cost is minimal and it helps in some cases. Your changelog edit makes it clear.

Go for it.

sloria:
> I agree it would be a shame to not do this if it works in 98% cases and we can't to anything for the 2% hypothetical edge cases anyway. This is an improvement. Cost is minimal and it helps in some cases. 

my thinking exactly. this PR is strictly better than the status quo, even though it doesn't solve for all user cases.

this is the last PR i wanted to get in for 4.0. i've made a [PR](https://github.com/marshmallow-code/marshmallow/pull/2732) to add deprecation warnings for changes we've already made in 4.0. i'll merge that, cut a 3.x release, then merge 4.0 into `dev`. i'll hold off on releasing 4.0 for a few days.

lafrech:
Great! That was a nice sprint!

PR #2732, opened 2025-01-06
Deprecation warnings for marshmallow 4 changes

sloria:
raise warnings to help users migrate to marshmallow 4

- using base field classes: `Field`, `Number`, and `Mapping`
- returning `False` from validators

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -29,7 +29,10 @@ from marshmallow.utils import (
     missing as missing_,
 )
 from marshmallow.validate import And, Length
-from marshmallow.warnings import RemovedInMarshmallow4Warning
+from marshmallow.warnings import (
+    ChangedInMarshmallow4Warning,
+    RemovedInMarshmallow4Warning,
+)
 
 if typing.TYPE_CHECKING:
     from marshmallow.schema import Schema, SchemaMeta
@@ -135,6 +138,16 @@ class Field(FieldABC):
         "null": "Field may not be null.",
         "validator_failed": "Invalid value.",
     }
+
+    def __new__(cls, *args, **kwargs):
+        if cls is Field:
+            warnings.warn(
+                "`Field` should not be instantiated. Use `fields.Raw` or  "
+                "another field subclass instead.",
+                ChangedInMarshmallow4Warning,
+                stacklevel=2,
+            )
+        return super().__new__(cls)
 
     def __init__(
         self,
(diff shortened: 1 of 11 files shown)


src/marshmallow/fields.py
```diff
@@ -136,6 +139,16 @@ class Field(FieldABC):
...
 
+    def __new__(cls, *args, **kwargs):
+        if cls is Field:
+            warnings.warn(
+                "`Field` should not be instantiated. Use `fields.Raw` or  "
+                "another field subclass instead.",
+                ChangedInMarshmallow4Warning,
+                stacklevel=2,
+            )
+        return super().__new__(cls)
```
lafrech, line 150: @sloria we just had an issue when running our app in a fresh virtualenv with latest marshmallow 3 versions. I'm still trying to figure out what the problem is.


```py
class UserSchema(AutoSchema):
    class Meta(AutoSchema.Meta):
        model = User
        exclude = ("_is_active", "_is_admin")

    id = msa.auto_field(dump_only=True)
    name = msa.auto_field(validate=ma.validate.Length(1, 80))
    email = ma.fields.Email(required=True)
    password = msa.auto_field(validate=ma.validate.Length(1, 80), load_only=True)
    is_admin = ma.fields.Boolean(dump_only=True)
    is_active = ma.fields.Boolean(dump_only=True)
```

`load_only` and `validate` are somehow ignored in the password field.

```
<fields.String(dump_default=<marshmallow.missing>, attribute=None, validate=[<Length(min=None, max=200, equal=None, error=None)>], required=True, load_only=False, dump_only=False, load_default=<marshmallow.missing>, allow_none=False, error_messages={'required': 'Missing data for required field.', 'null': 'Field may not be null.', 'validator_failed': 'Invalid value.', 'invalid': 'Not a valid string.', 'invalid_utf8': 'Not a valid utf-8 string.'})>
```

`git bisect` pointed me here and indeed, commenting the whole `def __new__` block solves it.

It seems to be related with `auto_field` because it works if I use a "normal" field.

```py
    password = ma.fields.String(validate=ma.validate.Length(1, 80), load_only=True)
```

```
<fields.String(dump_default=<marshmallow.missing>, attribute=None, validate=<Length(min=1, max=80, equal=None, error=None)>, required=False, load_only=True, dump_only=False, load_default=<marshmallow.missing>, allow_none=False, error_messages={'required': 'Missing data for required field.', 'null': 'Field may not be null.', 'validator_failed': 'Invalid value.', 'invalid': 'Not a valid string.', 'invalid_utf8': 'Not a valid utf-8 string.'})>
```


I don't think we're doing anything fancy in our code. OTOH, I think the `auto_field(..., load_only)` case is tested in marshmallow-sqlalchemy, so I really don't know where the issue may come from.

I'll continue to investigate tomorrow but I thought you might have an idea.

sloria: @lafrech are you using the latest marshmallow-sqlalchemy==1.1.1? I released a fix yesterday that affects usage of `auto_field` with marshmallow>=3.24

in case you're curious, this was the fix: https://github.com/marshmallow-code/marshmallow-sqlalchemy/pull/628/files#diff-ac1f3e987a72b794c7c4318fcbd3bb325974b73fa46c3c511fdba347f60948acR60

lafrech: Oh, you're right. marshmallow-sqlalchemy was locked to <0.30.

I performed a pip install using the eager update strategy, I thought everything would be updated regardless.

Thanks. Sorry about the noise. It works indeed.

PR #2738, opened 2025-01-08
Context.__enter__ returns self

sloria:
it's a common pattern with context managers to return `self` in `__enter__` to allow method chaining.

in particular, it nicely allows the value of `get()` to be correctly typed

```python
from marshmallow.experimental.context import Context

with Context[bool](True) as context:
    # Don't need to do Context[bool].get().
    # value is inferred to be bool.
    value = context.get()
```

### DIFF
--- a/src/marshmallow/experimental/context.py
+++ b/src/marshmallow/experimental/context.py
@@ -28,6 +28,8 @@ Example usage:
         # {'name_suffixed': 'foobar'}
 """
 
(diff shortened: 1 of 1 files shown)


lafrech (approved):
Absolutely. I think this was what I had in mind but I forgot to do it. And since that use case wasn't tested because in practice the value is accessed in an inner function, I didn't notice.

PR #2766, opened 2025-01-12
Remove utils.isoformat, to_iso_time, to_iso_date

lafrech:
Set the standard lib method directly in fields.py. We did the same for `from_isoformat`.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1274,8 +1274,8 @@ class DateTime(_TemporalField[dt.datetime]):
     """
 
     SERIALIZATION_FUNCS: dict[str, typing.Callable[[dt.datetime], str | float]] = {
-        "iso": utils.isoformat,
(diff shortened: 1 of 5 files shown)


src/marshmallow/fields.py
```diff
@@ -1387,7 +1387,10 @@ class Time(_TemporalField[dt.time]):
     :param kwargs: The same keyword arguments that :class:`Field` receives.
     """
 
-    SERIALIZATION_FUNCS = {"iso": utils.to_iso_time, "iso8601": utils.to_iso_time}
+    SERIALIZATION_FUNCS = {
+        "iso": dt.time.isoformat,
+        "iso8601": dt.time.isoformat,
+    }
```
lafrech, line 1393: Could fit in one line but I find it more readable this way.

sloria (approved):

PR #2767, opened 2025-01-12
Remove utils.from_rfc, rfcformat

lafrech:
Same as #2766.

Removed tests are covered in serialization/deserialization tests.


sloria (approved):
👍 👍 it's nice that we can use the stdlib for all datetime parsing and formatting now

PR #2768, opened 2025-01-12
Remove unused utils.is_keyed_tuple

lafrech:
This doesn't seem to be used anymore. Should we keep it?


sloria (approved):
Good catch! no need to keep dead code

PR #2771, opened 2025-01-13
Validate unknown with type checking rather than at runtime

sloria:
Add type annotations such that invalid values for `unknown` raise an error in type checkers like mypy. Removes the run-time checking added in https://github.com/marshmallow-code/marshmallow/pull/1989

For each of the underlined cases above, mypy gives an error like this:

```
Argument "unknown" to "MySchema" has incompatible type "Literal['bad']"; expected "Literal['exclude', 'include', 'raise'] | None"
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -95,6 +95,8 @@ Thanks :user:`ddelange` for the PR.
 - Methods decorated with `marshmallow.pre_load`, `marshmallow.post_load`, `marshmallow.validates_schema`,
   receive ``unknown`` as a keyword argument (:pr:`1632`).
   Thanks :user:`jforand` for the PR.
+- Passing invalid values for ``unknown`` will cause an error in type checkers.
+  Runtime checking is removed (:pr:`2771`).
 
 Deprecations/Removals:
 
@@ -130,8 +132,8 @@ Documentation:
 
 Deprecations:
 
-- The ``ordered`` `class Meta <marshmallow.Schema.Meta>` option is deprecated (:issue:`2146`, :pr:`2762`). 
-  Field order is already preserved by default. Set `marshmallow.Schema.dict_class` to `collections.OrderedDict` 
+- The ``ordered`` `class Meta <marshmallow.Schema.Meta>` option is deprecated (:issue:`2146`, :pr:`2762`).
+  Field order is already preserved by default. Set `marshmallow.Schema.dict_class` to `collections.OrderedDict`
   to maintain the previous behavior.
 
 3.25.1 (2025-01-11)
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -506,7 +506,7 @@ class Nested(Field):
         only: types.StrSequenceOrSet | None = None,
         exclude: types.StrSequenceOrSet = (),
         many: bool = False,
-        unknown: str | None = None,
+        unknown: types.UnknownOption | None = None,
         **kwargs: Unpack[_BaseFieldKwargs],
     ):
         # Raise error if only or exclude is passed as string, not list of strings
@@ -656,7 +656,7 @@ class Pluck(Nested):
         field_name: str,
         *,
         many: bool = False,
-        unknown: str | None = None,
+        unknown: types.UnknownOption | None = None,
         **kwargs: Unpack[_BaseFieldKwargs],
     ):
         super().__init__(
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -201,7 +201,7 @@ class SchemaOpts:
         self.include = getattr(meta, "include", {})
         self.load_only = getattr(meta, "load_only", ())
         self.dump_only = getattr(meta, "dump_only", ())
-        self.unknown = typing.cast(types.UnknownOption, getattr(meta, "unknown", RAISE))
+        self.unknown = getattr(meta, "unknown", RAISE)
         self.register = getattr(meta, "register", True)
         self.many = getattr(meta, "many", False)
 
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -36,7 +36,6 @@ from marshmallow.utils import (
     is_instance_or_subclass,
     missing,
     set_value,
-    validate_unknown_parameter_value,
 )
 
 
@@ -202,7 +201,7 @@ class SchemaOpts:
         self.include = getattr(meta, "include", {})
         self.load_only = getattr(meta, "load_only", ())
         self.dump_only = getattr(meta, "dump_only", ())
-        self.unknown = validate_unknown_parameter_value(getattr(meta, "unknown", RAISE))
+        self.unknown = typing.cast(types.UnknownOption, getattr(meta, "unknown", RAISE))
         self.register = getattr(meta, "register", True)
         self.many = getattr(meta, "many", False)
 
@@ -372,7 +371,7 @@ class Schema(metaclass=SchemaMeta):
         """Fields to exclude from serialized results"""
         dump_only: typing.ClassVar[tuple[Field] | list[Field]]
         """Fields to exclude from serialized results"""
-        unknown: typing.ClassVar[str]
+        unknown: typing.ClassVar[types.UnknownOption]
         """Whether to exclude, include, or raise an error for unknown fields in the data.
         Use `EXCLUDE`, `INCLUDE` or `RAISE`.
         """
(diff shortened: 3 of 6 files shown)


sloria:
attn @sirosen , since you wrote the initial [run-time checking PR](https://github.com/marshmallow-code/marshmallow/pull/1989)

lafrech:
I'm a bit lukewarm about this.

If a typo made the code crash at runtime, I'd think alright, mypy users will get a warning earlier, others will deal with it.

In this case, a typo may be silently ignored, so the check does have added value.

Not saying I'm totally opposed. After all, users should use the "Enum" values from utils.

Should we make it a real `Enum`, btw? This would also trigger type checker warnings.

sirosen:
I hadn't gotten to coming back to this yet, but since @lafrech seems to be having similar thoughts I figured I should weigh in.

I think some folks may have layers of indirection which make this less of an obvious win.
For a rather extreme example,
```python
schema_builder = MyCoolSchemaBuilder()
builder_settings: dict[str, typing.Any] = {"unknown": ma.EXCLUDE, ...}
new_schema = schema_builder.build(settings=builder_settings)
```

In these sorts of usages, there are all sorts of refactors which the user can pursue to get back type-safety (e.g., make that settings object a dataclass rather than a dict, where `unknown` is a slot with a specific type), but their out of the box code will not have that safety, and it won't be obvious to them what they've lost.

_That said_, I think exclude/include/raise should probably be a str-enum and we should treat the current (string) values as legacy/deprecated and update in v4.
i.e.
```python
def a_thing_handling_unknown(unknown: UnknownValue = UnknownValue.INCLUDE):
     if not isinstance(unknown, UnknownValue):
        warnings.warn(
            "Passing values to `a_thing_handling_unknown()` without converting them to `UnknownValue` "
            "is deprecated behavior. Use the `INCLUDE`, `EXCLUDE`, or `RAISE` constants. If you must "
            "pass a string, convert it with `UnknownValue(value)` before passing."
        )
        unknown = UnknownValue(unknown)
    ...
```

That has the advantage that anyone who is using the constants will continue to have their code work with no warnings, but anyone passing `"INCLUDE"` by hand will get a clear warning which tells them what they need to fix.

The thing which makes me uneasy about this is that it's a change which is silent to the user.

sloria:
> Should we make it a real Enum, btw? This would also trigger type checker warnings.

> That said, I think exclude/include/raise should probably be a str-enum and we should treat the current (string) values as legacy/deprecated and update in v4.

i don't see an advantage of using a StrEnum over this PR and there's 2 disadvantages:

1. it's more keystrokes for `marshmallow.UnknownValue.EXCLUDE` vs `marshmallow.EXCLUDE` 
2. it would cause type-check errors for current users' code. this would be breaking change, even with @sirosen's deprecation suggestion as written


>  After all, users should use the "Enum" values from utils.

exactly. the [documented usage](https://marshmallow.readthedocs.io/en/latest/quickstart.html#handling-unknown-fields) is the constants. passing strings was never explicitly supported. this feels different to me than https://github.com/marshmallow-code/marshmallow/pull/896 , where we validate `only` to be a collection of strings rather than a string because passing a string is common typo. there should be no reason for users to pass magic strings to `unknown`.

sirosen:
> 1. it's more keystrokes for `marshmallow.UnknownValue.EXCLUDE` vs `marshmallow.EXCLUDE`

I was thinking we'd define both, so you'd still use `marshmallow.EXCLUDE` (`marshmallow.UnknownValue` would be the type of `marshmallow.EXCLUDE`).

But I don't love python enums, so if we want to avoid them I'm okay with the idea that we stick with a `Literal` type.

> 2. it would cause type-check errors for current users' code. this would be breaking change, even with @sirosen's deprecation suggestion as written

I've struggled a bit (for a while, not just this case) to think through what kinds of breakage are acceptable from a type checker perspective. Even changing from `str` to `Literal[...]` breaks this, since a strictly typed helper which handles the `unknown` value is typed today as `str`.

i.e. I can have the following helper defined:
```python
def build_my_schema(*, unknown: str = marshmallow.RAISE) -> marshmallow.Schema:
    class MySchema(marshmallow.Schema): ...

    return MySchema(unknown=unknown)
```

This isn't purely hypothetical either. `webargs` is using a `str` type and relies on this to detect when a user passes a value, vs an explicit `None`, vs the library default:
https://github.com/marshmallow-code/webargs/blob/dbde72fe5db8a999acd1716d5ef855ab7cc1a274/src/webargs/core.py#L190

So changing to `Literal` as the type is 100% a breaking change for type checking.
IMO that's not a strong argument against it -- we're just as likely to be broken by a new `mypy` version.

---

I think I'm missing what the goal of this change is. Do we want to remove the check because it has a runtime performance cost?

I can get onboard with it if the position of `marshmallow` is changing to be that "users must obey typing contracts, other usages are never supported, and we strongly recommend using a checker like mypy or pyright". But I don't think that the contract with users is so clear today.

sloria:
> marshmallow.UnknownValue would be the type of marshmallow.EXCLUDE

oh i see now. i misunderstood your original proposal. that would work, but it sounds like we're in agreement about eschewing an `Enum`.

> Even changing from str to Literal[...] breaks this, since a strictly typed helper which handles the unknown value is typed today as str.

sorry, to clarify: this PR _is_ an intentional breaking type change, which is why this is for 4.0.

> I think I'm missing what the goal of this change is. Do we want to remove the check because it has a runtime performance cost?

it's a) reducing runtime cost, b) less code, and c) better type support. admittedly, a and b are minimal, and c) could be done independently, i.e. keep the annotations but add back the runtime validation. i'd prefer to do all three, but was curious to your feedback.

sirosen:
> which is why this is for 4.0.

Oh! Okay, that makes a big difference. I didn't realize that `4.0` had merged to `dev`.
In that case, I feel way differently. I have no objection.

I'm onboard with dropping a runtime check as long as the message to users is "you need to obey the type contracts in v4.0". In principle, I think changes like this are a great way that type-checking can be used to justify more efficient code with fewer "paranoid" type and value checks. I was only really concerned about it under v3, where I think the messaging around type checking is less clear.

---

It's worth thinking through what the `webargs` code I pointed at will look like for v4.

Do you expect it to be annotated to be...
```python
unknown: typing.Literal["include", "exclude", "raise"] | _Unset | None = UNSET,
```
?

(I'm presuming that since the _type_ is no longer `str`, we'll use a new missing-like type to capture the default.)

If we go the enum path -- which I am still hesitant about -- the annotation would become
```python
unknown: marshmallow.UnknownType | _Unset | None = UNSET
```
which I think is slightly nicer.

The advantage of having `marshmallow.UnknownType` available as a symbol is that writing anything which wraps a marshmallow API doesn't need to worry as much about accurately reproducing the type.

This is possible even without the enum if it's a type alias:
```python
# marshmallow/types_.py or similar module
UnknownType: typing.TypeAlias = typing.Literal["include", "exclude", "raise"]
```

sloria:
there's already a type alias https://github.com/marshmallow-code/marshmallow/pull/2771/files#diff-630db869873f99f6928a65684132f8ed4ca0de6cab0b9857fe60ed7b83b79721R18 , though it's not annotated with `typing.TypeAlias`. not sure if it's necessary; tbh i didn't know about `TypeAlias` until now 🤔

so in webargs we can do

```python
unknown: UnknownOption | _Unset | None = UNSET
```

sloria:
> I'm onboard with dropping a runtime check as long as the message to users is "you need to obey the type contracts in v4.0". In principle, I think changes like this are a great way that type-checking can be used to justify more efficient code with fewer "paranoid" type and value checks.

yeah, that's the principle being implicitly applied here. i suppose we could some docs (maybe in the upgrading guide) that states this explicitly

sirosen:
I'd like a statement in the upgrading doc for this -- happy to help write that content, when I get some time, since I care about it.

As for `TypeAlias`, I think it's an important disambiguator in order to guarantee good behaviors. I'm trying to remember the kind of case which triggers this, and not able to recall exactly right now. But type checkers and IDEs can get tripped up between "a type alias which exists for documentation / clarity purposes" and "a type assignment which is meant for use as a runtime value"; that's the core of what I remember about this.

sloria:
> I'd like a statement in the upgrading doc for this -- happy to help write that content, when I get some time, since I care about it.

that'd be great! send a PR whenever you have time

@lafrech can i get a ✅ if you're good with this change?

sloria:
as a follow up, i'm also planning to move `EXCLUDE`, et. al. and `missing` to a `marshmallow.constants` module (and still re-export them in `__init__.py`). `marshmallow.utils` is an odd place for them.

lafrech (approved):
I have no real objection to this.

I'm not fully convinced about not using a real `Enum`. As @sirosen wrote, we can still publish enum value to let the user import and use values rather than the enum. I wouldn't mind this being a breaking change since it is 4.0. But I don't really mind.

sloria:
so long as we keep the constants, the additional `Enum` would provide no additional value and IMO would be unnecessary API surface (even if it were private), so i'm going to pass on that for now.

lafrech:
Just to clarify, I was thinking of publishing the enum items not the enum values (strings).

But in fact, as a user, I'd rather use the enum. I'm fine with `ma.Unknown.EXCLUDE` instead of `ma.EXCLUDE`.

But I don't mind either way. Maybe I'm wrong about using enums just because they exist.

PR #2772, opened 2025-01-13
Check for fields declared as instances at class declaration rather than at instantiation

sloria:
in the spirit of erroring earlier. also obviates the `is_instance_or_subclass` util


lafrech (approved):
Great.

Issue #1391, opened 2019-09-11
Add a way to trim whitespace characters


245967906:
Like the trim_whitespace parameter in the django rest framework.

sloria:
You could use a custom field.

```python
from marshmallow import fields, Schema


class TrimmedString(fields.String):
    def _deserialize(self, value, *args, **kwargs):
        if hasattr(value, 'strip'):
            value = value.strip()
        return super()._deserialize(value, *args, **kwargs)


class ArtistSchema(Schema):
    name = TrimmedString(required=True)


print(ArtistSchema().load({"name": " David "}))
# {'name': 'David'}
```

You could also use field composition if you want to trim other field types.

```python
from marshmallow import fields, Schema


class Trim(fields.Field):
    def __init__(self, inner, *args, **kwargs):
        self.inner = inner
        super().__init__(*args, **kwargs)

    def _bind_to_schema(self, field_name, parent):
        super()._bind_to_schema(field_name, parent)
        self.inner._bind_to_schema(field_name, parent)

    def _deserialize(self, value, *args, **kwargs):
        if hasattr(value, 'strip'):
            value = value.strip()
        return self.inner._deserialize(value, *args, **kwargs)

    def _serialize(self, *args, **kwargs):
        return self.inner._serialize(*args, **kwargs)


class ArtistSchema(Schema):
    name = Trim(fields.String(), required=True)
    email = Trim(fields.Email())


print(ArtistSchema().load({"name": " David ", "email": " david@hotmail.com "}))
# {'name': 'David', 'email': 'david@hotmail.com'}
```


That said, I'm going to keep this issue open because this may be a common enough use case to justify adding to marshmallow core. Feedback welcome.

sloria:
Should we add a `strip_whitespace` parameter to `Field`? @lafrech @deckar01

245967906:
@sloria  thanks for your quick reply, the example you gave above can solve my problem, it's great.
but as you said, adding a strip_whitespace parameter might be a better way.

245967906:
@sloria hi, I understand what you mean. I will compare the implementation of similar libraries and make changes later if I find a better solution.

sloria:
Copying https://github.com/marshmallow-code/marshmallow/pull/1397#issuecomment-531323116 here for discussion:

> @sloria hi, I go to see the implementation of DRF and typesystem. summarized as follows:
> 
[...]

PR #2786, opened 2025-01-16
Allow validators to transform values

sloria:
Still need to update the docs, but this proves the concept mentioned in https://github.com/marshmallow-code/marshmallow/issues/1391#issuecomment-2596318335 , where we require that validators return a validated and possibly modified value, allowing users to apply a pipeline of transformations to any field

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -264,14 +264,14 @@ class Field(typing.Generic[_InternalT]):
         check_key = attr if self.attribute is None else self.attribute
         return accessor_func(obj, check_key, default)
 
-    def _validate(self, value: typing.Any) -> None:
+    def _validate(self, value: _InternalT) -> _InternalT:
         """Perform validation on ``value``. Raise a :exc:`ValidationError` if validation
         does not succeed.
         """
-        self._validate_all(value)
+        return self._validate_all(value)
 
     @property
-    def _validate_all(self) -> typing.Callable[[typing.Any], None]:
+    def _validate_all(self) -> typing.Callable[[_InternalT], _InternalT]:
         return And(*self.validators)
 
     def make_error(self, key: str, **kwargs) -> ValidationError:
@@ -373,7 +373,7 @@ class Field(typing.Generic[_InternalT]):
         if self.allow_none and value is None:
             return None
         output = self._deserialize(value, attr, data, **kwargs)
-        self._validate(output)
+        output = self._validate(output)
         return output
 
     # Methods for concrete classes to override.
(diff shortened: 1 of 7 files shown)


lafrech (commented):
Looks like this doesn't require many changes to existing validators because they already return the value.

It sounds a bit wrong to use a validator to modify the value, since it is not their intended use. That's not a strong argument, they could be called post_processors. It just makes me wonder if some day we'll realize that we took a shortcut that seemed fine but doesn't actually work for all cases.

Pydantic does it. But in a more complete way ("before", "after", "plain", "wrap"). The implementation in this PR is what Pydantic would call "after". What if tomorrow we want to implement "before", to modify the data before deserialization?

Are we trying to implement pre/post - load/dump but at field level rather than schema level? Should we do that instead?

sloria:
fwiw sqlalchemy also allow validators to change the value: https://docs.sqlalchemy.org/en/20/orm/mapped_attributes.html#simple-validators

> A quick way to add a “validation” routine to an attribute is to use the [validates()](https://docs.sqlalchemy.org/en/20/orm/mapped_attributes.html#sqlalchemy.orm.validates) decorator. An attribute validator can raise an exception, halting the process of mutating the attribute’s value, or can change the given value into something different.

i agree it isn't immediately apparent given the name "validators", but perhaps that's preferable than to adding more API surface? i'm not sure yet.

> Are we trying to implement pre/post - load/dump but at field level rather than schema level? Should we do that instead?

that's an interesting idea. something like

```python
class ArtistSchema(Schema):
    name = fields.Str(post_load=(strip_whitespace, uppercase))
    email = fields.Str(post_load=(strip_whitespace, uppercase))

# and/or

class ArtistSchema(Schema):
    name = fields.Str()
    email = fields.Str()

    @field_post_load("name", "email")
    def strip_whitespace(self, field_name: str, value: str) -> str:
        return value.strip()
```

i like this idea, but will have to give it some thought

sloria:
i suppose one downside of that would be that `validate`/`validates` becomes somewhat redundant with `post_load`/`field_post_load`, right?

sloria:
otoh, it's nice that field_pre/post_load/dump could be added as a non-breaking feature

lafrech:
> i suppose one downside of that would be that `validate`/`validates` becomes somewhat redundant with `post_load`/`field_post_load`, right?

Possibly. But the call is not exactly the same since `validate` expects a validator, not a function.

sloria:
closing this in favor of discussing https://github.com/marshmallow-code/marshmallow/issues/2787

Issue #2787, opened 2025-01-17
Add field-level pre/post-load API


sloria:
As suggested by @lafrech in https://github.com/marshmallow-code/marshmallow/pull/2786#pullrequestreview-2557439185, we can enable pre/post-load on fields to address use cases like whitespace trimming (https://github.com/marshmallow-code/marshmallow/issues/1391) via an API similar to the schema-level pre/post_load/dump API.

Something like:

```python
from marshmallow import Schema, fields

class ArtistSchema(Schema):
    name = fields.Str(post_load=(str.strip, str.upper))
    email = fields.Str(post_load=(str.strip, str.upper))

```

* pro: allows field-specific transformations without having to implement a custom field
* pro: non-breaking change
* con: overlaps with existing functionality provided by custom fields and the schema-level `@pre/post_load`

### [Allow validators to transform values](https://github.com/marshmallow-code/marshmallow/pull/2786)

* same as [pydantic](https://docs.pydantic.dev/latest/concepts/validators/#field-validators) and [sqlalchemy](https://docs.sqlalchemy.org/en/20/orm/mapped_attributes.html#simple-validators and https://github.com/marshmallow-code/marshmallow/issues/2787#issuecomment-2615157038)
* pro: minimal API surface added
* pro: minimal internal code changes
* con: it's not intuitive that validators can mutate values
* con: doesn't solve for pre-validation transformations or dumping/serialization
* con: breaking change

### Decorator API

update: this was considered as part of the original proposal but rejected/tabled (https://github.com/marshmallow-code/marshmallow/issues/2787#issuecomment-2603337689 and https://github.com/marshmallow-code/marshmallow/issues/2787#issuecomment-2615157038)

```python
from marshmallow import Schema, fields, field_post_load

class ArtistSchema(Schema):
    name = fields.Str()
    email = fields.Str()

    @field_post_load("name", "email")
    def strip_whitespace(self, field_name: str, value: str) -> str:
        return value.strip()
```

llucax:
Why is this considered a breaking change?

sloria:
sorry for the confusion. The breaking change refers to the 'validators transform values' alternative, not the proposed approach

sloria:
on second thought, i don't think field-level pre/post-dump makes as much sense as pre/post-load. field-level transformations are much less common on serialization than in deserialization, and custom fields already address the rare cases when they're necessary.

sloria:
[...]

PR #2799, opened 2025-01-20
Add pre/post_load parameters to Field

sloria:
~~no docs yet, this is to prove the concept in #2787~~ docs finished


sloria:
changing the base to dev since v3 is about to EOL

PR #2810, opened 2025-02-14
Meta Decorator

### LINKED ISSUE #1879: Meta Decorator
## Motivation

I often have lots of small schemas containing unique `Meta` classes with only a few attributes. The `Meta` class sometimes requires more lines than the schema definition and makes a file containing multiple schemas harder to read. This is especially true when using `marshmallow-jsonapi` due to `type_`.

## Proposal

Provide a `meta` decorator that will inject its kwargs into a `Meta` class for the class it is wrapping.

## Example

Before:

```py
class TestSchema(Schema):
    foo = fields.String()
    bar = fields.String()

    class Meta:
        type_ = 'tests'
        ordered = True
```

After:

```py
@meta(type_='tests', ordered=True)
class TestSchema(Schema):
    foo = fields.String()
    bar = fields.String()
```

deckar01:
Add an experimental decorator for declaring meta attributes that inherits by default.

Resolves #1879

Screenshot 2025-02-13 at 4 05 28 PM

### DIFF
--- a/src/marshmallow/experimental/meta.py
+++ b/src/marshmallow/experimental/meta.py
@@ -1,0 +1,70 @@
+from __future__ import annotations
+
+import typing
+
+if typing.TYPE_CHECKING:
+    from marshmallow.fields import Field
+    from marshmallow.schema import SchemaMeta
+    from marshmallow.types import UnknownOption
+
+
+@typing.overload
+def meta(
+    *bases: SchemaMeta,
+    fields: tuple[str, ...] | list[str] | None,
+    additional: tuple[str, ...] | list[str] | None,
+    include: dict[str, Field] | None,
+    exclude: tuple[str, ...] | list[str] | None,
+    many: bool | None,
+    dateformat: str | None,
+    datetimeformat: str | None,
+    timeformat: str | None,
+    render_module: typing.Any | None,
+    index_errors: bool | None,
+    load_only: tuple[str, ...] | list[str] | None,
+    dump_only: tuple[str, ...] | list[str] | None,
+    unknown: UnknownOption | None,
+    register: bool | None,
+    **kwargs,
+):
+    """
+    :param *bases: The meta classes to inherit from. Inherits from the decorated schema's
+        Meta class by default. Pass `None` to prevent inheritance.
+    :param fields: Fields to include in the (de)serialized result
+    :param additional: Fields to include in addition to the explicitly declared fields.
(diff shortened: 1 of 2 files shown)


src/marshmallow/experimental/meta.py
```diff
@@ -0,0 +1,70 @@
...
+        by class name in `Nested` fields. Only set this to `False` when memory
+        usage is critical. Defaults to `True`.
+    """
+
+
+@typing.overload
+def meta(*bases, **kwargs): ...
+
+
+def meta(*bases, **kwargs):
```
sloria, line 64: the `@overload` usage seems a bit odd...does the approach with `TypedDict` and `total=False` not work? like what we do here? https://github.com/marshmallow-code/marshmallow/blob/f0c6afb430a6d032300658c898752d6c7732b188/src/marshmallow/fields.py#L87-L98

or perhaps

```python
def meta(
    *bases: SchemaMeta,
    fields: tuple[str, ...] | list[str] | None,
    additional: tuple[str, ...] | list[str] | None,
    # ...
    **kwargs,
):
    def wrapper(schema):
        mro = bases if bases else (schema.Meta,)
        meta = type(schema.Meta.__name__, mro, {
            "fields": fields,
            "additional": additional,
             # ...
             **kwargs
        })
        return type(schema.__name__, (schema,), {"Meta": meta})

    return wrapper
```

deckar01: `Unpack[TypedDict]` does not allow accepting extra kwargs, but we allow custom `Meta` attributes.

Explicitly copying the args into a dictionary defines missing attributes as `None`, which breaks inheritance.

One `@overload`is all that is needed (and is valid), but [mypy requires a sacrificial overload](https://github.com/python/mypy/issues/5047#issuecomment-898048105).

sloria: Ah I see. In that case, maybe just add comments about why this approach is necessary. Could even include a link to that gh comment

src/marshmallow/experimental/meta.py
```diff
@@ -0,0 +1,70 @@
...
+
+if typing.TYPE_CHECKING:
+    from marshmallow.fields import Field
+    from marshmallow.schema import SchemaMeta
+    from marshmallow.types import UnknownOption
+
+
+@typing.overload
+def meta(
+    *bases: SchemaMeta,
```
sloria, line 13: Should this be 

```suggestion
    *bases: Schema.Meta,
```

? Surprised mypy doesn't raise an error in the tests tho 🤔

PR #2823, opened 2025-04-23
Fix incorrect export by removing 'pprint' from __all__.

Florian-Laport:
The import of `pprint` from `utils` was (correctly) removed from the main `__init__.py` by b393c1fba7bfc0220637399bd3452d179e808cbd, but `pprint` was never removed from the `__all__` list.

This breaks wildcard imports. Easily reproducible since the latest `4.0.0` PyPi release:
image


sloria:
Thanks!

PR #2835, opened 2025-07-15
Add missing partial to SchemaOpts

### LINKED ISSUE #2834: SchemaOpts misses partial
If I try to reproduce the solution in https://github.com/marshmallow-code/marshmallow/issues/2297#issuecomment-2284962125 with `partial` instead of `dateformat`, it fails. I guess this is because the `__init__` at https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/schema.py#L204 is missing the `partial` keyword.

number492:
Fixes https://github.com/marshmallow-code/marshmallow/issues/2834

### DIFF
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -216,6 +216,7 @@ class SchemaOpts:
         self.include = getattr(meta, "include", {})
         self.load_only = getattr(meta, "load_only", ())
         self.dump_only = getattr(meta, "dump_only", ())
+        self.partial = getattr(meta, "partial", False)
         self.unknown = getattr(meta, "unknown", RAISE)
         self.register = getattr(meta, "register", True)
         self.many = getattr(meta, "many", False)


lafrech:
Thanks for contributing.

I suppose the reason `partial` is not in `SchemaOpts` is because it is meant to modify the behaviour at runtime. If it is hardcoded in the schema to ignore missing required fields, there is no point adding required fields in the first place.

But there may be legit use cases:

- Setting partial as default and overriding it to `False` at runtime
- Create `SomethingSchema`, then subclass as `SomethingPartialSchema`

I see no objection to this. It is neither breaking nor surprising.

A unit test would be nice.

number492:
I have a codebase that does something really hacky. If I remember correctly, it constructs schemas from plain classes, and it needs to do normal schemas and "incomplete" schemas where required fields are not required. This is necessary to be able to save incomplete forms on a UI, while still keeping track of fields that will eventually have to be specified by the user.

I wanted to replace this logic with partial, but failed on this missing line. It's been a while, I'll dig into the issue again and come back with a more detailed use case, and possibly some unit tests.

PR #2838, opened 2025-07-24
Docs fix tuple spelling mistake

MarkoM-dot:
Doc examples are actually tuples not "duples".

### DIFF
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -550,7 +550,7 @@ class Schema(metaclass=SchemaMeta):
         :return: Serialized data
 
         .. versionchanged:: 3.0.0b7
-            This method returns the serialized data rather than a ``(data, errors)`` duple.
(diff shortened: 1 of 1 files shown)


lafrech:
I think duple is a relatively uncommon word meaning 2-tuple.

MarkoM-dot:
> I think duple is a relatively uncommon word meaning 2-tuple.

Uncommon and has a meaning of "having two elements" at least according to this [reference](https://www.merriam-webster.com/dictionary/duple). Regardless, technical documentation should use technical terms familiar to Pythonistas. Technically, the method returns a tuple type.

The changelog also agrees [here](https://github.com/marshmallow-code/marshmallow/blob/dev/CHANGELOG.rst#300b7-2018-02-03).

sloria (approved):
i'm fine with the change, if only for the sake of consistency with the changelog. thanks!

PR #2854, opened 2025-09-29
feat: improve consistency of many arg with nested schema

jafournier:
For nested field, we can get some inconsistencies between `field.many` and `field.schema.many`.

Here is a proposal to handle inconsistency edge cases (see tests).

PS: Such inconsistencies have some effects on automatic  swagger generation in `flask_appbuilder` (and thus in projects using it, as `superset`).
This will solve https://github.com/apache/superset/issues/33884.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -179,3 +179,4 @@ Contributors (chronological)
 - Dharanikumar Sekar `@dharani7998 <https://github.com/dharani7998>`_
 - Nicolas Simonds `@0xDEC0DE <https://github.com/0xDEC0DE>`_
 - Florian Laport `@Florian-Laport <https://github.com/Florian-Laport>`_
+- jfo `@jafournier <https://github.com/jafournier>`_
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,15 @@
 Changelog
 =========
 
+
+4.2.0 (unreleased)
+------------------
+
+Other changes:
+
+- ``many`` argument of ``Nested`` properly overrides schema instance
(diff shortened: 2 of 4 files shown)


sloria (commented):
Sorry for the delay in reviewing this. I'm curious how users are hitting this case (reading the superset issue didn't make it immediately obvious to me.

Users should really do `List(Nested(...))`, as `Nested(instance, many=True)` [might go away in the future](https://github.com/marshmallow-code/marshmallow/issues/779). That said, I do think this change makes sense given the existing behavior of the other kwargs to `Nested`, so I think we can go ahead with this. If none of the other maintainers object, will plan to merge and release this on Sunday

deckar01 (approved):

jafournier:
Thanks for reviewing and merging!

@sloria the issue was indeed not obvious to diagnose.

In `superset` there are ORM relation fields (see [owners](https://github.com/apache/superset/blob/master/superset/models/slice.py#L100)).
`flask_appbuilder` converts ORM class to marshmallow class, and relation fields are handled using `Nested(instance, many=...)` (see [here](https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/api/convert.py#L185)).
`_column2relation` is called by `_column2field` with `nested=True` (see [here](https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/api/convert.py#L205)).

Consequently I was confronted to `owners = Nested([owner class], many=True)`.

Changing `Nested(instance, many=True)` [here](https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/api/convert.py#L185) can be considered too, though I do not know the consequences of such a change in this codebase yet.

PR #2857, opened 2025-10-06
Disable RUF043 in tests: allow metacharacters in match patterns

lafrech:
This disables [RUF043](https://docs.astral.sh/ruff/rules/pytest-raises-ambiguous-pattern/) in tests.

This should fix #2856.

I don't think this rule is that interesting. It requires cumbersome code for all those string that contain a dot.


sloria (approved):
no strong opinion. either disabling the rule or using escaping with `error occurred\\.` seem fine to me

lafrech:
OK, I'll go on with this. I don't find the added value of the test worth the string changes.

PR #2861, opened 2025-10-20
(fix) missing constant with len validation

agentgodzilla:
Currently if you utilize the `missing` constant as fallback value for a field in a schema alongside a length validator it fails with the following error `"object of type '_Missing' has no len()`

I'm using missing like this because I have some function fields that parse some optional data (in an annoying list of dictionaries), and if you include a standard pythonic default value of `None, [], {}, ""` the blank field is still included in `load()` results instead of being excluded, where as if you use missing it gets dropped from result as expected. 

If you'd like to know more about my use case or my schema just let me know


sloria:
Makes sense to me! added a regression test. good to merge when CI is green.

PR #2882, opened 2025-12-23
Fix CVE reference format in CHANGELOG

zware:
The `cve` role adds the `CVE-` prefix on its own, so the link as is points to `CVE-CVE-2025-68480`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ Changelog
 
 Bug fixes:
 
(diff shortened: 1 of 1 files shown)


zware:
Of course, *now* I see #2881.  This is the same for the 4.x line; I'll double-check for other references that need the fix in this branch.

sloria:
Thanks! I had already merged `3.x-line` into dev but hadn't pushed yet

PR #2892, opened 2026-01-10
fix: handle uppercase `file` URLs 

### LINKED ISSUE #2891: ValidationError with uppercase `file` URLs
## Issue Description
I noticed that when a file URL uses an uppercase scheme (FILE), fields.Url raises a ValidationError, even though the exact same URL with a lowercase scheme (file) works fine. Since URL schemes are  case-insensitive, I’d expect either both to be accepted or both to be rejected.

For example, HTTPS:// and https:// both work without any issues. This seems related to marshmallow issue #2249 and PR #2800. I’m happy to open a PR to fix this if it makes sense.

## Reproducing Code Example

```

class MySchema(Schema):
    url = fields.Url(schemes={"file"})

MySchema().load({"url": "FILE:///path/to/somefile.txt"})
```

## Expected Result
```
{"url": "FILE:///path/to/somefile.txt"}
```
## Actual Behavior
```
  File "test.py", line 6, in <module>
    MySchema().load({"url": "FILE:///path/to/somefile.txt"})
  File ".../site-packages/marshmallow/schema.py", line 730, in load
    return self._do_load(
  File ".../site-packages/marshmallow/schema.py", line 938, in _do_load
    raise exc marshmallow.exceptions.ValidationError: {'url': ['Not a valid URL.']}
```

thanhlecongg:
Fix #2891

### DIFF
--- a/src/marshmallow/validate.py
+++ b/src/marshmallow/validate.py
@@ -214,8 +214,8 @@ class URL(Validator):
 
         # Hostname is optional for file URLS. If absent it means `localhost`.
         # Fill it in for the validation if needed
-        if scheme == "file" and value.startswith("file:///"):
-            matched = regex.search(value.replace("file:///", "file://localhost/", 1))
+        if scheme == "file" and value.lower().startswith("file:///"):
+            matched = regex.search("file://localhost/" + value[8:])
         else:
             matched = regex.search(value)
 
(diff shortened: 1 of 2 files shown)


lafrech (approved):
Thanks. Looks good to me.

I'll let a few days pass for other maintainers to chime in.

Feel free to ping me if I don't merge this before next week.

thanhlecongg:
@lafrech Just pinging to see if this can be merged. Thanks!

lafrech:
Fix merged and released.

Thanks again.

PR #2894, opened 2026-01-27
Fix Constant field rejecting None values during load

### LINKED ISSUE #2868: `fields.Constant` rejects `None` constant values during load
When that constant is `None`, loading any payload that contains null triggers `ValidationError: Field may not be null`. The stems from the initialization order: `Field.__init__` sets `allow_none` based on the (yet-unknown) `load_default`, and only afterwards `Constant.__init__` assigns the actual constant/load_default.

```python

class DemoSchema(Schema):
    # Should always load/dump None, independent of input
    sentinel = fields.Constant(None)

schema = DemoSchema()

assert schema.dump({"sentinel": "anything"})["sentinel"] is None

schema.load({"sentinel": None})
```

```
  File "/data/src/test.py", line 11, in <module>
    schema.load({"sentinel": None})
  File "/home/hdd/miniconda3/envs/py312/lib/python3.12/site-packages/marshmallow/schema.py", line 730, in load
    return self._do_load(
           ^^^^^^^^^^^^^^
  File "/home/hdd/miniconda3/envs/py312/lib/python3.12/site-packages/marshmallow/schema.py", line 938, in _do_load
    raise exc
marshmallow.exceptions.ValidationError: {'sentinel': ['Field may not be null.']}
```

emmanuel-ferdman:
# PR summary: 
`Constant(None)` was rejecting null values during load with "Field may not be null" because `allow_none` was being inferred before the constant value was assigned. Moving the default assignments before `super().__init__()`, lets the base class correctly infer `allow_none=True` when the constant is `None`.

Fixes #2868.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -183,3 +183,4 @@ Contributors (chronological)
 - Xiao `@T90REAL <https://github.com/T90REAL>`_
 - jfo `@jafournier <https://github.com/jafournier>`_
 - thanhlecongg `@thanhlecongg <https://github.com/thanhlecongg>`_
+- Emmanuel Ferdman  `@emmanuel-ferdman  <https://github.com/emmanuel-ferdman>`_
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+unreleased
+----------
+
+Bug fixes:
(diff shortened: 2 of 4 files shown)


lafrech (approved):
LGTM

Thanks!

sloria (approved):

PR #2895, opened 2026-01-30
Fix Constant(None) field rejecting null input

### LINKED ISSUE #2868: `fields.Constant` rejects `None` constant values during load

wingding12:
## Summary

When `Constant` field is initialized with `None` as the constant value, loading a payload containing null would incorrectly raise:

```
ValidationError: {'sentinel': ['Field may not be null.']}
```

### Root Cause

The issue was in the initialization order:
1. `Constant.__init__` calls `super().__init__(**kwargs)`
2. `Field.__init__` computes `allow_none` based on `load_default` (line 213)
3. At this point, `load_default` is still `missing_` (not passed in kwargs)
4. So `allow_none` is computed as `False`
5. Only afterwards does `Constant.__init__` set `self.load_default = constant`
6. But `allow_none` has already been incorrectly set

### Fix

Pass `load_default` and `dump_default` to the parent constructor via `kwargs.setdefault()` so that `allow_none` is computed correctly when the constant is `None`.

## Test Plan

Added test `test_constant_none_allows_null_input` that verifies:
- Loading `{"sentinel": None}` works without raising ValidationError
- Loading `{"sentinel": "anything"}` also works (constant always wins)
- Dumping also works correctly

Fixes #2868

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -2065,10 +2065,13 @@ class Constant(Field[_ContantT]):
     _CHECK_ATTRIBUTE = False
 
     def __init__(self, constant: _ContantT, **kwargs: Unpack[_BaseFieldKwargs]):
+        # Set load_default and dump_default before calling super().__init__
+        # so that allow_none is correctly computed when constant is None
+        # (fixes issue #2868)
+        kwargs.setdefault("load_default", constant)
+        kwargs.setdefault("dump_default", constant)
         super().__init__(**kwargs)
         self.constant = constant
-        self.load_default = constant
-        self.dump_default = constant
 
     def _serialize(self, value, *args, **kwargs) -> _ContantT:
         return self.constant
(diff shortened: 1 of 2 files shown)


sloria:
Done in https://github.com/marshmallow-code/marshmallow/pull/2894

PR #2897, opened 2026-02-07
fix: prevent OneOf.options()from emitting phantom pairs

### LINKED ISSUE #2869: `OneOf.options()` emits extra pairs when labels outnumber choices
`marshmallow.validate.OneOf.options()` is supposed to “return the (value, label) pairs, where value is a string associated with each choice.” However the implementation zips `choices` and `labels` with `zip_longest(..., fillvalue="")`, so it produces as many pairs as the longer input. When `labels` is longer than `choices`, `options()` yields phantom entries whose value is the empty string, even though no such choice exists. 
Better to raise a ValueErorr?

```python

oneof = OneOf(choices=[], labels=["extra"])
print(list(oneof.options()))
# Expected: []
# Actual: [('', 'extra')]  # bogus option with empty string value
```

JaeHyuckSa:
### Description
Replaced it with zip bounded by the number of choices, added a warning when labels exceed choices, and included the related test coverage.

### Related Issue
Fix: #2869

### DIFF
--- a/src/marshmallow/validate.py
+++ b/src/marshmallow/validate.py
@@ -4,8 +4,8 @@ from __future__ import annotations
 
 import re
 import typing
+import warnings
 from abc import ABC, abstractmethod
-from itertools import zip_longest
 from operator import attrgetter
 
 from marshmallow.exceptions import ValidationError
@@ -594,6 +594,16 @@ class OneOf(Validator):
         self.labels = labels if labels is not None else []
         self.labels_text = ", ".join(str(label) for label in self.labels)
         self.error: str = error or self.default_message
+        if labels is not None:
+            try:
+                if len(self.labels) > len(self.choices):
+                    warnings.warn(
+                        "The number of labels exceeds the number of choices. "
+                        "Extra labels will be ignored in options().",
+                        stacklevel=2,
+                    )
+            except TypeError:
+                pass
 
     def _repr_args(self) -> str:
         return f"choices={self.choices!r}, labels={self.labels!r}"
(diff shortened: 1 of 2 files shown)


src/marshmallow/validate.py
```diff
@@ -594,6 +594,16 @@ def __init__(
...
         self.error: str = error or self.default_message
+        if labels is not None:
+            try:
+                if len(self.labels) > len(self.choices):
+                    warnings.warn(
+                        "The number of labels exceeds the number of choices. "
+                        "Extra labels will be ignored in options().",
+                        stacklevel=2,
+                    )
+            except TypeError:
```
lafrech, line 605: In which case is this exception raised?

tests/test_validate.py
```diff
@@ -725,15 +726,26 @@ def test_oneof_options():
...
     oneof = validate.OneOf([1, 2])
     expected = [("1", ""), ("2", "")]
     assert list(oneof.options()) == expected
 
 
+def test_oneof_options_labels_exceed_choices():
+    with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter("always")
+        oneof = validate.OneOf([1, 2], ["one", "two", "three"])
+        assert len(w) == 1
```
lafrech, line 744: We might want to check the warning message, kinda like `pytest.raises`'s `match` argument.

lafrech (commented):
Hi. Sorry about the delay. Looks good.

Just minor comments.

Thanks.

sloria:
addressed in https://github.com/marshmallow-code/marshmallow/pull/2909 . thanks anyway for the PR ❤️

Issue #2893, opened 2026-01-10
`utils.get_value` raise TypeError with out-of-range `int` index


thanhlecongg:
## Issue Description
When passing an out-of-range `int` index to `utils.get_value`, a `TypeError` is raised instead of returning the provided `default` value. Based on the documentation and function’s behavior in `string` cases, I’d expect it to fall back to default when the index is invalid. FYI, this issue seems to occur because handling `IndexError` requires falling back to `getattr(obj, key, default)`. This logic was introduced in PR #1063 following git blame. I think this issue can be fixed by handling this `TypeError` to return `default` value. I can make an PR if this change make sense to you. 

## Reproducing Code Example

```py
from marshmallow import utils

### Case 1: List
lst = [0, 1, 2, 3, 4, 5]
print(utils.get_value(lst, 999, default=3))

### Case 1: Dictionary with integer keys
dictionary = {1: 'a', 2: 'b', 3: 'c'}
print(utils.get_value(dictionary, 4, default='z'))

### Case 3: List of object
class PointClass:
    def __init__(self, x, y):
        self.x = x
        self.y = y

list_obj = [[PointClass(24, 42),{"x": 24, "y": 42}]]
print(utils.get_value(list_obj, 3, default=None))
```

## Expected Result
```
Case 1: 3
Case 2: z
Case 3: None
```

## Actual Behavior
```
Traceback (most recent call last):
  File "test.py", line 3, in <module>
    print(utils.get_value(lst, 999, default=3))
  File ".../site-packages/marshmallow/utils.py", line 108, in get_value
    return _get_value_for_key(obj, key, default)
  File ".../site-packages/marshmallow/utils.py", line 126, in _get_value_for_key
    return getattr(obj, key, default)
TypeError: getattr(): attribute name must be string
```

deckar01:
`get_value` is only used to serialize data from the output of `Schema.load()`. Since enveloping the output with classes is a documented use case, it also handles class attributes. `Schema.get_attribute()` was designed to be overridden to provide an accessor for other types of envelopes (like lists).

We should consider breaking the attribute accessor out into a separate utility that users have to opt into. Toasted dropped attribute access as one of its primary optimizations.

We should probably also change the list test from (https://github.com/marshmallow-code/marshmallow/pull/340). That PR didn't add support for lists.

deckar01:
> `get_value` works fine for accessing integer keys on dicts. `get_value` has not supported lists for the last 13 years. ...

[...]

PR #2898, opened 2026-02-08
Fix get_value TypeError with non-string keys

veeceey:
## Summary

- Fix `utils.get_value` raising `TypeError` instead of returning the `default` value when an integer key triggers the `getattr` fallback path (since `getattr` requires string attribute names)
- Catch `TypeError` from `getattr()` in both branches of `_get_value_for_key` and return `default` instead
- Add regression tests covering out-of-range list indices, missing integer dict keys, and integer keys on non-subscriptable objects

Fixes #2893

## Test plan

- [x] Added `test_get_value_out_of_range_int_index_returns_default` covering all three cases from the issue (list, dict with int keys, nested list of objects)
- [x] Added `test_get_value_int_key_on_non_subscriptable_returns_default` for the `not hasattr(obj, "__getitem__")` branch
- [x] Full test suite passes (1134 tests)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -118,12 +118,18 @@ def _get_value_for_keys(obj, keys, default):
 
 def _get_value_for_key(obj, key, default):
     if not hasattr(obj, "__getitem__"):
-        return getattr(obj, key, default)
+        try:
+            return getattr(obj, key, default)
+        except TypeError:
+            return default
 
     try:
         return obj[key]
     except (KeyError, IndexError, TypeError, AttributeError):
-        return getattr(obj, key, default)
+        try:
+            return getattr(obj, key, default)
+        except TypeError:
+            return default
 
 
 def set_value(dct: dict[str, typing.Any], key: str, value: typing.Any):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -79,6 +79,42 @@ def test_get_value():
     assert utils.get_value(lst, MyInt(1)) == 2
 
 
+def test_get_value_out_of_range_int_index_returns_default():
+    """get_value should return default when an int index is out of range.
+
+    Regression test for https://github.com/marshmallow-code/marshmallow/issues/2893
+    """
+    # Out-of-range index on a list
+    lst = [0, 1, 2, 3, 4, 5]
+    assert utils.get_value(lst, 999, default=3) == 3
+
+    # Missing int key on a dict with int keys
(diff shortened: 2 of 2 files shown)


deckar01:
`utils.get_value` is the default for `Schema.get_attribute`.

https://marshmallow.readthedocs.io/en/latest/extending/overriding_attribute_access.html

The field `attribute` only supports strings, so using ints would error in other ways before getting to this code.

https://marshmallow.readthedocs.io/en/latest/marshmallow.fields.html#marshmallow.fields.Field

veeceey:
Thanks for the review, @deckar01. You raise a fair point that in the standard schema serialization path, `Field.attribute` only accepts strings, so `Schema.get_attribute` would never pass an integer key through to `get_value`.

That said, I think the fix is still warranted for a couple of reasons:

1. **`get_value` explicitly accepts integer keys in its signature (`key: int | str`) and there are existing tests on main that exercise this** (`test_get_value` passes `int` and `MyInt` keys into lists). The function clearly intends to support integer keys, but that support is incomplete: valid indices work, while out-of-range indices crash with a `TypeError` instead of returning `default`. That's an inconsistency in the function's own contract.

2. **`get_value` is a public utility, not just an internal detail of `Schema.get_attribute`.** Users can (and do, per issue #2893) call `marshmallow.utils.get_value` directly or pass it as a custom `accessor` to `Field.get_value`. In that context, the caller controls the key type, and hitting a `TypeError` instead of getting the `default` back is surprising.

3. **The fix is minimal and safe** -- it just catches `TypeError` from `getattr()` when the key isn't a string and returns `default`, which is already the documented fallback behavior. It doesn't change any behavior for string keys.

I do take your point from the issue discussion that the broader question of whether `get_value` should support attribute access at all is worth considering separately (re: Toasted's optimization). But until that's decided, I think it's reasonable to make the existing integer-key support consistent rather than leaving a partial implementation that crashes on edge cases.

Happy to hear if you still think this should be approached differently.

deckar01:
`get_value` works fine for accessing integer keys on dicts. `get_value` has not supported lists for the last 13 years. The PR that added the list tests was not adding list support.

veeceey:
Thanks for the clarification, @deckar01. You're right -- `get_value` hasn't supported lists for 13 years, and the existing list tests were not adding list support. Since integer keys on dicts work fine and the `Field.attribute` path only accepts strings, this fix isn't warranted. Closing this PR. Thanks for the review\!

PR #2901, opened 2026-02-18
Fix Constant field with required=True raising ValueError

### LINKED ISSUE #2900: `fields.Constant` with `required=True` flag raised `load_default` warning
I have a schema containing a `Constant`, with `required=True` set, as the schema is ultimately exported to an OpenAPI spec, and the field must be marked as `required` there.

The fix from #2894 now triggers the validation failure in https://github.com/marshmallow-code/marshmallow/blob/2a3812d5049c83e98db60a0869919521f97cd77d/src/marshmallow/fields.py#L216

My feeling on the fix is to undo the changes from #2894 and instead reproduce the `allow_none` flag initialization in `Constant.__init__`

Happy to create a PR if you'd be happy with this fix.

worksbyfriday:
## Summary

Fixes #2900.

`Constant(42, required=True)` raises `ValueError: 'load_default' must not be set for required fields.` because #2894 moved `load_default` assignment into kwargs before `super().__init__()`.

The fix:
- Sets `load_default` and `dump_default` after `super().__init__()` (avoiding the conflict check)
- Overrides `_validate_missing` to skip the `required` check, since a Constant always provides a value regardless of input
- Preserves the #2894 fix for `Constant(None)` by explicitly setting `allow_none = True` when the constant is `None` and `allow_none` was not provided

## Test plan

- [x] Existing `test_constant_none_allows_none_value` still passes (#2894 regression)
- [x] New `test_constant_with_required` verifies `Constant(42, required=True)` works
- [x] All 1133 tests pass

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -2075,6 +2075,8 @@ class Constant(Field[_ContantT]):
             self.allow_none = True
 
     def _validate_missing(self, value):
+        # Omit check for value is missing_
+        # Below is just a paranoid check
         if value is None and not self.allow_none:
             raise self.make_error("null")
 
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -2065,10 +2065,18 @@ class Constant(Field[_ContantT]):
     _CHECK_ATTRIBUTE = False
 
     def __init__(self, constant: _ContantT, **kwargs: Unpack[_BaseFieldKwargs]):
-        kwargs["load_default"] = constant
-        kwargs["dump_default"] = constant
         super().__init__(**kwargs)
         self.constant = constant
+        self.load_default = constant
+        self.dump_default = constant
+        # If allow_none was not explicitly provided and the constant is None,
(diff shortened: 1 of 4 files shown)


src/marshmallow/fields.py
```diff
@@ -2065,10 +2065,18 @@ class Constant(Field[_ContantT]):
...
         self.constant = constant
+        self.load_default = constant
+        self.dump_default = constant
+        # If allow_none was not explicitly provided and the constant is None,
+        # None should be considered valid (mirrors Field.__init__ logic).
+        if kwargs.get("allow_none") is None and constant is None:
+            self.allow_none = True
+
+    def _validate_missing(self, value):
+        if value is None and not self.allow_none:
```
lafrech, line 2089: Shouldn't happen, right?
I mean if `constant` is `None`, `allow_none` is set at init.
Anyway, I understand the logic in keeping this here (only remove the `missing` check).
This is also beneficial to users modifying `constant` after init, although I don't think this is meant to be supported.

lafrech (approved):
Looks good to me.
Thanks!

PR #2903, opened 2026-02-20
Fix nested partial to use attr_name instead of data_key for prefix

bysiber:
## Summary

When using dot-delimited `partial` with nested schemas, the sub-partial prefix is derived from `field_name` (the `data_key`) instead of `attr_name`. This causes the nested partial to silently break when a field has `data_key` set.

## Problem

The `partial` parameter accepts dot-delimited attribute names:
```python
schema.load(data, partial=("address.zip_code",))
```

In `_deserialize`, the top-level partial check correctly uses `attr_name`:
```python
partial_is_collection and attr_name in partial
```

But the nested sub-partial extraction uses `field_name` (which is the `data_key` if set):
```python
prefix = field_name + "."  # field_name could be "homeAddress" (data_key)
```

If a field is declared as `address = Nested(AddressSchema, data_key="homeAddress")`, then:
- `attr_name` = `"address"`
- `field_name` = `"homeAddress"` (from data_key)
- User passes `partial=("address.zip_code",)`
- `prefix` = `"homeAddress."` — doesn't match `"address.zip_code"`
- Result: `sub_partial` is empty, nested schema doesn't get `partial`

## Fix

Use `attr_name` instead of `field_name` for the prefix, consistent with how the top-level partial check works.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -186,3 +186,4 @@ Contributors (chronological)
 - Emmanuel Ferdman  `@emmanuel-ferdman  <https://github.com/emmanuel-ferdman>`_
 - Fridayworks `@worksbyfriday  <https://github.com/worksbyfriday>`_
 - `@rstar327 <https://github.com/rstar327>`_
+- Kadir Can Ozden  `@bysiber  <https://github.com/bysiber>`_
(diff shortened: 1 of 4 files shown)


sloria (commented):
Can you please add a test to demonstrate the issue/fix?

PR #2908, opened 2026-02-27
Fix get_value TypeError with out-of-range int index

rstar327:
Fixes #2893

`utils.get_value` raises `TypeError` instead of returning the `default` value when given an out-of-range `int` index on a list. This happens because the fallback `getattr(obj, key, default)` requires a string attribute name, and passing an `int` raises `TypeError`.

This PR catches the `TypeError` from `getattr` and returns `default` instead.

```python
# Before
utils.get_value([0, 1, 2], 999, default=3)
# TypeError: attribute name must be string, not 'int'

# After
utils.get_value([0, 1, 2], 999, default=3)
# 3
```

All 1133 tests pass.

### DIFF
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -123,7 +123,10 @@ def _get_value_for_key(obj, key, default):
     try:
         return obj[key]
     except (KeyError, IndexError, TypeError, AttributeError):
-        return getattr(obj, key, default)
+        try:
+            return getattr(obj, key, default)
+        except TypeError:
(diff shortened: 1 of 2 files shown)


lafrech:
Please take the time to investigate bugtrackers about the changes your AI tool suggests rather than strafe GitHub with tens of pull requests.

This change was submitted 3 days ago in #2905, and as I answered there, the conversion in #2893, concluded that current behaviour is intended and could be clarified in a comment.

PR #2909, opened 2026-02-27
Fix OneOf.options() emitting phantom entries when labels outnumber choices

### LINKED ISSUE #2869: `OneOf.options()` emits extra pairs when labels outnumber choices

rstar327:
Fixes #2869

`OneOf.options()` used `zip_longest(choices, labels, fillvalue="")` which produced phantom `('', 'label')` entries when `labels` was longer than `choices`. These phantom entries have no corresponding choice and are misleading.

This PR replaces `zip_longest` with explicit padding/truncation so that `options()` always returns exactly as many pairs as there are choices:
- Extra labels beyond the number of choices are silently ignored
- Missing labels are padded with empty strings (preserving existing behavior)

```python
# Before
OneOf(choices=[], labels=["extra"]).options()  # [('', 'extra')]

# After
OneOf(choices=[], labels=["extra"]).options()  # []
```

All 1132 tests pass.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -174,7 +174,7 @@ Contributors (chronological)
 - Aditya Tewary `@aditkumar72 <https://github.com/aditkumar72>`_
 - Sebastien Lovergne `@TheBigRoomXXL <https://github.com/TheBigRoomXXL>`_
 - Peter C `@somethingnew2-0 <https://github.com/somethingnew2-0>`_
-- Marcel Jackwerth `@mrcljx` <https://github.com/mrcljx>`_
+- Marcel Jackwerth `@mrcljx <https://github.com/mrcljx>`_
 - Fares Abubaker `@Fares-Abubaker <https://github.com/Fares-Abubaker>`_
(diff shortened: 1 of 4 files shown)


lafrech (approved):
LGTM. Thanks.

lafrech:
Once this is merged, there will be only one call of `zip_longest` left, in `_invoke_processors`.

While removing it breaks a test, I find the use case questionable and I'd be tempted to suggest removing it in next major as allowing processors to modify the length of collections opens the door to unpredictable issues.

PR #2910, opened 2026-02-27
Fix Constant field with required=True raising ValueError

### LINKED ISSUE #2900: `fields.Constant` with `required=True` flag raised `load_default` warning

rstar327:
Fixes #2900

`fields.Constant` always sets `load_default` to the constant value internally. Since #2894 added a check that rejects `load_default` on required fields, `Constant(constant, required=True)` now raises `ValueError`.

This PR fixes the issue by:
1. Extracting `required` from kwargs before calling `super().__init__()`, then setting it after
2. Overriding `_validate_missing` in `Constant` to skip the required-field check, since the constant value is always available via `load_default`

```python
# Before
fields.Constant("foo", required=True)
# ValueError: 'load_default' must not be set for required fields.

# After
class MySchema(Schema):
    type_ = fields.Constant("foo", required=True)

MySchema().load({})       # {'type_': 'foo'}
MySchema().dump({})       # {'type_': 'foo'}
MySchema().fields["type_"].required  # True
```

All 1133 tests pass.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -2065,10 +2065,21 @@ class Constant(Field[_ContantT]):
     _CHECK_ATTRIBUTE = False
 
     def __init__(self, constant: _ContantT, **kwargs: Unpack[_BaseFieldKwargs]):
+        # Extract required before passing to super().__init__ because
+        # Constant always sets load_default (to the constant value),
+        # which would conflict with the required=True check in Field.__init__.
+        is_required = kwargs.pop("required", False)
         kwargs["load_default"] = constant
         kwargs["dump_default"] = constant
         super().__init__(**kwargs)
+        self.required = is_required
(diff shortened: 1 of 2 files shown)


sloria:
done in https://github.com/marshmallow-code/marshmallow/pull/2901

PR #2911, opened 2026-03-01
Allow passing a Field instance to `unknown` option

### LINKED ISSUE #853: Allow specifying output field when using `unknown=INCLUDE`
From https://github.com/marshmallow-code/marshmallow/issues/524#issuecomment-393057138

> We could even go further and let the user set a default Field when using *INCLUDE*, but I'm not sure it is so useful, and it is never too late to add it in a later step.

(edited to make the argument name correct)

I suggested [here](https://github.com/marshmallow-code/marshmallow/issues/524#issuecomment-397430467) that a field could be passed to unknown, e.g. `unknown=fields.Str()`.

Br1an67:
## Summary

Allow a `Field` instance to be passed as the `unknown` option (in addition to the existing `RAISE`, `INCLUDE`, `EXCLUDE` constants). When a `Field` is provided, unknown field values are deserialized through that field, enabling validation and type coercion.

Closes #853

## Example

```python

class MySchema(Schema):
    name = fields.String()

schema = MySchema(unknown=fields.Int())
schema.load({"name": "Joe", "age": "42"})
# => {"name": "Joe", "age": 42}
```

The `Field` can be set via:
- `Schema(unknown=fields.Int())`
- `class Meta: unknown = fields.Int()`
- `schema.load(data, unknown=fields.Int())`
- `fields.Nested(ChildSchema, unknown=fields.Int())`

## Changes

- **`src/marshmallow/types.py`**: Updated `UnknownOption` type alias to include `Field` (under `TYPE_CHECKING` to avoid circular imports)
- **`src/marshmallow/schema.py`**: Added `isinstance(unknown, Field)` branch in `_deserialize` that routes unknown values through the field's `deserialize()` method, with proper error collection
- **`tests/test_schema.py`**: Added 7 tests covering: basic deserialization, validation errors, Meta config, many=True, load() kwarg, nested schemas, and passthrough with `Field()`

### DIFF
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -690,7 +690,21 @@ class Schema(metaclass=SchemaMeta):
                 }
                 for key in set(data) - fields:
                     value = data[key]
-                    if unknown == INCLUDE:
+                    if isinstance(unknown, ma_fields.Field):
+
+                        def getter(val, unknown_field=unknown, field_name=key):
+                            return unknown_field.deserialize(val, field_name, data)
+
+                        deserialized = self._call_and_store(
+                            getter_func=getter,
+                            data=value,
+                            field_name=key,
+                            error_store=error_store,
+                            index=index,
+                        )
+                        if deserialized is not missing:
+                            ret_d[key] = deserialized
+                    elif unknown == INCLUDE:
                         ret_d[key] = value
                     elif unknown == RAISE:
                         error_store.store_error(
--- a/src/marshmallow/types.py
+++ b/src/marshmallow/types.py
@@ -9,14 +9,25 @@ from __future__ import annotations
 
 import typing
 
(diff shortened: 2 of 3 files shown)


themavik (commented):
What this PR does: Allows passing a Field instance to the unknown option to deserialize unknown keys through that field.

Done well: The isinstance(unknown, ma_fields.Field) check before INCLUDE/RAISE is clean. Using _call_and_store for deserialization reuses existing error handling. The TYPE_CHECKING guard for UnknownOption avoids circular imports. Test coverage for schema, meta, many=True, load kwarg, nested, and fields.Field passthrough is thorough.

Suggestion: When unknown is a Field and deserialized is missing, the key is not added to ret_d—consistent with EXCLUDE. Consider documenting this in the docstring.

PR #2916, opened 2026-03-12
fix(utils): handle TypeError from getattr with non-string keys in get_value

Dhi13man:
## Summary

- `utils.get_value` raises `TypeError` when given an out-of-range integer key for a list or a missing integer key for a dict, because the `getattr` fallback requires a string attribute name
- Wrapped the `getattr` fallback in `_get_value_for_key` with `try/except TypeError` to gracefully return the default value for non-string keys
- Added regression test covering list out-of-range, dict missing int key, and in-range cases

Closes #2893

## Test plan

- [x] `test_get_value_with_out_of_range_int_key` — 4 assertions covering list OOR, dict missing key, and both in-range
- [x] Full test suite passes (1133 tests, 0 failures)
- [x] `ruff check` and `ruff format --check` clean

Generated with [Dhiman's Agentic Suite](https://github.com/Dhi13man)

### DIFF
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -123,7 +123,12 @@ def _get_value_for_key(obj, key, default):
     try:
         return obj[key]
     except (KeyError, IndexError, TypeError, AttributeError):
-        return getattr(obj, key, default)
+        try:
+            return getattr(obj, key, default)
+        except TypeError:
+            # getattr requires a string attribute name; if key is
+            # a non-string (e.g. int), fall back to the default.
+            return default
 
 
 def set_value(dct: dict[str, typing.Any], key: str, value: typing.Any):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -79,6 +79,24 @@ def test_get_value():
     assert utils.get_value(lst, MyInt(1)) == 2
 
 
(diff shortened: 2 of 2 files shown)


lafrech:
I should train an AI to open issues on other AIs to ask them to lurk a little on bugtrackers before posting PRs that are exact copies of other PRs that have already been rejected, especially when they were rejected after a discussion concluding that current behaviour was fine.

It is kinda rude (and spammish) to send stuff here for us to check without taking the time to review it correctly first. And a bit naive to assume the AI actually knows what it is doing. Surprisingly, it is capable of spotting what looks like easy issues in backlogs and understand enough to produce a change that looks right, but it doesn't seem to be able to realize the issue and PRs on the very same bugtracker explain otherwise. Go figure.

Maintainer work time should not be burdened by the review of sloppy contributions essentially meant to increment a PR counter on a GH profile. Does anybody seriously care about GH stats anyway?

Dhi13man:
Understood. My apologies.

Dhi13man:
I had leftover Claude Max and wanted to contribute to backlog clearing in open source projects before it reset, but I see that this could be a nuisance. My apologies.

lafrech:
Understood is all that matters.

We've been seeing a lot of this lately.

PR #2917, opened 2026-03-13
Fix TypeError in utils.get_value with out-of-range int index

oaksprout:
Fixes #2893. When passing an out-of-range int index to utils.get_value, a TypeError was raised because it tried to fallback to getattr() with an int key. This PR ensures getattr() is only called if the key is a string.

### DIFF
--- a/reproduce_issue.py
+++ b/reproduce_issue.py
@@ -1,5 +1,5 @@
 from marshmallow import utils
-import pytest
+
 
 def test_get_value_out_of_range_int():
     lst = [0, 1, 2, 3, 4, 5]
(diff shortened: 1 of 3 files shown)


sloria:
this is getting old

https://github.com/marshmallow-code/marshmallow/pull/2916#issuecomment-4049754966
https://github.com/marshmallow-code/marshmallow/pull/2908#issuecomment-3972920093
https://github.com/marshmallow-code/marshmallow/pull/2898#issuecomment-3874371182
https://github.com/marshmallow-code/marshmallow/pull/2905#issuecomment-3955048373

PR #2918, opened 2026-03-13
Fix TypeError in utils.get_value with out-of-range int index

oaksprout:
Fixes #2893. When passing an out-of-range int index to utils.get_value, a TypeError was raised because it tried to fallback to getattr() with an int key. This PR ensures getattr() is only called if the key is a string.


sloria:
seriously? a duplicate slop PR after [i pointed out](https://github.com/marshmallow-code/marshmallow/pull/2917#issuecomment-4056580524) that the previous PR was a duplicate of previous slop PRs?

Issue #533, opened 2016-09-21
Bytes field and RawString field


asmodehn:
Not sure if these would make sense for marshmallow as a default fields, but while working with marshmallow and existing library doing network serialization/deserialization ([ROS](http://wiki.ros.org/msg), I came to need a field that would do : 

Py2 : str --serialize--> str --deserialize--> str
Py3 : bytes --serialize--> bytes --deserialize--> bytes

So something similar to the string field, but for bytes/rawstring...

I implemented a quick prototype for my usecase like : 

``` python
def ensure_binary_type(val):
    if isinstance(val, marshmallow.compat.text_type):
        val = val.encode('utf-8')
    return marshmallow.compat.binary_type(val)


class RosString(marshmallow.fields.Field):
    """A ros string, serializing as ros python field type :
    Python <3.0 implies str, python >3.0 implies bytes

    If you need unicode serialization, have a look at RosTextString.

    No marshmallow field class for this, so we're declaring it here.

    :param kwargs: The same keyword arguments that :class:`Field` receives. required is set to True by default.
    """
    default_error_messages = {
        'invalid': 'Not a valid binary string.'
    }

    def __init__(self, **kwargs):
        super(RosString, self).__init__(required=True, **kwargs)

    def _serialize(self, value, attr, obj):
        if value is None:
            return None
        return ensure_binary_type(value)

    def _deserialize(self, value, attr, data):
        if not isinstance(value, marshmallow.compat.basestring):
            self.fail('invalid')
        return ensure_binary_type(value)

```

Also I came to need another field that would serialize as unicode, and deserialize as str/bytes : 

Py2 : str --serialize--> unicode --deserialize--> str
Py3 : bytes --serialize--> str --deserialize--> bytes

prototype : 

``` python
class RosTextString(RosString):
    """A ros string, serializing into unicode.
     Python <3.0 means unicode --deserialize--> str, python >3.0 means str --deserialize--> bytes

    For planning for python3, we serialize this as unicode string
    For using with ROS we deserialize as str / bytes

    No marshmallow field class for this, so we're declaring it here.

    :param kwargs: The same keyword arguments that :class:`Field` receives. required is set to True by default.
    """
    default_error_messages = {
        'invalid': 'Not a valid text string.'
    }

    # we got this from marshmallow.fields.String
    def _serialize(self, value, attr, obj):
        if value is None:
            return None
        return marshmallow.utils.ensure_text_type(value)
```

[...]

PR #2921, opened 2026-03-15
Implement new `Bytes` field

rrad5409:
I would like to be able to use the native `bytes` type when deserialising. Right now the closest approximation is `marshmallow.fields.String`, but that requires it to be UTF8, which is too restrictive if I want pure binary data.

I added a new `Bytes` field, which supports deserialising from:
- `bytes`
- `bytearray`
- `str`
- `int`
- `typing.SupportsBytes`
- `typing.Iterable` (i.e. `list[int]`)

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import abc
+import base64
 import collections
 import copy
 import datetime as dt
@@ -48,6 +49,7 @@ __all__ = [
     "AwareDateTime",
     "Bool",
     "Boolean",
+    "Bytes",
     "Constant",
     "Date",
     "DateTime",
@@ -880,8 +882,33 @@ class String(Field[str]):
 
 class Bytes(Field[bytes]):
     """
-    Marshmallow field type for any bytes array.
+    A field for deserializing strings into byte arrays.
+
+    :param encoding: Specifies the string encoding used when encoding/decoding to/from strings.
+    :param errors: Error behaviour when converting to/from a :class:`str`, inherited from it's constructor.
+    :param serialize: Specifies the return type when serializing.
+        `base64` and `str` use the value of `encoding` for the string.
+    :param kwargs: The same keyword arguments that :class:`Field` receives.
+
+    .. versionadded:: 4.3.0
     """
+    #: Default error messages.
+    default_error_messages = {
+        "not_bytes": "Not a bytes-like object.",
+        "unicode": "Invalid unicode string.",
+    }
+
+    def __init__(
+        self,
+        encoding: str = "utf-8",
+        errors: str = "strict",
+        serialize: typing.Literal["int", "str", "bytes", "base64"] = "base64",
+        **kwargs: Unpack[_BaseFieldKwargs],
+    ):
+        super().__init__(**kwargs)
+        self.encoding = encoding
+        self.errors = errors
+        self.serialize = serialize
 
     def _deserialize(
         self,
@@ -892,15 +919,11 @@ class Bytes(Field[bytes]):
     ) -> bytes:
         try:
             match value:
-                case bytes() as b:
-                    return b
-                case bytearray() as ba:
-                    return bytes(ba)
                 case str() as s:
                     return bytes(
                         s,
-                        encoding="utf-8",
-                        errors="ignore",
+                        encoding=self.encoding,
+                        errors=self.errors,
                     )
                 case int() as i:
                     return i.to_bytes(
@@ -908,12 +931,40 @@ class Bytes(Field[bytes]):
                         byteorder="big",
                         signed=i < 0,
                     )
-                case obj:
-                    if isinstance(obj, (typing.SupportsBytes, typing.Iterable)):
-                        return bytes(obj)
-            raise ValidationError("not a bytes-like object")
+                case _:
+                    return bytes(obj)
         except TypeError as e:
-            raise ValidationError("not a bytes-like object") from e
+            raise self.make_error("not_bytes")  from e
+        except UnicodeError as e:
+            raise self.make_error("unicode")  from e
+
+    def _serialize(self,
+        value: bytes,
+        attr: str | None,
+        obj: typing.Any,
+        **kwargs: typing.Any,
+    ) -> str | int | bytes:
+        try:
+            match self.serialize:
+                case "str":
+                    return str(
+                        value,
+                        encoding=self.encoding,
+                        errors=self.errors,
+                    )
+                case "base64":
+                    return base64.standard_b64encode(value)
+                case "int":
+                    return int.from_bytes(
+                        value,
+                        byteorder="big",
+                    )
+                case "bytes":
+                    return value
+                case _:
+                    typing.assert_never(self.serialize)
+        except UnicodeError as e:
+            raise self.make_error("unicode")  from e
 
 
 class UUID(Field[uuid.UUID]):
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -933,7 +933,7 @@ class Bytes(Field[bytes]):
                         signed=i < 0,
                     )
                 case _:
-                    return bytes(obj)
+                    return bytes(value)
         except TypeError as e:
             raise self.make_error("not_bytes") from e
         except UnicodeError as e:
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -878,6 +878,44 @@ class String(Field[str]):
             raise self.make_error("invalid_utf8") from error
 
 
+class Bytes(Field[bytes]):
+  """
+  Marshmallow field type for any bytes array.
+  """
+
+  def _deserialize(
+    self,
+    value: typing.Any,
+    attr: str | None,
+    data: typing.Mapping[str, typing.Any] | None,
+    **kwargs: typing.Any,
+  ) -> bytes:
+    try:
+      match value:
+        case bytes() as b:
+          return b
+        case bytearray() as ba:
+          return bytes(ba)
+        case str() as s:
+          return bytes(
+            s,
+            encoding="utf-8",
+            errors="ignore",
+          )
+        case int() as i:
+          return i.to_bytes(
+            length=max(1, (7 + i.bit_length()) // 8),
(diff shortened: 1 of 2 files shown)


deckar01:
This was previously determined to be out of scope for a core field. https://github.com/marshmallow-code/marshmallow/issues/533#issuecomment-251571144

If you are interested in publishing this field as a package and maintaining it, consider adding it to the [ecosystem page](https://github.com/marshmallow-code/marshmallow/wiki/Ecosystem).

rrad5409:
What's the reason for it being out of scope? To me it seems like something perfectly on-par with having UUID, IPv4, and string support - is there something I'm missing here?

Like it's a dead-simple implementation, and it saves users having to make their own package or class.

deckar01:
It's not clear from that discussion what was out of scope. It is possible it was the suggested implementation that was rejected rather than a bytes field in general.

I suspect there hasn't been much demand for bytes, because serialization formats generally enforce a specific encoding (like UTF-8) and don't support raw bytes without additional encoding (like base64). Where is the object containing bytes coming from in your use case?

rrad5409:
I'm using YAML config files, and I want to be able to specify a sequence of `bytes` for when I'm processing some binary data.

```py
# this is the python class
class Config:
  header: bytes
  footer: bytes
```

```yaml
# and here is how I want to write my config values
header: 0xFFAABBFF
footer: !!binary vu8=
```

I already have the YAML loader correctly doing the deserialisation and giving me the dictionary, but now I want marshmallow to allow me to pass that dictionary into the schema.
```py
# EDIT: `header` was mistakenly a byte string before, fixed
config = { 'header': 0xFFAABBFF, footer: b'\xBE\xEF' }
```

I wasn't trying to add support the decoding part, all I wanted was for marshmallow to _not_ try and interpret the bytes as a `str`, and just leave them as-is. I'm focusing on the schema field validation side, as opposed to the deserialisation side.

src/marshmallow/fields.py
```diff
@@ -878,6 +878,44 @@ def _deserialize(self, value, attr, data, **kwargs) -> str:
...
+                        s,
+                        encoding="utf-8",
+                        errors="ignore",
+                    )
+                case int() as i:
+                    return i.to_bytes(
+                        length=max(1, (7 + i.bit_length()) // 8),
+                        byteorder="big",
+                        signed=i < 0,
+                    )
```
deckar01, line 934: Marshmallow often leans on the builtin constructor for type coercion to keep the implementation simple, then excludes types that might be confusing.

```py
if isinstance(value, (bool, int)):
    raise ...
try:
    return bytes(value)
except TypeError:
    ...
```

Big int to bytes is probably out of scope.

src/marshmallow/fields.py
```diff
@@ -878,6 +878,44 @@ def _deserialize(self, value, attr, data, **kwargs) -> str:
...
+            match value:
+                case bytes() as b:
+                    return b
+                case bytearray() as ba:
+                    return bytes(ba)
+                case str() as s:
+                    return bytes(
+                        s,
+                        encoding="utf-8",
+                        errors="ignore",
```
deckar01: Silently discarding data should not be the default behavior.

deckar01 (commented):
The YAML use case seems like a good argument for supporting a bytes field. I would recommend passing the `encoding` and `errors` args through the field's constructor and preserving the builtin defaults.

src/marshmallow/fields.py
rrad5409: The problem is that YAML doesn't natively support writing a `bytes` value using hex/octal/binary notation, only encoded base64 strings. So I'm not able to write `val: 0xDEADBEEF`, get that deserialised to an `int`, and then convert it to `bytes` in the field. _(sorry I made an error in my sample earlier, the header should have been an `int` not `bytes)_

Especially when writing binary-related values (like registers, addresses, bitmasks etc), writing it as `0xABAA` or `0b11001010` is far more convenient to write and more understandable.

The same reasoning applies to `str` - being forced to encode a string using base64 is pure inconvenience, when I could just write `something: "Hello World"`.

src/marshmallow/fields.py
rrad5409: I did this because it's already a string, and hence we know that it's already valid. However, I'm not sure if my assumption is actually correct anymore - Rust strings are always UTF8, but I don't know about Python.

I read through the [unicode](https://docs.python.org/3/howto/unicode.html) and [encoding](https://docs.python.org/3/library/codecs.html#encodings-and-unicode) documentation, but they are unclear. [PEP 393](https://peps.python.org/pep-0393/) states they switch between multiple encodings (UCS-4, ASCII, etc) for efficiency.

Should I change the behaviour to `"error"` and wrap any unicode exceptions in `ValidationError`?

src/marshmallow/fields.py
```diff
@@ -878,6 +878,44 @@ def _deserialize(self, value, attr, data, **kwargs) -> str:
             raise self.make_error("invalid_utf8") from error
 
 
+class Bytes(Field[bytes]):
+    """
+    Marshmallow field type for any bytes array.
+    """
```
sloria, line 894:
```suggestion
    """A field for deserializing strings into byte arrays."""
```

src/marshmallow/fields.py
```diff
@@ -878,6 +878,44 @@ def _deserialize(self, value, attr, data, **kwargs) -> str:
...
+                case int() as i:
+                    return i.to_bytes(
+                        length=max(1, (7 + i.bit_length()) // 8),
+                        byteorder="big",
+                        signed=i < 0,
+                    )
+                case obj:
+                    if isinstance(obj, (typing.SupportsBytes, typing.Iterable)):
+                        return bytes(obj)
+            raise ValidationError("not a bytes-like object")
```
sloria: ValidationErrors use a sentence.

```suggestion
            raise ValidationError("Not a bytes-like object.")
```

though you should define `default_error_messages` and use `self.make_error("invalid")` as in the other fields.

src/marshmallow/fields.py
sloria: > Should I change the behaviour to "error" and wrap any unicode exceptions in ValidationError?

this seems right to me

src/marshmallow/fields.py
```diff
@@ -878,6 +878,44 @@ def _deserialize(self, value, attr, data, **kwargs) -> str:
             raise self.make_error("invalid_utf8") from error
 
 
+class Bytes(Field[bytes]):
```
sloria, line 883: Make sure to add this to `__all__` at the top of the file

src/marshmallow/fields.py
sloria: i also lean towards omitting this. i plan to finish up and release [field-level pre_load](https://github.com/marshmallow-code/marshmallow/pull/2799) soon, which should simplify your use case here.

```python
  def int_to_bytes(value):
      if isinstance(value, int):
          return value.to_bytes(
              length=max(1, (7 + value.bit_length()) // 8),
              byteorder="big",
              signed=value < 0,
          )
      return value

class MySchema(Schema):
    foo = fields.Bytes(pre_load=int_to_bytes)
```

so to build off of @deckar01's suggestion, i think the final implementation should look something like:

```python
  def _deserialize(self, value, attr, data, **kwargs) -> bytes:
      if isinstance(value, (bool, int)):
          raise self.make_error("invalid")
      if isinstance(value, str):
          try:
              return value.encode("utf-8")
          except UnicodeEncodeError as error:
      try:
          return bytes(value)
      except TypeError as error:
```

sloria (commented):
Some driveby suggestions. 

shouldn't this also have serialization behavior? for one, msgspec [serializes to base64](https://jcristharif.com/msgspec/supported-types.html#bytes-bytearray-memoryview). should we do the same?

rrad5409:
I added the `encoding` and `errors` fields for the `str` constructor, and pass them through to it.

I also added the serialisation behaviour as requested, and I included another parameter to allow for different return types for flexibility. So it can return `bytes`, `int`, `str` using the above encoding, or `str` as base64. I can also add deserialisation for base64 if needed, I just wasn't sure if it was wanted.

I'd really like to keep the `int` deserialisation if possible, because it's super useful, and should be completely transparent if it's not needed.

src/marshmallow/fields.py
```diff
@@ -878,6 +880,95 @@ def _deserialize(self, value, attr, data, **kwargs) -> str:
...
+                    return int.from_bytes(
+                        value,
+                        byteorder="big",
+                    )
+                case "bytes":
+                    return value
+                case _:
+                    typing.assert_never(self.serialize)
+        except UnicodeError as e:
+            raise self.make_error("unicode") from e
```
sloria, line 969: serialization should not perform validation, since deserialized values are assumed to be valid. this was an [explicit design choice](https://marshmallow.readthedocs.io/en/latest/upgrading.html#validation-does-not-occur-on-serialization).

sloria (commented):
sorry for the delay. i'm still ok with adding this, but i think the remaining work is making sure this stays consistent with the rest of the fields. for example the `serialize` doesn't exist on other fields. we have `format` on time-related fields, but that applies to both deserialization and serialization. i'm hesitant to add a new pattern.

maybe it makes sense to drop `serialize` and rely on the recently-added field-level `pre_load`/`post_dump` API rather than hardcoding a set of formats here.

```python
class MySchema(Schema):
    data = fields.Bytes(
        pre_load=int_to_bytes, 
        post_dump=base64.standard_b64encode,
    )
```

that will significantly reduce the added complexity and give this a better chance of landing.

src/marshmallow/fields.py
rrad5409: Oh this was me just wrapping the error in case the `str()` call threw it. I missed that part of the docs sorry.

rrad5409:
So you are proposing that we pass in the transformation function into the field constructor, as opposed to passing in a parameter that then gets selected inside the field serialiser? That's actually be a lot simpler, I like that.

Would that also remove the necessity for the `encoding` and `errors` parameters, as we'd be using `pre_load = functools.partial(str, encoding=..., errors=...)`?

sloria:
> Would that also remove the necessity for the encoding and errors parameters, as we'd be using pre_load = functools.partial(str, encoding=..., errors=...)?

yes, i think so 👍

PR #2923, opened 2026-03-19
Handle TypeError in _get_value_for_key when key is non-string

Hassad674:
Fixes #2893

When get_value receives an out-of-range int index for a list, the IndexError fallback calls getattr(obj, key, default) which raises TypeError because getattr requires a string attribute name. The same happens with dict int keys that don't exist.

Wrapped the getattr fallback in a try/except TypeError so it returns the default value instead of crashing. Added tests for the list and dict cases from the issue.


deckar01:
You seem to have fallen into an AI honeypot. Reported.

https://github.com/marshmallow-code/marshmallow/issues/2893#issuecomment-4058717845

PR #2927, opened 2026-03-26
Add IDN (Internationalized Domain Name) support to URL validator

### LINKED ISSUE #2821: There is not way to make `marshmallow.validate.URL` accept IDNs (Internationalized Domain Names)
IDN: https://en.wikipedia.org/wiki/Internationalized_domain_name

Valid URLs like https://তৌহিদুর.বাংলা are always rejected and there is not way to fix this.

rstar327:
Add support for Internationalized Domain Names (IDN) in the `URL` validator.

URLs with Unicode hostnames like `https://তৌহিদুর.বাংলা` and `https://münchen.de` are currently always rejected because the hostname regex only accepts ASCII characters.

The fix converts Unicode hostnames to punycode (IDNA encoding) before regex validation using Python's built-in `str.encode('idna')`. The original URL value is preserved and returned — only the validation step uses the encoded form.

All 1149 existing tests pass.

Fixes #2821

### DIFF
--- a/src/marshmallow/validate.py
+++ b/src/marshmallow/validate.py
@@ -195,6 +195,46 @@ class URL(Validator):
     def _format_error(self, value) -> str:
         return self.error.format(input=value)
 
+    @staticmethod
+    def _encode_idn(value: str) -> str:
+        """Encode internationalized domain names (IDN) to punycode.
+
+        Converts Unicode hostnames to their ASCII-compatible encoding
+        so that the existing ASCII-only regex can validate them.
+        """
+        try:
+            from urllib.parse import urlsplit, urlunsplit
+
+            try:
+                parsed = urlsplit(value)
+            except ValueError:
+                return value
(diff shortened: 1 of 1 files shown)


sloria:
@rstar327 are you a user of marshmallow and do you have any stake in this particular feature being implemented?

rstar327:
> @rstar327 are you a user of marshmallow and do you have any stake in this particular feature being implemented?

Not a heavy user, but I saw the open issue and the fix was clean, just IDNA encoding before the existing regex. Happy to take feedback on the approach.

sloria:
looking into this a bit more, i think we can take Django's simpler approach which is to relax the regex to allow unicode chars. See https://github.com/django/django/blob/f6167b8bc881babd19b67c004e8f37954afc192e/django/core/validators.py#L71-L85 . I've made a PR with this approach in #2928

PR #2930, opened 2026-03-30
Migrate to uv and use dependency groups

sloria:
given the recent high-profile supply chain attacks, seems a good time as any to lock our dev environment deps 😬 

to migrate your current dev environment from pip/venv to uv:

```
deactivate
rm -rf .venv
uv sync
uv run pre-commit install --allow-missing-config
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,9 +1,12 @@
 version: 2
 updates:
-- package-ecosystem: pip
+- package-ecosystem: uv
   directory: "/"
   schedule:
-    interval: daily
+    interval: monthly
+  groups:
+    all-dependencies:
+      patterns: ["*"]
   open-pull-requests-limit: 10
 - package-ecosystem: "github-actions"
   directory: "/"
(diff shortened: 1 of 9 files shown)


.github/dependabot.yml
```diff
@@ -1,9 +1,12 @@
 version: 2
 updates:
-- package-ecosystem: pip
+- package-ecosystem: uv
   directory: "/"
   schedule:
-    interval: daily
+    interval: monthly
```
sloria, line 6: so we don't get too many lockfile update PRs

lafrech:
I'll be ooo for a few days and I have no prior knowlege of uv, so don't
wait for my review.

sloria:
@lafrech sounds good. just didn't want to change too much tooling without your awareness. sounds like you're good with this moving forward without your review though, so i'll just let this sit for a day or so in case Jared or Stephen want to take a look

deckar01:
`uv run tox` on a fresh clone passed. :+1:

PR #2933, opened 2026-04-01
fix: return default for out-of-range int key in get_value

nuglifeleoji:
Fixes #2893.

## Problem

When an integer `key` is passed to `utils.get_value` and `obj[key]`
raises an exception (e.g. `IndexError` for a list, `KeyError` for a
dict with integer keys), the code falls back to
`getattr(obj, key, default)`. However, `getattr()` requires the
attribute name to be a string, so it raises:

```
TypeError: getattr(): attribute name must be string
```

## Fix

In `_get_value_for_key`, guard the `getattr` fallback with
`isinstance(key, str)`. When the key is not a string, return `default`
directly instead of calling `getattr`.

## Examples (now work correctly)

```python
# Out-of-range integer index → returns default
utils.get_value([0, 1, 2], 999, default=42)   # 42

# Missing integer key in dict → returns default
utils.get_value({1: "a"}, 99, default="z")     # "z"
```

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -120,11 +120,15 @@ def _get_value_for_keys(obj, keys, default):
 
 def _get_value_for_key(obj, key, default):
     if not hasattr(obj, "__getitem__"):
+        if not isinstance(key, str):
+            return default
         return getattr(obj, key, default)
 
     try:
         return obj[key]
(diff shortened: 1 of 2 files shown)


sloria:
[looks familiar](https://github.com/marshmallow-code/marshmallow/pull/2917#issuecomment-4056580524)

PR #2937, opened 2026-04-02
Fix validate.Email to accept IDNs

### LINKED ISSUE #2936: Update Email validator to support IDNs
Follow-up to [#2821](https://github.com/marshmallow-code/marshmallow/issues/2821#issuecomment-4174424998): fix support for IDNs in the `Email` validator

sloria:
closes #2936 
same approach as https://github.com/marshmallow-code/marshmallow/pull/2928

### DIFF
--- a/src/marshmallow/validate.py
+++ b/src/marshmallow/validate.py
@@ -13,6 +13,7 @@ if typing.TYPE_CHECKING:
     from marshmallow import types
 
 _T = typing.TypeVar("_T")
+_UNICODE_LETTERS = "\u00a1-\uffff"
 
 
 class Validator(ABC):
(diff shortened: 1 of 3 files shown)


src/marshmallow/validate.py
```diff
@@ -292,13 +295,6 @@ def __call__(self, value: str) -> str:
 
         if domain_part not in self.DOMAIN_WHITELIST:
             if not self.DOMAIN_REGEX.match(domain_part):
-                try:
-                    domain_part = domain_part.encode("idna").decode("ascii")
```
sloria, line 296: this fallback is no longer necessary and previously resulted in a bug with accepting invalid IDN emails with leading or trailing hyphens

PR #2944, opened 2026-04-07
Security: Shared ContextVar causes context type confusion across all Context parameterizations

barttran2k:
## Problem

The `_CURRENT_CONTEXT` ContextVar is a single module-level variable shared by all `Context[T]` parameterizations. Since `get()` is a classmethod that always reads from the same ContextVar, nesting different Context types causes the inner context to silently overwrite the outer one. For example, `Context[AuthContext]` and `Context[UserContext]` share state, so `with Context[AuthContext](auth): with Context[UserContext](user): Context[AuthContext].get()` would return the `user` value, not `auth`. This can lead to authorization bypass if contexts are used for access control decisions.

**Severity**: `medium`
**File**: `src/marshmallow/experimental/context.py`

## Solution

Use a per-class ContextVar by storing it as a class variable initialized in `__init_subclass__` or using a dictionary keyed by the concrete class, so each parameterized Context type has its own isolated ContextVar.

## Changes

- `src/marshmallow/experimental/context.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/src/marshmallow/experimental/context.py
+++ b/src/marshmallow/experimental/context.py
@@ -42,6 +42,7 @@ except ImportError:  # Python<3.10
 _ContextT = typing.TypeVar("_ContextT")
 _DefaultT = typing.TypeVar("_DefaultT")
 _CURRENT_CONTEXT: contextvars.ContextVar = contextvars.ContextVar("context")
+_CONTEXT_CLASSES: dict[typing.Any, type] = {}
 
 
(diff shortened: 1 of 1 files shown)


deckar01:
If a human finds anything useful in this PR, please cherry pick it.

PR #2945, opened 2026-04-07
Security: Thread-unsafe global mutable class registry

barttran2k:
## Problem

The `_registry` module-level dictionary in `class_registry.py` is modified by the `register()` function without any thread synchronization (e.g., locks). The `register()` function performs check-then-act operations (e.g., `if classname in _registry ... _registry[classname].append(cls)`) that are not atomic. In multi-threaded applications where schemas are defined dynamically, this can lead to race conditions, lost registry entries, or inconsistent state.

**Severity**: `low`
**File**: `src/marshmallow/class_registry.py`

## Solution

Add a threading.Lock to protect read-modify-write operations on `_registry`, or document that schema class registration must happen at import time before spawning threads.

## Changes

- `src/marshmallow/class_registry.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/src/marshmallow/class_registry.py
+++ b/src/marshmallow/class_registry.py
@@ -11,6 +11,7 @@ class:`fields.Nested <marshmallow.fields.Nested>`.
 
 from __future__ import annotations
 
+import threading
 import typing
 
 from marshmallow.exceptions import RegistryError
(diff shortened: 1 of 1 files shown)


deckar01:
If a human finds anything useful in this PR, please cherry pick it.

sloria:
this isn't a security issue. no user input touches the registry (dynamic creation with `Schema.from_dict` bypasses it).

@barttran2k please go tell your owner to stop wasting maintainers' time by sending unreviewed "security" PRs

PR #2947, opened 2026-04-13
Switch to renovate

sloria:
Renovate has sensible defaults and allows sharing for sharing configs across repos easily. This uses the config I made [here](https://github.com/marshmallow-code/renovate-config/blob/main/default.json). The behavior is very similar to the current dependabot behavior (monthly grouped PRs), with the addition of

- GHA digest pinning
- 7-day [dependency cooldown](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns)

If no objections to this in the next ~48 hours, I'll go ahead and merge this and add the same setup to the rest of the marshmallow-code repos


lafrech:
No objection.

Note that dependabot also provides [cooldown](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-).

One feature I miss from old dependabot is automerge. Looks like renovate has automerge.

sloria:
> Note that dependabot also provides [cooldown](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-).

i know. the main benefit of renovate for our use case is the shared org-level config. the ancillary benefits are the built-in presets (keeps config small), cross-ecosystem grouping (keeps PR noise down), and maturity of features that are relevant to us (digest pinning and uv support were only recently added to dependabot, whereas these have been in renovate for a long time).

> One feature I miss from old dependabot is automerge. Looks like renovate has automerge.

i haven't enabled this as part of the migration. given the PRs will only be monthly, i think the review burden should be quite low

lafrech:
OK, I see.

:+1:  then.

PR #2951, opened 2026-04-13
Fix: ensure metadata is defensively copied in Field

### LINKED ISSUE #2949: Field(metadata=...) keeps caller mapping by reference (copy semantics changed since #2701)
### Summary
 After #2701 ("Remove passing metadata as kwargs"), `Field.__init__`
 now assigns `self.metadata = metadata` directly.

 This changes behavior from defensive-copy semantics to
reference/alias semantics for `metadata=...`.

 ### Why this is surprising
 Before #2701, metadata was built with:
 ```python
 self.metadata = {**metadata, **additional_metadata}

which created a new dict.

Now:

 metadata = metadata or {}
 self.metadata = metadata

so later mutations to the caller-provided mapping are reflected in
field.metadata.

Repro (latest release)

 import importlib.metadata as md


 meta = {"k": 1}
 f = fields.Field(metadata=meta)
 print(f.metadata is meta)  # True

 meta["k"] = 2
 print(f.metadata["k"])     # 2 (changed externally)

Expected

Field(metadata=...) should not be affected by later external
mutation of the caller mapping (or this behavioral change should be
explicitly documented as intentional).

Actual

field.metadata aliases caller mapping and changes when caller
mutates it.

If preserving prior semantics is desired while keeping #2701's
kwargs removal, use a defensive copy, e.g.:

 metadata = metadata or {}
 self.metadata = dict(metadata)

(or equivalent behavior-preserving approach for mappings).

lequangdung2005:
Description: This PR ensures that metadata passed to a Field is defensively copied. Previously, it was assigned by reference, causing external mutations to the original dictionary to affect the field's metadata (a regression from #2701).

Fixes #2949

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -230,7 +230,7 @@ class Field(typing.Generic[_InternalT]):
         self.required = required
 
         metadata = metadata or {}
-        self.metadata = metadata
+        self.metadata = dict(metadata)
         # Collect default error message from self and parent classes
         messages: types.ErrorMessages = {}
         for cls in reversed(self.__class__.__mro__):
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -245,7 +245,37 @@ class TestMetadata:
             metadata={"description": "foo", "widget": "select"},
         )
         assert field.metadata == {"description": "foo", "widget": "select"}
+    def test_metadata_is_defensively_copied(self):
+        original_meta = {"key1": "value1", "key2": "value2"}
+        field = fields.Field(metadata=original_meta)
 
+        original_meta["key1"] = "modified_value"
+        original_meta["key3"] = "new_key"
+
+        assert field.metadata.get("key1") == "value1"
+        assert "key3" not in field.metadata
+
+    def test_empty_metadata_is_defensively_copied(self):
+        empty_meta = {}
+        field = fields.Field(metadata=empty_meta)
+
+        empty_meta["injected"] = "value"
+
+        assert "injected" not in field.metadata
+
+    def test_metadata_defensive_copy_with_other_params(self):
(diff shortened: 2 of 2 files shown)


sloria:
i'm lukewarm on this. for one, it doesn't prevent deeply-nested objects from getting mutated. but i'm not even sure users should rely on libraries to copy data like this.

did you run into a real-world use case where this is an issue?

lequangdung2005:
Thanks for the feedback — totally fair concern. 

I agree this doesn’t protect deep nested objects; this PR is only meant to restore the prior top-level isolation that existed before #2701. 

I hit this in a real case with shared OpenAPI metadata (`apispec`): reusing one base `metadata` dict across fields, then changing a top-level key like `description` or `example` for one field, unexpectedly changed others due to aliasing. 

Since `Field` instances are long-lived, a shallow defensive copy (`dict(metadata)`) avoids that footgun at negligible cost and restores predictable behavior.

sloria:
the previous copy behavior via (`self.metadata = {**metadata, **additional_metadata}`) was incidental rather than a deliberate choice/API guarantee. generally one shouldn't mutate dicts that are passed down.

the closest analogue is sqlalchemy's `Column(info={...})`, which doesn't copy: https://github.com/sqlalchemy/sqlalchemy/blob/d3a85fbd07fbd88e5872df39290db050c0a6f0a9/lib/sqlalchemy/sql/schema.py#L992 .

gonna pass on this one. thanks anyway though

lafrech:
Good point about the deepcopy.
I was lukewarm too but had no absolute argument.
I agree the user should be responsible for not mutating (or (deep)copying when passing the argument).
My concern was that if we engage in this we could be asked to do the same everywhere and even deepcopy. On second thought this sounds like bad practice.

PR #2957, opened 2026-04-22
fix(utils): handle TypeError from getattr in get_value for non-string keys

armorbreak001:
Fixes #2893

When `obj[key]` raises `IndexError` or `TypeError` for an out-of-range int key, the fallback to `getattr(obj, key, default)` also raises `TypeError` because attribute names must be strings. Wrap the `getattr` call in a try/except to catch this and return the default value instead.


lafrech:
image

PR #2958, opened 2026-04-25
fix: utils.get_value return default for out-of-range int keys

Jah-yee:
Fixes marshmallow-code/marshmallow#2893

Handle TypeError from getattr when key is an int and obj has no such attribute (e.g., list with out-of-range index, dict with missing int key).


lafrech:
image

Jah-yee:
PR still active and ready for review. 🙏

PR #2963, opened 2026-05-03
Refactor: don't pass attr, obj, data to fields (de)serialize methods

### LINKED ISSUE #2039: [RFC] Don't pass attr, obj, data to fields (de)serialize methods

lafrech:
#2039 is a long standing issue I've been wanting to address for a while but it implies a lot of changes that can hardly be done incrementally so I never got down to it.

[Disclaimer: The company I work for wants us to leverage AI power because everyone does it (FOMO) so we asked for Claude tokens and I tried agent coding on our projects to give management feedback about it and I figured why not try it on this issue as well. Obviously, it is not as simple as telling the tool to fix #2039. It took me a few tries to get the steps in the right order. And although I was supervising, there may be a few hallucinations I haven't spotted yet. The tool was not always clever at guessing what I didn't tell, but it helped me facing the fear of rewriting all calls to serialization functions in lib and tests.]

I'm sharing this as it is a nice POC. It still needs to be polished around the edges and there may be possibilities for further improvements, but some of them could be done later. Changes certainly need to be double-checked (but this would also be true if I had done all this manually, TBH). Consider it an illustration of what I have in mind.

The [de|]serialization_getter arguments added to `Field` could be added in MA4 while deprecating the use of `Function`/`Method` to do fancy accessing. The rest is breaking changes that would have to wait until 5.0.

Happy to get feedback before doing any more work on this.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -2019,8 +2019,8 @@ class Method(Field):
     """A field that takes the value returned by a `Schema <marshmallow.Schema>` method.
 
     :param serialize: The name of the Schema method from which
-        to retrieve the value. The method must take an argument ``obj``
-        (in addition to self) that is the object to be serialized.
+        to retrieve the value. The method must take an argument ``value``
+        (in addition to self) that is the value to be serialized.
     :param deserialize: Optional name of the Schema method for deserializing
         a value. The method must take a single argument ``value``, which is the
         value to deserialize.
@@ -2028,8 +2028,6 @@ class Method(Field):
     .. versionchanged:: 3.0.0
         Removed ``method_name`` parameter.
     """
-
-    _CHECK_ATTRIBUTE = False
 
     def __init__(
         self,
@@ -2061,7 +2059,7 @@ class Method(Field):
 
     def _serialize(self, value, attr, obj, **kwargs):
         if self._serialize_method is not None:
-            return self._serialize_method(obj)
+            return self._serialize_method(value)
         return missing_
 
     def _deserialize(self, value, attr, data, **kwargs):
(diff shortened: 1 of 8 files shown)


lafrech:
I don't remember the rationale for not having mypy run in pre-commit.

sloria:
only skimmed the code changes (haven't read the tests), but this is looking really nice. i do think this is the right direction overall. maybe rename to `load_getter` and `dump_getter` for consistency.

> The [de|]serialization_getter arguments added to Field could be added in MA4 while deprecating the use of Function/Method to do fancy accessing. The rest is breaking changes that would have to wait until 5.0.

👍 that seems like a good release plan

lafrech:
> maybe rename to `load_getter` and `dump_getter` for consistency.

Yes, definitely. If only for the shortness of it. I did the renaming.

The first two commits should be non-breaking and could be copied into another PR to make it to MA 4.x once docs are added.

The third (`Constant` rework) should be non-breaking as well, although there may not be a huge interest in adding this to 4.x. It can be done while modifying `Function` and `Method` in 5.0.

PR #2971, opened 2026-05-28
fix: return default instead of raising TypeError for out-of-range int index in get_value (#2893)

MohammadYusif:
## Summary

- `utils.get_value(lst, 999, default=3)` raised `TypeError: getattr(): attribute name must be string` instead of returning `3`
- Root cause: `_get_value_for_key` catches `IndexError` from an out-of-range integer index then falls back to `getattr(obj, key, default)` — `getattr` only accepts string attribute names, so integer keys cause it to raise
- Fix adds an `isinstance(key, int)` guard in the except block so integer keys return `default` directly without calling `getattr`

## Changes

- `src/marshmallow/utils.py`: return `default` immediately when `key` is an `int` and the subscript lookup fails
- `tests/test_utils.py`: added `test_get_value_out_of_range_int_index_returns_default` covering out-of-range list index, missing integer dict key, and empty list

Fixes #2893

### DIFF
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -125,6 +125,8 @@ def _get_value_for_key(obj, key, default):
     try:
         return obj[key]
     except (KeyError, IndexError, TypeError, AttributeError):
+        if isinstance(key, int):
+            return default
         return getattr(obj, key, default)
 
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -79,6 +79,21 @@ def test_get_value():
     assert utils.get_value(lst, MyInt(1)) == 2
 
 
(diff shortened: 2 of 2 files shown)


lafrech:
@MohammadYusif you obviously mean to be very helpful. Do you think you could try to reach out one way or another to the people developing the AI tool you're using to ask them to teach that tool to read the conversations below the issues the tool pretends to address? There's a blatant shortcoming in the way AI browses bugtrackers.

MohammadYusif:
I am actually utilizing my own tool that I made myself to eventually create something robust. I appreciate the critisim. And will be used to enhance my tool's way of doing research.

Thank you.

PR #2979, opened 2026-06-10
Materialize single-use iterables in OneOf/NoneOf validators

IgnazioDS:
`OneOf` and `NoneOf` are annotated to accept any `typing.Iterable`, but they store the argument directly and consume it once while building the error text (`choices_text` / `values_text`), then re-iterate it in `__call__`:

```python
def __init__(self, choices: typing.Iterable, ...):
    self.choices = choices
    self.choices_text = ", ".join(str(c) for c in self.choices)  # consumes a generator
...
def __call__(self, value):
    if value not in self.choices:  # re-iterates -- empty if a generator was passed
        raise ValidationError(...)
```

So passing a generator silently breaks validation:

```python
>>> OneOf(iter([1, 2, 3]))(2)          # rejects a valid value
ValidationError: ['Must be one of: 1, 2, 3.']
>>> NoneOf(iter([1, 2, 3]))(1)         # accepts an invalid value (returns it)
1
```

`OneOf` rejects everything; `NoneOf` accepts everything.

## Fix

Materialize only true single-use iterators (detected via `iter(x) is x`) into a list, leaving sequences and strings untouched so their membership semantics are preserved — e.g. `OneOf("abc")` keeps treating `"abc"` as a string (`"b" in "abc"`) rather than `["a", "b", "c"]`. Regression tests added for both validators.

Full test suite passes (1180 passed); `ruff check`/`format` clean.

### DIFF
--- a/src/marshmallow/validate.py
+++ b/src/marshmallow/validate.py
@@ -566,7 +566,11 @@ class NoneOf(Validator):
     default_message = "Invalid input."
 
     def __init__(self, iterable: typing.Iterable, *, error: str | None = None):
-        self.iterable = iterable
+        # Materialise single-use iterators (e.g. generators) so they are not
+        # exhausted while building ``values_text`` and can be reused in
+        # ``__call__``. Sequences and strings are kept as-is to preserve their
+        # membership semantics (e.g. substring checks for ``str``).
+        self.iterable = list(iterable) if iter(iterable) is iterable else iterable
         self.values_text = ", ".join(str(each) for each in self.iterable)
         self.error: str = error or self.default_message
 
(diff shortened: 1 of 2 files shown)


lafrech:
Thanks for contributing.
Seems like an easy fix but I wonder whether we should open that door of just state explicitly we don't accept generators.
Regarding the implem, I find the comments a bit too claudishly verbose.

IgnazioDS:
Fair point, I can go either way. The thing that bugs me is that today it fails silently: OneOf with a generator rejects every value and NoneOf accepts every value, with no error, so it's easy to hit and annoying to debug. Since the params are already typed Iterable, materialising one-shot iterators felt like the smallest fix that keeps str and sequence behaviour identical.

If you'd rather not support generators at all, happy to switch it to narrowing the type to Sequence, or raising a clear error so it at least fails loudly. Whatever you prefer.

Comments trimmed.

lafrech:
I'll let other maintainers express their opinion.

deckar01:
We have previously used `utils.is_iterable_but_not_string()` for list casts.

I don’t think this code needs any comments, nor do the tests. The regression test documents the intention, and the test name is usually sufficient to explain the behavior.

deckar01:
We should consider using set rather than list for O(log n) `in` lookups. I don’t think we would need a type check for that since it would handle strings correctly.

deckar01:
> Full test suite passes (1180 passed); ruff check/format clean.

This account has opened 5 PRs and pushed ~100 commits to their own repos today. Looks like an unsupervised bot.

sloria:
>  Looks like an unsupervised bot.

certainly does. let's pass on this one. if there's a real user with a real use case that needs this, we can reconsider

PR #2983, opened 2026-06-18
Respect uv.lock in tox envs

### LINKED ISSUE #2981: Use uv.lock in tox testenv
See discussion in https://github.com/marshmallow-code/flask-smorest/pull/860.

Creating this issue so we keep track of it here rather than in a closed PR in flask-smorest.

@sirosen wrote:

> There's a separate config we need to use in any project where we want to use `uv.lock` in a tox testenv:
https://github.com/tox-dev/tox-uv/tree/main#uvlock-support
> 
> I'd be game to turn this on anywhere, but recommend [setting `requires`](https://tox.wiki/en/stable/reference/config.html#requires) in tox config if we do, to make sure that `tox-uv` is always installed.

sloria:
using the uv-venv-lock-runner
NOTE: dev-only envs that aren't run in CI like `docs-serve` still keep unpinned deps and use `runner = uv-venv-runner` to keep the old behavior

closes https://github.com/marshmallow-code/marshmallow/issues/2981

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,15 @@ docs = [
   "sphinxext-opengraph",
 ]
 tests = ["pytest", "simplejson"]
-dev = [{ include-group = "tests" }, "tox", "tox-uv", "pre-commit>=3.5,<5.0"]
+lint = ["pre-commit>=3.5,<5.0"]
+mypy = ["mypy>=1.14.1", "types-simplejson"]
+dev = [
+  { include-group = "tests" },
+  { include-group = "lint" },
+  { include-group = "mypy" },
+  "tox",
+  "tox-uv",
+]
 
 [tool.uv]
 default-groups = ["dev"]
--- a/tox.ini
+++ b/tox.ini
@@ -1,19 +1,19 @@
 [tox]
+requires = tox-uv
 envlist = lint,mypy,py{310,311,312,313,314},docs
 
 [testenv]
+runner = uv-venv-lock-runner
 dependency_groups = tests
 commands = pytest {posargs}
 
 [testenv:lint]
-deps = pre-commit>=3.5,<5.0
+dependency_groups = lint
 skip_install = true
 commands = pre-commit run --all-files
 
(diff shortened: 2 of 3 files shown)


tox.ini
```diff
@@ -23,15 +23,18 @@ commands = sphinx-build --fresh-env docs/ docs/_build {posargs}
 ; Below tasks are for development only (not run in CI)
 
 [testenv:docs-serve]
+runner = uv-venv-runner
```
sloria, line 26: I'm keeping these dev-only envs on the old behavior, which is to leave these CLIs (sphinx-autobuild, restview) unpinned.

i'm not 100% this is the right thing to do...mostly following [attr's lead](https://github.com/python-attrs/attrs/blob/89fae8300f484544c1b7678cea5efe58c551fbb9/tox.ini#L103-L108) here

sloria:
attn @lafrech @sirosen if you want to take a look. will plan to merge on Saturday if no objections

pyproject.toml
```diff
@@ -43,7 +43,15 @@ docs = [
   "sphinxext-opengraph",
 ]
 tests = ["pytest", "simplejson"]
-dev = [{ include-group = "tests" }, "tox", "tox-uv", "pre-commit>=3.5,<5.0"]
+lint = ["pre-commit>=3.5,<5.0"]
+mypy = ["mypy>=1.14.1", "types-simplejson"]
```
sloria, line 47: diff makes this change look bigger than it actually is cuz of autoformatting. `lint` and `mypy` are the relevant additions here

lafrech (approved):
Looks good. Thanks for tackling this.

No opinion either about the dev-only envs but following attr is good enough for me.

tox.ini
sirosen: 👍 I think this is good or at least okay.

I wonder if there's a way to setup a dependency cooldown for these... a good project for me for later. 😉

sirosen (approved):

PR #2986, opened 2026-06-24
fix: Pluck._serialize crashes with KeyError on missing nested field

gaoflow:
## Problem

`Pluck._serialize` raises a bare `KeyError` when the plucked field is absent from the nested object. This happens for both `many=False` and `many=True`.

**Reproducer:**
```python

class AuthorSchema(Schema):
    id = fields.Int()
    name = fields.Str()

class BookSchema(Schema):
    title = fields.Str()
    author_name = fields.Pluck(AuthorSchema, "name")

s = BookSchema()
s.dump({"title": "Test", "author_name": {"id": 1}})  # KeyError: 'name'
```

The crash occurs because both code paths use `dict[key]` instead of `dict.get(key)`:
- Non-many (fields.py:721): `ret[self._field_data_key]`
- Many (utils.py:91): `d[key]` inside `utils.pluck`

Setting `dump_default` on the Pluck field does not help, because the base
`Field.serialize` only applies it when the entire nested object is missing,
not when a specific field within it is absent.

## Fix

- `Pluck._serialize`: use `ret.get(self._field_data_key, missing_)` so the field is gracefully omitted when the key is absent.
- `utils.pluck`: use `d.get(key)` so individual items with missing keys produce `None` instead of crashing.

## Behavior after fix

- **Non-many**: field is omitted from serialized output (consistent with other missing-value behavior).
- **Many**: missing keys produce `None` in the output list.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -718,7 +718,7 @@ class Pluck(Nested):
             return None
         if self.many:
             return utils.pluck(ret, key=self._field_data_key)
-        return ret[self._field_data_key]
+        return ret.get(self._field_data_key, missing_)
 
     def _deserialize(self, value, attr, data, partial=None, **kwargs):
         self._test_collection(value)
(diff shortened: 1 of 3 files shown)


lafrech:
This might be a case of "data being serialized should be valid, fields are not meant to deal with invalid data".

PR #2987, opened 2026-06-25
Add skip_on_field_errors to field validators

### LINKED ISSUE #2961: Inconsistent value type when validating nested schema fields
**Problem:**

When a Nested schema has a validation error its `@post_load` hooks are not invoked bypassing any transformations done there. This means the parent schema’s `@validates` and `@validate_schema` hooks have to deal with both the transformed and untransformed child values.

Originally I encountered this using `marshmallow-sqlalchemy` with `load_instance=True`.

**Steps to reproduce:**

The code below shows the problem.

```python
import dataclasses
import uuid

import marshmallow as ma
import pytest


@dataclasses.dataclass
class ChildModel:
    uuid: uuid.UUID
    name: str


class Child(ma.Schema):
    uuid = ma.fields.UUID(required=True)
    name = ma.fields.String(required=True)

    @ma.post_load
    def transform(self, data, **_kwargs):
        return ChildModel(**data)


class Parent(ma.Schema):
    child = ma.fields.Nested("Child")

    @ma.validates("child")
    def validate_child(self, value, **_kwargs):
        if value:
            # This fails when the child has a validation error as then its @post_load is not called.
            assert isinstance(value, ChildModel)


def test_load_valid_uuid():
    data = {"child": {"uuid": "c81505b4-258b-4912-b8bc-8ac913d56736", "name": "test"}}
    result = Parent().load(data)
    assert isinstance(result["child"], ChildModel)


def test_load_invalid_uuid():
    data = {"child": {"uuid": "invalid-uuid", "name": "test"}}
    with pytest.raises(ma.exceptions.ValidationError):
        result = Parent().load(data)
```

When the child schema validates then `Parent.validate_child` is called with a ChildModel instance, but if the child data is invalid it gets called with a dict instance instead.

lin-hongkuan:
Fixes #2961.

## Summary
- add an opt-in `skip_on_field_errors` parameter to `@validates`, matching the existing schema-validator option while preserving the current default behavior
- skip only validators for fields that already have field errors, so validators for other valid fields can still run
- add regression coverage for nested fields whose `post_load` transformation should not be bypassed into a parent field validator when the nested field already has errors

## Tests
- `uv run --with tzdata pytest tests\test_decorators.py tests\test_schema.py -q`
- `uv run ruff check src\marshmallow\decorators.py src\marshmallow\schema.py tests\test_decorators.py tests\test_schema.py`
- `uv run ruff format --check src\marshmallow\decorators.py src\marshmallow\schema.py tests\test_decorators.py tests\test_schema.py`
- `uv run mypy --show-error-codes`

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -14,6 +14,7 @@ from marshmallow import (
     Schema,
     class_registry,
     fields,
+    post_load,
     validate,
     validates,
     validates_schema,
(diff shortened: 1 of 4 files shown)


tests/test_schema.py
```diff
@@ -1991,6 +1992,48 @@ def validates_inner(self, data, **kwargs):
...
+
+        validated_values = []
+
+        class Outer(Schema):
+            inner = fields.Nested(Inner)
+
+            @validates("inner", skip_on_field_errors=True)
+            def validates_inner(self, value, **kwargs):
+                assert isinstance(value, ChildModel)
+                validated_values.append(value)
```
lafrech, line 2018: This doesn't seem useful to me.
The `assert isinstance` is enough to demonstrate we went there only on valid input.

lafrech (changes requested):
Thanks for contributing.

This looks good to me.

Just a nitpick about the test.

PR #2989, opened 2026-06-30
Fix schema validators after collection pre-load changes length

nyxst4ck:
## Summary

- Avoid zipping deserialized `many=True` items with the original raw collection when a per-item `@validates_schema` hook did not request `pass_original`.
- Add a regression test for `@pre_load(pass_collection=True)` changing collection length before per-item schema validation.
- Add `@nyxst4ck` to `AUTHORS.rst` per the contribution guidelines.

## Root cause

For `many=True` schema validation, `_invoke_schema_validators` always used `zip(data, original_data, strict=True)` for per-item validators. If a collection-level pre-load hook filtered or otherwise changed the number of items, per-item schema validators that did not use `pass_original` still failed with `ValueError` from the strict zip, even though `original_data` is unused in that case.

## Test Plan

- `uv run pytest tests/test_decorators.py -k collection_pre_load_can_change_length_before_item_schema_validation -q`
- `uv run pytest tests/test_decorators.py -q`
- `uv run pytest -q -k "not test_from_timestamp_with_overflow_value"`
- `uv run tox -e lint`

Also ran `uv run pytest -q`; it had one unrelated Windows/Python assertion failure in `tests/test_utils.py::test_from_timestamp_with_overflow_value` because the platform error message was `Error converting value to datetime` instead of the expected regex.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -188,3 +188,4 @@ Contributors (chronological)
 - `@rstar327 <https://github.com/rstar327>`_
 - Kadir Can Ozden  `@bysiber  <https://github.com/bysiber>`_
 - Dhruvil Darji  `@dhruvildarji  <https://github.com/dhruvildarji>`_
+- `@nyxst4ck <https://github.com/nyxst4ck>`_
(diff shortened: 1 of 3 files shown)


nyxst4ck:
Daily triage follow-up: this PR is mergeable and its public checks pass. The only GitHub Actions run is `action_required`, which means the fork workflow has not been approved to run. Could a maintainer approve the run (or re-run it after approval) so the required build check can complete?

PR #2992, opened 2026-07-06
Auto-set allow_none when an Enum member has value None

### LINKED ISSUE #2985: Question: Should `fields.Enum()` accept `None` when an enum member has value `None`?
I’d like to clarify whether the following behavior is intentional or whether Marshmallow would consider supporting it.

#### Minimal example
```
import marshmallow as ma
from enum import Enum

class MyEnum(Enum):
    A = 'a'
    NONE = None

class S(Schema):
    sentinel = fields.Enum(MyEnum, by_value=True)

data = {'sentinel': None}
result = S().load(data)
```

Currently, the execution of this example raises: `ValidationError: {'sentinel': ['Field may not be null.']}`.

However, in Python, enum members can have `None` as values. From that perspective, it might be expected that `None` would be treated as part of the field's value domain when using `by_value=True`. According to the documentation, if this parameter is set to `True`, it delegates deserialization to `fields.Raw`, which has its own mechanism to deal with None values before the enum value resolution occurs.

We noticed that explicitly adding `allow_none=True` to the field allows the example to run successfully . However, this behavior is not documented and might not respect other design decisions. 

Would this be considered this a valid use case to support natively without requiring an explicit `allow_none=True`?

Thank you in advance for any clarification :) 

Similar to #2868

#### Tool Disclosure
This observation was raised by an automated testing tool currently being developed by our research team. Before submitting this issue, we tested the example in the last development version of marshmallow and reviewed the related documentation.

youdie006:
Closes #2985.

`fields.Enum` now defaults `allow_none=True` when the passed Enum has a member whose value is `None`, mirroring `fields.Constant(None)` (as suggested on the issue). An explicit `allow_none=False` is preserved, so loading `None` then raises a `ValidationError`.

Added `TestEnumAllowNone` covering by-name and by-value, the explicit-`allow_none=False` error case, and regression guards (an enum without a None member stays `allow_none=False`; an explicit `allow_none=True` is honored).

---

*AI disclosure: this PR was drafted with Claude Code (AI-assisted). The change was verified locally — the added `TestEnumAllowNone` cases fail without the fix and pass with it, and the full test suite (1186 tests) passes.*

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1947,6 +1947,12 @@ class Enum(Field[_EnumT]):
         self.enum = enum
         self.by_value = by_value
 
+        # If allow_none was not explicitly provided and the enum has a member
+        # whose value is None, None should be considered valid
+        # (mirrors fields.Constant(None) behavior). See issue #2985.
+        if kwargs.get("allow_none") is None and any(m.value is None for m in enum):
+            self.allow_none = True
+
         # Serialization by name
         if by_value is False:
             self.field: Field = String()
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -1,3 +1,5 @@
+from enum import Enum
+
 import pytest
 
 from marshmallow import (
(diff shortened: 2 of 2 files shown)


src/marshmallow/fields.py
```diff
@@ -1947,6 +1947,12 @@ def __init__(
         self.enum = enum
         self.by_value = by_value
 
+        # If allow_none was not explicitly provided and the enum has a member
+        # whose value is None, None should be considered valid
+        # (mirrors fields.Constant(None) behavior). See issue #2985.
+        if kwargs.get("allow_none") is None and any(m.value is None for m in enum):
```
deckar01, line 1953: ~Won’t `kwargs.get("allow_none")` error when the `allow_none` arg is unset and cause the tests to fail?~ nm. It does not.

deckar01:
> Drafted-by: Claude Code (Opus 4.8) (no human review before posting)

https://github.com/apache/airflow/pull/69431

Any reason no AI disclosure was added to this PR?

youdie006:
Fair question, and apologies for the inconsistency. Yes — like the linked PR, this was drafted with Claude Code; I've added the same AI disclosure to the description here. The change is verified locally (the added `TestEnumAllowNone` cases fail without the fix and pass with it, and the full suite passes), but you're right that the disclosure should have been there from the start.

PR #2994, opened 2026-07-07
Enum Allow None Default

### LINKED ISSUE #2985: Question: Should `fields.Enum()` accept `None` when an enum member has value `None`?

deckar01:
Fixes #2985

> If `by_value` is enabled on an enum with a `None` value `allow_none` defaults to `True`.

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1928,6 +1928,7 @@ class Enum(Field[_EnumT]):
     If `by_value` is `False` (default), enum members are (de)serialized by symbol (name).
     If it is `True`, they are (de)serialized by value using `marshmallow.fields.Raw`.
     If it is a field instance or class, they are (de)serialized by value using this field.
+    If `by_value` is enabled on an enum with a `None` value, `allow_none` defaults to `True`.
 
     .. versionadded:: 3.18.0
     """
@@ -1968,6 +1969,8 @@ class Enum(Field[_EnumT]):
             self.choices_text = ", ".join(
                 str(self.field._serialize(m.value, None, None)) for m in enum
             )
+            if "allow_none" not in kwargs and any(m.value is None for m in enum):
+                self.allow_none = True
 
     def _serialize(
         self, value: _EnumT | None, attr: str | None, obj: typing.Any, **kwargs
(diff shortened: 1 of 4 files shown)


sloria (approved):

src/marshmallow/fields.py
```diff
@@ -1928,6 +1928,7 @@ class Enum(Field[_EnumT]):
     If `by_value` is `False` (default), enum members are (de)serialized by symbol (name).
     If it is `True`, they are (de)serialized by value using `marshmallow.fields.Raw`.
     If it is a field instance or class, they are (de)serialized by value using this field.
+    If `by_value` is enabled on an enum with a `None` value `allow_none` defaults to `True`.
```
lafrech: Perhaps add a comma here:

> If `by_value` is enabled on an enum with a `None` value, `allow_none` defaults to `True`

lafrech (approved):

sloria:
@deckar01 feel free to update the changelog then merge when you're finished with this

PR #2995, opened 2026-07-08
Return default from get_value for out-of-range int index

AnayGarodia:
Fixes #2893

`get_value(obj, key, default)` is documented to fall back to `default` when a key is missing, and it does for string and dict keys. But an out-of-range integer index against a sequence raises `TypeError: attribute name must be string, not 'int'` instead of returning the default:

```python
>>> utils.get_value([0, 1, 2, 3], 999, default=42)
```

`_get_value_for_key` catches the `IndexError` from `obj[key]` and falls back to `getattr(obj, key, default)`, but when `key` is an int, `getattr` itself raises `TypeError` because attribute names must be strings. A non-string key can never be an attribute name, so this guards that path to return `default` directly instead of calling `getattr`.

Added a regression test covering the out-of-range int index and a missing int dict key; the int case raises `TypeError` on `dev` and passes with the change.

### DIFF
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -125,6 +125,8 @@ def _get_value_for_key(obj, key, default):
     try:
         return obj[key]
     except (KeyError, IndexError, TypeError, AttributeError):
+        if not isinstance(key, str):
+            return default
         return getattr(obj, key, default)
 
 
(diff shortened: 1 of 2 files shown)


lafrech:
Should this be reported as spam?

AnayGarodia:
Sorry for the noise, that's fair. I was going through open issues and #2893 looked like a straightforward gap, but you're right that an out-of-range integer index into `get_value` isn't a real-world scenario and this isn't worth carrying. No objection to the close; I'll be more selective about which issues actually warrant a change.

lafrech:
> I was going through open issues

You didn't even read the issue comments. Neither did you AI.

PR #2997, opened 2026-07-08
Return default from get_value for non-string keys (#2893)

apoorvdarshan:
## Summary

Fixes #2893. `utils.get_value` falls back to `getattr(obj, key, default)` whenever the container lookup fails, but `getattr` requires a **string** attribute name. So an out-of-range integer index (or any non-string key) raises `TypeError: attribute name must be string` instead of returning the default:

```python
utils.get_value([0, 1, 2], 999, default=3)          # TypeError (expected: 3)
utils.get_value({1: "a", 2: "b"}, 4, default="z")    # TypeError (expected: "z")
```

## Fix

Only call `getattr` when the key is a string; a non-string key can't be an attribute name, so return the `default`.

## Tests

- Added `test_get_value_with_out_of_range_int_key_returns_default`; it fails on `dev` (`TypeError`) and passes here.
- The full test suite passes (1179); `ruff` clean; changelog updated.

*Disclosure: prepared with AI assistance; reviewed and verified locally.*

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,15 @@
 Changelog
 =========
 
+Unreleased
+----------
+
+Bug fixes:
+
+- ``utils.get_value`` now returns the ``default`` for a non-string key such as
+  an out-of-range integer index, instead of raising a ``TypeError`` from
+  ``getattr`` (:issue:`2893`).
+
 4.3.0 (2026-04-03)
 ------------------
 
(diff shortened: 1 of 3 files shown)


lafrech:
>  reviewed and verified locally

Really? Did you read the issue you're pretending to fix (hint: and the discussion below the OP)?

apoorvdarshan:
Fair criticism, and I apologize. Two mistakes on my part: I didn't engage properly with the discussion under the issue before opening this, and I've since realized this org doesn't accept AI-assisted contributions — I shouldn't have submitted it here at all. Withdrawing it and I won't send more to marshmallow-code. Sorry for the wasted time.

PR #3002, opened 2026-07-16
fix: propagate dotted only/exclude paths correctly when Nested schema is an instance

Praveenkumar76:


sloria:
just checking: have you read our AI policy? https://github.com/marshmallow-code/.github/blob/main/AI_POLICY.md

sloria:
are you are marshmallow user? the PRs you've sent read to me as unattended LLM usage. per our AI policy we're not policing against LLMs per se, but before maintainers invest their limited time reviewing these, i'd like to understand the motivation here

Praveenkumar76:
To clarify, i've recently worked on marshmallow and encountered these edge cases while implementing and testing schemas, after debugging the root causes and checking the open issues, i've implemented the fixes and added regression tests, so i wanted to contribute the fixes back to the project, i respect the project's AI policy and i'll close this PR

PR #3004, opened 2026-07-20
Raise ValidationError (not ValueError) for NaN in TimeDelta

vidigoat:
### Problem

A `TimeDelta` field raises a bare `ValueError` (instead of `ValidationError`) when it receives `NaN`, so loading untrusted input can crash with an exception that isn't part of marshmallow's public error contract:

```python

class S(Schema):
    duration = fields.TimeDelta()

S().load({"duration": "nan"})
# ValueError: cannot convert float NaN to integer   <- leaks; expected ValidationError
```

`"inf"` / `"-inf"` already behave correctly (they raise `ValidationError`), which makes the `NaN` case an inconsistency rather than intended behaviour.

### Cause

`TimeDelta._deserialize` only catches `OverflowError` around `dt.timedelta(**kwargs)`:

```python
try:
    return dt.timedelta(**kwargs)
except OverflowError as error:
```

`float("nan")` passes the earlier `float(value)` guard, but `dt.timedelta(nan)` raises `ValueError` — not `OverflowError` — so it escapes uncaught. (`inf`/`-inf` work because `timedelta` raises `OverflowError` for those.)

### Fix

Catch `ValueError` alongside `OverflowError`, so `NaN` produces the field's standard `"Not a valid period of time."` error like every other invalid input:

```python
except (OverflowError, ValueError) as error:
```

Extended the existing parametrized `test_invalid_timedelta_field_deserialization` with `"nan"`, `"inf"`, `"-inf"`. Full test suite passes locally.

---
*Disclosure: I used an LLM to help find and draft this fix; I've reviewed every line, reproduced the bug, and run the suite myself.*

### DIFF
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1581,7 +1581,10 @@ class TimeDelta(Field[dt.timedelta]):
 
         try:
             return dt.timedelta(**kwargs)
-        except OverflowError as error:
+        except (OverflowError, ValueError) as error:
+            # OverflowError: value out of timedelta's range (e.g. infinity).
+            # ValueError: value is NaN (float("nan") passes the float() guard
+            # above but timedelta() rejects it).
             raise self.make_error("invalid") from error
 
 
--- a/src/marshmallow/fields.py
+++ b/src/marshmallow/fields.py
@@ -1583,8 +1583,7 @@ class TimeDelta(Field[dt.timedelta]):
             return dt.timedelta(**kwargs)
         except (OverflowError, ValueError) as error:
             # OverflowError: value out of timedelta's range (e.g. infinity).
-            # ValueError: value is NaN (float("nan") passes the float() guard
-            # above but timedelta() rejects it).
+            # ValueError: value is NaN.
             raise self.make_error("invalid") from error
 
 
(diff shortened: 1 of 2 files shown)


src/marshmallow/fields.py
```diff
@@ -1581,7 +1581,10 @@ def _deserialize(self, value, attr, data, **kwargs) -> dt.timedelta:
 
         try:
             return dt.timedelta(**kwargs)
-        except OverflowError as error:
+        except (OverflowError, ValueError) as error:
+            # OverflowError: value out of timedelta's range (e.g. infinity).
```
lafrech, line 1585: This comment is a bit too verbose. I'd stop at "value is NaN".

(Likewise, the commit message is pretty long for our standards. It could be one line.)

lafrech (changes requested):
Thanks.
Just a few minor comments. LLM tend to be too verbose.

vidigoat:
Comment shortened, thanks.

On the commit message: I wasn't able to rewrite the original commit here. If you squash-merge, feel free to use just the PR title as the message.

lafrech (approved):

lafrech (approved):
Thanks.

PR #3008, opened 2026-07-23
fix(tests): handle OSError exception message in test_from_timestamp_with_overflow_value

LaRoyBot:
Updates the exception matching regex in `test_from_timestamp_with_overflow_value` to handle platform-specific OSError messages ("Error converting value to datetime" / "Timestamp is too large") raised by OS C library timestamp conversions on Windows/Python 3.14.


sloria:
there were already several PRs sent for this same issue. also, please see our [AI policy](https://github.com/marshmallow-code/.github/blob/main/AI_POLICY.md).

PR #3009, opened 2026-07-25
Return default from get_value for out-of-range int keys

deepakganesh78:
Fixes #2893

### Root cause
`utils.get_value` delegates to `_get_value_for_key`, which on any exception from `obj[key]` falls back to `getattr(obj, key, default)`. `getattr` requires a *string* attribute name, so when `key` is an integer (an out-of-range list index, or a missing integer dict key) the fallback itself raises:

```
TypeError: getattr(): attribute name must be string
```

instead of returning the caller-supplied `default`. This contradicts the documented behaviour and the way string keys already degrade to `default`.

### Fix
Handle non-string keys explicitly in `_get_value_for_key`: attempt the item lookup and, on failure, return `default` directly — skipping the attribute-access step that can never succeed for a non-string key. String keys retain their existing `__getitem__` -> `getattr` fallback.

```python
from marshmallow import utils
utils.get_value([0, 1, 2, 3, 4, 5], 999, default=3)        # -> 3
utils.get_value({1: 'a', 2: 'b', 3: 'c'}, 4, default='z')  # -> 'z'
```

### Tests
Added a regression test `test_get_value_with_out_of_range_int_key_returns_default` in `tests/test_utils.py`.

Full suite: `1184 passed, 1 failed` — the single failure is the pre-existing, Windows-only `test_from_timestamp_with_overflow_value` (issue #2999), unrelated to this change.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,11 @@ Features:
 - If `by_value` is enabled on an enum with a `None` value `allow_none` now defaults to `True`.
   Thanks :user:`GeraldineGalindo` for the suggestion (:issue:`2985`).
 
+Bug fixes:
+
+- `marshmallow.utils.get_value` returns the ``default`` instead of raising a
+  ``TypeError`` when given an out-of-range or missing integer key (:issue:`2893`).
+
 4.3.0 (2026-04-03)
 ------------------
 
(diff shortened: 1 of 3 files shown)


sloria:

PR #3010, opened 2026-07-31
Return default from get_value for missing non-string keys

adarshsm:
Fixes #2893.

`utils.get_value` fell back to `getattr()` when item access failed, but `getattr()` rejects non-string attribute names — so an out-of-range list index or an absent integer dict key raised `TypeError` instead of returning the default:

```python
>>> utils.get_value([0, 1, 2], 999, default=3)
>>> utils.get_value({1: "a"}, 4, default="z")
```

A non-string key can only be resolved through item access, never attribute access, so this resolves non-string keys via item lookup only and returns the `default` when it misses. Present integer keys still work as before. Added a regression test and a changelog entry.


lafrech:
https://github.com/marshmallow-code/marshmallow/issues/2893#issuecomment-4058717845

PR #3013, opened 2026-08-02
Add partial keyword support to SchemaOpts (#2834)

### LINKED ISSUE #2834: SchemaOpts misses partial

rudrakumar07:
Fixes #2834.

Added `partial` keyword to `SchemaOpts.__init__` so that `class Meta: partial = True` works for schema-level partial loading.

Also updated `_deserialize` to check `self.opts.partial` as a fallback when a field is missing.

### DIFF
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -220,6 +220,7 @@ class SchemaOpts:
         self.unknown = getattr(meta, "unknown", RAISE)
         self.register = getattr(meta, "register", True)
         self.many = getattr(meta, "many", False)
+        self.partial = getattr(meta, "partial", False)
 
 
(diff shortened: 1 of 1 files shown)


lafrech:
> Also updated _deserialize to check self.opts.partial as a fallback when a field is missing.

Shouldn't be needed if self.partial = partial.

Also this PR lacks tests.

See #2835.

PR #3015, opened 2026-08-03
Pre commit autoupdate

lafrech:
Replaces #3014.

I took it upon myself to add two ruff rules to the ignore list.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@ ci:
   autoupdate_schedule: monthly
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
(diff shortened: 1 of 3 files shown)


sloria:
thanks!

on a related note: i wonder if we should [use renovate for pre-commit updates](https://docs.renovatebot.com/modules/manager/pre-commit/) to drop a service dependency?

lafrech:
IIUC, pre-commit CI would still run the checks, like it does on all PRs, 
so we would still depend on it, just not for the auto-update part, 
right?

I don't mind either way.

sloria:
i guess we'd use github actions to run the checks, then renovate for updates.

it's working well as is, though, so no need to change

PR #3016, opened 2026-08-05
Fix URL validator rejecting a fragment after an empty path

hdimer:
`validate.URL` rejects an absolute URL whose fragment follows an empty path with no query:

```python
from marshmallow import validate
validate.URL()("https://example.com#frag")   # ValidationError: Not a valid URL.
```

Adding a path or a query makes it pass (`https://example.com/#frag`, `https://example.com?a=1#frag`), so only the bare-host + fragment case is affected. Per RFC 3986 a fragment may follow an empty path, and anchor links to a domain root (`https://example.com#section`) are common and valid.

The tail regex after the host allowed `/`, `/…` or `?…`, but a `#fragment` could only appear inside the path/query part. This adds `#` to the set of characters that may introduce the URL tail. As a consequence a bare `#frag` is now also accepted in relative mode, which matches the existing handling of a bare `?query` and is an RFC 3986 same-document reference.

### Testing

Added cases to the existing parametrized URL tests in `tests/test_validate.py` (absolute `#fragment` / `user@host#fragment`, and a bare `#frag` in both relative lists). Full test suite passes (`1188 passed`); `ruff check`/`ruff format` clean.

---
Used AI assistance on this; I reviewed and tested the change myself.

### DIFF
--- a/src/marshmallow/validate.py
+++ b/src/marshmallow/validate.py
@@ -157,7 +157,8 @@ class URL(Validator):
                     r"(?::\d+)?",
                 )
             )
-            relative_part = r"(?:/?|[/?]\S+)\Z"
+            # `#` may lead a fragment right after the host, e.g. `https://example.com#frag`.
+            relative_part = r"(?:/?|[/?#]\S+)\Z"
 
             if relative:
                 if absolute:
--- a/src/marshmallow/validate.py
+++ b/src/marshmallow/validate.py
@@ -157,7 +157,6 @@ class URL(Validator):
                     r"(?::\d+)?",
                 )
             )
-            # `#` may lead a fragment right after the host, e.g. `https://example.com#frag`.
             relative_part = r"(?:/?|[/?#]\S+)\Z"
 
             if relative:
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -188,3 +188,4 @@ Contributors (chronological)
 - `@rstar327 <https://github.com/rstar327>`_
 - Kadir Can Ozden  `@bysiber  <https://github.com/bysiber>`_
 - Dhruvil Darji  `@dhruvildarji  <https://github.com/dhruvildarji>`_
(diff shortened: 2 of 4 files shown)


src/marshmallow/validate.py
```diff
@@ -157,7 +157,8 @@ def _regex_generator(
                     r"(?::\d+)?",
                 )
             )
-            relative_part = r"(?:/?|[/?]\S+)\Z"
+            # `#` may lead a fragment right after the host, e.g. `https://example.com#frag`.
+            relative_part = r"(?:/?|[/?#]\S+)\Z"
```
lafrech, line 160: I'd remove the comment.

lafrech (approved):
Thanks.

I suck at regexes but I think Django ORM does that too: https://github.com/django/django/blob/stable/6.0.x/django/core/validators.py#L130.

lafrech:
BTW anyone knows why I only get "rebase and merge" and "squash and merge" options anymore on the green merge button above ? What happened to the "normal merge" option?

sloria:
@lafrech Allow merge commits is unchecked in settings. Feel free to re-enable it, but is there an advantage to that over the other options?


Screenshot 2026-08-07 at 18 45 36@2x

hdimer:
Done, removed.

lafrech:
> @lafrech Allow merge commits is unchecked in settings. Feel free to re-enable it, but is there an advantage to that over the other options?

Ah OK. Fine, then.

Typically, in my repos, I rebase CI PRs and merge non-trivial PRs. A side-benefit of the merge is that the commit contains the PR number. Also, I have projects for which merges are fine but rebases nearly impossible, although this shouldn't happen here, as we expect users to rebase PRs.

Overall, this setting is fine for marshmallow.

lafrech:
LGTM. I'll be AFK for a few days. Holding this in case another maintainer wants to review. I can merge and release when I'm back.

PR #3017, opened 2026-08-09
Return the default for an out-of-range int key in `get_value`

dchaudhari7177:
Closes #2893.

## Problem

```python
>>> utils.get_value([0, 1, 2], 999, default=3)
```

`get_value` documents a `default` for exactly this case, and returns it correctly for string keys. With an int key a miss raises instead, so callers cannot tell a missing index from a real error.

Three shapes are affected, all from the issue:

| call | before | after |
|---|---|---|
| `get_value([0, 1, 2], 999, default=3)` | `TypeError` | `3` |
| `get_value({1: "a"}, 4, default="z")` | `TypeError` | `"z"` |
| `get_value(point_obj, 0, default=None)` | `TypeError` | `None` |

## Cause

`_get_value_for_key` falls back to `getattr(obj, key, default)` when the subscript raises:

```python
try:
    return obj[key]
except (KeyError, IndexError, TypeError, AttributeError):
    return getattr(obj, key, default)
```

`getattr`'s third argument suppresses a *missing* attribute, not an invalid attribute *name* — a non-string name is a `TypeError` before the lookup happens. So the guard that makes the string path safe does nothing on the int path.

The same applies to the earlier `if not hasattr(obj, "__getitem__")` branch, which hands the key straight to `getattr`.

## Fix

Route both `getattr` calls through a helper that returns the default for a non-string key. An int key can only ever have been an index or a mapping key, and the caller has already tried both, so the default is the answer.

String keys are untouched — they still reach `getattr` exactly as before.

## Tests

- Five parametrized cases: index past the end, negative index past the start, absent int key in an int-keyed mapping, a nested sequence (the issue's third case), and an object with no `__getitem__`.
- One asserting the no-default behaviour is `missing_`, not an exception.
- One control asserting a string key still resolves a real attribute and still honours its default — i.e. the fix does not disable attribute fallback.

Revert-verified: with `utils.py` reverted, 6 of the 8 fail; the control passes on both.

Full suite on Windows / Python 3.13: **1194 passed, 1 failed**. The failure is `test_from_timestamp_with_overflow_value`, the pre-existing Windows message mismatch in #2999 (already addressed by #3000) — it fails identically on a clean `dev`.

## AI disclosure

Written with Claude Code (Claude Opus 5): it diagnosed the `getattr` name-vs-value distinction, wrote the change in `utils.py` and the tests in `test_utils.py`, and drafted this description. I ran the suite and the revert check by hand.

### DIFF
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -118,14 +118,24 @@ def _get_value_for_keys(obj, keys, default):
     )
 
 
+def _getattr_or_default(obj, key, default):
+    # Attribute names are always strings, so a non-string key cannot name one:
+    # `getattr` would raise TypeError rather than simply miss and return the
+    # default. Such a key was only ever an index or a mapping key, and both
+    # have already been tried by the caller, so the default is the answer.
+    if not isinstance(key, str):
+        return default
+    return getattr(obj, key, default)
+
+
 def _get_value_for_key(obj, key, default):
     if not hasattr(obj, "__getitem__"):
-        return getattr(obj, key, default)
+        return _getattr_or_default(obj, key, default)
 
     try:
         return obj[key]
     except (KeyError, IndexError, TypeError, AttributeError):
-        return getattr(obj, key, default)
+        return _getattr_or_default(obj, key, default)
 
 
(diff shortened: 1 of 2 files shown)


lafrech:
Read the issues you pretend to address.

PR #3018, opened 2026-08-09
Add GitHub action to ensure PR authors are in AUTHORS.rst

sloria:
To reduce the chore of adding to AUTHORS.rst.

To finish this up, I'll need to set a `AUTHORS_TOKEN` with a PAT that has public repo access only

### DIFF
--- a/.github/workflows/authors.yml
+++ b/.github/workflows/authors.yml
@@ -1,0 +1,62 @@
+name: authors
+on:
+  # zizmor: ignore[dangerous-triggers] nothing from the PR is fetched or executed
+  pull_request_target:
+    types: [opened, reopened]
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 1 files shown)


lafrech:
I agree on the principle.

Can't review thoroughly in the coming days, so if you're confident on the implementation, go for it.

sloria:
alternatively we could just get rid of AUTHORS.rst. it's rather redundant with github contributors list, and `pyproject.toml` has the list of maintainers.

lafrech:
Dunno. I get the point of minimizing maintenance effort. But could it be displayed in the docs? Also contributors may delete their GH account.

An alternative could be a simpler action that only checks the contributor is in the list but doesn't add it. But since you already wrote the action to add it...

PR #3020, opened 2026-08-13
fix(utils): allow negative POSIX timestamps in from_timestamp()

### LINKED ISSUE #3019: [Bug] Negative POSIX timestamps cannot be deserialized after timestamp serialization
## Bug description

`fields.DateTime(format="timestamp")` can serialize valid datetimes before the Unix epoch (1970) into negative POSIX timestamps, but `marshmallow` then rejects those same negative values when deserializing them.

This creates a round-trip failure where a value produced by Marshmallow cannot be read back by the same field.

## Reproduction

```python
from datetime import datetime, timezone

field = fields.DateTime(format="timestamp")

value = datetime(1950, 5, 3, tzinfo=timezone.utc)

serialized = field.serialize("date", {"date": value})
print(serialized)
# -620568000.0

field.deserialize(serialized)
# marshmallow.exceptions.ValidationError:
# Not a valid datetime.
```

The underlying Python `datetime.fromtimestamp()` supports negative timestamps, so the negative value itself is valid.

## Expected behavior

A timestamp produced by `fields.DateTime(format="timestamp")` should be accepted by the corresponding deserializer, including valid negative POSIX timestamps representing dates before 1970.

## Actual behavior

Negative POSIX timestamps are rejected during deserialization, even though the serializer can produce them from valid pre-1970 datetimes.

This means timestamp serialization is not reliably round-tripable for dates before the Unix epoch.

## Root cause

`marshmallow.utils.from_timestamp()` explicitly rejects values below zero before calling Python's timestamp conversion.

The underlying `datetime.fromtimestamp()` implementation can correctly handle these negative values, so the explicit negative-value restriction is unnecessarily broad.

Existing `OSError` and `OverflowError` handling can continue to handle genuine platform-specific timestamp conversion failures.

## Proposed fix

Remove the explicit rejection of negative timestamp values from `from_timestamp()` and rely on the existing timestamp conversion and exception handling.

Add regression coverage that:

1. Serializes valid pre-1970 datetimes using both `timestamp` and `timestamp_ms`.
2. Confirms the serialized values are negative.
3. Deserializes those values.
4. Confirms the original datetimes are recovered.

## Validation

I have reproduced the issue on the current development branch.

The regression test fails on the unfixed implementation and passes after the proposed fix.

The full test suite passes with the fix.

I would like to contribute a fix for this issue.

Sohel2309:
## Summary

Fixes #3019.

`from_timestamp()` currently rejects negative POSIX timestamps even though negative timestamps are valid and are produced by `fields.DateTime(format="timestamp")` when serializing datetimes before the Unix epoch.

This breaks serialization/deserialization round-tripping for valid pre-1970 datetimes.

## Reproduction

```python
from datetime import datetime, timezone

field = fields.DateTime(format="timestamp")
value = datetime(1950, 5, 3, tzinfo=timezone.utc)

serialized = field.serialize("date", {"date": value})
# -620568000.0

field.deserialize(serialized)
# ValidationError: Not a valid datetime.
```

The same field can therefore produce a value that it cannot deserialize.

## Root cause

`from_timestamp()` explicitly rejected values below zero before calling `datetime.fromtimestamp()`.

Python's timestamp conversion already supports negative POSIX timestamps, so this explicit restriction was unnecessarily preventing valid dates from being deserialized.

## Fix

Remove the explicit negative-value rejection and rely on the existing timestamp conversion and `OSError` / `OverflowError` handling for genuine conversion failures.

Regression coverage was added for negative timestamps, including timestamp and timestamp-millisecond formats.

## Validation

- Regression tests pass.
- Full test suite: 1197 passed.
- Ruff formatting passes.
- Ruff lint passes.
- Mypy passes.
- No unrelated changes are included.

Fixes #3019

### DIFF
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -41,8 +41,6 @@ def from_timestamp(value: typing.Any) -> dt.datetime:
     if value is True or value is False:
         raise ValueError("Not a valid POSIX timestamp")
     value = float(value)
-    if value < 0:
-        raise ValueError("Not a valid POSIX timestamp")
 
     # Load a timestamp with utc as timezone to prevent using system timezone.
     # Then set timezone to None, to let the Field handle adding timezone info.
--- a/tests/test_deserialization.py
+++ b/tests/test_deserialization.py
@@ -569,8 +569,14 @@ class TestFieldDeserialization:
                 dt.datetime(2013, 11, 10, 0, 23, 45, 123456),
             ),
             ("timestamp", 1, dt.datetime(1970, 1, 1, 0, 0, 1)),
+            # Negative timestamps (pre-1970-01-01 dates) are valid POSIX
+            # timestamps and must be accepted; see
+            # test_timestamp_field_deserialization_round_trip_negative below.
+            ("timestamp", -1, dt.datetime(1969, 12, 31, 23, 59, 59)),
+            ("timestamp", -86400, dt.datetime(1969, 12, 31, 0, 0, 0)),
             ("timestamp_ms", 1384043025000, dt.datetime(2013, 11, 10, 0, 23, 45)),
             ("timestamp_ms", 1000, dt.datetime(1970, 1, 1, 0, 0, 1)),
+            ("timestamp_ms", -1000, dt.datetime(1969, 12, 31, 23, 59, 59)),
         ],
     )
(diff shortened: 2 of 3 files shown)


lafrech:
Replaced by #3026.

PR #3021, opened 2026-08-13
fix(schema): validate dotted nested options on non-nested and non-existent fields (#2180)

### LINKED ISSUE #2180: Are nested options applied correctly?
Experimenting with `only`, I noticed it is possible to assign an invalid value to a Field's `only` attribute. 

Looking at the code in the `__apply_nested_option` method of the `schema.Schema` class, there is no validation done when propagating the `only` value to a field.

```python

class S(Schema):
    f = fields.Str()

S(only=['f.a.b'])
<S(many=False)>  # actual output.
ValueError: Option: 'only' with value: 'a.b' is not valid for Field: 'f'.  # expected output.
```

All comments are welcome.

harmehak0173:
### Summary of Changes
Fixes #2180.

When `only` or `exclude` options contained dotted paths (e.g. `only=["foo.bar"]`), marshmallow previously:
1. Silently stripped the dotted segment when `foo` was a non-nested scalar field (e.g. `fields.Str()`), converting `only=["foo.bar"]` into `only=["foo"]` and dumping the entire scalar field without error or warning.
2. Crashed with an unhandled `KeyError` when `foo` was a non-existent field name.

### Solution
1. In `Schema.__apply_nested_option`, validate that dotted parent field names exist in `self.declared_fields`. If the parent field does not exist, raise a clean `ValueError(f"Invalid fields for {self}: {{'{key}'}}.")`.
2. Inspect the target field (unwrapping container fields like `List`) without invoking `@property` getters prematurely, ensuring the field supports nested options (`schema`, `only`, `exclude`, or `nested`). If the parent field is a non-nested scalar field (e.g. `fields.Str()`), raise a clean `ValueError(f"Cannot apply nested option '{option_name}' to non-nested field '{key}'.")`.
3. Added unit tests in `tests/test_schema.py` (`test_nested_options_on_non_nested_or_non_existent_fields`).

### Testing
- `ruff check` & `ruff format`: Passed cleanly.
- `pytest`: 1,188 passed.

### DIFF
--- a/src/marshmallow/schema.py
+++ b/src/marshmallow/schema.py
@@ -970,14 +970,31 @@ class Schema(metaclass=SchemaMeta):
             nested_options[parent].append(nested_names)
         # Apply the nested field options.
         for key, options in iter(nested_options.items()):
+            if key not in self.declared_fields:
+                raise ValueError(f"Invalid fields for {self}: {{'{key}'}}.")
+            field_obj = self.declared_fields[key]
+            inner_obj = field_obj
+            while hasattr(inner_obj, "inner"):
+                inner_obj = inner_obj.inner
+            if not (
+                hasattr(type(inner_obj), "schema")
+                or hasattr(type(inner_obj), "nested")
+                or hasattr(type(inner_obj), "only")
+                or hasattr(type(inner_obj), "exclude")
+                or "only" in inner_obj.__dict__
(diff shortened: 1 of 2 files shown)


deckar01:
This diff is mostly non-sense, please see our [AI policy](https://github.com/marshmallow-code/.github/blob/main/AI_POLICY.md)

PR #3022, opened 2026-08-14
fix: get_value returns default for non-string keys

Sohel2309:
Fixes #2893.

`utils.get_value()` should return the supplied default when a key cannot
be resolved. When a non-string key such as an integer is used, the
`getattr()` fallback currently raises:

This happens both when `obj[key]` raises an exception and when the object
does not implement `__getitem__`.

Fix both paths by only using `getattr()` when the key is a string.

Added regression coverage for:
- out-of-range list index
- missing integer dictionary key
- object without `__getitem__`

The new regression test passes. The full `tests/test_utils.py` suite
has 20 passing tests; one unrelated existing timestamp test fails due
to an existing error-message mismatch.


sloria:

PR #3023, opened 2026-08-14
fix: get_value returns default for out-of-range/missing int keys

Sohel2309:
Fixes #2893.

`utils.get_value()` raised `TypeError: attribute name must be string` instead of returning the caller-supplied `default` when passed an out-of-range or missing int key.

This happened because `_get_value_for_key()` falls back to `getattr(obj, key, default)` after `obj[key]` fails. However, `getattr()` requires a string attribute name, so when `key` is an `int`, that fallback raises `TypeError` instead of returning `default`.

Fix: only attempt the `getattr()` fallback when `key` is a `str`; for other key types, return `default` directly.

Added a regression test covering the reproduction cases from the issue.

The full test suite passes with no regressions.

### DIFF
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -120,12 +120,12 @@ def _get_value_for_keys(obj, keys, default):
 
 def _get_value_for_key(obj, key, default):
     if not hasattr(obj, "__getitem__"):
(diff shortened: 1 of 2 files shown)


deckar01:

PR #3024, opened 2026-08-14
Simplify `get_value` to match actual usage

deckar01:
The `str | int` typehint and accompanying test were incorrect, because the only two calls to `get_value` in marshmallow have type hints that require the `attr` key to be a string.

Closes #2893

## Benchmark

Looping instead of recursion reduces dump time ~3%. Unconditionally splitting would further reduce the line count, but the unnecessary list allocation has a significant performance penalty.

| python version | dev (usec/dump) | this (usec/dump) | 
|-----|-----|------|
| 3.14  | 8.9    | 8.6 (-3.4%) | 
| 3.13 | 11.8   | 11.4 (-3.4%) | 
| 3.12 | 14.5   | 14.5 (-0.0%) | 
| 3.10 | 16.5   | 15.9 (-3.6%) |

### DIFF
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -94,7 +94,7 @@ def pluck(dictlist: list[dict[str, typing.Any]], key: str):
 # Various utilities for pulling keyed values from objects
 
 
-def get_value(obj, key: int | str, default=missing):
+def get_value(obj, key: str, default=missing):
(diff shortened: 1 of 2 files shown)


sloria (approved):
great catch!

lafrech (approved):
Great job closing the AI-slop trap (#2893).

PR #3025, opened 2026-08-17
Return default for out-of-range int index in get_value (#2893)

maharshi-coding:
Fixes #2893.

## The bug

`utils.get_value` falls back to `getattr(obj, key, default)` when `obj[key]` raises. But `getattr` requires a string attribute name, so a non-string key — e.g. an out-of-range `int` index — raises `TypeError: attribute name must be string` instead of returning the `default`:

```python
from marshmallow import utils

utils.get_value([0, 1, 2], 999, default=3)   # TypeError  (expected: 3)
utils.get_value({1: "a"}, 4, default="z")    # TypeError  (expected: "z")
```

This is inconsistent with the string-key path and with the documented `default` behavior.

## The fix

Guard the `getattr` fallback so a non-string key returns `default` directly (a non-string can never name an attribute). Valid indexing and string-key attribute fallback are unchanged. Added a regression test covering lists, int-keyed dicts, and lists of objects.

The full suite passes locally. (One pre-existing, Windows-only `test_from_timestamp_with_overflow_value` failure reproduces on a clean checkout too and is unrelated to this change.)

Happy to add a `CHANGELOG.rst` entry in whatever version/format you prefer.

---

*Disclosure: this change was prepared with AI assistance; I reviewed it and verified the tests before submitting.*

### DIFF
--- a/src/marshmallow/utils.py
+++ b/src/marshmallow/utils.py
@@ -120,11 +120,17 @@ def _get_value_for_keys(obj, keys, default):
 
 def _get_value_for_key(obj, key, default):
     if not hasattr(obj, "__getitem__"):
+        # ``getattr`` requires a string attribute name; a non-string key (e.g. an
+        # int index) cannot name an attribute, so fall back to ``default``.
+        if not isinstance(key, str):
+            return default
         return getattr(obj, key, default)
 
     try:
         return obj[key]
     except (KeyError, IndexError, TypeError, AttributeError):
+        if not isinstance(key, str):
+            return default
         return getattr(obj, key, default)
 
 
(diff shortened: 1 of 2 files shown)


lafrech:
Congratulations, thanks to this contribution, you just obtained the "certified AI slop" badge.

image


Before rushing for more, please see our [AI policy](https://github.com/marshmallow-code/.github/blob/main/AI_POLICY.md).

You may also be interested in the conversation in the issue your AI pretends to fix, and in #3024, in which a maintainer fixes it properly, hopefully ending a long-lasting streak of such PRs.

maharshi-coding:
You're right, and I'm sorry for the noise. This violated your AI policy, and per the discussion in the issue the change was wrong anyway. I've withdrawn my other PRs too and I'll stop.

PR #3028, opened 2026-08-19
tests: fix tautological assertion in test_containsonly_custom_message

aoright:
### Summary
- In `tests/test_validate.py::test_containsonly_custom_message`, replace tautological `assert expected in str(expected)` with `pytest.raises(ValidationError, match=...)` to actually verify the exception message against the expected string.


deckar01:
https://github.com/marshmallow-code/.github/blob/main/AI_POLICY.md

PR #3033, opened 2026-08-26
Fix dot notation handling for Nested fields given schema instances

abhijeet117:
## Summary

`fields.Nested` given a callable returning a `Schema` instance bypassed the dot-aware resolution that `Schema.__init__` applies to its arguments, so dot-qualified `only`/`exclude` entries (e.g. `b.a2`) were intersected against top-level field names and silently produced empty selections. The instance branch of `Nested.schema` now resolves options through `_normalize_nested_options` before merging them with the instance constraints, matching the class path.

## Testing

Reproduced with the issue snippet: `DSchema(only=("d.c.b.a2",)).dump(...)` returned `{'d': {'c': {}}}` instead of the selected subfield; it now dumps correctly. Added regression tests covering ancestor dot notation and dotted `only`/`exclude` passed to instance-backed `Nested`; all three fail without the fix and pass with it. Full pytest run, plus `ruff check`, `ruff format --check`, and `mypy`, pass locally.

## Checklist

- bug reproduced before fix
- root cause identified
- bug fixed
- tests passed

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -190,3 +190,4 @@ Contributors (chronological)
 - Dhruvil Darji  `@dhruvildarji  <https://github.com/dhruvildarji>`_
 - Haïm Dimer `@hdimer  <https://github.com/hdimer>`_
 - `@vidigoat <https://github.com/vidigoat>`_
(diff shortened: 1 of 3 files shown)


sloria:

Issue #2194, opened 2023-10-05
Dot in key is not parsed correclty on loads()


dchirikov:
Hey, devs!
First, thanks for the great project! Really useful and helpful. 
I found a weird behavior while experimenting with dictionary keys containing dots (".") in it. Here is the example:
```
>>> from marshmallow import Schema, fields
>>> TestSchema = Schema.from_dict({"something": fields.Str(), "some.thing": fields.Str()})
>>> TestSchema().loads('{"something": "data"}')
{'something': 'data'}
>>> TestSchema().loads('{"some.thing": "data"}')
{'some': {'thing': 'data'}}
```
Is it a correct behavior? How to work around it?

lafrech:
I don't see an explicit test for this. @sloria was this intended?

deckar01:
`from_dict` lets you use field names that would otherwise not be valid class attributes. The observed behavior is consistent with using dotted keys for `data_key` and `attribute`.

```py
from marshmallow import Schema, fields


Test = Schema.from_dict({'foo.bar': fields.Str()})
schema = Test()
obj = schema.load({'foo.bar': 'baz'})
print('from_dict', 'load', obj)
data = schema.dump(obj)
print('from_dict', 'dump', data)


class Test(Schema):
    foo_bar = fields.Str(data_key='foo.bar', attribute='foo.bar')
schema = Test()
obj = schema.load({'foo.bar': 'baz'})
print('Schema', 'load', obj)
data = schema.dump(obj)
print('Schema', 'dump', data)
```

```
from_dict  load  {'foo': {'bar': 'baz'}}
from_dict  dump  {'foo.bar': 'baz'}

Schema     load  {'foo': {'bar': 'baz'}}
Schema     dump  {'foo.bar': 'baz'}
```

This behavior is not documented for `attribute` thought. It appears to be a side effect of performing dotted name resolution in `get_value` at load time.

A workaround would be to explicitly define an `attribute` for the field without dots so that no nesting occurs.

```py
from marshmallow import Schema, fields


Test = Schema.from_dict({'foo.bar': fields.Str(attribute='foo_bar')})
schema = Test()
obj = schema.load({'foo.bar': 'baz'})
print('from_dict', 'load', obj)
data = schema.dump(obj)
print('from_dict', 'dump', data)


class Test(Schema):
    foo_bar = fields.Str(data_key='foo.bar', attribute='foo_bar')
schema = Test()
[...]

PR #3042, opened 2026-09-07
Document that a dot in `attribute` is a path separator

dchaudhari7177:
Refs #2194.

The conclusion on that issue was that the behaviour is intended — a side effect of dotted-name resolution — but, as @deckar01 noted, *"This behavior is not documented for `attribute`"*. This documents it and pins the untested half.

`get_value` walks a dotted key on dump and `set_value` nests it on load. Both say so in their own docstrings, but the `attribute` parameter does not, and `attribute` is where a user meets the behaviour. `Schema.from_dict({"some.thing": fields.Str()})` names the field `some.thing`, an unset `attribute` defaults to the field name, and the load result is `{"some": {"thing": ...}}` with nothing in the reference to predict it.

The docstring now states that a dot is a path separator in both directions, and that there is no escape — a key that genuinely contains a dot needs `data_key` with a plain `attribute`, which is the workaround the reporter asked for.

### Tests

`test_serializer_can_specify_nested_object_as_attribute` already pins the dump direction. The load direction — the one that surprised the reporter — was untested, so it could have changed silently. Added:

- `test_dotted_attribute_is_a_path_on_load_too` — load nests, and the value round-trips back to the flat external key.
- `test_from_dict_key_containing_a_dot_is_treated_as_a_path` — the reported case exactly, both directions.

### Verified

- `pytest tests/` — 1191 passed, 1 failed. The failure is `test_utils.py::test_from_timestamp_with_overflow_value`, which fails identically on pristine `dev` here (platform-dependent overflow bound); unrelated to this change.
- `ruff check` and `ruff format --check` clean on both files at the pinned v0.16.1.

Docs and tests only — no behaviour change.


sloria:

PR #3044, opened 2026-09-08
docs: drop duplicated 'the' in quickstart

simpleqt:
`docs/quickstart.rst`: "if **the the** value is missing from the input" → single "the". Docs-only.


lafrech:
Fixed in dev branch.

Also, please see our [AI policy](https://github.com/marshmallow-code/.github/blob/main/AI_POLICY.md).

PR #3052, opened 2026-09-18
Reject bool for Decimal(places=...)

Pitchfork-and-Torch:
## Summary
`bool` is a subclass of `int`, so `fields.Decimal(places=True)` silently quantized values to **1** decimal place (and `places=False` to 0 places). Raise `TypeError` instead when `places` is a `bool`.

## Test plan
- [x] Added `test_decimal_places_rejects_bool`
- [x] Existing `test_decimal_field` still passes


lafrech:
I'd rather not perform this kind of checks.
Besides, this should be caught by typing checks.

